PR #7, opened 2019-11-07
Settings menu for manually setting the time/date

jasnell:
It's very rudimentary, but it works.


gfwilliams:
Thanks! Just thinking - maybe in 'settings-init' we could check the time and if it's 1970, set it to the start date of Nodeconf. At least then only hour/minute/day need setting

PR #8, opened 2019-11-08
fun run app and openlocation icon

conoro:


gfwilliams:
Thanks! Please can you spliit nceu-funrun.js up into the icon/json parts as in https://github.com/espruino/BangleApps#developing-your-own-app though?

conoro:
Sorry, missed that step! On it.

PR #10, opened 2019-11-08
NodeConfEU Out of the Box Startup

jasnell:
Modifies bootloader to look for an app +start at load. If it exists, it runs it first. 

The start up app runs then needs to clean itself up after by deleting itself, ensuring that the start sequence only runs once. Startup app can be reloaded onto the watch from the app store at any time


jasnell:
May end up iterating on this a bit more but this is the initial experience at least

PR #17, opened 2019-11-11
add party parrot w000

thekemkid:


gfwilliams:
Totally what Bangle.js needed!

gfwilliams:
Just made a few tweaks as it seems there were some issues with it - works great now. It'd be even better if it went faster when you waved your arms around like a madman :)

PR #18, opened 2019-11-11
oops too

thekemkid:


gfwilliams:
Sorry - just made some similar tweaks. Actually doing the decompress for the render was quite good as it saved a bunch of memory.

If you want the screen clear I'd just remove `transparent : 254,` - that way you'd avoid flicker.

gfwilliams:
Tweaks already applied :)

PR #20, opened 2019-11-12
Added the connect button

gdorsi:


gfwilliams:
Thanks!

gfwilliams:
Just a note that if you click 'upload' on an app right away the connect button doesn't disappear - it also doesn't reappear if the watch disconnects. I guess a quick hack would be to add something to `comms.js` that just polls `Puck.isConnected()` every second when connected. Not nice, but maybe in future the Puck library could be modified to add connected/disconnected events

PR #22, opened 2019-11-12
update the connection button at every connection change

gdorsi:


index.js
```diff
@@ -247,23 +247,28 @@ function refreshMyApps() {
...
     refreshLibrary();
   }).catch(err => {
     showToast("Getting app list failed, "+err,"error");
   });
 }
 
+var connectMyDeviceBtn = document.getElementById("connectmydevice");
+
+function handleConnectionChange(connected) {
+  connectMyDeviceBtn.style.display = connected ? 'none' : '';
```
gdorsi: Would make sense to clean the library on disconnection?

gdorsi:
If you want to try the PR before the merge this is the pages link: https://gdorsi.github.io/BangleApps/

index.js
gfwilliams: I don't know... Maybe if it was an accidental disconnection it makes sense to keep it?

gfwilliams:
This looks great - thanks!

No worries if you don't feel like it, but actually as there's a connect button, you could replace it with a 'Disconnect' one when the page is connected to the watch :)

index.js
gdorsi: Ok, also the implementation may rise the complexity bar of the script

PR #26, opened 2019-11-12
Quick fixes and performance improvement.

RichardBoyewa:
Some performance improvement updates. Suspends timer when LCD is off, and resume when it comes back on.
Hourly buzz also added.


gfwilliams:
Nice - thanks! Just checking - but do you not get an `lcdPower` event if you manually use `setLCDPower`? If you do, you might be getting 2 copies of each interval when you perform a gesture

RichardBoyewa:
@gfwilliams I have updated it. So, it now relies on the `lcdPower` and it works.

gfwilliams:
Thanks!

PR #27, opened 2019-11-12
added the disconnect button

gdorsi:


index.js
```diff
@@ -250,24 +250,35 @@ function refreshMyApps() {
...
 
-document.getElementById("myappsrefresh").addEventListener("click", getInstalledApps);
-connectMyDeviceBtn.addEventListener("click", getInstalledApps);
+document.getElementById("myappsrefresh").addEventListener("click", () => {
+  getInstalledApps().catch(err => {
+    showToast("Getting app list failed, "+err,"error");
+  });
+});
+connectMyDeviceBtn.addEventListener("click", () => {
+  if (connectMyDeviceBtn.classList.contains('is-connected')) {
```
gdorsi: Here i have used a derived connection state and not the Puck.isConnected() state to avoid toggle/panic click race-conditions.
  
Would be better to manage this with a pending state but it would require state centralization and a bigger refactoring.

gfwilliams:
This looks spot on - thank you!

PR #37, opened 2019-11-13
Add GPS Info application

chrisvel:
I quickly set up an app to show altitude, GPS [lat, lon], time and satellites in one screen.


gfwilliams:
Thanks - looks great! Would you be able to revert the big whitespace changes in `apps.json` though?

If you wanted to add to it, in the `GPS-raw` event there's some really cool data about the number and signal strength of each of the satellites :)

gfwilliams:
Thanks!

chrisvel:
Reverted - sorry about that, it's automatically formatted on save by editor. 
Nice idea about `GPS-raw` data, they could be added in a table (GPS-info v.2). 
Thanks for this awesome gadget!

PR #42, opened 2019-11-14
[app] promodoro

best4him:


aivopaas:
Looks like you accidentally copied apps.json file into apps folder

gfwilliams:
Thanks! It might be an idea to test this by enabling GitHub pages on your fork, because I think there are some small gotchas:

* `promodoro` is bigger than 7 chars, so the menu item uploaded to the watch wouldn't load the app
* `{"name": "*promodoro","url": "promodoro.js","evaluate": true}` and `"icon":"*promodoro",`  should reference `promodoro-icon.js` if you had an icon file in JS, but it seems like you don't have an icon

Trivial stuff to fix though so I'll just tweak it now and merge in - thanks for removing the extra `apps.json` :)

best4him:
Thanks! :)

PR #44, opened 2019-11-15
Skeleton

MaBecker:


MaBecker:
Please check if this is ok like this.

ra100:
@MaBecker could you fix indentation in skeleton app to match with other apps (2 spaces)?
And for the parts like `<your main function>();;`, I'd maybe put it in comment, so the final JS file is actually valid, like:
```js
Bangle.on('lcdPower', (on) => {
  if (on) {
    drawWidgets();
    // <your main function>();
  }
});
```
Anyway, that's just my personal preference.

apps/clock-mixed.js
```diff
@@ -4,6 +4,8 @@
     const Radius = { "center": 8, "hour": 78, "min": 95, "dots": 102 };
     const Center = { "x": 120, "y": 132 };
 
+    try{ E.setTimeZone(require("Storage").readJSON("@setting").timezone);} catch(e){}
```
gfwilliams: This should be part of `settings-init.js` already? I think I mentioned in an email but if that's not applying then we need to change `bootloader.js` rather than messing with every watch face

skeleton/add_my-great-app_to_apps.json
```diff
@@ -0,0 +1,12 @@
+// Create an entry in apps.json as follows:
```
gfwilliams: I wonder whether we could make this a README.md that describes what `skeleton` is and where to put the files?

apps/clock-mixed.js
MaBecker: Yes you did, chaning it in bootload.js would probably the best choice.

Look's like settings-init.js is not processed before a clock is started.

skeleton/add_my-great-app_to_apps.json
MaBecker: Good point will add one

apps/clock-mixed.js
MaBecker: > @MaBecker could you fix indentation in skeleton app to match with other apps (2 spaces)?
> And for the parts like `<your main function>();;`, I'd maybe put it in comment, so the final JS file is actually valid, like:
> 
> ```js
> Bangle.on('lcdPower', (on) => {
>   if (on) {
>     drawWidgets();
>     // <your main function>();
>   }
> });
> ```
> 
> Anyway, that's just my personal preference.

Good point, will do some shaping later

PR #48, opened 2019-11-16
Add Large Digit Clock to apps

knolleary:
Adds a large digit clock app.

image


apps/clock-blob.js
```diff
@@ -0,0 +1,108 @@
...
+        digits = [-1,-1,-1,-1,-1,-1];
+        intervalRef = setInterval(redraw,1000);
+        redraw();
+    }
+    startTimers();
+    Bangle.on('lcdPower',function(on) {
+        if (on) {
+            g.clear();
+            startTimers();
+            try { if (drawWidgets) { drawWidgets();} } catch(err) {}
```
knolleary: Not sure if I should be calling `drawWidgets` like this. I've seen some other clocks do it and some don't. Do the widgets refresh themselves as needed?

MaBecker:
Nice watch

Maybe you like to remove the extra `buf` to spare memory. 

\+ As this display is very fast there will be no flicker.
\+ Get a nice animation at start for free :)
\+ A kind of morphing when numbers change
\- Have to add some move/translation to get it back centered 

Just tested a version without `buf`

clock-blob


Will definitely add this to my code, nice hint.

```
try { if (drawWidgets) { drawWidgets();} } catch(err) {}
```

apps/clock-blob.js
gfwilliams: That looks good - to be honest I'd skip the `try` as it really shouldn't throw (maybe drawWidgets should catch stuff?).

Basically the widgets should refresh themselves, but if you clear the whole screen then you need to force a redraw. Because sometimes (at least at the moment) BLE connection events cause the console to scroll, doing a redraw when the LCD first goes on is a good way to fix it ;)

gfwilliams:
Thanks - this looks awesome!

> Will definitely add this to my code, nice hint.

Please don't! If something is throwing exceptions we really ought to just fix that (and maybe drawWidgets) rather than hacking around it in every clock app.

I wouldn't worry about the buffer in case you're hitting memory issues - it's only 3kB the way you have it and you have ~40k total.

Only thing I would say is if you want to optimise, `NUMBERS` could be one big Uint8Array, which would be a bit more efficient and faster. Definitely merging this in as-is though :)

knolleary:
Thanks @gfwilliams

I added the try/catch around the drawWidgets call as I was sometimes getting an undefined error with it. Mute in more familiar with the various moving parts I'll go back and see if I can figure out what was happening.

From what you say, I only really need to call it after clearing the screen. I currently also call it when I refresh the clock, even though that doesn't clear things. Will tidy that up.

Will look at the NUMBERS array. What's the trick to checking memory usage?

gfwilliams:
Thanks!

For memory usage checking, you can figure out what's available with `process.memory()`, and `E.getSizeOf(...)` will return the number of variable slots used by something: http://www.espruino.com/Reference#l_E_getSizeOf

If you're interested in the specifcs on making stuff fast/efficient there's http://www.espruino.com/Performance - Espruino's a bit different so some of the stuff you might do with a normal JS engine may not work as well.

MaBecker:
> Please don't! If something is throwing exceptions we really ought to just fix that (and maybe drawWidgets) rather than hacking around it in every clock app.

So what about using this few lines to handle lcd turns on?

```
    Bangle.on('lcdPower', (on) => {
        if (on) {
         	<call your main function>     
    }});
```

gfwilliams:
Yes, that's fine - no worries there - it's just the error handling :)

PR #49, opened 2019-11-16
Add Select Clock option to settings app

knolleary:
Having created my clock app, it took a bit of digging to work out how to get it to appear as the default clock face. It was a case of either deleting all other clocks or gaming the `sortorder`  property - but that ends up being an arms race of who can commit the most negative number.

In doing this, I dug into the firmware and settings apps to see how they worked. In doing so, I thought I'd try adding an option to select the preferred clock to show on the main screen.

Having got things working, I wanted to share.

The settings menu now has a 'Select Clock' option. This takes you to a sub menu that lists all apps of type 'clock' with the current one marked with a `*`. The user can select their preferred clock app from the menu. It stores the src property of the selected clock as `settings.clock`.

The bootloader app checks this property and tries to load that app. If the property is not set, or the app not found, then it reverts to the current behaviour of using the first clock app it finds.

I also ran `bin/firmwaremaker.js` to update `firmware.js` - which appears to have updated pulled in updates to some of the other apps as well. Not sure if that's something you want contributors to update, or that you'll do as needed.


apps/bootloader.js
```diff
@@ -101,12 +101,28 @@ if (startapp) {
...
-  if (clockApps && clockApps.length > 0) eval(require("Storage").read(clockApps[0].src));
+  var settings;
+  try {
+    settings = require("Storage").readJSON('@setting');
+  } catch (e) {
+    settings = {}
+  }
+  var clockApp = settings.clock;
+  if (clockApp) {
+      clockApp = require("Storage").read(clockApp)
```
ra100: What happens when you have some app as default in settings, but you deleted it?

apps/bootloader.js
```diff
@@ -101,12 +101,28 @@ if (startapp) {
...
+  }
+  if (!clockApp) {
+    var clockApps = require("Storage").list().filter(a=>a[0]=='+').map(app=>{
```
ra100: I'd suggest using `===` as it's safer and a bit faster than `==`

apps/bootloader.js
```diff
@@ -101,12 +101,28 @@ if (startapp) {
...
+      try { return require("Storage").readJSON(app); }
+      catch (e) {}
+    }).filter(app=>app.type=="clock").sort((a, b) => a.sortorder - b.sortorder);
```
ra100: Same here with `===`,
and I see mixed `"` and `'` in this file, can we choose one of those, but that's minor, just for consistency :)

apps/bootloader.js
```diff
@@ -101,12 +101,28 @@ if (startapp) {
...
-    catch (e) {}
-  }).filter(app=>app.type=="clock").sort((a, b) => a.sortorder - b.sortorder);
-  if (clockApps && clockApps.length > 0) eval(require("Storage").read(clockApps[0].src));
+  var settings;
+  try {
+    settings = require("Storage").readJSON('@setting');
+  } catch (e) {
+    settings = {}
+  }
+  var clockApp = settings.clock;
```
ra100: Can we use 2 different vars, one for clock name from settings, and another for clock app object?
Something like `defaultClockApp` and `clockApp`?

apps/bootloader.js
```diff
@@ -101,12 +101,28 @@ if (startapp) {
...
+      try { return require("Storage").readJSON(app); }
+      catch (e) {}
+    }).filter(app=>app.type=="clock").sort((a, b) => a.sortorder - b.sortorder);
+    if (clockApps && clockApps.length > 0) {
+        clockApp = require("Storage").read(clockApps[0].src);
+    }
+  }
+  if (clockApp) eval(clockApp);
   else E.showMessage("No Clock Found");
+
```
ra100: I think, here we can delete also clockApp: `delete clockApp`

apps/settings.js
```diff
@@ -174,6 +176,37 @@ function makeConnectable() {
     showMainMenu();
   });
 }
+function showClockMenu() {
+  var clockApps = require("Storage").list().filter(a=>a[0]=='+').map(app=>{
```
ra100: `clockApps` seem to not be reassigned, so it can be `const`

apps/settings.js
```diff
@@ -174,6 +176,37 @@ function makeConnectable() {
...
+    catch (e) {}
+  }).filter(app=>app.type=="clock").sort((a, b) => a.sortorder - b.sortorder);
+  const clockMenu = {
+    '': {
+      'title': 'Select Clock',
+    },
+    '< Back': showMainMenu,
+  };
+  clockApps.forEach((app,index) => {
+    var label = app.name;
```
ra100: I suppose, we can use `let` instead of `var`

apps/bootloader.js
knolleary: Happy to update it - although this was copied directly from the existing code in `firmware.js` so didn't want to rewrite it if particular choices had already been made.

apps/bootloader.js
knolleary: Again - this code was from `firmware.js`. Happy to go for consistency, but that needs considering across all of the existing code.

apps/bootloader.js
knolleary: It could, but for what reason? `clockApp` is the clock app we're going to load. First we check to see if settings provides it, otherwise we fall back to the existing code to get the first clock app we find. Adding another var would only increase memory usage for little benefit.

apps/bootloader.js
knolleary: If it has been deleted then `clockApp` will be `undefined` after line 112, so the check on line 114 kicks in and we go back to loading the first clock app we can find.

apps/bootloader.js
gfwilliams: >  it's safer and a bit faster than ==

@ra100 have you benchmarked this on Espruino? I'd be tempted to say that it was almost identical, if not marginally slower. In this case I fail to see how it is any safer as we know we're dealing with a string.

apps/bootloader.js
gfwilliams: @ra100 I'm really not sure this is helpful. It just feels like you're suggesting changes for the sake of picking holes in it.

apps/bootloader.js
ra100: @gfwilliams point taken, I didn't test it on espruino, just browser :) and you're right, it's marginally slower ~1% or so. And in "safer", I mean, that we can use ===, as we know for sure, we're dealing with strings, so we don't need to rely on == type conversion.

apps/bootloader.js
ra100: @gfwilliams I'm sorry, I'll try to refrain from these not helpful comments. I guess, I can create PR if it "pains" me so much :)

apps/bootloader.js
ra100: Point taken, didn't think about memory usage.

gfwilliams:
Thanks! Looks great - to be honest I'd just been deleting all but the clock app I wanted, but it's nice to be able to change it :)

apps/bootloader.js
gfwilliams: :) Yeah, I agree it's definitely good practice to use `===` regardless of the minor speed difference (which may change in the future anyway). I think there's a bit of a balance of code quality vs friendliness to make in each project, and with something like Bangle.js everybody is contributing for fun so I think it's important not to be too picky.

But also given Nick's background I have quite a lot of confidence in his code :) Checking up on stuff like 'what happens if the clock app is removed' is super helpful though - so thanks for going through this.

PR #50, opened 2019-11-17
Change the apps folder structure

ra100:
Reorganizes apps, every app is in it's own folder named by the 7-char ID.


gfwilliams:
Thanks - looks great!

PR #53, opened 2019-11-25
Skeleton files and some app fixes

MaBecker:


gfwilliams:
I think there are still merge issues because the apps got moved into their own folders. Also I'd have thought that having drawWidgets in the LCD 'on' handler wasn't such a bad idea? If you're ok with it I'll merge this but just not the removal of drawWidgets in the clocks? Then it should be pretty easy to merge in

MaBecker:
Ok with me

gfwilliams:
Huh - actually it merged just fine :) I've just moved stuff around as I figured you could just copy the skeleton folder in order to create a new app

MaBecker:
Thanks

> I figured you could just copy the skeleton folder in order to create a new app

Nice

PR #57, opened 2019-12-03
some small fixes

MaBecker:
fix simple clock tags
add  drawWidgets()
add color to battery widget


gfwilliams:
Neat, thanks! Just so you know, new firmwares allow you to setColor with a String like `"#123456"` which I guess might be more readable?

PR #60, opened 2019-12-05
GPS needs more speed!

AkosLukacs:
Some tweaks to the GPS info app:
 - Speed is shown on the screen
 - "Animation" on the "Waiting for GPS" screen
 - With 10px padding feels a bit nicer :)


gfwilliams:
Nice - thanks!

PR #61, opened 2019-12-06
A small change to 3x2 Pixel Clock to make it work in the emulator

sebi5361:
I have encapsulated the drawWidgets function into a try block to allow the app to run in the Emulator.
https://www.espruino.com/ide/emulator.html?codeurl=https://banglejs.com/apps/apps/clck3x2/clock3x2.js&upload
You can add the _run in the Emulator_ icon next to the app if you wish.

### DIFF
--- a/apps/clck3x2/clock3x2.js
+++ b/apps/clck3x2/clock3x2.js
@@ -60,7 +60,6 @@
 
   function drawTime() {
     g.clear();
-    //drawWidgets();
 
     let d = Date();
     let h = d.getHours();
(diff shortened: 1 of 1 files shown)


gfwilliams:
Actually please can we not add hacks to work around the emulator? It'd be much better to modify the emulator than to have every single app needing a workaround in it.

sebi5361:
OK I'll remove my pull request.
I just copied this hack from the the Large Digit Clock: https://github.com/espruino/BangleApps/blob/master/apps/blobclk/clock-blob.js#L97

gfwilliams:
Thanks - realistically that probably shouldn't be in there either.  Just added https://github.com/espruino/BangleApps/issues/63

PR #66, opened 2019-12-14
Berlin clock

eska-muc:
Hi, I've implemented an app which is displaying time similar to the Berlin Clock (see https://en.wikipedia.org/wiki/Mengenlehreuhr ). Please consider to add this app to the main project.
Kind regards, Stefan


gfwilliams:
Nice - thank you!

PR #67, opened 2019-12-15
Added Espruino Rotating Cube as an App

digitaIfabric:
I found this code in the Espruino IDE and think it would look cool on the Bangle as an app.

Next app: Tenet.js


gfwilliams:
Thanks! Yes, that's probably a good idea :)

However to make it appear in the Bangle.js loader you need a JSON description file and an icon. There's an example you can just copy over and modify as needed - check out https://github.com/espruino/BangleApps#example-applications

MaBecker:
Hi @digitaIfabric,

If you like use this png, it is much smaller

cube

gfwilliams:
Thanks!

Issue #68, opened 2019-12-18
Allow non-integer timezones


ryzokuken:
Hello! 👋 

I just got my Bangle.js from NodeConf EU this morning and started playing around with it. Apparently, it's not possible to set non-integer time-zones on the thing (I'm currently in UTC+5:30 and it became quite evident that this hasn't been addressed yet).

I thought changing the step size from 1 to 0.5 here (https://github.com/espruino/BangleApps/blob/master/apps/setting/settings.js#L101) will work, but maybe there's more nuance to it than that.

gfwilliams:
Thanks! Yes, changing the step should fix it. I'm away at the moment but if you add a pr I'm happy to merge it

MaBecker:
@ryzokuken, check this line too

https://github.com/espruino/BangleApps/blob/d306209da77cfc7c706f26054d59baaa422aff77/apps/setting/settings.js#L103

ryzokuken:
@MaBecker Yeah, I guess I'll just remove it.

PR #69, opened 2019-12-18
feature: all thirty-minute timezones

ryzokuken:
/cc @gfwilliams.

I tested it locally, it works. It'll be nice to work upstream on an alternate representation for this menu though (one that can go 5:30 instead of 5.5) but this should be as functional as anything.

Fixes: #68

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -41,7 +41,7 @@
   { "id": "setting",
     "name": "Settings",
     "icon": "settings.png",
-    "version":"0.01",
+    "version":"0.02",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! The `0|v` bit was there to cope with what happens if timezone wasn't defined in settings though. Might be worth just using `+v` instead?

Having time format would be nice as you say

ryzokuken:
@gfwilliams PTAL.

gfwilliams:
Thanks!

PR #70, opened 2019-12-18
app: add a simple 12-hr clock

ryzokuken:
/cc @gfwilliams wdyt?


ryzokuken:
I wonder if a better option here would be to add a setting called "12-hour clock" and expect clock devs to handle that (then I'd remove the new app and just add a condition in the main `sclock`)/

gfwilliams:
Looks good - thanks!

I'm in two minds about the 24 hour thing - in stuff like this it makes a lot of sense, but then I'm also a big fan of keeping stuff as simple as possible.

I'll merge this now and I've filed an issue on https://github.com/espruino/BangleApps/issues/76

PR #71, opened 2019-12-19
app: add nyan cat

ryzokuken:
@gfwilliams how does this look? I constantly fought with the program running out of memory (I am using 2 less frames and a much smaller color palette than I originally planned), but still cannot run it on my watch. It's stuck on "Loading..." it works fine on the emulator though. Any idea what's wrong?


ra100:
Maybe the memory issue is that you already have some pieces of other apps/widgets in the memory on you watch and this is not simulated in emulator. But take this only as a wild guess.

ryzokuken:
@ra100 well, your guess is as good as mine, if not better.

ryzokuken:
P.S. I'm all up for ideas regarding how to shrink the size of things here. I've already compressed the images, chosen a pretty small encoding as well as  skipped a few frames.

ra100:
@ryzokuken what if you sliced one image into multiple pieces and "rendered" one "frame" out of 4 or 8 pieces? Would it cause too much flicker?

gfwilliams:
Hi! Sorry for the delay replying - just been on paternity leave.

I reckon if you moved `atob(...)` from the drawImage call to around each string - so:

```
var images = [
  atob("qoAi/okjHf4..."),
  atob("..."),
 ...
];
....
buffer: require("heatshrink").decompress(images[i])
```

Then that might help you enough, since the binary version of the base64 string will be a bit more compact.

Are you using a beta watch, or a Nodeconf one? The beta firmware should have been a bit more efficient with memory, so doing a firmware update might really  help you out

gfwilliams:
Do you think you'd get a chance to tweak this, or shall I close it?

Switching the `atob`s got it working for me, but as you say it's only greyscale.

Looking at what's displayed though you're actually outputting a pre-pixellated image. You could just use images that were half the size (using 25% of the memory) and then use `scale:2` when rendering?

PR #72, opened 2019-12-24
feat: "my files" tab to list and download files

feugy:
### What's in there?

- a new tab to list files on your watch
- it displays a table of files, indicating files type when possible
- when clicking on row, trigger download as a file on your host
- refresh button that works similarly to "My Apps" refresh button

image

### Notes to reviewers

As application may create files on the watch, it would be nice to have a simple way to download them on your computer. In my case, I would like to retrieve the GPX files of my activities.

This implementation supports textual and binary files, and in particular, files in append mode, which the final octal sequence in name makes communication a bit hairy.
Without encoding accessed file names (with `encodeURIComponent`), Puck experiences either JSON parsing error (when listing the files) or general failure (when accessing file content)

One thing really deserve improvement: accessing content will crash if it does not fit the bangle available RAM.
I would like to read chunks one by one and collect them on browser side, but the Storage implementation does not support such thing (as far as I know).
I reused a pretty old-yet-reliable [3rd party](https://eligrey.com/blog/saving-generated-files-on-the-client-side/) lib to turn the file content into a downloadable file, but this piece could easily be re-written if needed (especially since we don't need to support IE/Edge browsers)

Finally, I apologize in advance if I broke the style guide. I am a little bit lost without eslint.

### DIFF
--- a/comms.js
+++ b/comms.js
@@ -95,5 +95,35 @@ watchConnectionChange : cb => {
   return () => {
     clearInterval(interval);
   };
+},
+listFiles : () => {
+  return new Promise((resolve,reject) => {
+    Puck.write("\x03",(result) => {
+      if (result===null) return reject("");
+      //use encodeURIComponent to serialize octal sequence of append files
+      Puck.eval('require("Storage").list().map(encodeURIComponent)', (files,err) => {
+        if (files===null) return reject(err || "");
+        files = files.map(decodeURIComponent);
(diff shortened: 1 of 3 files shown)


gfwilliams:
This is great - thanks!

It's a feature I've been looking at adding into the Web IDE for a while, but it's great to have this on the App Loader.

PR #74, opened 2020-01-03
clock: Centerclock - Simple digital clock with date

brainfoolong:
Just a clock interface that match my preferred style.


gfwilliams:
Nice - thank you!

PR #75, opened 2020-01-05
Add Bold clock app, Pedometer and Digital clock widgets

aerialist:
Add Bold clock app
Add Pedometer widget
Add Digital clock widget


gfwilliams:
Looks awesome - thanks! I'll add some code to ensure that it's easy to continue step counting when changing apps soon as well.

PR #80, opened 2020-01-20
app: stethoscope

blechdom:
app that adds audio to the hrm app


gfwilliams:
This looks great - thanks!

PR #94, opened 2020-02-10
✨ gbridge: increase contrast

ra100:
It was quite difficult to read small text with contrast of #b0b0b0 text on #404040 background.


gfwilliams:
Thanks!

PR #95, opened 2020-02-10
add missing class

MaBecker:


gfwilliams:
Thanks! this would never have worked!

PR #103, opened 2020-02-15
Clock: Pipboy

detached:
This is a Fallout pipboy styled clock.

pipboyClock

As I don't own a bangle.js yet, I was only able to test the app in the emulator.

https://detached.github.io/BangleApps/


gfwilliams:
This looks awesome - thanks!

PR #112, opened 2020-03-04
New widget  widid

MaBecker:
Display the last two tuple of your Bangle.js address in the widget section

Very helpful if you have more than one device ;-)


MaBecker:
Wow, so many updates for such a simple widget.......

Please take a closely look before you merge.

gfwilliams:
This is a great idea!

If you take a look at the travis build status and it fails, you can click on `Details` and it'll tell you what the problem was.

Right now at https://travis-ci.org/espruino/BangleApps/builds/658541879?utm_source=github_status&utm_medium=notification it's saying `ERROR: App widid icon doesn't exist`. So if you change `"icon": "widid.png",` to `"icon": "widget.png",` that should be sorted.

But I'll just do that now after merging as it's a trivial fix. Also, minor thing but I'd put `g.reset()` before drawing anything just to return everything (eg font, bg color) to a known state - it might be that a previous widget changed font alignment for example.

MaBecker:
Thanks!

PR #116, opened 2020-03-06
clock2x3, setting clock type, simplifying app code

sebi5361:
Dear Gordon,
This pull request is about the 2x3 pixel digit clock:
- I set "clock" as its type, so it can be selected as a clock face,
- I changed its name from 3x2 to 2x3 to be consistent with the way you name fonts,
- I simplified the code by removing an unnecessary operation.


sebi5361:
I have also adjusted the french locales as asked me @MaBecker.

gfwilliams:
Neat - thanks!

PR #120, opened 2020-03-09
New App: Heart Rate Recorder

sebi5361:
Dear Gordon,
This new app is a heart rate recorder.
It is a brut copy of your GPS recorder App where everything related to GPS is now related to HRM.
The only difference in the code has to do with the date:
It allows to read the same hour for a recording on the watch and on the PC (contrarily to the GPS app where there is a time zone difference). The tweak I use is certainly not the best as I didn't really understand the reason for this bug on the GPS app, but it seems to work.


gfwilliams:
Looks great - thanks!

PR #130, opened 2020-03-18
Added app to handle fullscreen notifications

AverageMarcus:
Added an application that can be triggered by other Bangle.js apps to display fullscreen notifications and allows the wearer to look back at previous notifications (configurable).

Screenshots:

image

image

image


MaBecker:
think  about adding g.reset() to save.

AverageMarcus:
@MaBecker Not sure what you mean by "to save". Doesn't `g.reset()` put all the graphics settings back to the default as on load?

MaBecker:
just update my first comment

gfwilliams:
This looks like a really neat solution - thanks!

Please could you just tweak the changelog into the text-only format used for the other logs though? The loader looks for a `ChangeLog` file and loads it when the version is clicked on and I'd rather not have to check for a bunch of different styles :)

AverageMarcus:
Updated. Is this the right format @gfwilliams?

gfwilliams:
Yes, that's perfect - thanks!

Xaseron:
Hey, greetings from germany.
Can we change the font to `Dennis8` http://www.espruino.com/Fonts?
This would help to display notfication messages with special characters like äöüß

MaBecker:
> Hey, greetings from germany.
> Can we change the font to `Dennis8` http://www.espruino.com/Fonts?
> This would help to display notfication messages with special characters like äöüß

Well, the 6x8 font has all you need.

https://github.com/espruino/Espruino/issues/1770

Xaseron:
Awesome. 
I've haven't tested the gadgebridge releated PRs. 
I just can state that the current available packages in app loaders have problems with special characters. Maybe the cause is gadgetbridge on my phone. I don't know.

gfwilliams:
Sorry it's taken so long to sort this, it seemed tricky to integrate this nicely. 

I'm afraid I haven't merged this, but have added a simple `notify` (top of screen as before) and `notifyfs` app, which create something that allows you to `require("notify").show(...)`. Apps can also set a dependency so that a notification library gets loaded automatically if one doesn't exist.

There's no stored notification list but hopefully the new one can run at the same time as most other apps. The issue with this PR (as far as I could see) was once it's fired it doesn't load widgets, so you wouldn't actually get a second notification if one came in?

If you'd like to PR the new `notifyfs` with any changes that'd be great though!

PR #131, opened 2020-03-18
Gadgetbridge fullscreen notifications

AverageMarcus:
> ⚠️  This requires #130 to first be merged

Added support for fullscreen notifications to the Gadgetbridge app.

Settings page updated to support enabling fullscreen notifications or keeping current slide-down notifications.

Also added the ability to turn off notifications completely and to hide the widget icon.

image


gfwilliams:
This looks good - I guess potentially rather than checking for `notify.app.js` you could do `Bangle.loadWidgets` and then see if `global.ShowFullscreenNotification` exists?

That way, it'd be pretty easy for others to create different apps to display notifications in different ways.

In fact -  there are some other options here...

* Add `Bangle.showNotification`, and then make the gadgetbridge top bar implementation into another app that just implements `Bangle.showNotification` - so you can install an app that displays notifications whatever way you want.
* Turn your fullscreen notifications into a "fullscreen gadgetbridge" app. Realistically that might actually be more sensible - since I guess to make it a really good experience you'll want quite a tight coupling between gadgetbridge and the notifications - maybe displaying icons depending on the service type (whatsapp/hangouts/etc)

gfwilliams:
So this one basically includes `notify.js` as well?

Not sure if you've seen https://github.com/espruino/BangleApps/pull/155 but it seems to be abstracting out the 'draw notification' and looks like a good fit.

I'll leave this open and when I get a sec I'll try and do what we did for `locale`, which is to set it up so you can have apps that create a `notify` module in storage.

More info on https://github.com/espruino/BangleApps/issues/162

gfwilliams:
Closing this one for now - see my comment on  https://github.com/espruino/BangleApps/pull/130#issuecomment-638923204

> have added a simple `notify` (top of screen as before) and `notifyfs` app, which create something that allows you to `require("notify").show(...)`. Apps can also set a dependency so that a notification library gets loaded automatically if one doesn't exist.

It should provide a nice way for anything to create a notification, and notifications should now be able to be customised by installing a new app - to make them fullscreen or otherwise.

However there is no notification history at the moment - I guess that could be added though? Maybe the library could create a `Bangle.on('notification'...)` event that we could then hook onto with a notification history app?

PR #135, opened 2020-03-22
locales -> added Hungarian and fixed some typos

AkosLukacs:
Fixes:
Distance: based on the snippet `distance: n => (n < 1000) ? Math.round(n) + locale.distance[0] : Math.round(n/1000) + locale.distance[1]`, distance[0] should be 'meter', locale.distance[1] should be 'kilometer' (or the equivalent). It was swapped the other way in some cases
Temperatur(e): The `e` was missing in some cases


MaBecker:
thanks

gfwilliams:
Awesome - thanks!

PR #138, opened 2020-03-24
localize mclock date display

DerGuteWolf:


gfwilliams:
Thanks! Just so you know locales is built in (it is just overridden by what you upload later) so you don't even need to test for it's existence

PR #139, opened 2020-03-25
localize miclock

DerGuteWolf:


gfwilliams:
Thanks! Please could you also add a note to the ChangeLog and update the version in apps.json to people know it's a new version? http://www.espruino.com/Bangle.js+App+Loader#making-changes

DerGuteWolf:
Version is updated (to 0.03 despite the type in the commit message...) and Changelog is updated.

DerGuteWolf:
@MaBecker : I had first used `locales` in my change for the mclock, cf. https://github.com/espruino/BangleApps/commits/master/apps/mclock/clock-morphing.js exactly because of this problem, I am not sure what to do best?
```javascript
var locales = require("locale"); 
```
looks a bit unusual, here I agree with @gfwilliams .

MaBecker:
this code needs some change to allow 

    var locale = require("locale"); 

https://github.com/espruino/BangleApps/blob/a874d5838c1dee5a5dfb4e71844d961506dd29e9/apps/locale/locale.html#L63-L78

DerGuteWolf:
@MaBecker: `var locale = require("locale");`  works in apps! This is not the problem...

gfwilliams:
Thanks - I just fixed the `locale` app to just add `var` so it defines a local variable. I'd changed `mclock` from locales->locale and so had broken it, and not tested with a custom locale installed :)

This looks great - thanks!

MaBecker:
@DerGuteWolf:  Thanks for adding the  language specific part.

PR #140, opened 2020-03-25
update local.html

MaBecker:
make this work
var locale = require('locale');


MaBecker:
Just did a local test - looks ok.

Will update http://forum.espruino.com/conversations/345155/#comment15170910 after merge

gfwilliams:
Sorry - only just saw this for some reason, I think I was too busy packing boxes :) Actually changing to `loc` is a good idea, but the underlying issue here was that there was no `var` so it was being defined as a global variable - and I have just fixed that.

I also removed the indenting as it saves a few bytes when uploaded :)

MaBecker:
Thanks!

PR #143, opened 2020-03-26
Auto activeFilter using url

brainfart-be:
Hi !
I was tired of clicking on the tag after a refresh of the page.
I decided to put the tag in the url.

Let me know if this need any modification

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ httpGet("apps.json").then(apps=>{
   }
   appJSON.sort(appSorter);
   refreshLibrary();
+  refreshFilter();
 });
 
 // Status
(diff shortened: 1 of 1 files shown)


gfwilliams:
Awesome - that was annoying me too. Thanks!

brainfart-be:
Thanks!

Unfortunately, i just found some bugs.
I 'm quiet new to github, should i do another pull request since this one is closed or is there a way to handle that.

The issues are when you try to go back to all, it doesn't work, if the #hash is not a proper filter, it doesn't fallback to "all"

ps: I just got my bangle 5min ago, can't wait to play with it !

gfwilliams:
Thanks! Yeah, it's just easier to do another PR to fix it :)

PR #145, opened 2020-03-26
add function meridian to locale

MaBecker:


gfwilliams:
This sounds good, thanks!

Just so you know though, when changes get added to the `locale` library they also need to be added to https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/locale.js which is the one that gets used if there's no locale installed (or in the emulator)

gfwilliams:
I'll add that to it now

gfwilliams:
Also - while it doesn't show as an error at the moment (just a warning), if you bump the version in the changelog it also needs to go in apps.json :)

MaBecker:
Thanks!

PR #149, opened 2020-03-26
Fix autoFilter behavior

brainfart-be:
Added a check to validate the tag from the url, fallback to "all"
Fix the behavior when click on all

Sorry again for that double PR


gfwilliams:
No problem at all - thanks for checking it over and getting it fixed :)

PR #150, opened 2020-03-26
Add Mario Clock application

paulcockrell:
Mario Clock is a low-res (80x80) watch face with Mario running
through a level. He jumps to hit the bricks, which changes the
time displayed.

Pressing BTN1 will make Mario jump, and extend the display for
a futher 10 seconds.

marioclock-withdatefixed


gfwilliams:
This looks awesome, thanks! Just merging now.

Should be on https://espruino.github.io/BangleApps/ and banglejs.com/apps next time I do and update

PR #152, opened 2020-03-27
Update of lap log feature stopwatch app 

Red-The-Hunter:
- made lap logs count up from 1
- made it such that logs scroll to next column after 18th entry
- can now display 36 entries without going off-screen

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -355,7 +355,7 @@
   { "id": "swatch",
     "name": "Stopwatch",
     "icon": "stopwatch.png",
-    "version":"0.01",
+    "version":"0.02",
     "description": "Simple stopwatch with Lap Time recording",
     "tags": "health",
     "allow_emulator":true,
--- a/apps.json
+++ b/apps.json
@@ -356,7 +356,7 @@
     "name": "Stopwatch",
     "icon": "stopwatch.png",
     "version":"0.03",
-    "description": "Simple stopwatch with Lap Time recording",
+    "description": "Simple stopwatch with Lap Time logging to a JSON file",
     "tags": "health",
     "allow_emulator":true,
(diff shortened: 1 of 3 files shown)


Red-The-Hunter:
screenshot (1)
screenshot (2)
screenshot

Red-The-Hunter:
To Do (V0.03)
Remove lap function for BN3 when timer is stopped
Add Save Lap log to file for BN3 when timer is stopped

gfwilliams:
Nice - save laps sounds like a great idea!

Red-The-Hunter:
V0.03
- Remove lap function for BN3 when timer is stopped
- Add Save Lap log to file for BN3 when timer is stopped (Saves as StpWch-[DATE TIME]

Red-The-Hunter:
Screenshot is from Web IDE
screenshot (4)

Because Web IDE cannot use the storage functions, these are files generated by my own watch, thus different amount of entries/different timings
Capture
Capture2

Red-The-Hunter:
Realised something... I should swap BN1 and BN3 because BN3 is used to exit applications...

Red-The-Hunter:
screenshot (5)
screenshot (6)
Swapped functions of BN1 and BN3 so long-pressing BN3 to exit would not save the lap log.
It is fine if the stopwatch resets upon BN 3 press as stopwatch app is being exited.

Red-The-Hunter:
To-Do (V0.04)
- Add total time elapsed counter
- Contemplate if total time elapsed should be main counter displayed or lap time should be main display

gfwilliams:
This is great - thanks! I'll add a tutorial soon on making an `interface.html` so there can be a nice way to load the lap times

PR #153, opened 2020-03-27
App: Grocery

brainfart-be:
Create a grocery list and track what's in your cart.

I'm not certain about my Storage management, but it seems to work.


gfwilliams:
Looks great - thanks! Is it worth renaming the `grocery` storage file to `grocery.json` just so it's more obvious that it's just JSON?

PR #154, opened 2020-03-27
App: Dev clock

oli-sanders:
Add Digital Clock including timestamp (tst), beats(@), days in the current month (dm) and days since new moon (l)
image


MaBecker:
nice watch, i like the idea about having new moon information.

Do you like to add locales like implemented in miclock, so d and m names are can be displayed in users language when they have installed the languages app.

oli-sanders:
Thanks, great Idea - updated to use locales now.

gfwilliams:
Thanks!

Issue #147, opened 2020-03-26
Deleting all clock apps prevents the start of the app menu


profriemke:
If all existing clock apps are deleted (e.g. accidentially) the watch only shows a message that no clock app is present. The app menu is then not accessible.
Suggestion: If no clock is present, the app menu should be shown as standard.

gfwilliams:
Thanks - that sounds like a great idea.

brainfart-be:
Seeing the launcher directly felt weird, i decided to just add the BTN2 click to open the menu.

MaBecker:
I suggest to start the About app.

brainfart-be:
I personally removed it from my watch, i would end up with the same issue.
We can maybe make the smallest watch app possible with a message saying "to install a clock go to https://banglejs.com/apps#clock" or something like this.
(The #clock part is not yet live but the features was merged already)

gfwilliams:
I think your current solution is nice and tidy for the moment. If someone doesn't have a clock it's extremely unlikely they'll have the About app I guess :)

PR #158, opened 2020-03-29
Fixing issue147

brainfart-be:
Add an event on BTN2 to open launcher when no clock detected
See #147 .


brainfart-be:
Sorry , i did something wrong with my branch... I'll clean that and redo my PR

PR #165, opened 2020-03-30
Analog clock updates

nikmartin:
Refactored analog clock to get my feet wet with Bangle.js and Espruino:

- simplify timers to a single second timer
- fix draw widgets order of call 
- make seconds ticks more readable
- add elapsed seconds display
- add date

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -133,7 +133,7 @@
   { "id": "aclock",
     "name": "Analog Clock",
     "icon": "clock-analog.png",
-    "version":"0.02",
+    "version":"0.10",
     "description": "An Analog Clock",
     "tags": "clock",
     "type":"clock",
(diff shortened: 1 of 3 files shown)


MaBecker:
if you like add locales too 😉

nikmartin:
@MaBecker My search-fu is failing me, does Espruino have a localization system, or does each app have to come up with their own locale scheme?

MaBecker:
Well,  a full set of locales  would unnecessary blow up the firmware, so every one can pick a locale with app Languages to be available for apps on his Bangle.js

Sneak into[ Mixed Clock](https://github.com/espruino/BangleApps/blob/master/apps/miclock/clock-mixed.js) app to figure out how this work and look at this forum [conversion](http://forum.espruino.com/conversations/345155/#comment15172813)

nikmartin:
Done!

gfwilliams:
Thanks! Please could you add/update the ChangeLog so users know what has changed? https://github.com/espruino/BangleApps/blob/master/apps/aclock/ChangeLog

nikmartin:
ah, yes, i did but didn't bring it to this branch.

gfwilliams:
Thanks!

PR #166, opened 2020-03-30
Create Bar Clock

rigrig:
A simple clock with hh:mm time, date, and a bar to show seconds:
screenshot


MaBecker:
Nice idea, if you like add locale to your watch.

Sneak into[ Mixed Clock](https://github.com/espruino/BangleApps/blob/master/apps/miclock/clock-mixed.js) app to figure out how this work and look at this forum [conversion](http://forum.espruino.com/conversations/345155/#comment15172813)

rigrig:
Thanks, realised I forgot about that almost as soon as I made the pull request ;-)
And then I figured it really should respect the 12h setting as well, so fixed that too.

gfwilliams:
Looks great - thanks!

PR #167, opened 2020-03-30
marioclock: use short date format from locale, take timeout from sett…

DerGuteWolf:
…ings

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -855,7 +855,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
-    "version":"0.02",
+    "version":"0.03",
     "description": "Animated Mario clock, jumps to change the time!",
     "tags": "clock,mario,retro",
     "type": "clock",
--- a/apps/marioclock/ChangeLog
+++ b/apps/marioclock/ChangeLog
@@ -1,2 +1,3 @@
 0.01: Create mario app
 0.02: Fix day of the week and add padding
+0.03: use short date format from locale, take timeout from settings
--- a/apps/marioclock/marioclock-app.js
+++ b/apps/marioclock/marioclock-app.js
@@ -6,6 +6,9 @@
 **********************************/
 
 var locale = require("locale");
+const storage = require('Storage');
+const settings = (storage.readJSON('setting.json',1)||{});
+const timeout = settings.timeout||10;
 
 // Screen dimensions
 let W, H;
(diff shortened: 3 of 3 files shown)


gfwilliams:
Thanks - it's tricky about the timeout... Bangle.js keeps the screen on whenever there's a `.flip` - but if you're animating then that's going to keep the screen on.

Maybe I could make that one of the `Bangle.setOptions` so your clock can turn it off when it boots, in order to allow the clock to sleep?

DerGuteWolf:
Unfortunatly the few bytes more apparently means we gethit by "Upload Failed, got UNFINISHED STRING expected EOF" now.

gfwilliams:
I'll fix that issue with the App Uploader today, so I wouldn't be too concerned.

paulcockrell:
@DerGuteWolf How do I go about selecting the original date format?

DerGuteWolf:
The setting is taking from the locale app, ie what is selected during installation/update of this app.

paulcockrell:
Oh right, I might be missing something here, in `Settings->Locale` only see time-zone and clock-style, but nothing to do with date formatting.

DerGuteWolf:
@paulcockrell : Yes, the language to be used is not in settings but in the locale app.

paulcockrell:
@DerGuteWolf Looks to me like I can't get the original date format back in that case, and am stuck with this short date.

gfwilliams:
Not sure I understand your problem? You install the `Language` app in the App Loader to set the language, and if you don't like it you can just remove the app.

In terms of Date format, ` locale.date(new Date(), true);` gives you a short date, but ` locale.date(new Date(), false);` will give you a long one

DerGuteWolf:
@paulcockrell : locale defines only two variants: short and long (and long is really long in some locales - but only some - and thus won't fit). The original format was a mixture between both and there is no support in locale to localize such an intermediate variant. Perhaps you should suggest adding a third variant to  @gfwilliams .
Alternativy we could first try long and if it fits, use this and otherwise use the short version.

paulcockrell:
My problem is I want to see the Day name and day number as I originally wrote it. The year is not something I need reminding of.

DerGuteWolf:
@gfwilliams : Part of the problem here comes also from the fact, that esp the long versions are not consistent in the various locales, eg en_gb "Feb 28 2020" (which I would consider an intermediate variant as no weekday and month abbrev) vs de_de "Sonntag, 1. März 2020" (which is really the long variant, full weekday and full month).
en

DerGuteWolf:
@paulcockrell: I understand, but this is unfortunatly currently not localisable....

paulcockrell:
@DerGuteWolf It's fine, I understand making it valid for various locales important for other users. So I appreciate the change you put in. Maybe it's just me, but I do need reminding what the day of the week is, they all blend in to one  :-)

DerGuteWolf:
see #183

PR #173, opened 2020-03-31
Documents gadgetbrigde protocol

detached:
This shows currently possible commands and events


gfwilliams:
Thanks! Did you know there are already some docs of the protocol on http://www.espruino.com/Gadgetbridge ?

detached:
ups.. I didn't know that. Thanks for the link

PR #176, opened 2020-04-01
Barclock fix and cleanup

rigrig:
1. Fix date screen area not being cleared at midnight
2. Code cleanup


gfwilliams:
Neat - thanks!

PR #178, opened 2020-04-01
Bar clock: fix SyntaxError

rigrig:
Whoops, sorry about that.

Seems the app still works though
(Snuck in during code cleanup, I'm blaming autoformat)


gfwilliams:
Thanks :) If you enable GitHub pages for your GitHib account it's always worth trying to load the app yourself before PRing :)

PR #180, opened 2020-04-01
widget hand wash timer

MaBecker:
Inaktive:

IMG_9979

Active:

IMG_9980

Please  let me know if the widget function is clearly described  - Thanks.


gfwilliams:
Very cool, thanks - that's really nice and tidy.

In terms of description `Any swip will start`, I think we'd normally say `swipe`. But maybe `Swipe on the screen to...`

From a purely practical point of view though, you've just swiped your watch screen with a dirty hand. Maybe saying something explicit like `Swipe your wrist over the watch face`?

MaBecker:
Thanks, so let me update that.

Can Bangle.setOptions() disable event twist or should there be a flag isWashing to ignore further twist events?

gfwilliams:
> Can Bangle.setOptions() disable event twist 

You mean twist to turn on? It can.. but does it matter if Bangle.js turns on while hands are being washed?

MaBecker:
It is just about another event during washing and drying.

MaBecker:
>If it now responds to just twisting, doesn't just twisting your wrist to see the time cause the timer to come on?
 
Wow, now it's getting tricky, any suggestion how to avoid that?

gfwilliams:
Well, swipe works ok? You don't have to swipe with a finger, you can swipe your wrist over the watch face (that's what I meant originally)

MaBecker:
Ok, got you wrong, yes swipe works, even swiping with the wrist.

MaBecker:
I would say, good to go

gfwilliams:
Thanks :)

PR #181, opened 2020-04-01
Battery widget: color by level and percentage as text

rigrig:
This has two commits:
0.05: Change color depending on battery level
0.06: Show battery percentage as text

Because I think the coloring is probably nice, but not sure if everybody would like the percentage? If not I could remove that (or maybe add settings?).

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -326,7 +326,7 @@
   { "id": "widbat",
     "name": "Battery Level Widget",
     "icon": "widget.png",
-    "version":"0.04",
+    "version":"0.05",
     "description": "Show the current battery level and charging status in the top right of the clock",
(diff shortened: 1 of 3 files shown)


MaBecker:
Nice, have thought about creating a new battery widget with this features instead of changing this one?

Purple-Tentacle:
Percentage and color is great.
Settings to show/hide percentage would be wonderful, as you may have to save space when using many widgets.

gfwilliams:
Thanks! Just tried this and it's really neat, but I think some folks may still prefer the old one.

I've just merged this, but have done so as a new application `widbatpc` which'll appear as `Battery Level Widget (with percentage)`

PR #183, opened 2020-04-02
Take into account discussion from #167

DerGuteWolf:
Using now weekday abbrev, space, short dateformat with full-year removed.
This should work for most locales (I hardcoded also replacing a trailing slash esp for en_IL and fr_FR)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -892,7 +892,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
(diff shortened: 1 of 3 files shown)


DerGuteWolf:
@paulcht: Could you take a look?

paulcockrell:
Looks great! I think that is a really neat solution to our discussion. Top stuff 👍

gfwilliams:
Thanks!

PR #184, opened 2020-04-02
Astrocalc, Sun and Moon info calculator

paulcockrell:
# Astrocalc

View information about the Sun and Moon, such as position in the sky, rise & set times and much more based on your GPS data.

This is quite a beefy project, so apologies if it goes against any community guidelines that I may have missed.

If anyone wants to chip in there are a few outstanding enhancements that could be made:

* Language is English only
* Doesn't render moon phases for southern hemisphere, this is simple to correct, and doesn't require new images.
* It could store your last GPS fix, and use that when it boots as an option, rather than always waiting for a valid GPS lock.
* Better Moon images, the current ones are rather uninspiring. 
* And anything else we can come up with.

## Images
IMG_20200402_122356
IMG_20200402_122414
IMG_20200402_122451
IMG_20200402_122500


gfwilliams:
Looks great, although do you think you'd be able to do the images a bit differently? The current method will be a bit slow and while I'm sure it works fine, I'd prefer others didn't see it and copy it :)

You could either include the images in your program code (if it's inside a function, it won't use any RAM), or you could do like we do here: https://github.com/espruino/BangleApps/tree/master/apps/animals

So, you export with compression, but as "image string", and then you shove that in a file much like you have done (but with `.img` extension), but set `evaluate:true` in `apps.json`.

Then when you load you literally just have to load the file and put it into drawImage - no messing about at all.

paulcockrell:
Ah I can't believe I missed the Animal app 🤦‍♂ Update on way, will test shortly.

paulcockrell:
Thanks for the image loading tip. It appears to be working as before.

gfwilliams:
Thanks, that's awesome!

Purple-Tentacle:
Check my app moonphase. It has some images in there.

PR #185, opened 2020-04-02
switch_to_local_name

MaBecker:
just found one app using local.lang


gfwilliams:
Thanks! Sorry, I forgot to do a search :)

PR #187, opened 2020-04-02
Balltastic, a simple ball eats dots game

ArPhil:
I think the title is speaking for it self. I created a simple game with a ball buy using the acceleration sensors. So you can control the ball by moving the watch. Eating a ret dot brings you points. After some amount of eating, you get to the next level where the time pressure of eating the red dot increases. Have Fun! :)


gfwilliams:
That's awesome - thanks!

PR #188, opened 2020-04-02
App: Toucher

brainfart-be:


gfwilliams:
Awesome - thanks! Did you know there's a `Bangle.on('swipe'` event when you drag your finger left-right or right-left?

brainfart-be:
I didn't know :D
I will add that tonight :D 
Thanks

PR #189, opened 2020-04-02
Gadgetbridge notification optimization

Xaseron:
The 10ms timeout delay for the animation was kinda laggy on my watch.
i've changed the timeout delay to 15ms and it looks more fluently now.
Some notifications like Whatsapp group chat have very long titles because they consists of chat name + name of the person. These caused the title to go out of bound. Therefore i've limited the title length to 18 characters.


gfwilliams:
Neat - thanks!

PR #192, opened 2020-04-02
Enhancement: Add better touch and swipe support to Toucher

brainfart-be:


gfwilliams:
Neat - thanks!

Purple-Tentacle:
@brainfart-be 
How can you activate the launcher?

PHDPeter:
@Purple-Tentacle that comment actually helped me as I never got this app to work without that lol

PR #194, opened 2020-04-02
Settings: Add support for app/widget settings

### LINKED ISSUE #182: Settings: widgets
I was thinking that maybe it would be nice if we could add widget settings to the settings app, instead of having separate "apps" just to configure a widget?

Something like
* A widget can set `WIDGETS["mywidget"].settings` to a function to handle settings
* The settings app gets a new submenu "Widgets", containing (only) all widgets that have a `settings` property
* Ideally widgets also set `name`, which is used as menu entry
    - If not set, fall back to the `WIDGETS` key? So it would just show as "mywidget"
    - Or should `.settings` contain both a name and a callback?
    - Maybe we could get the name from `apps.json`? Seems like not all widgets use their  `id` as `WIDGETS` key though

So you'd have something like this:
```js
  function settings(back) {
    const mySettingsMenu = {
        '': { 'title': 'My Widget' },
        'Sound': {
            value: false,
            onchange: s => {s ? Bangle.beep() : false} // TODO: save new value
        },
        '< Back': back // provided by caller (settings app)
    };
    E.showMenu(mySettingsMenu)
  }
  WIDGETS["mywidget"]={
    /* other properties go here */
    name:"My widget",
    settings:settings,
  };
```

rigrig:
Fixes #182

Apps and widgets can add a `<appid>.settings.js` file which can be opened from the "App/widget settings" submenu.
This file should define a single function to configure the app/widget.
The function is passed a `back` argument, which can be used to return to the settings menu by calling `back()`.

Example `<appid>.settings.js`:
```
(function(back) {
  const mySettingsMenu = {
    '': { 'title': 'My Widget' },
    '< Back': back, // go back to settings menu
    'Sound': {
      value: false,
      format: s => s ? 'on' : 'off',
      onchange: s => {if(s) Bangle.beep()} // TODO: save new value
    },
  };
  E.showMenu(mySettingsMenu)
})
```

Full example for battery widget: #195


rigrig:
* Could maybe do with some bikeshedding about where in the list of settings to put it?
* Going back and forth between menus seems to run out of memory pretty fast, maybe there is a better way to do that?

gfwilliams:
> Going back and forth between menus seems to run out of memory pretty fast, maybe there is a better way to do that?

Do you have an example? It shouldn't... something must be hanging on to a reference somewhere.

This looks great though! Positioning looks fine IMO. Only things I'd say are:

* For `app.settings.js` just do `(function(back) { ... })"` instead of `function settings() {}` since the latter actually defines a function called `settings` in that scope :)
* It'll need some docs so it gets used! :)

rigrig:
(Looks like the memory issue is unrelated to this, I'll open a new issue.)

Changed it to `eval('('+appSettings+')');`, as I figured it would be easier to do that here instead of needing it in every settings script.

gfwilliams:
This is great - thanks! Now to modify some apps. Gadgetbridge seems like a good option. Any thoughts about others?

rigrig:
Maybe the welcome app? It could still write to global settings, but would only show up if actually installed.

gfwilliams:
> Maybe the welcome app? 

Yes, could be a good idea... https://github.com/espruino/BangleApps/issues/221

PR #195, opened 2020-04-02
Add settings to `widbatpc`

rigrig:
Depends on #194, so I figured this could serve as example how that would work.

Adds settings to the battery level widget (with percentage) to
* turn on/off the percentage
* turn on/off the charging indicator icon
* switch colors between "by level"/green/monochrome

Commit 39621bfae0c659e354840b31ff092e51ab6dca26 adds the settings


gfwilliams:
Looks great!

PR #198, opened 2020-04-03
MarioClock [Bug Fix] - 12/24hr setting to be respected

paulcockrell:
Use user setting of a 12 or 24 hour clock  when rendering the time in MarioClock


gfwilliams:
Thanks!

Issue #136, opened 2020-03-23
Add Internationalisation to more apps


gfwilliams:
We now have a great 'locale' module for handling language-specific units in apps - but virtually nothing uses it right now. We should have a push to add it to as many apps as possible - and also to document how to use it.

MaBecker:
Just started writing a conversation with title 

`What does the Languages app and how can it be used in apps`

gfwilliams:
... which is at http://forum.espruino.com/conversations/345155

MaBecker:
Yes

MaBecker:
Looks like there is no meridian  function  that returns a value from  `ampm`

I guess that will do:

```
 meridian: d => (d.getHours() <= 12) ? locale.ampm[0]:locale.ampm[1],
```

just posted pr https://github.com/espruino/BangleApps/pull/145

stephenrjohnson:
I'm just wondering we have access to the GPS so would it be worth on first start when we get the time, getting the country and from that guessing the locale / timezone. We can prompt the user for input on changing this.

gfwilliams:
I think the Web Browser itself would have a much better idea of locale than we could get from GPS. But that's a good thought - we could try and figure out what the locale was and pre-select it from the list

stephenrjohnson:
There are two things, one is the locale which I suppose we could guess from the web browser settings (good idea). The other is the timezone, that is more location based. The locale and timezone don't really relate at all, just because your in the UK doesn't mean you want the language to be English. I'm more on about the timezone here (sorry for not being clear), as when the user switches on the device for the first time we set the clock using the GPS, but we could also get a long / lat from that guess the country and from that guess the timezone for this.
We would need to add the timezone.db and have a system that gets s country from a long / lat, it's just an idea not sure if you want me to have a crack at it. This would be good when travelling as to set your watch to the correct timezone you just select the menu item (update timezone etc).

gfwilliams:
Well, the timezone can also be got from the browser too - and is already: https://github.com/espruino/BangleApps/blob/master/js/comms.js#L87

And when using Gadgetbridge it should be updated automatically too.

So really this would just be for the case where someone went to a new country, without a phone/pc, and didn't want to set the timezone manually. I'm not sure it warrants adding a potentially quite large database of timezones to every watch (it'd have to know the actual borders, for instance https://www.timeanddate.com/time/map/)

[...]

PR #202, opened 2020-04-03
Internationalization of alarm app

DerGuteWolf:
As I wrote in the private messages, this is the first step: localizing a single app.

### DIFF
--- a/apps/alarm/alarm.html
+++ b/apps/alarm/alarm.html
@@ -31,30 +31,34 @@
           xhr.send();
       });
 
-      let appjsTranslated;
-      let translate = () => {
+      let translated; = {};
+	  let translate = () => {
+	    Promise.all(translateFile("app.js"), translateFile("alarm.js")).then(() => {
+          document.getElementById("upload").style.visibility = 'visible';
+        }).catch( error => {
+          console.log(error);
+          document.getElementById("upload").style.visibility = 'hidden';
+        });
+	  };
+      let translateFile = (fileName) => {
         var languageSelector = document.getElementById("languages");
         var lang = languageSelector.options[languageSelector.selectedIndex].value;
         console.log(lang);      
-        request("app.js").then( appjs => 
+        request(fileName).then( fileContent => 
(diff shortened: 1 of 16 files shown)


apps/alarm/alarm.html
```diff
@@ -0,0 +1,103 @@
...
+        var lang = languageSelector.options[languageSelector.selectedIndex].value;
+        console.log(lang);      
+        request(fileName).then( fileContent => 
+          request("app_" + lang + ".json")
+          .catch( () => request("app_" + lang.substring(0,2) + ".json"))
+          .catch( () => request("app.json"))
+          .then( langjson => {
+            var trans = JSON.parse(langjson);
+            fileContent = fileContent.replace(/"([^"]*?)"\/\*LANG\*\//g, function(m, p1) { return '"' + (trans[p1]||p1) + '"' + (trans[p1]?'':'/*LANG*/'); });
+            fileContent = fileContent.replace(/'([^']*?)'\/\*LANG\*\//g, function(m, p1) { return "'" + (trans[p1]||p1) + "'" + (trans[p1]?'':'/*LANG*/'); });
```
srl295, line 54: this is clever, but I think it would be better to just use a function in the callsite, something like https://www.i18next.com/translation-function/essentials#passing-a-default-value (and many others) use:

`locale.t('new', 'New Alarm')`

Using the English (here) as the key is not good, if the English text “New Alarm” changes, then you can't reuse the translations properly.

Actually, there's already a locale.translate function: https://www.espruino.com/Bangle.js+Locale#translation
but it uses the English text, and doesn't support Application resources.

Needs some work!

srl295 (changes requested):

gfwilliams:
Closing this as we now have a global solution: https://github.com/espruino/BangleApps/issues/136

I've just merged in the translations and made the relevant changes to the alarm app: https://github.com/espruino/BangleApps/commit/26a8755da01c4817733970fd770476cc87f9b403

It's still not possible to enable via the UI but all the basics are there now

Issue #206, opened 2020-04-03
Settings: running out of memory


rigrig:
Repeatedly opening a submenu and going back eats up memory, until there isn't enough left.

#### Steps to reproduce:
1. Open Settings
2. Select "Locale"
3. Select "< Back"
4. Repeat from 2.

##### Expected behavior:
Infinite back-and-forth through the submenus

##### Actual behavior:
After a few times "weird" things start happening, and `New interpreter error: LOW_MEMORY,MEMORY` shows up in the console.

My guess is this happens because every time you open "Locale" we create another locale menu, and every time you go "< Back" a new main menu is created, but the old menus are also still around?

Log of me typing in what I am about to select, and checking memory after selecting it:
```js
>//Settings
=undefined
>
 ____                 _
|  __|___ ___ ___ _ _|_|___ ___
|  __|_ -| . |  _| | | |   | . |
|____|___|  _|_| |___|_|_|_|___|
         |_| espruino.com
 2v05 (c) 2019 G.Williams
{"t":"status","bat":72}
>process.memory()
={ free: 1034, usage: 1066, total: 2100, history: 0,
  gc: 0, gctime: 5.40161132812, blocksize: 16, "stackEndAddress": 536923400, flash_start: 0,
  "flash_binary_end": 483740, "flash_code_start": 1073741824, flash_length: 524288 }
>//Locale
=undefined
>process.memory()
={ free: 901, usage: 1199, total: 2100, history: 6,
  gc: 0, gctime: 5.79833984375, blocksize: 16, "stackEndAddress": 536923400, flash_start: 0,
  "flash_binary_end": 483740, "flash_code_start": 1073741824, flash_length: 524288 }
>//Back
=undefined
>process.memory()
={ free: 599, usage: 1501, total: 2100, history: 8,
  gc: 0, gctime: 6.56127929687, blocksize: 16, "stackEndAddress": 536923400, flash_start: 0,
  "flash_binary_end": 483740, "flash_code_start": 1073741824, flash_length: 524288 }
>//Locale
=undefined
>process.memory()
={ free: 483, usage: 1617, total: 2100, history: 8,
  gc: 0, gctime: 6.89697265625, blocksize: 16, "stackEndAddress": 536923400, flash_start: 0,
  "flash_binary_end": 483740, "flash_code_start": 1073741824, flash_length: 524288 }
>//Back
=undefined
>process.memory()
={ free: 187, usage: 1913, total: 2100, history: 8,
[...]

PR #208, opened 2020-04-03
Fix for #206  Settings: running out of memory

rigrig:
Added benefit: "< Back" now restores your position in the main menu.

Duplicates a bunch of code from [E_showMenu.js](https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showMenu.js), maybe `restoreMenu` could be added as `l.restore` in there?

### DIFF
--- a/apps/setting/settings.js
+++ b/apps/setting/settings.js
@@ -31,7 +31,11 @@ if (!settings) resetSettings();
 
 const boolFormat = v => v ? "On" : "Off";
 
+let main
 function showMainMenu() {
+  if (main) {
+    return restoreMenu(main)
+  }
   var beepV = [ false,true,"vib" ];
   var beepN = [ "Off","Piezo","Vibrate" ];
   const mainmenu = {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - but I feel like I really need to fix this inside `E.showMenu`/Espruino if at all possible...

The whole idea is `E.showMenu` should be nice and easy to use - folks shouldn't have to worry about this

rigrig:
Yeah, I expected as much, but didn't really feel up to messing with the Espruino code.

Maybe you could add this to `E_showMenu.js`:
```js
restore : function() {
  g.clear(1);g.flip();
  Bangle.drawWidgets();
  l.draw();
  if (Bangle.btnWatches) {
    Bangle.btnWatches.forEach(clearWatch);
  }
  Bangle.btnWatches = [
    setWatch(function() { l.move(-1); }, BTN1, {repeat:1}),
    setWatch(function() { l.move(1); }, BTN3, {repeat:1}),
    setWatch(function() { l.select(); }, BTN2, {repeat:1}),
  ];
}
```

And then settings could do something like
```js
let menus = {}
function showMainMenu() {
  if (menus.main) {menus.main.restore(); return}
  const mainmenu = {
    '': { 'title': 'Settings' },
    'Locale': showLocaleMenu,
  };
  menus.main = E.showMenu(mainmenu);
}

function showLocaleMenu() {
  if (menus.locale) {menus.locale.restore(); return}
  const localemenu = {
    '': { 'title': 'Locale' },
    '< Back': showMainMenu,
  };
  menus.locale = E.showMenu(localemenu);
}

showMainMenu();
```

rigrig:
Hmm, I guess keeping more references around to all menus isn't really a great solution actually...

PR #213, opened 2020-04-03
Improvement: Close launcher when lcd turn off

brainfart-be:
The lcd stay on because of the load function.
I don't have workaround yet


gfwilliams:
Thanks - looks like a good plan.

I think right now, turning the LCD on at reload might be built in to the firmware :(

PR #215, opened 2020-04-03
Gadgetbridge: move config to settings menu

rigrig:
Not happy about having to leave `{"name":"gbridge.app.js"},` under `apps.json` `storage`, but otherwise its left behind when updating...


rigrig:
Travis isn't happy about it either :-(
```
> node bin/sanitycheck.js

ERROR: App gbridge file gbridge.app.js has no contents
```

gfwilliams:
I think it's still referenced at `{"name":"gbridge.app.js"},` in `apps.json`? If you take that out it'll be ok?

gfwilliams:
Ok, just spotted the top comment. I still think it should be removed.

Your https://github.com/espruino/BangleApps/issues/216 should fix it though?

gfwilliams:
Thanks!

PR #218, opened 2020-04-04
Workout HRM

msdeibel:
Adds the Workout HRM (wohrm) app to BangleApps.


gfwilliams:
Looks great - thank you!

gfwilliams:
Just a note on this - I'll swap it not, but in future please can you put the ChangeLog in the opposite order?

We have a script that checks the version numbers are correct, and this breaks if the ChangeLog is different :)

PR #220, opened 2020-04-04
Mario Clock [Enhancement] - Better performance / graphics

paulcockrell:
# Enhancement
After seeing @gfwilliams suggestions on how to do graphics in a more performant manner I thought I would apply this to the Mario Clock to speed it up.
* Store images inside functions for better performance and reduce ram usage.
* Re-draw all the graphics, make multi-tone for better visuals. Make it more true to the original Gameboy visuals
* Multi-layered background for 'depth' or 'parallax' effects
* Mario jumping and hitting blocks (by way of time change or BTN1 press) will eject a coin like in the game!

# Images:
maioclock-0 06

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -893,7 +893,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
-    "version":"0.05",
+    "version":"0.06",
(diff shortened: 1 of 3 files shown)


gfwilliams:
That looks awesome!

PR #222, opened 2020-04-04
Add minion themed clock

singintime:
Adding a simple, minion themed clock.

screenshot


gfwilliams:
Nice! Looking at the date, if you wanted to make it a bit more readable you could maybe change the font to `setFont("6x8",2)` or something like that?

PR #226, opened 2020-04-04
Refacor & Optimized string sizes and positions

Purple-Tentacle:
Made variable sizes and positions depending on digits


Purple-Tentacle:
Forgot version and ChangeLog

gfwilliams:
Thanks! I think in future, if you've created a pull request, you can actually just keep updating it by changing whatever branch you used (even `master` if you never created a branch) - so you don't have to close and reopen if you don't want to :)

PR #203, opened 2020-04-03
Added PWA and Browser Theme Support

OmegaRogue:
Add more favicon resolutions
Add Progressive Web App Support
Add Theme and Icon support for Safari
Add Theme and Icon Support for Windows Metro
Add Theme and Icon Support for Android Chrome

Signed-off-by: OmegaRogue <thatomegarogue@gmail.com>

PR #228, opened 2020-04-04
Add PWA Support

OmegaRogue:
now working
#203


OmegaRogue:
With the current configuration it only works for the production site at banglejs.com/apps

gfwilliams:
Nice - thanks! Just a quick question before merge - is the service worker actually required for this?

Right now I'm not sure it really does anything? We can't realistically get the loader running offline (without downloading every app), and I've found that the service worker on the IDE (while great in allowing it to work offline) was quite a big source of pain for a while, as once it's loaded by a bunch of users it can be really difficult to change it on their browsers.

OmegaRogue:
> Nice - thanks! Just a quick question before merge - is the service worker actually required for this?
> 
> Right now I'm not sure it really does anything? We can't realistically get the loader running offline (without downloading every app), and I've found that the service worker on the IDE (while great in allowing it to work offline) was quite a big source of pain for a while, as once it's loaded by a bunch of users it can be really difficult to change it on their browsers.

The service worker i included doesn't have offline functionality, i found that at least a simple service worker is required for a PWA, the service worker just logs the events to console, it doesn't cache the site.

gfwilliams:
Thanks!

PR #229, opened 2020-04-04
Aclock face

nikmartin:
clock face was interrupting the widget area, so i shifted it down and maximized the space it can occupy while still leaving room in the widget area


nikmartin:
woops, this was one the wrong branch to do a PR from! not sure why it's including all my history. i'll reclone the repo and start again

PR #230, opened 2020-04-04
Move loading of welcome to boot.js

### LINKED ISSUE #221: Welcome: Move welcome loader to `welcome.boot.js`
Discussed here: https://github.com/espruino/BangleApps/pull/194

rigrig:
Fixes #221 

Having to add
```js
// clear out all other apps/widgets
g.clear(1)
clearTimeout()
clearInterval()
clearWatch()
Bangle.removeAllListeners()
```
Makes me a bit sad, but I can live with it.
(Needed because the bootloader no longer chooses between running the welcome or the clock)

Also added settings for `ncstart`, because the "run welcome" setting was removed from the regular settings. (installing two welcome apps still only runs one of them though)


gfwilliams:
Thanks! Did you have any success with `setTimeout(()=>load("welcome.js"));` instead?

The whole `clear*` is a bit of a hack as you say, and even so that won't fix everything (I think the Bluetooth widget (if installed) would redraw on connect/disconnect) and all widgets would still be in memory.

rigrig:
I tried `setTimeout(()=>load("welcome.app.js"));`, but that gets you into a bootloop, because `boot0.js` runs all `*.boot.js` again, including `welcome.boot.js`.

Unless there is some way to see which file is being loaded that I missed?

gfwilliams:
Argh, that's frustrating. Right now there isn't, and even if I added it to the firmware it wouldn't help with watches running original firmwares.

I guess one option is to change the way welcome works, so it only appears once (eg it flips the setting at the beginning, not the end?):

```
var s = require('Storage').readJSON('setting.json', 1) || {};
if (!s.welcomed) {
  s.welcomed = true;
  require('Storage').write('setting.json', s);
  setTimeout(()=>load("welcome.app.js"));
}
```

It might be a more sensible way of doing it anyway, then you could just long-press out of it.

rigrig:
Changed it like your suggestion

gfwilliams:
Thanks! This really cleans things up

PR #232, opened 2020-04-04
widbatpc: Draw percentage as inverted on monochrome battery

rigrig:


gfwilliams:
Wow! That's an impressive solution!

That's going to end up being pretty slow, but given it'll only be for a few pixels it's not a big deal.

(just to add, another option is [`g.setClipRect`](http://www.espruino.com/Reference#l_Graphics_setClipRect), but your way is more fun!)

rigrig:
Thanks, I hadn't noticed `g.setClipRect` or I probably would've used that.

Issue #210, opened 2020-04-03
Bangle.js Launcher improvements


ps-igel:
Wearing the Bangle for a week now, these issues pop up frequently:

1. BTN2 often accidentally presses and it ends up showing the launcher
2. while in launcher widget (i.e. GadgetBridge) don't work because they are probably not loaded
3. returning to the clock using BTN3 takes a bit too long for my taste

suggestions:
- introduce setting to choose BTN1 for showing launcher
- make launcher return to default clock when LCD times out
- reduce the amount of time BTN3 needs to return to clock (maybe only a tiny bit)

gfwilliams:
Hi,

It sounds like the main issue here is BTN2 accidentally opening the launcher? One option would be to disable waking the watch up on BTN2 press, so you had to press some other button to have the watch displaying first, (*then* press BTN2?)

That's actually something that can be done in code right now: `Bangle.setOptions({wakeOnBTN2:false})`

So realistically it's just the ability to change that from setting that would be needed?

Right now the clocks themselves are responsible for using BTN2 for the launcher, so you could make your own clock app that used BTN1.

> make launcher return to default clock when LCD times out

That sounds like a great idea.

> reduce the amount of time BTN3 needs to return to clock (maybe only a tiny bit)

Sadly it's that long precisely because someone else complained it wasn't long enough! Can't please everyone :)

I guess it could be configurable though...

ps-igel:
> One option would be to disable waking the watch up on BTN2 press

I already have, still happens though. Probably when turning the wrist in combination with tight jacket arms.

> Right now the clocks themselves are responsible for using BTN2 for the launcher, so you could make your own clock app that used BTN1.

Right, didn't think on that. That'll be enough for the moment I guess.

> I guess it could be configurable though...

That'll be fine. No hurry though.

Thanks!

brainfart-be:
@ps-igel
If you want to give it a try, i developed another laucher - Touch launcher 
( https://espruino.github.io/BangleApps/#tool ) 
If you like it, i can easily add the feature to close the launcher when the display turn off.

ps-igel:
@brainfart-be 
I like the launcher. Looking forward for the new feature.

brainfart-be:
Thank's, it has been merged today. (v0.3)
Due to a limitation (probably on the firmware), when the LCD turn off and the launcher quit, the lcd turn on again on the clock, but at least it quit the launcher

ps-igel:
@brainfart-be 
[...]

PR #233, opened 2020-04-05
Rewrite and add asked feature  on issue #210

brainfart-be:
Rewrite and add asked feature  on issue #210 
Not sure about the perf, i didn't tried lcd mode 80x80 though;


gfwilliams:
Nice - thanks!

PR #234, opened 2020-04-05
Adds LCD brightness setting

### LINKED ISSUE #186: Settings missing brightness
In the settings app there is no brightness selector. For now I made an app for managing brightness:

I created a widget that doesn't draw anything but set the brightness taking the value from a value save in the storage by my new brightness app.
Then I had to edit the launch.app.js because widgets are not loaded there.

But I think it would be easier to have the brightness selector inside the settings app :)

detached:
Fixes #186


gfwilliams:
Thanks - this'll need something in `settings.boot.js` as well I imagine? I'll make a note in #186

PR #235, opened 2020-04-05
Add support for type:"storage" files

### LINKED ISSUE #216: Removing an app should look at `app.info` `files`
When removing/updating an outdated app, the list of files on the device might no longer match the list in `apps.json`
See for example #215, where `gbridge.app.js` was changed to `gbridge.settings.js`.

I think removing an app should get `app.info` from the device to determine which files to remove.

rigrig:
Fix #216 
Add support for type:"storage" files

On install, these files are only uploaded if they actually have content
On update, these files are *not* overwritten if they already exist
On remove, these files *are* deleted

This is meant for e.g. `<appid>.settings.json` files, so updating does
not reset user configuration.

Also: install/update/remove now check `<appid>.info` on the device
* install: if app.info exists perform update instead
* update: only delete files that won't be overwritten with a new version
* remove: delete files listed in `<appid>.info`
(If we fail to load `<appid>.info` these all fall back to `apps.jaon`)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -346,7 +346,7 @@
     "storage": [
       {"name":"widbatpc.wid.js","url":"widget.js"},
       {"name":"widbatpc.settings.js","url":"settings.js"},
-      {"name":"widbatpc.settings.json","content": "{}"}
+      {"name":"widbatpc.settings.json","type": "storage"}
     ]
   },
   { "id": "widbt",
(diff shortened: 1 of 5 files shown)


rigrig:
Maybe it would be better to just hardcode the no-overwrite behavior for `<appid>.settings.json` instead of this `"type":"storage"` stuff?

Purple-Tentacle:
What about files storing data? Like steps etc.?

gfwilliams:
Just saw these comments - but I added a comment on https://github.com/espruino/BangleApps/issues/216 about this.

rigrig:
I figured deleting settings is a safe cleanup, as in the worst case people just need to reconfigure the app if they accidentally remove it, while actual data would be more bothersome to lose.

With this implementation apps like the pedometer widget can add `{"name":"wpedom.json","type":"storage"}` to have it cleaned up, but e.g. the gps track recorder creates `".gpsrc"+n` files, so it wouldn't work there.
(I guess that's a +1 for `type:"storage"` over hardcoding `<appid>.settings.json`)

Maybe there should be something similar to `interface` for uninstalling? It could check for data files and ask the user what to do about them,

PR #236, opened 2020-04-06
Mario Clock [Enhancement] - Multi characters and Night Mode

paulcockrell:
# Enhancements
* You can now `SWIPE-RIGHT` to toggle the character between `Mario` and `Toad`! (On a loop)
* You can now `SWIPE-LEFT` to toggle `DAY/NIGHT MODE` (The clock was burning my eyes out at night) (On a loop)
* Trees no longer high as clouds (didn't quite look right)
* Minor performance enhancements

# Images
### Mario [Default]
screenshot (7)
### Toad
screenshot (15)
### Mario
screenshot (13)
### Toad
screenshot (10)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -893,7 +893,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
-    "version":"0.06",
+    "version":"0.07",
     "description": "Animated Mario clock, jumps to change the time!",
     "tags": "clock,mario,retro",
     "type": "clock",
--- a/apps.json
+++ b/apps.json
@@ -894,7 +894,7 @@
     "name": "Mario Clock",
     "icon": "marioclock.png",
     "version":"0.07",
-    "description": "Animated Mario clock, jumps to change the time! Swipe right to change the character, and swipe left to toggle night mode.",
+    "description": "Animated Mario clock, jumps to change the time! Swipe right to change the character, and swipe left to toggle night mode. Press BTN1 to jump.",
     "tags": "clock,mario,retro",
     "type": "clock",
     "allow_emulator":true,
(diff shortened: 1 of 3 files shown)


gfwilliams:
That's great - thanks!

Purple-Tentacle:
@paulcockrell 
I like the watch more and more.
But I am not using it yet, because I miss the widgets.
Any plans on showing the widgets?

gfwilliams:
Sadly widgets aren't possible with the 80x80 mode at the moment. There's an issue open for new modes that'd allow widgets here: https://github.com/espruino/Espruino/issues/1794

paulcockrell:
@Purple-Tentacle I am glad you like the watch face. 

Would you be happy with me adding a `Battery` level icon to the app? I have been meaning to add this any way. I don't know if this would be enough for you.

paulcockrell:
I have made this update, but I am not sure about it. I like having the info, but the screen is becoming too busy. I'm thinking I might make BTN3 cycle through to another view that shows such info. What do you think @Purple-Tentacle ?
screenshot (9)

Purple-Tentacle:
@paulcockrell 
I would like all the widgets to be displayed, but this is no option with the screen mode.
Maybe later with the new ones.

I am thinking of integrating the battery information into the background graphic somehow?
For example a cloud which get less filled when battery percentage drops?
https://www.pcgames.de/screenshots/1280x/2018/07/Super_Mario_Land_Special_Retro_2018.07.19-2--1-pc-games.jpg

The temperature in your screenshot is the current one, obtained through GadgetBridge weather plugin?

paulcockrell:
Yeah I understand. But I haven't designed this watch face to show widgets, I picked the 80x80 on purpose.

I think if you want to have all the widgets, low res just doesn't make sense, there aren't enough pixels for everything.

I won't add any new objects to this app now. It can just about handle a good frame rate with what's on the screen, so I think this is soon going to be frozen apart from bug fixes.

What you could do is clone this and make a new version in default mode (and have widgets), with higher res images etc, but I don't know what frame rates would be like.

PR #242, opened 2020-04-06
Improve update process

### LINKED ISSUE #216: Removing an app should look at `app.info` `files`

rigrig:
Fix #216
Cleaner PR than #235: just improves the update process without worrying about special settings/data files.

- Always fill `appsInstalled` before uploading
- Install: if `appid.info` already on device, update instead
- Update: get files to remove from `appid.info`, only remove files which
          won't be overwritten anyway
- Remove: get files to remove from `appid.info`

### DIFF
--- a/js/comms.js
+++ b/js/comms.js
@@ -8,7 +8,7 @@ reset : (opt) => new Promise((resolve,reject) => {
     setTimeout(resolve,500);
   });
 }),
-uploadApp : (app,skipReset) => {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks for this! It looks really tidy...

PR #245, opened 2020-04-07
Mario Clock [Enhancement] - Extend info displayed + add Daisy

paulcockrell:
# Details
Press BTN3 to toggle what was displayed on the date panel (now referred to as the info panel) between the `Date`, `Battery level %`, and `Temperature`.

Add new character (this will be the last on I add) which is Princess Daisy.

## Info

This feature has been added as requests have come in to display widgets which is not possible in 80x80 display mode, nor was this watch face designed to show anything but what it currently does.

So as a compromise, I have updated the app to let you display the date, battery, and temperature in the one panel, by cycling through them on BTN3 press.

I think this works really well, and is not distracting nor taking up valuable space that wasn't already being used.

#### Date
screenshot (11)
#### Battery
screenshot (12)
#### Temperature
screenshot (14)
#### Princess Daisy
screenshot (16)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -914,7 +914,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
-    "version":"0.07",
+    "version":"0.08",
     "description": "Animated retro Mario clock, with Gameboy style 8-bit grey-scale graphics.",
     "tags": "clock,mario,retro",
     "type": "clock",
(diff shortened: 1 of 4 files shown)


paulcockrell:
cc @Purple-Tentacle

Purple-Tentacle:
Great :)
Problem is, I am losing connection to my phone when the GadgetBridge widget is not loaded.

paulcockrell:
@Purple-Tentacle Ah right that's a bit different to what I understood by 'seeing widgets in the clockface'. I certainly can take a look at that in another branch of work.

This PR covers everything I intended.

gfwilliams:
Thanks- looks great!

I guess there's a possibility of manually doing the 80x80 graphics:

```
var g2 = Graphics.createArrayBuffer(80,64,8);
...draw on g2
...copy to the screen with:
g.draw(g2,0,24,{scale:3});
```

But I think that'll end up being noticeably slower :(

PR #246, opened 2020-04-07
gbridge: don't turn on screen at start of every song

rigrig:
Do draw the notification, so you can still look up what is playing.


gfwilliams:
Cool, thanks!

PR #247, opened 2020-04-07
Added tabata high-intensivity app

kmcs:
Hi,
this is just a small app that helps you to do a interval training.
Thanks,
Timo


gfwilliams:
Thanks! Just so you know, there's a minor issue in your icon (a semi-colon where not needed). When you do the PR, you should see something like this after a few minutes:

image 

(Or hopefully a green tick!) If you click the red X and Details you'll get a page where it shows you what issue it thinks it found.

I'll just do the tweak now when merging though as it's super minor.

PR #248, opened 2020-04-08
Hebrew fixes and translation

yarons:
Hoping there will be RTL and font support soon.

If there's any way I can check it before merging it would be super.

### DIFF
--- a/apps/locale/locales.js
+++ b/apps/locale/locales.js
@@ -210,7 +210,7 @@ var locales = {
         int_curr_symbol: "ILS",
         speed: "kmh",
         distance: { 0: "m", 1: "km" },
-        temperature: "°F",
+        temperature: "°C",
         ampm: {0:"am",1:"pm"},
         timePattern: { 0: "%HH:%MM:%SS ", 1: "%HH:%MM" },
         datePattern: { 0: "%A, %B %d, %Y", "1": "%d/%m/%Y" }, //  Sunday, 1 March 2020  // 01/03/2020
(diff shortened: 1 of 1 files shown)


MaBecker:
Thanks, at the moment there is no font to display the chars you used. that's the reason for en_IL. 

Can you please check en_IL, to be sure it has  correct entries.

gfwilliams:
Thanks! But I'm afraid this is unlikely to work with the current fonts. You can see what we have available at https://www.espruino.com/Fonts#bitmap-font

You can test with the instructions at: https://www.espruino.com/Bangle.js+App+Loader

If there was a font with these characters, would it work for you? https://en.wikipedia.org/wiki/ISO/IEC_8859-8

yarons:
@MaBecker Sadly I've added the en_IL fixes as well but it's pretty basic, in Israel we're using Celsius, not Farnheit and I'll probably create an additional pull request to display the full date the right way.

@gfwilliams Thank you for all the info, I'm looking for a good bitmap editor to try and possibly design some bitmap Hebrew fonts myself.

Regarding ISO8859-8 it might do the trick but I think ISO8859-8-I is even better and from my experience Windows-1255 or CP1255 is the best in that case because it has a relatively the widest support range in addition to the word reversing algorithms (Hebrew is written from right to left).

Thank you very much!

MaBecker:
May be this Espruino `Bitmap Font Converter` can help, to create the wanted font.

http://ebfc.mattbrailsford.com/

yarons:
Yes, it looks awesome but I need to design the fonts first 😄 (draw the bitmap and put the character in the right place).

gfwilliams:
Closing this for now as it won't work, to try and de-clutter the repo. I changed the temperature in en_IL though

PR #252, opened 2020-04-08
GPS Info: Show number of satellites while waiting for fix

myxor:
Shows number of satellites while waiting for GPS fix like Speedo app does


gfwilliams:
Nice - thanks!

PR #253, opened 2020-04-08
Force reload of app list after successful upload to prevent exception

myxor:
I am forcing a reload of the app list after successful upload to prevent the following exception to occurr if one wants to remove an app right after installing it without reloading the apps or the web page in between:

> $SomeAppName removal failed, TypeError: Cannot read property 'split' of undefined

The error happens here:
https://github.com/espruino/BangleApps/blob/82790d5f38f288d50af7a089dea380fd12ebad4d/js/comms.js#L85


gfwilliams:
Thanks - but this seems like overkill for the problem.

The upload process should have added an app to `appsInstalled` along with a `files` element. I just added a fix that avoids the reload.

Issue #219, opened 2020-04-04
What about a App to customise my Banglejs?


MaBecker:
A app with options and value inputs for Bangle.setOptions() plus a free text input for optional stuff.

* Upload: store on device only option different to default and create cookie
* Default: rest values on app screen to default
* ...

Allows non developers to change default setting via app.

gfwilliams:
IMO the setOptions stuff should just be in Settings by default. There is an issue open for that...

But a custom app that allowed JS to be pasted into a boot.js would be cool, and would handily avoid your app removal issue if combined with favourite apps

MaBecker:
Ok, so let me try this.

gfwilliams:
Great! It'd be cool to use a syntax highlighted editor like CodeMirror, and maybe start off with some example code commented out that folks could uncomment and change?

MaBecker:
Have you planed to extend js/utils.js  by setCookie() and get getCookie()?

MaBecker:
```
function setCookie(cname, cvalue, exdays) {
  var d = new Date();
  d.setTime(d.getTime() + (exdays*24*60*60*1000));
  var expires = "expires="+ d.toUTCString();
  document.cookie = cname + "=" + cvalue + ";" + expires + ";path=/";
}

function getCookie(cname) {
  var name = cname + "=";
  var decodedCookie = decodeURIComponent(document.cookie);
  var ca = decodedCookie.split(';');
  for(var i = 0; i <ca.length; i++) {
    var c = ca[i];
    while (c.charAt(0) == ' ') {
      c = c.substring(1);
    }
    if (c.indexOf(name) == 0) {
      return c.substring(name.length, c.length);
    }
  }
  return "";
}
```

Does that save them local?

MaBecker:
```
function checkCookie() {
  var username = getCookie("username");
  if (username != "") {
   alert("Welcome again " + username);
  } else {
    username = prompt("Please enter your name:", "");
    if (username != "" && username != null) {
      setCookie("username", username, 365);
    }
  }
}
```

MaBecker:
Is storing CodeMirror in  css/codemirror and lib/codemirror like this ok ?

![Bildschirmfoto 2020-04-06 um 10 11 09](https://user-images.githubusercontent.com/15932646/78537008-0ef44500-77ef-11ea-8ad7-6591e2ee29f1.jpg)
[...]

PR #254, opened 2020-04-08
new app custom

MaBecker:
Wow that was very interesting, enjoyed it much! 

link to issue https://github.com/espruino/BangleApps/issues/219


gfwilliams:
Thanks! That'll be really helpful :)

PR #255, opened 2020-04-08
Update locales.js

stanzinofree:
Added it_IT copied by it_CH and modified for EUR as int_curr_symbol and \x80 as Euro symbol


MaBecker:
Thanks for it_IT

PR #258, opened 2020-04-08
Mario Clock [Enhancement] - Integrate GadgetBridge

paulcockrell:
# Summary

Integrate GadgeBridge (partially any way) in to Mario Clock.

# Details

If your BangleJS is connected to you mobile phone via GadgetBridge you will be notified of incoming calls and any generic notification.

Mario (or any of the characters) will show the notification type in a speech bubble, while the relevant details of the notification are scrolled along the info panel.

Pressing BTN1 or BTN3 will cancel the notification

# Issues

## Connection

It works fine, but having connected to the IDE, my Apps store, and Gadget bridge sometimes it wont find the device / connect without restarting the watch (turn off and on again). Maybe someone can see if I'm not handling the BLE properly?

## Emulator

The emulator complains the graphics buffer is out of memory. Can anyone see where I can save a few bytes?

# Images

mario-gadget-bridge

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -914,7 +914,7 @@
   { "id": "marioclock",
     "name": "Mario Clock",
     "icon": "marioclock.png",
-    "version":"0.08",
+    "version":"0.09",
     "description": "Animated retro Mario clock, with Gameboy style 8-bit grey-scale graphics.",
     "tags": "clock,mario,retro",
     "type": "clock",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Wow, nice! As much as I prefer not to reinvent the wheel, actually implementing it in such a way that it's part of the clock is awesome.

In terms of memory usage, adding `E.setFlags({pretokenise:1})` to the front may help on the emulator. It's turned on by default in Bangle.js's bootloader but since the emulator doesn't have that it might not do it?

gfwilliams:
Also, at some point I should add Storage into the emulator, then we can make this stuff save to flash.

PR #259, opened 2020-04-08
Toucher - Improve the rendering performance

brainfart-be:
I once again followed your tips ( #210 )

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1035,7 +1035,7 @@
     "name": "Touch Launcher",
     "shortName":"Menu",
     "icon": "app.png",
-    "version":"0.04",
+    "version":"0.05",
     "description": "Touch enable left to right launcher.",
     "tags": "tool,system,launcher",
     "type":"launch",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks good - thanks! Out of interest, do you find that the fillRect vs clear is faster?

brainfart-be:
I read somewhere that g.clear may glitch.
It was glitching on my bangle so i tried to put fillRect instead, it helped a bit i think.

But actually it's weird, when i was developing from the ide (on the watch), it felt a lot smoother than after installing it from the BangleApps.

Anyway it should be a bit faster than before :D

gfwilliams:
Ok - seems odd though as `.clear` is effectively just a fillRect under the hood.

> when i was developing from the ide (on the watch), it felt a lot smoother than after installing it from the BangleApps

This is something I could do with fixing - just created an issue at https://github.com/espruino/BangleApps/issues/264

PR #260, opened 2020-04-08
widpedom: Add goal setting

### LINKED ISSUE #250: Step counter goal
Most of the fitness tracker watches have some configurable step goal for each day (e.g. 10.000 steps).
So the watch can show the progress of reaching the set goal and can notify when the daily goal is reached.

Wouldn't it be great to have something similar here for the people who use this as a sports tracker as well?

rigrig:
Fix #250 
Draws progress as part of a circle
(default goal: 10.000 seems to be common-ish?)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -786,12 +786,13 @@
   { "id": "widpedom",
     "name": "Pedometer widget",
     "icon": "widget.png",
-    "version":"0.08",
+    "version":"0.09",
     "description": "Daily pedometer widget",
     "tags": "widget",
     "type":"widget",
     "storage": [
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks - this looks great! Do you think it makes sense to just use the existing json file for the goal though?

rigrig:
Drat, that would've been cleaner :-(
Not sure if changing that now is a good idea though? Or maybe mention in the upgrade message that people need to reconfigure and clean up the old settings file?

gfwilliams:
I'd just change it asap - it's only been up a few minutes right now so only a few people will have the extra file, which will disappear if they do `install default apps` anyway...

PR #265, opened 2020-04-09
widpedom: Fix daily goal, don't store settings in separate file

rigrig:
Looks like it actually didn't work at all, no idea how I managed to convince myself that it did :-(


gfwilliams:
Thanks - looks good!

PR #266, opened 2020-04-09
Fixed missing ampm property for some locales

Fabio286:
I've added `ampm` missing property in some locales (it_CH, it_IT, wae_CH), cause of following problem that make impossible to update Bangle's locale.

image


gfwilliams:
Thank you!

PR #269, opened 2020-04-09
Adds Battery Chart widget and app

msdeibel:
The icon.js needs fixing. Don't know why I never manage to get this working the first time. I would like to already release it, though. Since the data logging seems to work and people should be aware that the logger exists.


gfwilliams:
Thanks! Just done - I think it was just using `app.img.js` and not `.img` - changed, and now it passes :)

PR #271, opened 2020-04-10
Added Learn Nato Phonetic Alphabet app

unmotivagedgene:
First Bangle JS app. Based on the Morse Code app. Rather than provide a reference for Morse code it shows you the word used to communicate letters clearly from the NATO phonetic alphabet.


MaBecker:
Hi, normally I use names for spelling, but some friends  of mine always use this way, so this app will give me a chance to improve, nice.

Can you please add a ChangeLog file. You can start with the one from here
https://github.com/espruino/BangleApps/tree/master/apps/_example_app

If you like to tell more about your app you can add a Readme.md file like done for the Mario clock.

gfwilliams:
Looks great - thanks!

PR #273, opened 2020-04-10
new gpstrack app

jeffmer:
This app displays tracks recorded by gpsrec together with duration and distance on the bangle.js screen. It uses equirectangular projection which is accurate for small distances and faster than Mercator.


jeffmer:
Screen shot of app:

gfwilliams:
This looks great - thanks! Just before I merge, do you think this is something that might make sense being added to the GPS recorder app itself, since there is already a menu?

https://github.com/espruino/BangleApps/blob/master/apps/gpsrec/app.js

Up to you, but if you think it's a good idea then I'm happy to move your code over.

jeffmer:
Yes, very happy for you to merge this with the GPS Recorder - please go ahead. It was easier to develop separately. 

Jeff

gfwilliams:
Thanks! I just merged this in - it's really helpful!

PR #275, opened 2020-04-10
New clock - Numerals

ps-igel:
It's a simple clock only showing the time with big numerals in different colors.


MaBecker:
Hi, I like clocks.

Can you please add a ChangeLog file. You can start with the one from here

https://github.com/espruino/BangleApps/tree/master/apps/_example_app

MaBecker:
Close, sorry for not beeing precisely, please use the same text :)

ps-igel:
Sorry, I tried being original here. ;-)

gfwilliams:
Nice - thanks!

Issue #243, opened 2020-04-06
App update/removal: what to do with data/settings files?


rigrig:
Some apps create files on the device, how should update/remove handle those?

As I see it, there are multiple categories of these files:
* "settings": Used to save app settings, typically just one small file with known name
* "data": Used to save app data (gps tracks, heart rates, etc), these can be multiple files (e.g. `.gpsrc1`, `.gpsrc2`, etc), and might actually grow quite large
* "state": Used to save app state between reboots, e.g. the pedometer widget last known steps+date

I make this distinction because losing "settings" or "state" seems to me not as big a deal as losing "data":
You can always just reconfigure the app, but that data is gone forever.
On the other hand, having to reconfigure an app or seeing your step count reset after every update would be annoying.

Some solutions I can think of:
1. Have some way for apps to declare these files in `apps.json`, so the uninstaller can clean them up.
   Possibly this could include a filename prefix, to e.g. clean up all `.gpsrc*` files.
   This would be quite easy to implement, but I'm not too sure about the idea of just deleting all GPS tracks when someone uninstalls the app.
2. Apps should add created files to `<appid>.info` 
   This seems like a lot of hassle to implement in all apps though
3. Implement something similar to `interface.html`(or adapt that) to handle uninstalling. It could just silently clean up settings, or check for data files and ask the user what to do with them.
   This seems like a lot of hassle for all apps that only need to clean up one settings file though.

Maybe I'm being too paranoid, and option 1. would be fine?

gfwilliams:
> Have some way for apps to declare these files in apps.json

I think this is preferable. Maybe as you mentioned above we need separate categories for these kinds of files. Some kind of wildcard/regex might be a good idea as we're moving more towards having file extensions.

The other gotcha is right now, StorageFile and Storage are slightly different and have to be deleted in different ways :(

rigrig:
I mostly thought of these categories as different use cases, not sure if they need actual categorizing.
(Especially not if we are going to treat them all the same anyway)
<br>

> StorageFile and Storage are slightly different and have to be deleted in different ways

Uhm, so what happens if you mess this up? Could it just try both?
```js
storage.erase(filename); storage.open(filename, 'r').erase()
```
<br>

I also just noticed the documentation mentioning an 8-char limit on filenames, that doesn't seem to be true, or does it only apply to storageFiles?

gfwilliams:
> I also just noticed the documentation mentioning an 8-char limit on filenames

[...]

PR #281, opened 2020-04-11
Data files

rigrig:
Fix for #243

Adds a "data" section to apps.json, with data files to clean on uninstall   
These are added to `appid.info` as "dataFiles" or "storageFiles", and can contain wildcards.
If an app contains `<appid>.settings.js` and no "data" section, it is assumed to store its settings in `<appid>.settings.json`. So if all you want to do is save settings, you don't need to worry about adding a "data" section.

Also added commit to use this for the `gpsrec` app, as it seemed a good example app with pretty much every feature:
```json
"data": [
  {"name": "gpsrec.json"},
  {"wildcard": ".gpsrc?", "storageFile": true}
]
```
Which becomes this in `gpsrec.info`:
```json
,"dataFiles":"gpsrec.json","storageFiles":".gpsrc?"}
```
I made sanitycheck `WARN` if a wildcard pattern does not contain the appid, but that means it now complains about `gpsrec` :-(

Apps need to declare data files with "name" or "wildcard", and `sanitycheck` complains if you get it wrong. Both types are lumped together as-is in `<appid>.info` though, and removeApp simply checks if a filename contains `?` or `*` to decide how to handle it. (It could just handle all files as wildcards, but this seems cleaner and saves having to do `Storage.list()`)

the code handles storageFile wildcards by doing basically 
`Storage.list(/filename\u0001/).map(f => f.substr(0, f.length-1))`
 I don't think there is a nicer way to find those?

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,13 @@ and which gives information about the app for the Launcher.
   "files:"file1,file2,file3",
      // added by BangleApps loader on upload - lists all files
      // that belong to the app so it can be deleted
+  "dataFiles":"appid.data.json,appid.data?.json"
+     // added by BangleApps loader on upload - lists files that 
+     // the app might write, so they can be deleted on uninstall
+     // typically these files are not uploaded, but created by the app
+     // these can include '*' or '?' wildcards
+  "storageFiles":"
+     // same as "dataFiles", except the app handles these as storageFile
 }
 ```
 
@@ -240,16 +247,27 @@ and which gives information about the app for the Launcher.
      "evaluate":true          // if supplied, data isn't quoted into a String before upload
                               // (eg it's evaluated as JS)
     },
+  ]
+  "data": [                   // list of files the app writes to
+    {"name":"appid.data.json",  // filename used in storage
+     "storageFile":true       // if supplied, file is treated as storageFile
+    },
+    {"wildcard":"appid.data.*" // wildcard of filenames used in storage
+    },                         // this is mutually exclusive with using "name" 
+  ],
   "sortorder" : 0,            // optional - choose where in the list this goes.
                               // this should only really be used to put system
                               // stuff at the top
-  ]
 }
 ```
 
 * name, icon and description present the app in the app loader.
 * tags is used for grouping apps in the library, separate multiple entries by comma. Known tags are `tool`, `system`, `clock`, `game`, `sound`, `gps`, `widget`, `launcher` or empty.
 * storage is used to identify the app files and how to handle them
+* data is used to clean up files when the app is uninstalled
+  (If the app has settings but no data section, it is assumed settings are 
+  stored in `appid.settings.json`, so there is no need to add a data section 
+  containing only that file)
 
 ### `apps.json`: `custom` element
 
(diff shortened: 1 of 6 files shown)


gfwilliams:
Thanks - this looks really neat!

Only minor thing I'd say is could we change `dataFiles/storageFiles` to just `data/storage` or something shorter?

Seems stupid I know but as the app JSON for all apps is loaded into memory for the launcher/bootloader, anything we can do to avoid it using up more RAM seems like it'd be good.

rigrig:
I thought about naming it that, but I'm worried it would confuse things: `app.storage` would have different meanings depending on whether it came from `apps.json` or `<appid.info>`. I'll try to come up with some shorter names though.

rigrig:
Instead of separate `dataFiles` and `storagefiles`, I figured I could just put them together in a single `data` property, so you get e.g. `"data":"gpsrec.json;.gpsrc?"`
Right now it's separating regular and storageFiles with a semicolon, I was wondering if there are any characters which are forbidden from showing up in filenames? I'd prefer to use one of those then.
(I looked around a bit, but couldn't find any. If not maybe we could at least make sanitycheck complain about weird characters?)

Also, what would be the most convenient way to adjust existing apps? Add the changes to this PR, or one big new PR, or separate PRs for every app?

gfwilliams:
Thanks - this looks really neat - I'm a bit unsure about hard-coding settings but I guess it makes sense, especially to force use of the same filename.

Maybe we should have a quick think about the filename though. I think stuff I've made just uses `myapp.json` for settings so that should change, but I wonder whether `.settings.json` is a bit long given we only have 28 chars for the filename? Maybe just `.settings`?

There aren't any disallowed characters in filenames (apart from `\0`, but it seems pretty unfriendly). Please could we just use comma? That's what we currently use for the `files` list so I think it makes sense to use the same for `data`. It'd be great to make `sanitycheck` check filenames in `storage` for a comma though.

Am I right in thinking this PR would be totally fine being merged without app changes? If so I'd do that and then modify the apps in a separate PR

rigrig:
I figured most apps would just have the one settings file, and it would keep `apps.json` a bit cleaner if they don't all need to add a boilerplate `"data":[{"name":"myapp.json"}]`.  But leaving the hard-coded bit out would uncomplicate things, and it would be clearer from `apps.json` which apps actually use data files.
Probably best to leave out hard-coded "magic".

`myapp.json` does look like a better name, and it might contain non-settings data. I think I'll make `sanitycheck.js` suggest that.

> Please could we just use comma?

The way I did it now it needs two characters:
- comma between filenames
- semicolon between regular files and storageFiles

App with only regular files: `"data":"file1,file2"`
App with only storageFiles: `"data":";storageFile1,storageFile2"`
App with both: `"data":"file1,file2;storageFile1,storageFile2"`
So for `gpsrec` that turns into `"data":"gpsrec.json;.gpsrc?"` because it uses one regular file and one storageFile wildcard.

gfwilliams:
Ahh, gotcha. Sorry, I didn't read your post properly - yes, semicolon seems pretty sensible then.

And yes, I think probably removing the hard-coded is better - as you say it makes it more obvious if settings files are produced.

thanks!

gfwilliams:
This looks great - thanks for all the work you've put into this!

PR #282, opened 2020-04-11
Sanity check: fix warning about app without id, check for unknown keys

rigrig:


MaBecker:
what about checking if  ChangeLog is missing?

gfwilliams:
> what about checking if ChangeLog is missing?

It does, but only if the version isn't 0.01 :)

gfwilliams:
Thanks for this - really helpful for detecting mis-spelling/etc

PR #287, opened 2020-04-13
Add App Calculator

fredericrous:
Introduce the missing calculator for BangleJS!
You can navigate with tap right/left, btn top/bottom, select the key to press with the middle btn.

It's inspired from MacOS's one. It does not support Big Number. The current design doesn't let room for more than 14 digits.
There's room for improvement like adding a scientific layout, but the current one is pretty functional

Attached is the mockup I used before coding the thing
image


gfwilliams:
Wow, excellent - thanks! Just so you know, ` apps/calculator/calculator.info` isn't actually required - but I'll delete that here.

PR #289, opened 2020-04-13
Active Pedometer Widget

Purple-Tentacle:
Forum thread: http://forum.espruino.com/conversations/345754/#comment15208524


Purple-Tentacle:
I discovered a bug, deleting this PR

PR #290, opened 2020-04-13
Active Pedometer Widget

Purple-Tentacle:
See forum thread: http://forum.espruino.com/conversations/345754/#comment15209010


gfwilliams:
Nice - thanks! Just FYI if you create a PR from your GitHub repo and then you update that repo, it updates the pull request - so there's no need to close & reopen :)

PR #292, opened 2020-04-14
add favourite functionality

MaBecker:
* library select/unselect apps/widgets
* as library  section
* as upload
* deny unselect  for boot and setting


gfwilliams:
Looks awesome - thanks!

PR #293, opened 2020-04-14
New game Snake!

Fabio286:
A simple snake game for our Bangle.js.


gfwilliams:
Thanks! That was definitely needed :)

PR #302, opened 2020-04-14
Fix welcome apps running during settings update

### LINKED ISSUE #298: Welcome app loads while uploading apps!
Recent changes to run the welcome app on `.boot.js` really messed this up. When settings gets removed, the welcome app jumps up, which probably uses so much memory it makes it difficult for settings to update.

rigrig:
Welcome apps:
- Run again after update
- Don't run again when `setting` app was updated
- Add "Run Now" option to settings menu

This fixes #298

Not sure how much of a hurry there is to get a fix for this out? Once #281 is done this could be solved neater, but I didn't want to rush that.
(And I figured #298 would be easy to fix by just checking if settings was present. But that didn't work because the problem turns out to be that all settings are reset when the `setting` app is updated)


gfwilliams:
Wow, thanks for sorting this so quickly! I was going to have a poke around today and you'd already done it :)

This looks like a good fix to me. Apart from having a separate set of files that get called from `.bootcde` (eg. when no specific app is loaded) I think this is going to be the best solution.

gfwilliams:
Just to add, in the old case I guess we could have checked `welcomed===false` so if settings didn't exist it didn't fire, but that seems over the top

rigrig:
Well, I did kinda break it, so that annoyed me into wanting to fix it quickly ;-)   
And `welcomed===false` would fix it if settings doesn't exist, but it would still run again after reinstalling/updating the settings app, because that overwrites all settings with the default.

PR #303, opened 2020-04-15
Stopwatch, added total elapsed time

Red-The-Hunter:
I tried to mess with the interface.html but unfamiliar with HTML, reverted to the current one on main repo.
Changes:
Added total elapsed time
Changed display so total time is the main display with lap time underneath
The final lap that is left when the start/stop button is now saved into the array
Total run time now appends to first entry in the array

Could you please help me with the HTML to display the total time and make it only read laps from the 2nd entry of the array? Thanks!


Red-The-Hunter:
screenshot (7)

gfwilliams:
So just to be clear on this - you're adding an extra field to the JSON at the start, and realistically you just want what would have been:

```
0 : ...
1 : ...
2 : ...
3 : ...
...
```

to read:

```
Elapsed time : ...
0 : ...
1 : ...
2 : ...
...
```

?

Red-The-Hunter:
> So just to be clear on this - you're adding an extra field to the JSON at the start, and realistically you just want what would have been:
> 
> ```
> 0 : ...
> 1 : ...
> 2 : ...
> 3 : ...
> ...
> ```
> 
> to read:
> 
> ```
> Elapsed time : ...
> 0 : ...
> 1 : ...
> 2 : ...
> ...
> ```
> 
> ?

Heya, yeah, in the main app code I appended it to the array before the function call to map the array in reverse and return, I figured the total time would have been easier to handle if it was at the start of the array.

I've probably been making a mountain out of a molehill or done it in some roundabout way, sorry for that, saw the improvements you made to the earlier code, thanks! Has helped me learn. First time doing javascript, I wanted to use the bangle as a means to learn how to use js.

Would it be better to have it under the lap table, or have it put in as an entry above the lap table? 

eg.

Elapsed time | 0:20:34
-----------------------------------
Lap                  | Timing
-----------------------------------
1                      | 0:05:20
2                      | 0:05:15
3...
etc...

gfwilliams:
Ok, just added - I stuck it in next to the lap list

Red-The-Hunter:
Thank you!

PR #306, opened 2020-04-16
Active Pedometer: Distance calc and display

Purple-Tentacle:


gfwilliams:
Thanks! By the way you can do `require("locale").distance(dist_in_m)` to get a nice, formatted, internationalised distance (eg for europe <1km is in m, and > is in km)

PR #312, opened 2020-04-16
Remove "settings" from appid.info

rigrig:
Didn't see you already improved it a bit, but this also removes the need for appid.info.settings.
(So now `appid.settings.js` is hardcoded in the settings app instead of during upload)

So we could get rid of https://github.com/espruino/BangleApps/blob/242c35dfac89f2e962b807292f8e17dc6d0bdc80/js/appinfo.js#L63-L64 once enough(?) people have updated the settings app.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -120,7 +120,7 @@
   { "id": "setting",
     "name": "Settings",
     "icon": "settings.png",
-    "version":"0.16",
+    "version":"0.17",
     "description": "A menu for setting up Bangle.js",
     "tags": "tool,system",
     "storage": [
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks! I'm not sure memory usage there is such a big deal, but anything we can do to simplify things has got to be a bonus :)

IMO loading the JSON files on demand for sorting will probably be super slow for more than a handful of apps with settings though.

wrt mem usage: I had been working on the assumption that all apps' info files could be loaded into RAM in one go, but it seems that's not really sensible now. Once we load the JSON one bit at a time and keep only what we need, I don't think memory usage is an issue.

rigrig:
Updated the PR to just get rid of `settings` in `appid.info`.
I figured that people who are updating apps would also update the settings app, so it was OK to stop adding a `settings` field to `appid.info` right away.

gfwilliams:
Neat - thank you!

PR #315, opened 2020-04-17
Chrono Widget 0.02

Purple-Tentacle:
0.02: Setting to reset values, timer buzzes at 00:00 and not later (see readme)


gfwilliams:
Thanks!

PR #318, opened 2020-04-17
Data files

rigrig:
Make these apps use data files:

- widbatpc
- welcome
- setting
- alarm
- heart
- ncstart
- numerals
- gpsrec

gpsr**e**c uses `.gpsrcN` to store tracks, so sanitycheck still complains about the filename not containing the app Id :-(


apps/welcome/settings.js
```diff
@@ -1,13 +1,13 @@
 // The welcome app is special, and gets to use global settings
 (function(back) {
-  let settings = require('Storage').readJSON('welcome.settings.json', 1)
+  let settings = require('Storage').readJSON('welcome.sjson', 1)
```
gfwilliams, line 3: is this a typo?

gfwilliams:
Looks great - am I missing something about the settings though? Maybe it's fine but I'd be a little concerned about removing the default settings file in case some other app actually expects that there'll be values in it...

apps/welcome/settings.js
rigrig: Yes, well spotted.

rigrig:
I looked at all apps, and as far as I could tell they all handled missing `setting.json` properly.
The problem with including a default settings file is that it gets overwritten every time you update the app. Never deleting the global settings file seems like a good idea though.

gfwilliams:
Thanks! Definitely removing the default file seems like a good plan - there's so much duplication at the moment

gfwilliams:
Just to add, one other option for settings is a `overwrite:true` option that makes the IDE insert the code: `if (!storage.list(filename)) storage.write(filename, ...)`

rigrig:
We could do something like that, but do we actually want to upload default setting files?

I think apps should be able to handle missing settings. If only because a new app version might add a setting, so the updated app needs to handle the old file where that setting is missing.
The logical way to handle missing setting is to use the default value, so there is no point in uploading a default settings file: those values are coded in anyway so you would only duplicate them some more.

(Maybe apps could include a `settings-default` file which is never actually used, as documentation?)

gfwilliams:
Yes, good point :) Thanks!

PR #320, opened 2020-04-17
AstroCalc [Enhancement] - Store last GPS fix for optional use next time

paulcockrell:
# Problem:
The GPS lock can take a long time, and if indoors the app probably cannot be used.

# Solution:
Store the last successful GPS lock data to file, then when the app is next started, the user is presented with the option to use their last GPS position, with the date value updated.

# Images
astrocalc-screen-shot

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1018,7 +1018,7 @@
   { "id": "astrocalc",
     "name": "Astrocalc",
     "icon": "astrocalc.png",
-    "version":"0.01",
+    "version":"0.02",
     "description": "Calculates interesting information on the sun and moon cycles for the current day based on your location.",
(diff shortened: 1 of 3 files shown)


Purple-Tentacle:
As there are several apps using GPS data, maybe this could be done globally and not in each app seperately?

gfwilliams:
Looks great, thanks! I noticed `getCenterStringX`  - did you know you can do `setFontAlign(0,-1)`?

@Purple-Tentacle thanks - that could be a good idea. I think realistically it probably wants doing inside Espruino, so I filed an issue here: https://github.com/espruino/Espruino/issues/1803

PR #323, opened 2020-04-18
Fix App Calculator issues

fredericrous:
- Javascript rounded float issues
- Reset after equals wasn't done properly
- Introduce unit tests 🤗

image


gfwilliams:
Wow, thanks! I wonder if there's some way the emulator could be hooked up to a testing framework?

PR #324, opened 2020-04-18
Adding an app : taking pictures remotely with bluetooth HID

paul-chrlt:


gfwilliams:
This is great - thanks!

PR #327, opened 2020-04-19
Add option in settings to hide battery indicator widget when battery is over 20%.

lightnin:
In case the user only wants to know battery status when it is low enough to need charging soon, this setting will stop the widget from being drawn when the battery has >  20% capacity.  
It ain't too pretty here but it works. Suggestions / changes are of course welcome - I recognize my coding skills are pretty rusty. :)

### DIFF
--- a/apps/widbatpc/widbatpc.settings.js
+++ b/apps/widbatpc/widbatpc.settings.js
@@ -11,6 +11,7 @@
     'color': COLORS[0],
     'percentage': true,
     'charger': true,
+    'hideifmorethan20pct': false,
   }
   // ...and overwrite them with any saved values
   // This way saved values are preserved if a new version adds more settings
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Seems like a good idea. We just tweaked some other stuff so I'll have to rename some files to merge this in, and I'll just change `hideifmorethan20pct` to `hideifmorethan` and make the percentage configurable - I imagine some people might want to change the percentage level and I reckon it might look tidier too.

lightnin:
By all means please tweak! I'm noticing when I force reload / redraw many times in a row ( for my experimental weekend word clock) things slow down and memory errors show up in the console. It might be my WIP word clock app but it might be this widget. Not sure if I'm bailing out at the right point in draw.

PR #328, opened 2020-04-20
Added app for running

singintime:
Added an app for running sessions. It uses the embedded GPS, heart rate monitor and step counting features provided by Bangle.js to display:

- GPS status,
- [A CLOCK!!!](https://www.youtube.com/watch?v=sforhbLiwLA)
- total distance traveled,
- time elapsed since workout started,
- average pace (in minutes and seconds per km)
- total step count
- cadence (steps per minute)
- current speed
- current heart rate
- current cadence

screenshot (1)

GPS data goes through some basic cleaning: outliers are filtered out using the interquartile range technique, and total distance is not updated for movements smaller than 10 m to avoid taking into account unwanted noise.

Compared to the running app on my smartphone (Huawei Health) my distance count is pretty optimistic (about 10% greater). I'm planning on adding track recording on a file, using either CSV or GPX as a format, as a feature for the app and also to investigate this issue. I'm also looking into refining the graphical layout, which is a bit crude. In particular, I'd like to categorize info into multiple pages to have more readable and less crowded screens, given that they are supposed to be read while running.


gfwilliams:
Wow, that's great! Really nice having the filtering too.

In terms of layout, if you were able to keep the bottom 24px clear it might help as that can be used for widgets - not many use it right now, but one or two do.

For recording - maybe it'd be possible to just use the `GPS Recorder` widget? That should work alongside this and record to files (since it already has all the GPX Export stuff).

PR #329, opened 2020-04-20
Removed href="#" that scrolls pages on top

Fabio286:
I've replaced `href="#"` with `class="c-hand"` (available in Spectre framework) to avoid pages scrolling on top on every click on app changelogs or "Read mode..." links.


gfwilliams:
Cool - thanks!

PR #333, opened 2020-04-21
Minion clock v0.02

singintime:
Minor improvements on the Minion clock app, following @gfwilliams 's suggestions on my previous PR:

- Changed font to 6x8 with 2x zoom on the Date field,
- Fixed redraw of widgets on app startup / lcd screen on

screenshot (3)


gfwilliams:
Thanks! Do you think you could keep `Bangle.loadWidgets();` (just not `drawWidgets`) where you had it before? Just before `setInterval(draw, 1000);`.

The issue is you don't want to reload all the widgets from flash every time you redraw the clock. It'll be slow and it'll also break a lot of them :)

Also, I just noticed that this will redraw the clock every second regardless of whether the screen is on or not. It's not a big deal, but it might hurt battery life. In other clocks we tend to turn the setInterval on when the LCD turns on, and then turn it off when it turns off: https://github.com/espruino/BangleApps/blob/15a92c1b38006dfd908c91e69b060ca457d35e9b/apps/aclock/clock-analog.js#L130

singintime:
Yeah, makes sense :) Please let me know if you spot something else!

gfwilliams:
Thanks for doing that so quickly! I think you might still want a `timer = setInterval(draw, 1000);` just after `loadWidgets` though?

The LCD will be on when the clock is loaded, so `setInterval` wouldn't get called otherwise - and after first boot you probably still want the clock to run.

gfwilliams:
Thanks! Just merged.

PR #337, opened 2020-04-21
Active Pedometer 0.03 / Chrono Widget 0.03

Purple-Tentacle:
Active Pedometer:
- The graph module is now part of the source code.
- I implemented data logging to file and an app to view data as graph

Chrono Widget:
- Display only minutes:seconds (00:00) when less than 1 hour is left on timer


gfwilliams:
Thanks!

PR #338, opened 2020-04-21
Update apps.json entry for gpsnav

jeffmer:
Add a reference to the existing README.md file for gpsnav so that it can be read in the app loader - a nice feature that I have just become aware of.


gfwilliams:
Thanks!

PR #340, opened 2020-04-21
Weather widget+app

rigrig:
Based on http://forum.espruino.com/comments/15194626/, where 
NebbishHacker did most of the actual work :-)

I also had to update locale, because Gadgetbridge reports temperatures in Kelvin.


apps/weather/lib.js
```diff
@@ -0,0 +1,176 @@
...
+      g.setColor("#FFFFFF");
+      for(let i = 0; i<5; ++i) {
+        g.fillRect(x+layers[i][0]*r, y+(0.4*i-0.9)*r, x+layers[i][1]*r,
+          y+(0.4*i-0.7)*r-1);
+        g.fillCircle(x+layers[i][0]*r, y+(0.4*i-0.8)*r-0.5, 0.1*r-0.5);
+        g.fillCircle(x+layers[i][1]*r, y+(0.4*i-0.8)*r-0.5, 0.1*r-0.5);
+      }
+    }
+
+    function chooseIcon(condition) {
```
nebbishhacker, line 154: This method should work as long as the condition descriptions match the ones at https://openweathermap.org/weather-conditions, but it's kind of brittle and would break pretty easily if the descriptions are ever changed or localized. Long-term it would be good to see about getting the condition ID from gadgetbridge and choosing the icon based on that.

gfwilliams:
That's really cool, thanks! It's very cool how it'll just work with/without the gadgetbridge widget :)

Please could you remove the `locale` changes though and just subtract 273.15 before calling locale.temperature?

`locale` works with a built-in version inside Bangle.js firmware - so if you don't install any language you just get `en_GB`. So changing locale functions here means that anyone without a language installed would see 290ish degrees C, while with Language it'd be ok. Not to mention the docs not being updated :)

gfwilliams:
Thanks!

rigrig:
Hadn't thought of that, I figured there was still time to change it because no other apps used `locale.temp` yet. Changed it now. (Good thing that en_GB uses °C, or this would be even more tricky)

PR #343, opened 2020-04-23
BuffGym [Enhancement] - Add web interface for customisation

paulcockrell:
# Problem
The weights and weight-increments starting values were not customisable to the user, this would make the weights-tracking feature redundant.

# Solution
Add interface so the user can set their starting weights and weight-increment values and upload the customised workouts to the BangleJS

# Images
buffgym-webinterface


gfwilliams:
Wow, nice :)

Out of interest, is it up to the user to press BTN1 to indicate when they've done something, or can you use the accelerometer to detect when some things (eg weights) have been done?

paulcockrell:
That would be very neat, you could probably count reps with the accelerometer too, I looked at your tensor flow examples on this point. But yeah it's button interaction only. I think the way I've written it doesn't leave much memory spare.

Maybe one for another day!

gfwilliams:
Wow, yeah, I think tensorflow might be a bit over the top :)

I was imagining that if you're lifting Dumbbells it's going to be super easy to look at the accelerometer values and detect when the arm is up/level/down (or maybe even just the acceleration). Maybe first it's something for another app, then we could see about porting it over :)

PR #344, opened 2020-04-23
Toucher : Complete rewrite in 80x80, better perf, add settings

brainfart-be:
A week or two too late, sorry for those who waited.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1098,10 +1098,12 @@
     "description": "Touch enable left to right launcher.",
     "tags": "tool,system,launcher",
     "type":"launch",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Looks like the check is failing because you include `{"name":"toucher.json"}` with no contents?

Please can you add it like:

```
    "data": [
      {"name":"toucher.json"}
    ]
```

brainfart-be:
There you go!
I didn't know about that "data" config, thanks for pointing it out

gfwilliams:
Nice, thanks!

PR #348, opened 2020-04-25
wclock: Display in white during daytime, and hide digital time by default

lightnin:
Two minor changes for your consideration:
1. Use white for active time display during the daytime, and red at night. I find the red is difficult to see when outside in the daytime because of contrast / brightness. But it's nicer at night.

2. Much of the charm of a word based clock is in its casual imprecision. This changes the default behavior to not show the digital time, but show it whenever the user pressed the top button (and erase it when the button is released). 

I'm quite new to this so if you have any recommendations on formatting, etc. or better implementations, I'm open to feedback and thoughts. Thanks again for making such a cool watch and IDE! :)


apps/wclock/clock-word.js
```diff
@@ -59,10 +60,17 @@ function drawWordClock() {
     var midx;
     var midxA = [];
 
+    // Select time word display color based on hour
+    var activeColor = activeColorDay;
+    if(h < 7 || h > 19)
```
lixas, line 65: what about making a settings for sunrise/sunset time?

lightnin: I considered that, but didn't implement yet. My reasoning was that because this watch is designed to be "hackable" in such a way that the code for each app can be (relatively) easily viewed and edited, things that would normally need to be "settings" for an end user can instead be hard coded. The end-user in the case of Bangle.js is, in all probability, someone who bought the watch as a platform to learn and to hack.  One could consider hard coded values like this to be an invitation to dive in and make a simple change.  But I'm not sure how that view aligns with the goals of the project - and am happy to add a settings menu if the consensus is that it would be better that way. What do you think?

lixas: My opinion. Excuse my English

- Understand ability wise: your approach is good to understand the code

- Usability wise: settings would be better. A lot of peoples live where sun time at winter is totally different from sun time on summer. So as the sunrise/sunset times drifts over time. I guess noone wish to re-upload clock program each week with different times.

i guess next version of this clock could implement automatic sun rise/set calculation. I don't know is it good to implement sunrise/sunset at[ **locale** ](https://github.com/espruino/BangleApps/tree/master/apps/locale)so all apps could use that or at first try to implement in app, test it and then, maybe, implement at locale.
It is implemented in [**excel file**](https://www.esrl.noaa.gov/gmd/grad/solcalc/calcdetails.html) so it's not rocket science to migrate to JS. Just need to think about performance and resources.

lightnin: Everything you say makes good sense. I also considered automating the sunrise / sunset, perhaps using GPS, but it adds complexity and performance overhead that's probably not worth it. So let's go with adding a settings file for now. I'll work on that but it may not happen until the weekend. Thanks for giving feedback on this.

gfwilliams:
Hi - did you ever get around to adding settings, or shall I close this until that's sorted?

lightnin:
Yes, let's just close it for now. Not sure when I'll be able to get back to it.

PR #350, opened 2020-04-25
Add imprecise word clock app.

lightnin:
This is an: "Imprecise word clock for vacations, weekends, and those who never need accurate time." A remix of Gordon's excellent word clock, this click offers much less precision, giving the user only a vague picture of the time of day. Quite relaxing for those used to busy schedules!


MaBecker:
Cool -  I like it and will add it to my favourite list.

Have you thought about a readme to describe the functionality?

lightnin:
Good idea! All set. :)

gfwilliams:
That's great - thanks!

PR #352, opened 2020-04-25
bugfix: Bar Clock does not lauch if Languages app is not installed

lixas:
Fixed bug, that clock tries to use function `locale.meridian()` that may not exist if user does not installed Languages app. And the result is empty screen with no option.


gfwilliams:
Ahh - perfect, thanks!

PR #357, opened 2020-04-27
Pong app - improve ai, add 2 player local

fredericrous:
Fixes:
- sometimes the AI would be stuck in a loosing scheme, now its not but I hope the new AI won't be too difficult, it's possible to win but hard 🗡️ 
- sometimes the ball would bounce into a paddle instead of bouncing on to it, it should be fixed

Change/New:
- The "down" button is now BTN2 instead of BTN3, it was annoying to quit the game if we pressed BTN3 too long
- a screen that tells these buttons are inverted is shown when the game starts
- there is now a timer before the ball is thrown the first time (I don't know if it is of use, let's say it's convenient)
- 2 player mode over local (I know it's hard to play over one watch but why not ahahah)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1486,11 +1486,12 @@
     "name": "Pong",
     "shortName": "Pong",
     "icon": "pong.png",
-    "version": "0.01",
+    "version": "0.02",
     "description": "A clone of the Atari game Pong",
     "tags": "game",
     "type": "app",
     "allow_emulator": true,
+    "readme": "README.md",
     "storage": [
       {"name":"pong.app.js","url":"app.js"},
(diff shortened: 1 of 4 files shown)


gfwilliams:
Nice - thanks!

gfwilliams:
Just checked and the tests are failing right now because a README file is being referenced but it's not included. Did you forget to commit one, or was adding the `"readme"` line an accident?

fredericrous:
Indeed I forgot to commit the readme.
I also introduced an impure function, that's fixed

gfwilliams:
Great - thanks!

PR #363, opened 2020-04-29
keywords and readme update

paul-chrlt:
No changes in code, only enhancements for usability in app loader : added keywords and a read more


gfwilliams:
Nice - thanks!

PR #367, opened 2020-04-30
quick update class for Spectre.CSS

renaudgweb:


gfwilliams:
Thanks! If you're updating the ChangeLog, please could you bump the version number in `apps.json` to match?

Issue #162, opened 2020-03-30
Overhaul Notifications & Gadgetbridge


gfwilliams:
Merge:

* https://github.com/espruino/BangleApps/pull/130
* https://github.com/espruino/BangleApps/pull/131
* https://github.com/espruino/BangleApps/pull/155

Need to arrange that apps can create a `notify` module in storage, so you can then load an app from the app loader and have the scrolling notification, fullscreen notification, or any other custom notification (eg. custom look & feel)

Then remove notification rendering and just `require("notify")`, which should simplify the gadgetbridge module.

Also need docs :)

gfwilliams:
Ok, I could do with some input on this...

Looking into it, it feels like there's not a great deal of common code here just for Gadgetbridge - all the code (bar ~10 lines) is either handling user interaction or how stuff is rendered.

So personally I think this is somewhere where abstraction doesn't help us...

* Fullscreen notifications? Chances are we'll want new, bigger icons, different layout and different interaction methods. I think a complete new 'Gadgetbridge' app is needed.
* Different music controls? I think a new app makes sense there too.

Any thoughts?

detached:
Maybe we could implement a messaging system to loose coupling between gadgetbridge infrastructure and user interaction code. The gbrigde widget could just deal with the app communication and broadcast ipc messages. Other apps could subscribe to gbridge messages. 
With this architecture the user could choose the look and feel of the interaction with the phone by installing different apps and widgets. And the gbridge widget wouldn't be bloated by rendering code.

gfwilliams:
The problem is we have extremely limited memory, so we don't want to be unpacking and repacking the message.

The simplest way to work around this would be to just have a global `function GB(o){Bangle.emit('GB',o)}` - so just emit the gadgetbridge events and then apps can do `Bangle.on('GB',...)`.

However a super-simple thing to just build into each app would be:

```
var _GB = global.GB;
function GB(o) {
  // ...
  if (_GB)_GB(o);
}
```

which would work around that too, without needing any common code.

Purple-Tentacle:
I am up for more music controll. Things I want ( and am testing)

- Play / Pause
- Commands not only working when music notification is shown

detached:
@gfwilliams sry I don't get your last js example, but the first idea works like a charm.
I tried to reimplement the gbridge connection as a POC: I have broken down the gbridge app into three parts: connector, messages and music.
The connector just emits the GB messages.
[...]

PR #372, opened 2020-05-03
Split up gadgetbridge apps: gbnotify

detached:
This is the first of three apps for splitting up the gbridge widget: #162
The two other apps (call and music control) are on this branch https://github.com/detached/BangleApps/tree/new-gbridge-apps. I want to keep the pull request small, so I will contribute one after the other.

The gbnotify just handles notifications. The widget shows notifications in the same way as the gbridge app, just slightly redesigned.

The app allows swiping through the last five messages.


apps/gbnotify/widget.js
```diff
@@ -0,0 +1,89 @@
...
+  global.GB = (event) => {
+    if (event.t === "notify") {
+      store(event);
+      if (!Bangle.isLCDOn()) Bangle.buzz();
+      show(event);
+    }
+    if(_GB)setTimeout(_GB,0,event);
+  };
+
+  WIDGETS["mywidget"] = {
```
gfwilliams, line 84: Please could you change `mywidget` to `gbnotify` here? Otherwise other apps that forget too will just overwrite this

gfwilliams:
Nice - looks good. I like the idea of storing multiple notifications.

apps/gbnotify/widget.js
detached: Thanks for spotting this!

detached:
What do you think about register an event listener, so other widgets can show notifications with `Bangle.emit()`?

gfwilliams:
> What do you think about register an event listener,

I think probably not - realistically if we want a global way to send events we should probably build it into Bangle.js itself to try and get the memory usage down (and also to ensure that it's available for all apps).

detached:
Can I do something to get this merged?

gfwilliams:
Sorry for the delay - I've been a bit swamped with other stuff.

The thing that's made me think this is nontrivial is that this looks like it doesn't play well with the existing Gadgetbridge code at all, but then it doesn't seem to do the connection icon or battery level reporting. Having something 'modular' that doesn't work with the normal Gadgetbridge code seems like it's just going to really confuse the majority of users.

So I was trying to think of a way of handling this nicely, especially with several users complaining about being short on RAM - we really want to get things as efficient as possible.

It seems to me that there's a bit of duplication in https://github.com/detached/BangleApps/tree/new-gbridge-apps with how notifications are displayed/dismissed... 

I think in the long term I'll make a 'Notifications' library that can be used for displaying notifications. Maybe someone else makes different ones later (like fullscreen notifications, etc) - we could all call the notifications library, which should hopefully save some space and would also allow some customisation of look and feel later on.

But I guess for now we've got two options:

* Completely remove all notifications from the existing gadgetbridge widget and make it handle Battery level reporting and connection state only, and merge in all your widgets
* Document your widgets very clearly so folks know they don't work alongside the `gbridge` widget.

I'd be interested if others can weigh in here about what you think is best?

I like the modular idea and honestly I think the first option is better. I'll do the notification library now anyway as that should be pretty easy.

One minor point in your code.. Any chance you could remove the whole ` WIDGETS["gbnotify"] = {` completely so it doesn't just create a blank area on the widget bar? Same for `gbcall` - and you can actually change the width of your widget so the music widget could hide itself when not playing music and then pop up?

gfwilliams:
Ok, `notify` lib (and ability to specify dependencies) just gone in: https://github.com/espruino/BangleApps/commit/519262793d7ef91dfa91612ef72663e918ea1fe9

I nicked your different notification style - hope that's ok?

So that should tidy up the extra widgets quite a bit I think, if you were interested in implementing it?

detached:
Hi @gfwilliams,
sry that I didn't respond for that long time, I had quite a ruff time.

> I nicked your different notification style - hope that's ok?
That is great! :)

I just removed the notification rendering code from the widget and use the notify lib instead and also removed the `WIDGETS` code.

gfwilliams:
Sorry for the delay replying to this - and hope things are looking better now.

I think with the Messages/Android app this is no longer work keeping open. Gadgetbridge handling is at least Modular now as well.

PR #374, opened 2020-05-03
feat: add timer app

fbedussi:


gfwilliams:
Thanks! Just mergd, but I've renamed this app from `timer` to `simpletimer` though - the tutorial at https://www.espruino.com/Bangle.js+First+App uses `timer` for the example so I don't want folks to have troubles when they find there's already something with the same name :)

PR #375, opened 2020-05-03
numerals clock: fixed settings issue

ps-igel:
- fixed issue in settings where menu button would always jump back to BTN2 when entering the menu


gfwilliams:
Thanks!

PR #377, opened 2020-05-04
WIP - New Clock Face - Dotmatrix

paulcockrell:
New clock face, replicating the white on blue dotmatrix displays.

I don't have a watch atm, so if someone can check its okay (loads widgets etc) that would be ace.

dotmatrix-clock-screen-shot


detached:
The clock runs well, the widgets are loaded.



I was not able to open the launcher with BTN2. You might want to add this to your app.js:
```
setWatch(Bangle.showLauncher, BTN2, {repeat:false,edge:"falling"});
```

I am not sure about the battery drain of the clock because you are requesting the compass and HRM all the time. Maybe you could allow switching them on and off with BTN1?

A long month could overflow the screen size. Have you tested it with "January 21"?

gfwilliams:
Just to add, it'd be worth changing `g.clear()` to `g.clearRect(0,24,239,215)` so the blue background doesn't go right up to the widgets.

paulcockrell:
Thanks for checking it @detached yes I think i forgot to truncate the month string! Good spot. The display looks all off too, with the labels not lining up with the dotmatrix rows... strange.

Thanks for the tip @gfwilliams will have that added in.

I think its best I wait for my watch to do the QA on this watch face. 

Cheers all.

paulcockrell:
Closing until ready!

PR #378, opened 2020-05-04
Adding indication of HR

flautrup:
Adding visual indication of heart rate


gfwilliams:
I could be wrong, but does this not permanently turn the heart rate monitor on whenever showing the watch (even if the display is off?).

While fun, this would kill your battery in no time - so do you think it should be some kind of setting, or only on when the display is on?

flautrup:
Good feedback, I missed that I had to manage this as well. So I will close this pull request and do a new one when I have fixed this issue

PR #379, opened 2020-05-04
Stop impwclock from flashing every interval

Bzly:
This could also be used to stop https://github.com/espruino/BangleApps/tree/master/apps/wclock from flashing every interval.


gfwilliams:
Thanks - that seems great!

PR #383, opened 2020-05-05
Beebclock

tomgidden:
A nostalgic clock face.  BTN1 to enable/disable second hand, BTN3 to change clock-face style


gfwilliams:
Cool - thanks! Interested to see how this works as it's one of the first apps to include modules :)

tomgidden:
I'm having massive "Pebble" flashbacks from 7 years ago!

This version redraws the face entirely every minute, but I just noticed your post about drawing images direct from flash.  I'm going to do some tests and see whether that performs better.

gfwilliams:
I think having a 1 bit offscreen buffer would work well...

PR #387, opened 2020-05-06
Added support for indication of HR

flautrup:


flautrup:
Found a bug that needed to be fixed.

PR #390, opened 2020-05-07
Speedo app: Allow widgets

nxdefiant:
Enable Widgets in Speedo app

This allows e.g. GPS Recorder to work in the background while running Speedo.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -286,7 +286,7 @@
   { "id": "speedo",
     "name": "Speedo",
     "icon": "speedo.png",
-    "version":"0.01",
+    "version":"0.02",
     "description": "Show the current speed according to the GPS",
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! I'll merge now, but if you're adding widgets to more apps (which would be awesome!) please could you

* add a ChangeLog if there isn't one
* change `g.clear()` to `g.clearRect(0,24,239,215)` so widgets don't have to be redrawn each time the screen changes

I'll quickly tweak that for the speedo now as well.

gfwilliams:
Actually in this case it's not so simple because it uses double buffering, but you I guess you might have found that :)

nxdefiant:
Actually I did not know the clear() vs clearRect() part, thanks for the info.

PR #391, opened 2020-05-07
New App: Find Phone - find your phone via GadgetBridge

v1nc:
A simple app to find your phone via GadgetBridge.
Sends `findPhone` command if HID is enabled and phone is connected.
I will probably add a torch-like hotkey.


gfwilliams:
Neat - thanks! Just so you know it can be accessed from the Gadgetbridge settings on the phone, but this is much easier!

v1nc:
Thanks for the quick merge. I fell in love with BangleJs after one minute.

### Great work!

PR #392, opened 2020-05-07
New App: Get Up - reminds you to move after some time sitting

v1nc:
Sitting to long is dangerous!
This app reminds you to getup and move periodically. You can configure sitting and moving time in settings.

I am considering using movement to optimize the timer and add a widget.


gfwilliams:
Neat - thanks!

PR #394, opened 2020-05-07
Fix github links of apps for foked launchers

v1nc:
Github links of apps are statically linked to `github.com/espruino`.
These links are wrong if you run a forked App Launcher, probably with unpublished apps.

I added onClick eventListeners for every github link, which generate the right url, if the launcher is not accessed via the main domain.


gfwilliams:
Thanks - it's a good point.

However adding an onclick listener seems like overkill. Since we dynamically generate the HTML anyway, can't we just fill in the correct link when the HTML is created?

That way you can do right-click and copy link address - which is something I do quite a bit

v1nc:
Yeah you are right, I changed it.

gfwilliams:
Thanks. Just merged. Made some tweaks to it so it doesn't break when hosted at non-github URLs (eg localhost).

PR #401, opened 2020-05-10
App Manager: Add app sorting functionality

jl-cs:
Added sub-menu to "App Manager" app to sort apps:
- alphabetically ascending (by name)
- alphabetically descending (by name)
- manually

Since I think there might be "LOW MEMORY" issues when sorting manually very fast or when sorting asc/desc in rapid succession and then sorting manually, any suggestions, especially concerning these memory problems, are welcome!

### DIFF
--- a/apps/files/files.js
+++ b/apps/files/files.js
@@ -1,4 +1,4 @@
-const storage = require('Storage');
+const store = require('Storage');
 
 const boolFormat = (v) => v ? "On" : "Off";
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Looks great! Was there a reason for renaming storage->store though?

In terms of memory usage, best thing is to do what we do for launcher/etc - don't load all the app JSON into memory at one time, but just load the fields you need:

```
function setSortorder(app, val) {
  app = store.readJSON(app.id + '.info', 1); // load full app info before you save it
  app.sortorder = val;
  store.writeJSON(app.id + '.info', app);
}

function getAppsList() {
  return store.list('.info').map((a)=> {
    let app = store.readJSON(a, 1) || {};
    if (app.type !== 'widget') {
      return {id:app.id, name:app.name, sortorder:app.sortorder}; // <------------- extract only what's needed here
    }
  }).filter((a) => a).sort(sortHelper());
}
```

jl-cs:
Thanks a lot for the feedback! I'll update that, test again and create another pull request.

gfwilliams:
Great, thanks!

> just a convention I've used in other projects

Ok, no worries. If you contribute other code here I'd maybe stick with what the original author used though - refactoring existing code to personal taste at the same time as a PR makes it harder to see what your changes are, and also makes it tricky if the original author comes back to make some changes :)

jl-cs:
Ok I understand. I'll make sure to be careful and don't mix that in the future. Hopefully the original author won't be mad at me this time. 😅

Issue #381, opened 2020-05-05
Clock not updating after a while (RAM issue)


jl-cs:
Edit:
It seems the problem not only affects the Morphing Clock but all clocks. At least the different ones I've tried so far. The problem is the same everywhere, as described in the original report below. 

Currently I've got no idea how to tackle this. Any suggestions?

Original report:
Don't know if anyone else has the same problem, but when using the Morphing Clock and the LCD turns of for some time, the clock doesn't update itself when the LCD is turned on again. Event the seconds counter has stopped and it only returns to normal if I switch to menu and back by BTN2 or hold BTN3 to reset the clock.
Any ideas why this might be happening?

Bzly:
Which version of Morphing Clock do you have? The App Loader should tell you on the "My Apps" screen. 

From my cursory glance this looks similar to #279 - reportedly fixed in v0.04+

jl-cs:
I'm on v0.04 of Morphing Clock and also latest(?) firmware according to this post: [http://forum.espruino.com/conversations/346029/](url)

gfwilliams:
Could you try turning `Debug Info` to `Show`? And maybe also install the `RAM` widget.

It's possible you have enough stuff the clock is running out of memory, or maybe some other widget is causing an error which is stopping the clock from working.

jl-cs:
Yes, `Debug Info` is set to `Show`.
I'm also using the RAM Widget , but while daily usage to only other things running are the "Battery Level Widget (with percentage)", "Gadgetbridge" (with active bluetooth connection to my smartphone) and "Battery Chart". Then, when activating the watch it shows the RAM at 54% while clock + these widgets are in use. I don't know if that is already too much. Also, when the clocks hang, the RAM Widget doesn't show a strongly different amout of used RAM, at max. 2% or 3% more then the "normal"(?) 54%.

jl-cs:
Is there a way to log the running code even further? Despite using the latest versions of all apps from here [https://espruino.github.io/BangleApps](https://espruino.github.io/BangleApps) the problem persists and even got worse, since the freezes happen very often.

gfwilliams:
Yes, sure. The best way to debug is to connect with the Web IDE: https://www.espruino.com/ide/

Then hopefully if you can get the problem to occur, the IDE will display the full error message and stack trace that caused the issue.

jl-cs:
[...]

PR #402, opened 2020-05-11
App Manager Sorting: use less memory

jl-cs:
@gfwilliams I've updated the app by your suggestion, here's the pull request. Thanks again!

### DIFF
--- a/apps/files/ChangeLog
+++ b/apps/files/ChangeLog
@@ -1,2 +1,3 @@
 0.02: Fix deletion of apps - now use files list in app.info (fix #262)
 0.03: Add support for data files
+0.04: Add functionality to sort apps manually or alphabetically ascending/descending.
--- a/apps/files/files.js
+++ b/apps/files/files.js
@@ -207,6 +207,7 @@ function showSortAppsManually() {
 }
 
 function setSortorder(app, val) {
+  app = store.readJSON(app.id + '.info', 1);
   app.sortorder = val;
   store.writeJSON(app.id + '.info', app);
 }
@@ -215,7 +216,7 @@ function getAppsList() {
   return store.list('.info').map((a)=> {
     let app = store.readJSON(a, 1) || {};
     if (app.type !== 'widget') {
-      return app;
+      return {id: app.id, name: app.name, sortorder: app.sortorder};
     }
   }).filter((a) => a).sort(sortHelper());
 }


gfwilliams:
Thanks! Did this work ok?

I'd posted up the code but I didn't actually test myself :)

jl-cs:
Yes, it helps a bit. I've tested a it and now one has to try a little harder to get the memory warnings, but you can still get them. It's almost certain you get them when you sort ASC/DESC in succession a couple of times (3-4x) and then try to enter the "Sort: manually" sub-menu. Then there will almost always be the memory/low memory warning. Maybe it's possible to alleviate this be clearing/garbage collecting the memory when navigating back in the sub-menus, but I'm not sure.
(It also could just be me or my Bangle.js, see my question [here](https://github.com/espruino/BangleApps/issues/381#issuecomment-626365278).)
Since I'm not very used to programming as "close" to the hardware as with the Bangle.js, I don't quite understand why i. e. the garbage collector seems not to be helping here.
But maybe you could point me in a direction here to understand it better or provide a spot in the documentation where I could look to try to improve it even more.

gfwilliams:
Ok, thanks! The biggest issue with this kind of stuff is usually having a reference to old data hanging around - either via a variable, or with a function closure.

One big issue in settings was setting a menu item with `menu["foo"] = newMenu` - which cased newMenu to be called with an argument which was the old menu - and *that* meant there was a reference to it hanging around. I don;t think that's an issue here though

Looking at the source there's actually a global `m` variable that doesn't need to be there. I'll merge your changes and remove that  now, and that should help a lot.

jl-cs:
Thank you very much!

jl-cs:
Another short question: I've added to the changelog, but it doesn't show the new version number in the App Launcher at https://espruino.github.io/BangleApps/
I assume there has to be a value that has to be changed somewhere to make that happen, but I didn't see one in the App Manager folder in the repo.

gfwilliams:
Ok, just made some changes, but there's still some kind of issue. Actually just tracked this down to a firmware bug - `writeJSON` had a memory leak. I'll fix it now, but it'd need a new firmware update to work properly.

> I assume there has to be a value that has to be changed

Yep, it's in `apps.json` - we have to have it in a central place so the App Loader doesn't have to load all 100+ files each time it starts :)

jl-cs:
Thanks for the info.

Is there a schedule for when new firmware updates get released or is it done if you've got time? Not trying to pushy here, just curious. ;-)

gfwilliams:
The 'cutting edge' firmwares are built automatically every time there's a fix.

The 'official' releases are basically when I have time and when I think it'd stable!

PR #408, opened 2020-05-11
BatteryChart 0.10

msdeibel:
Remove widget icon and improve listener and setInterval handling for widget (might help with https://github.com/espruino/BangleApps/issues/381)


gfwilliams:
Just looking at this - if you're not drawing a widget (and it seems you don't update anything in 'reload') then you could just skip adding an item to WIDGETS alltogether. It'd save a bit of space.

PR #410, opened 2020-05-12
Update en_CA locale units

nebbishhacker:
We're pretty inconsistent when it comes to units here, but we are
theoretically a metric nation.


gfwilliams:
Neat - thanks!

PR #412, opened 2020-05-12
Switch from writeJSON to write to possibly mitigate memory leak in writeJSON

jl-cs:
Just changed from write to writeJSON as per recommendation for now.


README.md
```diff
@@ -1,5 +1,10 @@
-Bangle.js App Loader (and Apps)
-================================
+Bangle.js App Loader (and Apps) (FORK - not actively maintained)
```
gfwilliams, line 1: We probably don't want to commit this ;)

gfwilliams: ahh - just saw you closed it :)

jl-cs: Sorry, that was just me being stupid. ... 🤦

PR #413, opened 2020-05-12
Switch from writeJSON to write to possibly mitigate memory leak in writeJSON

jl-cs:
Just changed from write to writeJSON as per recommendation for now.


gfwilliams:
Thanks! Newer firmwares have already fixed the writeJSON issue but I guess people will still be using the old one :)

PR #415, opened 2020-05-12
Add Random Clock Loader widget

msdeibel:


gfwilliams:
That looks great - thanks!

PR #418, opened 2020-05-14
Adding Elegant Text Watchface

jpoles1:


gfwilliams:
Hi - please can you actually include the code in the repository rather than making it a submodule?

There was a bit of a discussion about this earlier, and it's just easier to keep things all in one place for now - especially as the version in apps.json and the ChangeLog are supposed to be in sync, so you couldn't actually make a change properly without a PR to this repo anyway.

PR #422, opened 2020-05-15
New App - Dotmatrix clock

paulcockrell:
New clock face, replicating the white on blue dotmatrix displays.

dotmatrixclock-screenshot


gfwilliams:
Very nice - thanks!

PR #424, opened 2020-05-15
Reduce invasiveness of RndmClk

msdeibel:


gfwilliams:
Thanks! This seems much more sensible!

I'd argue that perhaps you only need to change the clock face once - when the watch first starts up? Or maybe the *first* time the LCD turns off?

PR #428, opened 2020-05-17
Update gpsnav app to 0.03 with waypoint editor

jeffmer:
This version has a waypoint editor that uses the interface mechanism. Consequently, it is invoked by clicking on the download data icon from app icon in the app loader. This is a little misleading as the waypoint editor both downloads and then updates the edited waypoints.json data to the Bangle.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -331,10 +331,11 @@
   { "id": "gpsnav",
     "name": "GPS Navigation",
     "icon": "icon.png",
-    "version":"0.01",
-    "description": "Displays GPS Course and Speed, + Directions to waypoint and waypoint recording",
+    "version":"0.03",
+    "description": "Displays GPS Course and Speed, + Directions to waypoint and waypoint recording, now with waypoint editor",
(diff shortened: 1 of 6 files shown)


gfwilliams:
Wow, nice - thanks!

This has come up before - it seems the 'interface' system itself works ok, but the icon/'download' name is wrong? I'm very happy to get that changed, but spectre's out of the box icons aren't that great: https://picturepan2.github.io/spectre/elements/icons.html

jeffmer:
Yes, it’s the icon. What about using the same  ‘icon-menu’ used for the customised apps? It would not be confusing as it is in a different place and both actions result in interacting with a new pane/window.

gfwilliams:
I guess... I bet people will still get confused :) What about icon-edit or icon-more-horiz?

jeffmer:
Icon-edit is good for me, however, icon-more-horiz might be more general. Of course, it’s probably better to leave it as is - people do not like change:-)

PR #429, opened 2020-05-17
Report battery status to Gadgetbridge more often

nebbishhacker:
Make gbridge widget report battery status at regular intervals and shortly after connecting, instead of just when the widget is loaded.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -95,7 +95,7 @@
   { "id": "gbridge",
     "name": "Gadgetbridge",
     "icon": "app.png",
-    "version":"0.10",
+    "version":"0.11",
     "description": "The default notification handler for Gadgetbridge notifications from Android",
     "tags": "tool,system,android,widget",
     "type":"widget",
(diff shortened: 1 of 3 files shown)


gfwilliams:
This looks great, thanks! However any idea why there's the extra `Bluetooth.println("");` in there?

nebbishhacker:
That's explained in the commit message for that change - as far as I can tell, Gadgetbridge completely ignores the first line of text sent to it after connecting (at least on my phone). Prepending an empty line to each message to each line was the simplest workaround I could think of.

nebbishhacker:
Another weird issue - after being connected to my watch for a few hours Gadgetbridge seems to stop displaying the battery status updates changes until I disconnect and reconnect. Reloading the widget doesn't help, so I don't think the problem is in my code.

gfwilliams:
Thanks - sorry, missed that.

Strange about the lack of updates. If you were able to run `adb logcat` on your PC with the phone connected you'd get a bunch of log messages and I guess you might see if Gadgetbridge was actually getting the sent data or was mentioning it had dropped it somehow

PR #432, opened 2020-05-19
Widget Visibility Widget

jeffmer:
This is the widget as discussed [here](http://forum.espruino.com/conversations/347719/#comment15290140) in the forum. To allow BTN4 and BTN5 to be returned to the widget after, for example an alert, I have exposed a setup function in addition to the usual draw function.


jeffmer:
If the setup addition to the WIDGET interface is an issue, I am happy to remove it. Its is only necessary for my ANCS widget which only really works with modified firmware and so I will not submit it until it is supported by standard firmware. Not urgent as nearly all Bangle owners use Android phones. ANCS could help sell Bangles to iPhone users:-)

BTW: how long does it take for updated apps to appear in the App loader - the gpsnav  update you accepted 5 days or so ago is not there. I do appreciate you are very busy, however, I thought it was automatic.

gfwilliams:
No, it's fine - I've just been a bit busy and didn't get around to merging.

Did you consider using the `swipe` event for left & right? That might be neater?

jeffmer:
Will try with swipe - would be more intuitive.

PR #438, opened 2020-05-21
imgclock: Make text area wider

nebbishhacker:
This prevents imgclock from cutting off the date in locales with long date
formats such as "Wednesday, September 10, 2020". Also adjusts the position
and alignment of the date when necessary to prevent it from flowing off the edge of the
screen.


gfwilliams:
Thanks!

PR #439, opened 2020-05-22
Update of widviz to use swipe rather than buttons. 

jeffmer:
Many thanks for the suggestion Gordon.
 It no longer needs the setup interface and the operation is much more intuitive.


gfwilliams:
Thanks!

PR #440, opened 2020-05-23
gpstime 0.04: Fix shown UTC time zone sign

ikkala:
Like for example at [comms.js](https://github.com/espruino/BangleApps/blob/master/js/comms.js#L156), the `getTimezoneOffset()` return value must be inverted to get correct time zone.


gfwilliams:
Great - thank you!

PR #441, opened 2020-05-23
Add github link to My Apps

FransM:
Add a link to the github page for apps in the My Apps tab

This exists in the library tab, but not in My Apps.
Found it convenient to have the link there too.

Also boyscouted and fixed indening a bit


gfwilliams:
Nice - thanks! Yes, that looks like a really good idea.

PR #444, opened 2020-05-24
RandomClockWidget even less invasive

msdeibel:
No longer uses LCD off event, only widget reloads


gfwilliams:
Thanks! This looks really neat

PR #445, opened 2020-05-24
Update Berlin Clock to version 0.3

eska-muc:
Changed:

- Shrinked the whole clock-face to avoid cut-off edges on the physical device.

New features: 

- BTN1: switch on and off displayimg time also in decimal (helps to familarize with the clock-face)
- BTN3: show date below the actual clock-face


gfwilliams:
Thanks! Just so you know, there's usually 24px at the top and sometimes 24px at the bottom used for widgets. Since you're at 20px now, if you shrunk it by another 4 pixels you could be sure it wouldn't interfere with any of the widgets?

PR #446, opened 2020-05-24
Make : between hours and minutes flash

FransM:
Flash the : separator; even seconds off, odd seconds on


gfwilliams:
Neat idea, thanks! Just so you know, usually when we make changes we bump the version number and add a comment to the ChangeLog so that other users know they can update - but given this is a one-liner I'll just stick it in anyway.

PR #447, opened 2020-05-24
Added Binary Clock

Taffaz:
A binary clock that has hours and minutes in binary. 

There is a date string below the clock.

BTN1 toggles a standard digital clock for when your binary parsing skills are lacking.


Taffaz:
Bug found so Pull request closed and will submit new one

PR #448, opened 2020-05-25
Binary Clock

Taffaz:
A binary clock that has hours and minutes in binary.

There is a date string below the clock.

BTN1 toggles a standard digital clock for when your binary parsing skills are lacking.

Previous Pull Request closed by myself due to bug.


gfwilliams:
Thanks!

Issue #45, opened 2019-11-15
Add linter and/or code formatter


ra100:
I think this could improve readability if all files were formatted the same, for example just by using `prettier`.

gfwilliams:
I'd be interested to hear other people's input on this, but my feeling is that these apps are all relatively small and they're self-contained.

I feel like the interests of being friendly to newcomers and getting a wide variety of apps probably outweighs having them all looking identical.

ra100:
After a bit of thinking you're right. If someone wants to use formatter, let them use their own style, but don't enforce it in any way.

johanbove:
We could provide an eslint plugin. In my experience eslint is one of the most used linting tools; See https://eslint.org/docs/developer-guide/working-with-plugins

gfwilliams:
Sounds like a good plan. jslint is what I currently use in the Web IDE's editor but eslint sounds preferable, and it might be possible to tweak it such that it complained about features that Espruino doesn't support (like destructuring), as well as usage of `save()` in global scope and stuff like that.

If it could be a separate repo I could hopefully look at bringing it into the Web IDE as well.

In terms of BangleApps a script that ran through and checked the JSON and JS files would be great, but I could look at that. Eventually it'll have to do some smarter stuff (checking the files referenced by apps.json, maybe even checking images are valid and the right size).

To be honest so far at least 50% of the contributions I've had haven't even tested the code on the emulator, let alone checked that BangleApps still works, so I think this isn't something that most people will actually run. However if we could set it as a Travis task at least there'd be an easy way to check if something was obviously broken before merging.

gfwilliams:
Just added 'sanitycheck.js' and enabled Travis support so it's a big improvement now. Already caught quite a few issues (icons, upload problems, etc).

It would be reasonably easy to add eslint support to the tool - it could probably get a list of globals from http://www.espruino.com/json/espruino.json (although it'd fail a lot of stuff right now because of the pre-release nature of the Bangle.js firmware)

rigrig:
> If someone wants to use formatter, let them use their own style, but don't enforce it in any way.

I use a formatter anyway, so would prefer if there was some recommended/suggested coding style.
So don't check it, but maybe put a hint somewhere in the documentation along the lines of
"Any coding style is fine, but if you are looking to pick one we recommend &lt;coding style&gt;"

gfwilliams:
Actually that would be a good one to add to espruino in general as we tend to use the same format for everything. For now:

[...]

PR #450, opened 2020-05-26
Eslint

ikkala:
Added proposal for eslint configuration:

- For apps, add eslint with baseline .eslintrc.json
  - The "globals" section is based on https://banglejs.com/reference
    - Added `g` and `WIDGETS` to there too because they seemed to be in use in code of apps
    - Instead of using "globals" for all globals, in next phase, the "env" section could define for example "espruino" and "banglejs" for getting all the globals, like proposed in https://github.com/espruino/BangleApps/issues/45#issuecomment-583274072
  - The settings in "rules" section of this baseline .eslintrc.json are selected so that current apps will pass the checks
    - People don't have to immediately go and fix multiple places, or like said [here: "I feel like the interests of being friendly to newcomers and getting a wide variety of apps probably outweighs having them all looking identical."](https://github.com/espruino/BangleApps/issues/45#issuecomment-554951498)
    - As an exception to above, activated "no-mixed-spaces-and-tabs" (implements partly the "Double spaces for indent (no tabs)" from https://github.com/espruino/BangleApps/issues/45#issuecomment-607217736 nicely) and  "no-extra-semi" because they were affecting the current apps so lightly
    - Also, for example setting `"no-undef": "warn"` (currently [commented out](https://github.com/ikkala/BangleApps/blob/593a9240d799788248e9fd4ac5893531be2ec0a1/apps/.eslintrc.json#L147)) would be soft way to nudge people to good direction 
- buffgym already had its own .eslintrc.json, so made the app to honor it:
  - Pass redraw as parameter to Workout, instead of using global variable.
  - Make code to honor eslint linebreak-style (use Windows linebreak).
  - Change eslintrc indent rule to match code style.
  - Add missing semicolons
- npm test: Run eslint after sanitycheck.js
  - Added also separate npm run target: npm run target "lint-apps"


ra100:
Looks good to me.
Just one question about semis of course :D Would it make sense to not use, to "save" some space? But I think it's not too relevant, and anyway it's safer to have semis there, maybe even better for newcomers.

ikkala:
Saving space argument could also be used in case of tab vs space indentation. But, in general, I guess the space usage should be settled by minimizing the code instead.

And in more general, I think it's good idea to use the defaults that JS community in general use -- or the defaults the standard / defacto standard tools are using. In this case, this means using eslint:recommended as much as possible.  :)

ikkala:
If this pull request can be taken in some form to master I have next on in pipe: https://github.com/ikkala/BangleApps/commit/c6fd71be0333cc317d4b7f36dccd7460f67ccac3
It finalizes the implementation of "Double spaces for indent (no tabs)" of http://www.espruino.com/Code+Style.

ra100:
Yes, I agree, go with the default. And minimizing code is a good point. :)

apps/berlinc/berlin-clock.js
```diff
@@ -88,10 +88,9 @@ Bangle.on('lcdPower', (on) => {
   g.clear();
   if (on) {
     Bangle.drawWidgets();
-    // call your app function here
-    drawBerlinClock();
-  }
-});
+  // call your app function here
+  drawBerlinClock();
```
ra100, line 92: this one looks weird, I'd expect:
```js
  if (on) {
    Bangle.drawWidgets();
    // call your app function here
    drawBerlinClock();
  }
});
```

ikkala: Good catch -- I was doing the eslint automatic fix to [earlier version of this file originally](https://github.com/ikkala/BangleApps/commit/0a32e40b1abdf221b79228a8fb36659c18fa0a00#diff-e593f7fcdc3183a0484d87c313186faa) and after that just blindly auto-merged the newer changes from the author before making todays pull request.

I fix fix this.

gfwilliams:
Nice - thanks for this. Looks great!

atjn:
I have made PR's to fix some of the rules that were disabled to support existing apps:

- #3188
- #3187
- #3186

PR #451, opened 2020-05-26
Eslint indent

ikkala:
Force indent to 2 spaces as expected at http://www.espruino.com/Code+Style


ikkala:
Note: The code changes of this pull request are done by automatic eslint `-f` option.

gfwilliams:
Thanks! Does this cause the Travis check to fail if whitespace is wrong? If so I'm not sure how useful it'll be - I can pretty much guarantee nobody will bother getting their whitespace matching before doing a PR and then it might hide real errors :)

ikkala:
> Thanks! Does this cause the Travis check to fail if whitespace is wrong? If so I'm not sure how useful it'll be - I can pretty much guarantee nobody will bother getting their whitespace matching before doing a PR and then it might hide real errors :)

Good point. The current config for indent in pr is configured for "error" -- means that if indent problem => error => exit code != 0.  Shall I change that to warning level or will do the change?

gfwilliams:
Thanks - I'll push it in. I'll also add eslint for the js in the loader itself :)

ikkala:
Excellent.

> I'll also add eslint for the js in the loader itself :)

I was about to do that but then I realized it will need a bit different configuration.. and selecting proper config for it was not that trivial to me.  :)  You obviously know what you want from that.

gfwilliams:
> You obviously know what you want from that.

Honestly, not really. I'm not too bothered about the look but anything that'll pick up common pitfalls is good :)

ikkala:
> anything that'll pick up common pitfalls is good :)

Then, I suggest making the variable definition mandatory (or perhaps warn level in Apps to give mercy to people (`"no-undef": "warn"`), but error in App Loader?), and also `"no-redeclare": "warn"`. 

Consider also `"no-var": "warn"`-- people usually don't realize `var` scopes correctly -- using `let` and `const` is more natural for most.

PR #453, opened 2020-05-26
Require correct module in apploader.js  and fix typo in appinfo.js

detached:
Thank you for the apploader.js. That is really helpful!
For being able to use the apploader.js I had to change some code.

After installing noble with `npm install -g '@abandonware/noble'` I had to require '@abandonware/noble' instead of just 'noble'.

I can't test the code in a apple browseri, but I think the `ABuffer` is a typo and should be `Buffer`.

### DIFF
--- a/bin/apploader.js
+++ b/bin/apploader.js
@@ -15,6 +15,7 @@ var SETTINGS = {
 };
 var Utils = require("../js/utils.js");
 var AppInfo = require("../js/appinfo.js");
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! That's great.

Oops -  BufferA was an accidental commit - basically I couldn't get any debugging working with the iPad so I had to manually rename every Buffer instance to figure out where the issue was.

Issue #433, opened 2020-05-20
Alarms don't work


m0n5t3r:
Looks like the alarm works maybe once or twice, then in subsequent days they don't trigger any more... until I open the settings menu (I thought maybe I was sleeping too hard, but in the last 3 days I woke up before the alarm time, and there was no alarm...)

Where should I look in order to debug this further?

gfwilliams:
Hi - have you ever updated the firmware on your Bangle.js? It's possible it is related to a bug in the original firmware which could occasionally stop a file being changed.

To debug though, I'd look at `alarm.json` (you can view it using the Storage menu in the IDE). Hopefully you should be able to see it changing. There's:

* `hr` - the hour/minutes it should fire on
* `last` - the number of the last day the alarm fired on to avoid duplicates
* `on` - whether the alarm is enabled

I guess it's possible that the specific clock face you have is doing something `clearInterval()/etc` which may stop the alarm from being kept scheduled?

m0n5t3r:
Last I checked, it's the latest stable version (2v05); I don't see any `clearInterval` in https://github.com/espruino/BangleApps/tree/master/apps/sclock

The alarm isn't lost, btw, it shows up in alarms and gets triggered when I open the ~settings menu~ launcher (and I just looked in the IDE, alarm.json is there).

Everything else is up to date (I updated the bootloader now, it was at 0.16)

Going to try connecting with the IDE before the alarm is supposed to trigger tomorrow morning, just in case something gets printed.

gfwilliams:
You could also try running `eval(require("Storage").read("alarm.boot.js"))` if the alarm doesn't fire on time, and see if that does anything?

m0n5t3r:
only "undefined", which I guess is expected (looking at the code, the function doesn't return anything)

gfwilliams:
Yes - but when the alarm is due to fire it *should* show the alarm on the watch?

m0n5t3r:
ah, ok, I missed the last part of your message; if I understand correctly, unless there is an active alarm the app resets the whole thing every 100ms, triggering the bootloader? that seems a bit ... excessive, wouldn't a `setInterval` be more ... power efficient?

gfwilliams:
Not sure I understand? https://github.com/espruino/BangleApps/blob/master/apps/alarm/boot.js is what runs at startup. It uses `setTimeout` to schedule loading of the alarm when it's time for it to go off.

[...]

PR #455, opened 2020-05-27
Make the alarm scheduling more reliable

m0n5t3r:
Fixes https://github.com/espruino/BangleApps/issues/433: instead of only selecting the alarms that didn't already run during the current day for scheduling, get all active alarms, sort accordingly and add 24h to the timer if the alarm already ran.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -137,7 +137,7 @@
     "name": "Default Alarm",
     "shortName":"Alarms",
     "icon": "app.png",
-    "version":"0.07",
+    "version":"0.08",
     "description": "Set and respond to alarms",
     "tags": "tool,alarm,widget",
     "storage": [
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -5,3 +5,4 @@
 0.05: Add alarm.boot.js and move code from the bootloader
 0.06: Change 'New Alarm' to 'Save', allow Deletion of Alarms
 0.07: Don't overwrite existing settings on app update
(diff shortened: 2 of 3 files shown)


m0n5t3r:
should I do the version bump in here, or is it handled somewhere else / automatically?

gfwilliams:
Great, thanks! So the issue was when scheduling an alarm for tomorrow, today?

If you could bump the version in apps.json and add a note to ChangeLog that'd be great: https://www.espruino.com/Bangle.js+App+Loader#making-changes

I'm afraid it's not handled automatically :)

m0n5t3r:
done; would an auto-snooze function be considered useful? (as in, if there is no reaction to the first round of buzzing, schedule another round in 10 minutes...)

m0n5t3r:
The issue was that when you set up an alarm for the first time, `.last` is 0, the bootloader code would select it and schedule it somewhat correctly (`t` would be negative, get reset to 1s, so it would check every second until midnight, when it would finally schedule the run for the alarm time - I just pushed a commit to address this as well)

However, after the alarm runs and you dismiss it, `.last` becomes equal to the current day, and the bootloader code would just skip it and not schedule the next run unless something else happened to trigger a `load()` after midnight (say, restless user fidgetting with the watch instead of sleeping? :grin: )

gfwilliams:
Great, thanks! I'm not so sure about the auto-snooze. We could add it but I think it'd be good as an option rather than the default...

PR #459, opened 2020-05-28
New app SleepPhaseAlarm

nxdefiant:
Uses the accelerometer to estimate sleep and wake states with the principle of ([Estimation of Stationary Sleep-segments (ESS)](https://ubicomp.eti.uni-siegen.de/home/datasets/ichi14/index.html.en)). This app will read the next alarm from the alarm application and will wake you up to 30 minutes early at the best guessed time on first detected movement.

When SleepPhaseAlarm starts it searches for the next enabled alarm in alarm.json from the default alarm app. Beginning from 30 minutes before the alarm time it waits for a wake phase using the ESS algorithm to trigger an alarm.

Afterwards the default alarm app does trigger the scheduled alarm as usual.


gfwilliams:
Nice - thanks!

PR #461, opened 2020-05-28
Simpletimer: remember last set time, make BTN2 work

rigrig:
- Fix BTN2 to really open the launcher
- Remember last set time

I figured the fix for BTN2 doesn't need a separate version/changelog.


gfwilliams:
Thanks!

PR #462, opened 2020-05-28
gbridge: Add setting to show/hide icon

rigrig:
I noticed #131 also has this, but it looks like the notification overhaul stuff might take a while to do properly, and the icon was starting to bug me.

It's doing
```js
    Bangle.loadWidgets();
    Bangle.drawWidgets();
```
To make the show/hide take effect inside the settings, but that really eats RAM somehow.
I'll see if I can figure out why, and/or come up with a better way to show/hide the widget.
(maybe make hidden = width:0)

### DIFF
--- a/apps/gbridge/widget.js
+++ b/apps/gbridge/widget.js
@@ -1,7 +1,4 @@
 (() => {
-  const storage = require('Storage');
-  let settings;
-
   const state = {
     music: "stop",
 
@@ -14,6 +11,13 @@
     scrollPos: 0
   };
 
+  function settings() {
+    let settings = require('Storage').readJSON("gbridge.json", true) || {};
+    if (!("showIcon" in settings)) {
+      settings.showIcon = true;
+    }
+    return settings
+  }
 
   function gbSend(message) {
     Bluetooth.println("");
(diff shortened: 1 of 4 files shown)


rigrig:
I just realized reloading widgets eating up RAM is probably because they set up a bunch of event handlers, which keep hanging on to their callback even if the widget is reloaded.

rigrig:
Changed it to just reload the gbridge widget and redraw

apps/gbridge/widget.js
```diff
@@ -192,10 +196,24 @@
...
-  NRF.on("connect", changedConnectionState);
-  NRF.on("disconnect", changedConnectionState);
+  function reload() {
+    settings = storage.readJSON('gbridge.json', 1) || {};
+    if (!("showIcon" in settings)) {
+      settings.showIcon = true;
+    }
+    if (settings.showIcon) {
+      WIDGETS["gbridgew"].width = 24;
+      NRF.on("connect", changedConnectionState);
```
gfwilliams, line 206: If you happen to call 'reload' more than once without changing `settings.showIcon`, I think this'll just queue up multiple event listeners? Maybe add `NRF.removeListener` first to run in all cases?

apps/gbridge/widget.js
```diff
@@ -1,4 +1,6 @@
 (() => {
+  const storage = require('Storage');
+  let settings;
```
gfwilliams, line 3: Any chance of just storing the one setting you're interested in? By storing all settings here you're wasting quite a lot of memory for all apps, just to read the value of one boolean.

gfwilliams:
Thanks! Just added some suggestions though... I think especially the 'settings' is something we really ought to change as it could eat up a lot of memory

rigrig:
Changed to always clear listeners before re-adding them, and not keeping the setting in memory

gfwilliams:
Great - thanks!

PR #466, opened 2020-05-30
Stop the cliock face redrawing and running out of memory

hasmar04:
Removes the clear screen and widget loading from each wake so it only runs the first time. Prevents the watch from running out of memory due to redraws after approx. 3 wakes.


gfwilliams:
Thanks - that's great!

Just in future, please could you also bump the version in `apps.json`? I'll do that now though

PR #468, opened 2020-05-30
set to null to free memory

matjaz:


MaBecker:
did you tested the difference between `delete var` and  `var = null``

My understanding is that delete is the better way.

* delete removes the var 
* with = null the var still exist

matjaz:
that's not how JavaScript works.
see https://github.com/espruino/BangleApps/commit/f8ef03bade25ec8412691df6e96cd1ed99ade5c9#r39555100

gfwilliams:
Sorry - please can you explain in more detail, with some test results? Even in the emulator:

http://www.espruino.com/ide/emulator.html

As far as I'm concerned `delete X` is the correct way of doing it because it actually removed the variable name. Worse than that `X=null` actually keeps the variable name *AND* anothe rvariable for `null`.

```
>a = "Hello world"
="Hello world"
>process.memory().usage
=47
>delete a
=true
>process.memory().usage
=44
>a = "Hello world"
="Hello world"
>process.memory().usage
=47
>a = null
=null
>process.memory().usage
=46
```

gfwilliams:
Just to be super clear on this, you referenced: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Description

> Unlike what common belief suggests (perhaps due to other programming languages like delete in C++), the delete operator has nothing to do with directly freeing memory. Memory management is done indirectly via breaking references.

This is true. However MDN then says:

> The delete operator removes a given property from an object. 

It doesn't say explicitly here, but what's happening is `delete` is breaking a reference to the given object and so is causing memory to be freed - it's just not freeing the memory *itself*.

matjaz:
delete removes **property** from an object. But here we're talking about object/variable itself, so I would assume it's implementation specific.
I would prefer to find a way to remove this variables and free up memory.

Also in general I would prefer as less as possible global(per app) variables. So something like `init()/start()` function convention, wrapping variables in local scope which is destroyed once app is started. Not sure how does espruino/bangle handles loading/unloading apps.

matjaz:
MDN also says this.

> Any property declared with var cannot be deleted from the global scope or from a function's scope.
> As such, delete cannot delete any functions in the global scope (whether this is part from a function definition or a function expression).

gfwilliams:
> so I would assume it's implementation specific

Yes - if you do it in node.js it fails: `a="Hello";delete a` 

But if you do the same in Chrome *or* firefox then it works fine.

While `var a="Hello";delete a` doesn't work in firefox, it does in chrome.

> MDN: Any property declared with var cannot be deleted from the global scope

That's somewhat surprising because as above it seems implementation specific.

> MDN: delete cannot delete any functions in the global scope

fair enough - but we're not deleting functions.

> I would prefer to find a way to remove this variables and free up memory.

As I showed above, the original code is working just fine and is freeing memory.

> I would prefer as less as possible global(per app) variables.

Yes, however there's a balance between using the scarce memory efficiently, and having the code being 'nice'. The bootloader you wanted to modify is a great example - it's written that way to be fast and efficient with memory, it's not meant to look 'nice'.

However:

* Widgets do keep everything in their own scope
* Apps don't bother, because there's no point (see below).

> Not sure how does espruino/bangle handles loading/unloading apps.

The interpreter basically resets itself. While it'd be nice to be able to 'unload' apps, since they can hook onto Bluetooth, IO, sensors, etc it's just asking for trouble to get devs to write code in such a way that their app 'unloads' itself - Bangle.js would end up being super unstable.

matjaz:
It look's like it's different. In chrome this fails: 
```
var a="Hello";delete a;
> false
```

Also see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#Non-configurable_properties

> When a property is marked as non-configurable, delete won't have any effect, and will return false.
> var, let, and const create non-configurable properties that cannot be deleted with the delete operator

Also see strict mode. Is strict supported in espruino?

gfwilliams:
> Is strict supported in espruino?

No.

PR #471, opened 2020-06-01
Add auto-snooze option to alarms

m0n5t3r:
What the title says: this adds an "auto-snooze" option to the alarm editor which, if enabled, schedules a new round of buzzing 10 minutes later as if the snooze button was clicked.

I also added a behaviour fix for snoozing: the way it's implemented currently, snoozing adds 10 minutes to the alarm time and saves it; this means that, for repeating alarms, the next run will be 10 minutes * number of snoozes later, which is not something the user would expect.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -140,7 +140,7 @@
     "name": "Default Alarm",
     "shortName":"Alarms",
     "icon": "app.png",
-    "version":"0.08",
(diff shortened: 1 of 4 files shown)


gfwilliams:
This looks great - thanks!

PR #473, opened 2020-06-02
Fix position

bengwalker:
In combination with some widgets, the position of the bmp string was wrong on my watch. I fixed it.


gfwilliams:
Thanks! I think this might be due to the metronome not calling `g.reset()` before drawing to reset graphics state (so in your case, if you removed the widgets it'd break again).

I'll merge this and add `.reset()` and `setFontAlign` in at the start as well, as it's much better having it center aligned as you're doing.

PR #474, opened 2020-06-02
locale fixes + memory

matjaz:
Feel free to merge or cherry pick.


gfwilliams:
Thanks, this looks great - just fyi in future I wouldn't be too worried about replacing `substr` in code for Espruino - it's not going anywhere in Espruino any time soon.

PR #475, opened 2020-06-02
added Back translation

matjaz:


gfwilliams:
Thanks - has this actually been tested? It seems like in most cases the text is `> Back` and so the translation as-is may not work?

matjaz:
yeah. The idea is to explicitly call `locale.translate()`.
Actually, what if `E.showMenu()` would call it implicitly? and an option so app can extend locale.trans.

gfwilliams:
> Actually, what if E.showMenu() would call it implicitly?

That's what it does (and showMessage/etc). That's the whole point really - that apps don't generally need to be written with it in mind: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showMenu.js#L57

gfwilliams:
So was this tested at all? Literally the *only* app I see the string `Back` in is the GPS recorder. None of the others will work

matjaz:
no. I'll open new PR if you don't mind.

PR #476, opened 2020-06-03
Fix for TZ in Western hemisphere

chukharev:
In timezones in the Western hemisphere (UTC-1 etc.), the local year at UNIX epoch is 1969, not 1970. Therefore, instead of checking getFullYear()==1970, we should be checking getFullYear()<=1970 (I did < 2000 consistent with the existing code below on line 24).


gfwilliams:
Great - thanks!

PR #478, opened 2020-06-03
back translation

matjaz:
tested and working.


gfwilliams:
Great - thanks for these!

I'll change 'gpsrec' as well so it'll use the `< Back`

gfwilliams:
Hmm - actually it's not quite that easy for gpsrec as it's just using drawString, not a menu. I'll leave that one for later :)

matjaz:
what about `locale.translate('< Back')`

gfwilliams:
Yeah, but it's still got the arrow when that's not required. I guess `locale.translate('< Back').slice(2)` is an option, even though it's a bit of a hack

PR #481, opened 2020-06-04
en_AU updated to values metric and corrected date

grsutcliffe:
local "en_AU" had some incorrect values, Australia is fully metric, and date is %d/%m/%y.


gfwilliams:
Great - thanks!

PR #482, opened 2020-06-04
update to life app

jeffmer:
Fixed some start up issues
Changed period to 65ms - works better at start
Removed timing information as now always reads 1040ms
Added button labels for "launch" and "clock"


gfwilliams:
Thanks! Please could you change the name of `reset` though? It's a built-in function and overriding it may cause some issues uploading code since the watch will no longer reset :)

Just a thought about the timing info - you could just do `start=getTime()....;timetaken += getTime()-start` around your row calculation code if you still want it?

jeffmer:
OK, no problem, will get to it this evening.

jeffmer:
reset now called regen and in fact there was a bug, it seems that Espruino does not have default parameters as in `startdraw(init=false)` - it is easily simulated of course using undefined. I did not see it before as I only tested the minimised version which removed it:)

gfwilliams:
Interesting - thanks! I wonder if 'eslint' could be told to complain about default args so that the travis checks pick it up?

PR #487, opened 2020-06-06
locale improvements

matjaz:
- small number rouding to 1 dec. point
- number format
- dow, month single var


gfwilliams:
Thanks!

Issue #483, opened 2020-06-04
weather: Last update blinks


pasniak:
"weather received..." msg blinks (maybe it is refreshed too often?)

PR #492, opened 2020-06-09
Weather: Fix flickering last updated time

nebbishhacker:
Fixes https://github.com/espruino/BangleApps/issues/483


gfwilliams:
Thanks!

PR #493, opened 2020-06-09
magnav - Navigational Compass

jeffmer:
Compensates for tilt and roll and gives a surprisingly stable heading as long as calibration is diligent. The standard compass could easily be updated with the compensation routine however, I am not sure if the calibration data would be sufficiently robust to let it work.

One possibility is a calibration app as even though the data probably differs a lot between Bangles, it seems to remain stable for at least a few days. 

An improvement in this app would be some way of giving the users some feedback on the quality of calibration and even a buzz when sufficient waving about had been done:)


gfwilliams:
That's great - thanks! Really nice to have the calibration data saved.

PR #496, opened 2020-06-10
load only clock apps

matjaz:
Haven't done benchmarks, how internally Espruino works.

### DIFF
--- a/apps/boot/bootloader.js
+++ b/apps/boot/bootloader.js
@@ -2,7 +2,15 @@
 var clockApp=(require("Storage").readJSON("setting.json",1)||{}).clock;
 if (clockApp) clockApp = require("Storage").read(clockApp);
 if (!clockApp) {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - this looks great. I'd moved to doing this in other places but it seems I didn't spot it here

matjaz:
Maybe sth like

    clockApp = require("Storage").listAndRead(/\.info$/, file => {
        const app = require("Storage").readJSON(file,1);
        if (app && app.type == "clock") {
            return app;
        }
        // returns undefined, so it skips this file
    })
    .sort((a, b) => a.sortorder - b.sortorder)[0];

gfwilliams:
That's an idea - so without even allocating the array of filenames.

Although I guess since we're filtering with a regex anyway the list size will only ever be the amount of apps, which is probably <100?

matjaz:
yeah, array of all names, and more importantly contents of all files at the same in memory.

yeah, not sure if we could have more than 100 useful apps. maybe widgets/workers.

PR #500, opened 2020-06-13
Added Mixed Clock 2 fork of original mixed clock

nxdefiant:
Differences to original Mixed Clock:
- Color is all white for better readability in the bright sunlight
- Thicker clock hands for better readability in the bright sunlight
- Extra space under the clock for widgets
- Seconds in the digital clock


gfwilliams:
Great - thanks!

PR #507, opened 2020-06-17
Adds a one-button-tracker widget

gedankenstuecke:
The `1button` widget turns `BTN1` into a simple event tracker that can be used from the clock face and from basically everywhere where widgets can run. 

When pressed for at least 130 ms it will trigger the recording of when the button was pressed and when it was released, the timestamps of which are written into a CSV file that can be downloaded from the app loader. 

Many members in the _Quantified Self_ community are looking for simple ways of tracking events and this makes the Bangle a great device for doing this! 🙂


gfwilliams:
This is great - thank you!

PR #510, opened 2020-06-22
Added gpsautotime widget

nxdefiant:
A widget that automatically updates the Bangle.js time to the GPS time whenever there is a valid GPS fix. No need to press any extra buttons.


gfwilliams:
Nice - thanks!

PR #511, opened 2020-06-23
BugFix: Marioclock use of fillPoly adjustments

paulcockrell:
**Problem:**

Pyramids float above ground. Possibly due to changes to fillPoly function

**Solution:**

Adjust Pyramid Y values


gfwilliams:
Thanks!

PR #512, opened 2020-06-23
New MultiClock app

jeffmer:
As described in the forum [http://forum.espruino.com/conversations/349659/](http://forum.espruino.com/conversations/349659/).


gfwilliams:
Great - thanks!

PR #513, opened 2020-06-26
iPhone Notification Widget

jeffmer:
The widget now works reasonably well with the standard firmware and since it will work with **multiclock**, people can now try it before upgrading to the modified firmware if they need to improve connect times and range. Consequently, I think it is worth having in the main repository.


gfwilliams:
This is awesome - thank you!

PR #517, opened 2020-06-28
Simple Vector Clock

tebarius:
New clock.
(modification of the Simple Clock App to use and work with Vectorfont)


gfwilliams:
Great - thanks!

PR #518, opened 2020-06-30
gbridge: Don't keep LCD on while playing music

rigrig:
Looks like this fell through during the 'notify' upgrade.
Some apps can keep the LCD pretty much awake all the time with "position" updates. And even if they don't, turning on the LCD for every next song also isn't great for battery life.


gfwilliams:
Cool, thanks!

PR #519, opened 2020-07-02
Feat/added en_IN locale

gsin11:
Added new locale option for English India.


gfwilliams:
Thanks! Does this work for you?

I'm pretty sure ₹ isn't in the ISO8859 character set on the Bangle, so I'd have expected the locale loader to report an error to the dev console when it loads (and the currency symbol not to show)

gsin11:
> Thanks! Does this work for you?
> 
> I'm pretty sure ₹ isn't in the ISO8859 character set on the Bangle, so I'd have expected the locale loader to report an error to the dev console when it loads (and the currency symbol not to show)

I think you are right, I would then change it to previous symbol which was "Rs." which is still valid.

gfwilliams:
Thanks!

PR #520, opened 2020-07-04
Gpsrec OpenStreetMap track plotting

berkenbu:
I added an option to plot gpsrec tracks on top of OpenStreetMap tiles. 
The option only occurs when OpenStreetMap tiles are installed via the eponymous app. The tiles are centered on the 'center of gravity' of the track, and the track is scaled to the same scale as the OpenStreetMap tiles. The change is mostly a cut-and-paste job of the function drawopenstmap from the OpenStreetMap app.
Also includes a small change to change the color of the "gps signal acquired" icon in the widget from red to green - I found green to be much more visible in bright daylight than red.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -344,7 +344,7 @@
   { "id": "gpsrec",
     "name": "GPS Recorder",
     "icon": "app.png",
-    "version":"0.11",
+    "version":"0.12",
     "interface": "interface.html",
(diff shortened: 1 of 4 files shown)


gfwilliams:
This looks great, thanks!

I guess long term we should try and turn the tile (and maybe track?) rendering into some kind of module so it can be pulled into other apps.

PR #522, opened 2020-07-07
Gbridge

rigrig:
* notify/notifyfs: Add notification ID option, to only hide a specific notification 
* gbridge: Handle "notify-" events
* gbridge: Nicer display of alarm clock notifications

Should we worry about dependency versions at all? Or just assume that people will always just update everything?
In this case: if you only update Gadgetbridge but not `notify`/`notifyjs`, the notification `id` option is ignored, so nothing breaks.


gfwilliams:
This looks great - thanks!

Yes, I think we can assume that someone will just 'update all' - at least for the moment :)

PR #525, opened 2020-07-13
minionclk v0.03

singintime:
Fixing and improving the rendering under Espruino v2.06:

- Using bitmap font instead of vector one for hours and minutes
- Avoid rendering on top 24px, to leave space for widgets
- Increased background image resolution

screenshot (4)


gfwilliams:
Thanks! Do you think you'd be able to tweak the way you store the minion image back to the way it was before though? 

https://www.espruino.com/Code+Style#other-suggestions

I know it's a bit counter-intuitive, but it will save a bunch of RAM if you're able to define it in a function.

I guess either that or you could actually make it a file that you save to `Storage`? If you upload the original here then I could do the relevant tweaks for you?

singintime:
I created a getBackground() function, whose only scope is to return the image blob. Would it do the saving RAM trick?

gfwilliams:
Yes, that's perfect - thanks!

PR #526, opened 2020-07-14
notifyfs: fix custom render callback

rigrig:
Now draws properly in the middle of the screen, instead of outside it.


gfwilliams:
Thanks!

PR #530, opened 2020-07-19
Notify api

### LINKED ISSUE #527: Notify API improvements
I'm wondering if it might be a good idea to change the notify API a bit:
Turn `options.size` into a "suggested size", and pass the actual available area to `render`.

That way `render` could use e.g. a larger font for fullscreen notifications.

My idea would be that
- `area` starts out as the whole notification
- drawing `title`,`src` and/or `icon` shrinks `area`
- `body` text (if present) is drawn inside `area`
- `render` (if present) is called with `area`

So you *could* pass both `body` and `render` to e.g. draw a border around the text, but if you want to get fancy you need to render it yourself.

Or a simpler alternative: just an optional additional boolean argument passed to `render`, which indicates it *could* ignore the `y` parameter and instead draw to the whole screen.

rigrig:
Fixes #527 

(Having so much duplicate code in both `notify` and `notifyjs` really bugs me, but I guess it can't be helped.)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -93,7 +93,7 @@
     "name": "Fullscreen Notifications",
     "shortName":"Notifications",
     "icon": "notify.png",
-    "version":"0.03",
+    "version":"0.04",
     "description": "A handler for displaying notifications that displays them fullscreen. This may not fully restore the screen after on some apps. See `Notifications (default)` for more information about the notifications library.",
     "tags": "widget",
     "type": "notify",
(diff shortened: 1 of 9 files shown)


gfwilliams:
Nice, thanks! Yeah, the duplication is a pain but at least it's not on the watch itself as you have one or the other installed :)

I hope realistically it just means it should be easier for folk to create custom notification methods though.

Issue #239, opened 2020-04-06
Low on battery warning


veelco:
Would be awesome to have a warning if my battery is running low. It just shuts down right now. I'm using the Tiny Battery Widget currently which is some sort of an indication, but I like to have a pop-up message/vibration at 5% so I can recharge my watch.

Purple-Tentacle:
@gfwilliams 
That should be part of the discussed best practices documentation.
Every battery widget should use that imo.

gfwilliams:
What do you mean? Every widget should buzz for low battery?

Purple-Tentacle:
Every battery widget should.
Or not. 
But all should behave in the same way imo.

MaBecker:
There is a folder for snippets like the one you are talking about

https://github.com/espruino/BangleApps/tree/master/testing

To be honest: All apps and widgets are designed by individual persons in the way they like them to look and work. I suggest make a copy of tiny battery widget and change the new widget the way you like it.

 **Building Apps and Widgets the way YOU like them is the one of the biggest benefit you get with Bangle.js**

Purple-Tentacle:
> Building Apps and Widgets the way YOU like them is the one of the biggest benefit you get with Bangle.js
Agreed. And I am already doing such things in creating stuff I want and changing stuff the way I want it.
I am just thinking of a streamlined experience with all apps in the "official" app loader.
And that would mean that some things should be consistent in all apps.

In the end, I am not the one making the decision, I am just proposing.

gfwilliams:
Thanks - however in this case I'm with @MaBecker. I feel like the whole point of this is to put as few restrictions on what can be done as possible so the watch can be customised as much as needed.

I can imagine what might happen in a lot of cases is we end up with one 'basic' version of an app/widget, and one 'kitchen sink' one that has everything and a bunch of settings?

I'm not 100% sold on the low battery warning - for instance if it happens overnight it's just going to annoy a bunch of people. I could totally imagine something that catches your attention when you look at the watch though, like a flashing red X?

Purple-Tentacle:
>Thanks - however in this case I'm with @MaBecker. I feel like the whole point of this is to put as few restrictions on what can be done as possible so the watch can be customised as much as needed.

Ok, fair enough :)

>I'm not 100% sold on the low battery warning - for instance if it happens overnight it's just going to annoy a bunch of people

[...]

PR #531, opened 2020-07-20
Battery warning

rigrig:
Battery Warning, as discussed in #239

screenshot

## Settings:

* Percentage: Warn when battery reaches this level
* Buzz: Enable/disable buzzer

## Notes:
It never draws an actual widget, so I did't add the "widget" tag (because I figured then people would expect something to show up)


gfwilliams:
Looks great - thanks!

PR #532, opened 2020-07-21
Wait until GPS time is available

jjok:
Hi. I've been finding my Bangle switching off before I've had a chance to recharge it. The annoying thing is it's quite difficult to pick up the correct time, as it only checks the first GPS response, which is invariably `null`.
This change makes it wait until it gets a proper GPS response.

What do you think?

### DIFF
--- a/apps/boot/bootloader.js
+++ b/apps/boot/bootloader.js
@@ -20,24 +20,35 @@ setWatch(() => {
 }, BTN2, {repeat:false,edge:"falling"});)
 `;
 // check to see if our clock is wrong - if it is use GPS time
-if ((new Date()).getFullYear()<2000) {
-  E.showMessage("Searching for\nGPS time");
-  Bangle.on("GPS",function cb(g) {
-    Bangle.setGPSPower(0);
-    Bangle.removeListener("GPS",cb);
-    if (!g.time || (g.time.getFullYear()<2000) ||
-       (g.time.getFullYear()>2200)) {
-      // GPS receiver's time not set - just boot clock anyway
-      eval(clockApp);
-      delete clockApp;
-      return;
-    }
+function cancelTimeSearch() {
+  Bangle.setGPSPower(0);
+  Bangle.removeListener("GPS",cb);
+}
+function continueToClock() {
+  eval(clockApp);
+  delete clockApp;
+}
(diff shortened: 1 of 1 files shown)


gfwilliams:
I don't think this is something I'd want to include by default as it'll just annoy people who don't have a way to get a GPS fix (since it will happen every time the user tries to switch apps). Also I'd be pretty sure the implementation will leave a few functions allocated that'll use up RAM.

However what about making this a widget? There's already the "GPS Auto Time" widget (https://github.com/espruino/BangleApps/blob/master/apps/gpsautotime/widget.js) and you could either create a new one or add a setting to it which makes it turn the GPS on to get a fix if the time isn't set - nice thing with that is it'd happen in the background and your Bangle would still be usable while it was trying to get a fix...

jjok:
Thanks Gordon. This was just an idea, really, for a way of forcing it to pick up the time, which I didn't think existed already. Definitely open to any other ideas.

> the implementation will leave a few functions allocated that'll use up RAM.

Is there any more info about this anywhere? It would be nice if there was a Best Practices or something to explain that you need to be aware of RAM useage, when things are kept in RAM, and how to avoid that.

> However what about making this a widget?

Yep. I think I'll have a go at that.

> you could either create a new one or add a setting

Do you/the community have a preference here? Bigger apps/widgets with options to enable different functionality vs lots of small ones that each do one thing? Is there a technical benefit to either? Would one use more RAM, for example?

Thanks.

gfwilliams:
There's http://www.espruino.com/Code+Style and http://www.espruino.com/Bangle.js+Guidelines - but the bootloader is a bit of a special case. Basically it's executed from flash so you're fine there, but if you define a variable or function in there and done delete it then it's going to stay around even after the clock loads, which is probably not what you want.

> Do you/the community have a preference here?

Personally I think if you can have a nice small single-purpose widget then that's definitely better than settings. Less RAM/memory, and actually easier for people to find in the app loader if they search.

If what you're doing is going to duplicate *loads* of code into a new app then I think it's worth using settings, but in this case it's only 10 lines or so.

jjok:
I think GPS Time is exactly what I want.

https://github.com/espruino/BangleApps/tree/master/apps/gpstime

PR #535, opened 2020-07-25
Add is12Hour setting to cliock

brodyd795:
#### Description

The current CLI clock doesn't respect the 12hr vs 24hr setting in Settings and instead always renders the 24hr clock. This change makes the app aware of that setting and uses whichever the user has turned on.


brodyd795:
DO NOT REVIEW. 

Potential bug identified. Still investigating.

gfwilliams:
Did  you close this intentionally? Seems like you fixed the original issue?

brodyd795:
Yes, I intentionally closed it because of a weird state of my fork and the known bug(s). I'll just resubmit a new PR when I have a fix ready.

PR #542, opened 2020-08-06
BLE GATT Battery Service

jjok:
This is the battery service widget that we talked about here http://forum.espruino.com/conversations/351959/.

There are probably a few tweaks to be made.


gfwilliams:
Thanks! Looks great!

I guess it doesn't *need* to be a widget (could just be boot code) but at least this way you can extend it to display data if you want to later

apps/widgattbat/widget.js
```diff
@@ -0,0 +1,16 @@
+(() => {
+  function advertiseBattery() {
+    Bangle.bleAdvert[0x180F] = [E.getBattery()];
+    NRF.setAdvertising(Bangle.bleAdvert);
+  }
+
+  if (!Bangle.bleAdvert) Bangle.bleAdvert = {};
+  setInterval(advertiseBattery, 60 * 1000);
+  advertiseBattery();
+
```
jjok, line 10: Does it need to declare this stuff if it isn't drawing anything to the screen?

apps/widgattbat/widget.js
```diff
@@ -0,0 +1,16 @@
...
+  WIDGETS["gattbat"]={
+    area:"tl",
+    width:0,
+    draw:function() {}
+  };
```
jjok, line 15: Does it need to declare this stuff if it isn't drawing anything to the screen?

jjok:
> I guess it doesn't need to be a widget (could just be boot code)

Where would it live if it was boot code? Would it still be something people can choose to install, or would it always run for everyone?

jjok:
FYI This is what it looks like in Home Assistant.

image

apps/widgattbat/widget.js
gfwilliams: Nope, not at all... Actually that's probably the easiest option.

gfwilliams:
That's great! I'm going to have to give home assistant a go :)

> Where would it live if it was boot code?

Basically the same as right now: still in apps.json, but just `myapp.boot.js` rather than `myapp.widget.js`

However if you remove the `WIDGETS["gattbat"]` as you suggest then the difference will be small (it'll just get loaded regardless of whether the app in question requests widgets)

jjok:
I decided to change it to boot code. Should the type still be "widget", or should it be "app", or something else?

gfwilliams:
I'd make it `type:"bootloader"` - thanks!

jjok:
Should I rebase to fix the conflict? I know that can make the "Files changed" diff go a bit crazy once the PR has been opened.

gfwilliams:
No, that's fine - I just did the merge here. I'll just wait for the CI check to complete and then I'll merge this :)

gfwilliams:
Thanks! All sorted - it'll be live on the development apps repo now, and hopefully banglejs.com/apps when I push an update in a day or two

PR #546, opened 2020-08-13
ViewSTL app

berkenbu:
This is a first version of an STL file viewer. The code makes use of a few inlined C routines, therefore both a version with a base64 encoded binary blob (viewstl.min.js) and the full source (viewstl.app.js) are provided. I am too much of a JS/HTML amateur to figure out how to make a nice interface for uploading STL files, I am using the web IDE upload file functionality instead.


gfwilliams:
This is awesome - thanks!

gfwilliams:
... I wonder how long before someone does a raycasted game with inline C :)

PR #550, opened 2020-08-22
Add accelerometer/magnetometer rotation mode

berkenbu:
Add a third rotation mode: rotation follows accelerometer and magnetometer readings.


gfwilliams:
Nice, thanks!

PR #551, opened 2020-08-23
New app: World Clock

computermacgyver:
A fork of sclock that removes the year and UTC offset and replaces these with up to four customizable clocks (e.g., to see the time in four other locations in the world)


gfwilliams:
Thanks - that's a great idea!

PR #552, opened 2020-08-25
Fix great circle formula in distance calculation

berkenbu:
The distance to the next way point computation was not quite right, more markedly so over large distances (as noticed on a long road trip). The submitted fix can exhibit machine precision issues on short distances with 32 floats, but the 64 bit floats used by Espruino seem to do just fine when tested.


gfwilliams:
Nice - thanks! I'm not actually sure why I never used `Bangle.project` to compare the projected distances in the first place, but that looks great!

PR #553, opened 2020-08-31
banglerun v0.05

singintime:
screenshot (5)

This PR is a complete rewriting of the BangleRun app in TypeScript. Code has been separated into modules, to better handle the various features in the app. The main changes are:
- New layout of the displayed information on screen
- Kalman filtering on GPS and HRM input. The former fuses together position and speed readings from the GPS, significantly improving the tracking accuracy. I did two 10k runs, comparing results with two different smartphone apps (Huawei Health and Sports Tracker), and obtaining readings different by less than 1%
- Logging of the activities on CSV files

I did some practical tests on somewhat long runs (distance > 10k and time > 1hr), to ensure that old bugs affecting the display of time and distance are not present in this release. The project is already configured to support unit testing with Jasmine, and I plan to write some specs in the near future.


singintime:
I've got couple of questions about some minor issues that I couldn't solve:
1. I receive a lot of FIFO_FULL errors when pressing buttons and reading NMEA strings from the `GPS-raw` event. It doesn't seem to cause any malfunctioning in the app, but I was wondering if there's something I could do to avoid them
1. The GPS seems to take ages to lock, especially if compared to other bangle.js apps (e.g. gpstime). Is it just an impression of mine? Am I doing something wrong? Is there something I could do to improve the situation?
1. Is there a way to have a look at the CPU load of the watch, to make sure the app is not too computationally expensive?

gfwilliams:
> I receive a lot of FIFO_FULL errors when pressing buttons and reading NMEA strings from the GPS-raw event.

Ahh - that'll happen when a line of NMEA data comes in before you finished processing the last one. It's possible your processing itself is fast enough, but maybe some other functionality (likely a screen redraw?) is taking too long to complete.

What data are you needing to extract? Maybe it could be added into the Bangle.js firmware?

> It doesn't seem to cause any malfunctioning in the app
> The GPS seems to take ages to lock

These could be related :) But otherwise there shouldn't be a difference - the GPS runs independently so the only difference could be if there was some issue processing the data.

> Is there a way to have a look at the CPU load of the watch, to make sure the app is not too computationally expensive?

There's `setBusyIndicator` but it's not actually that helpful as it needs an IO pin to pulse, and about the only one that's usable is the speaker pin. You can take the back off the watch and do `setBusyIndicator(D18)` and then wire an LED up where the speaker is, but it's not great :)

singintime:
> What data are you needing to extract? Maybe it could be added into the Bangle.js firmware?

To be able to use the GPS parsed data instead of the NMEA strings, on top of latitude, longitude, altitude and speed (all of them extracted directly from the NMEA sentences, correct?), I would only need the PDOP (dilution of precision) of the GPS reading, namely field n. 15 of the GSA sentence. The field gives an estimate of the accuracy of the reading, which is one of the required inputs for the Kalman filter.

PR #554, opened 2020-08-31
Update worldclock with new customize.js location and add README

computermacgyver:
Sorry for missing the refactoring. I took this opportunity to quickly add a readme as well. Thanks for all the help!


gfwilliams:
That's great - thanks! I should have noticed when I pulled this in. I'll update the Espruino site and bangleapps in a few minutes.

PR #556, opened 2020-09-04
minionclk v0.04

singintime:
Clearing the screen on startDrawing(), to the date string on the bottom of the screen. This is currently not happening, unless the display is turned on at midnight.


gfwilliams:
Sorry for the delay - just been on holiday. Thanks!

PR #559, opened 2020-09-15
Add sensor battery indicator, small bugfixes

berkenbu:
Added a graphical indicator for the sensor battery. 
More robust tracking of moving time. 
Correct saving of total distance traveled on exit and reset.
Reconnect on sensor disconnect.


gfwilliams:
Nice - thanks!

PR #563, opened 2020-09-23
Screen flicker, speed stats and battery reporting improvements

berkenbu:
Reduce flicker on screen updates; make speed stats reporting (max and average) slightly more robust; continuously report sensor battery levels.


berkenbu:
Found another small issue tonight I would like to fix first, don't want to bother you with too many pull requests...

PR #568, opened 2020-09-29
Continuous battery monitoring, improved screen updates, manual reconnect

berkenbu:
Update sensor battery level icon continuously on 0x180f advertisements.
Reduce screen flicker with better screen update logic.
Allow for manually reconnecting to sensor with button 3 push.


gfwilliams:
Nice - thanks!

PR #570, opened 2020-09-29
New app: File manager

berkenbu:
A small file manager app to inspect (and manipulate) files in storage on the watch.


gfwilliams:
Great - thanks!

PR #578, opened 2020-10-15
New app: DSDRelay

berkenbu:
This is a small app to control bluetooth relays from the watch. In particular, it controls "DSD Tech" brand relays (single or multiple-relay boards can be found at the usual online retailers). Only tested on a single-relay board, but the manufacturer's documentation is clear enough that I am optimistic it will work with multi-relay boards, too.


gfwilliams:
That's great - thanks! I hadn't even seen those `DSD Tech` relays but they look really neat

PR #581, opened 2020-10-16
Mandelbrot app

berkenbu:
Small app that draws a Mandelbrot fractal on the watch screen. A region can be selected and zoomed.


gfwilliams:
Wow, that's great - thanks! It's great being able to use the Inline C for it as well :)

It'd actually make a really neat watch-face, having a new background each time you look at it.

PR #586, opened 2020-11-02
Add a watch face and a widget for NodeConf Remote

nicokaiser:


nicokaiser:
The bounds calculation still seems to be wrong, see the red dots in this picture

IMG_5968-Edit 1024

gfwilliams:
It's the second hand? I guess it could be a rounding issue as everything's a bit smaller than the original analog clock - the quick hack is just to increase the hand size for the bounding box:

```
addPt(r,handSizeSec+2,5);addPt(r,handSizeSec+2,-5);
```

Do you want me to merge this as-is, or do you want that fixed first?

Also, don't know if you're interested but you *could* have a smaller background image behind the clock, and could then render the NodeConf Remote text as a full-res 2 bit bitmap (which would get you nice greyscale).

gfwilliams:
Ready to merge?

nicokaiser:
Yes, ready, I rebased the commit.

gfwilliams:
Great - thanks!

PR #587, opened 2020-11-03
banglerun 0.06

dapgo:
You can cancel the merge, as I just wanted to include 4 files for banglerun (0.06)  but I did something wrong as I am seeing other commits....


gfwilliams:
Wow, strange.. Ok, cancelling.

If you're having trouble and you need me to manually take the banglerun files from https://github.com/dapgo/BangleApps-dapgo then let me know :)

dapgo:
I was trying to do everything from the github web because i had not installed the github desktop.
@gfwilliams  if you can add the 4 files from my github, it would be the faster way.

gfwilliams:
Ahh - I just started looking at this. Do you have the updated typescript source files? Or did you work directly by modifying the JavaScript that was created from the TypeScript?

dapgo:
I modified the javascript "banglerun.app.js" and  I tested that it works on my watch.
In fact just noticed about the src subfolder. Though have no previous experience with TS but i can try to modify and compile..

PR #589, opened 2020-11-12
Don't clear all intervals during initialisation of Large Clock

jjok:
The `clearInterval()` stops any boot code or widgets that use intervals from working.

See conversation here: http://forum.espruino.com/conversations/356341/

> Large Clock that line is totally pointless and could just be removed


gfwilliams:
Cool - thanks!

PR #594, opened 2020-11-15
Added maidenhead readout

Fscked-In-The-Head:
I wrote a new function specifically to calculate extended maidenhead squares on Bangle.js, and added it here because I like the functionality of this app better than hamloc. 
Most maidenhead calculators use only 6 characters of precision, which is only enough accuracy to determine a rough neighbourhood. That's fine for HF comms, but VHF can make use of extended accuracy. Eight characters brings that down to about a block, which is enough to meet someone when driving or on foot and sharing locators over radio. Ten would point at one building, and I may expand to that in the future.


gfwilliams:
Nice - thanks!

PR #603, opened 2020-12-03
Aviation GPS 

motorfiets:
Same app as the gpsnav by @jeffmer but with units changed for aviation use.


gfwilliams:
Thanks! Actually just a thought about this - there's a `locale` library in Bangle.js that handles converting units:

https://www.espruino.com/Bangle.js+Locale

Eg:

```
// Distance/Length (in meters) to formatted String
>require('locale').distance(150)
="150m"

// Automatic unit selection for larger distances
>require('locale').distance(15000)
="15km"
```

So is the unit used literally the only difference? If so maybe it makes sense to change `gpsnav` to use the `locale` library, and then to add an 'Aviation' locale that displays all the units in an Aviation-friendly way?

jeffmer:
I would be happy to implement this. I need to tidy up the display font sizes in any case.

motorfiets:
This makes great sense using the local. What I was thinking however was
tweaking the app further to make it more aviation focused. The gpsnav base
app is a great start to make those changes from. That being said it makes
sense to me to have a separate app that is an aviation focused navigation.

gfwilliams:
@jeffmer it'd be great if you could do that, thanks! I'm not sure if Marine is slightly different to Aviation as well - but maybe not...

@motorfiets Yes - I think long term a specific Aviation app makes a lot of sense. Maybe also even more specialised versions (I can imagine a Paramotor one might want to report fuel levels/RPM/etc). However for now if the *only* change is for the units and we can get what you want using @jeffmer's app then can we do that?

If you end up making a bunch more changes then sure, let's resubmit this - but it's a decent amount of code, so the longer we can keep from duplicating it the better as then any fixes/improvements that go in will go in for both apps :)

jeffmer:
I think navigation in general covers marine and aviation so would you prefer en_NV or en_NAV for the locale entry? I think en_NAV is more direct, however, I realise most entries are aa_AA.

I had not realise you added open map display to the gpsrec app - really neat!

gfwilliams:
I think `en_NAV` is better as it's obvious it is something different. I think I use the last 2 letters in order to show the flags but I think in this case a little hack to show an emoji for a boat or something would be appropriate. Totally happy to do that myself if you want.

Thanks! I should really pull the OpenStreetMap code into a library though (a bit like we do for locale) - it'd make it way easier for people to add the functionality to other things - it's not great effectively copy&pasting the map code everywhere :)

motorfiets:
Can other maps data be loaded? Say aviation maps?

gfwilliams:
If you have a source of that map data, then yes. Currently it supports OpenStreetMap but actually there are a bunch of different OpenStreetMap tile types. There may well be some set up for aviation:  https://github.com/espruino/BangleApps/blob/master/apps/openstmap/custom.html#L63

jeffmer:
Experimenting with locales, I loaded the en_IN locale which kept giving "NaN". I looked at the locale in Storage and it has the function: `  speed: n => round(n/undefined) + "kms",` which looks wrong!

gfwilliams:
Thanks - just merged and fixed `en_IN` - `locale.html` even prints a warning message to the console about it, so I guess the contribution was just never tested :(

PR #604, opened 2020-12-03
Refactoring, Improved Description

OmegaRogue:
Refactored and Reformatted some code and Improved the description by adding the ui framework the DANE Watchface is based on, which i apparently forgot to do before


gfwilliams:
Neat, thanks! However it looks like you disabled buttons 1 and 3 - was that intentional?

OmegaRogue:
in the previous version i didnt notice i forgot to comment the handlers out, they didnt do anything because i changed the big number from being a counter controlled by the buttons to a battery percentage display

gfwilliams:
Ok, good to know - thanks! :)

PR #605, opened 2020-12-05
update gpsnav to use locale + add en_NAV

jeffmer:
Relating to Aviation GPS #603, I have updated gpsnav to use locale for speed and distance and added en_NAV to locale. I have tested with en_GB, en_NAV and en_IN locales. As per comment in #603, en_IN appears to be broken when converting speed.


gfwilliams:
Thanks - that's great! I'll tweak en_IN as well - I wonder if 'kms' was just a typo - I did some quick googling and I can't see any real mention of it being `kms` rather than `kmh`

PR #606, opened 2020-12-07
Add Dane Touch Launcher

OmegaRogue:
Added a version of the Touch Launcher that fits with the Watchface "Digital Assistant, Not EDITH"


gfwilliams:
Nice - thanks!

gfwilliams:
Do you have an icon for this? Looks like the one in here is the placeholder

OmegaRogue:
> Do you have an icon for this? Looks like the one in here is the placeholder

Not yet, but I will add one soon, after moving the code for the shared UI of Watchface and Launcher into a module to avoid duplicates

PR #607, opened 2020-12-09
Add icon for Dane Launcher, move code to module

OmegaRogue:
Added an icon for Dane_tcr
Added module dane_arwes
Moved shared code from dane and dane_tcr to dane_arwes


gfwilliams:
Does this work now? You mentioned in the forum you were having issues.

OmegaRogue:
> Does this work now? You mentioned in the forum you were having issues.

Not yet, I'm having problems with getting my fork to change the submodule to master branch

OmegaRogue:
I checked again and apparently the core module is already up to date, but it still refuses to upload my apps

OmegaRogue:
it's trying to get dane_arwes.min.js but there only is dane_arwes.js

OmegaRogue:
and that was the problem, i manually minified it and added the minified version and it worked

gfwilliams:
Ok, great! That's odd though - it should try and get the minified version, and if that doesn't work it just included the unminified one and minifies it itself.

But it's working ok now?

gfwilliams:
Just a note that the .min.js file doesn't seem to be included here at the moment

OmegaRogue:
> Just a note that the .min.js file doesn't seem to be included here at the moment

I currently only have it on the branch that gets built into my apploader instance.
I noticed that the module doesnt actually get uploaded as a seperate file but to get added to each app javascript file that uses it individually, which seems redundant

OmegaRogue:
now the minified module is in this branch aswell

gfwilliams:
That's great - thanks! Merging :)

PR #608, opened 2020-12-09
Fix Font sizes in multiclock and magnav

jeffmer:
These were initially updated before the final version of the new Vector Font and I have now fixed them. In addition I have added a new clock face in multiclock.

Apologies for the large number of files affected but I though it best to fix both apps.


gfwilliams:
That's great - thanks for sorting this out!

PR #612, opened 2020-12-21
Addition of Heart Rate Variability monitor app

jabituyaben:
Added relevant app folder and ammended JSON.

The app first reads raw HRM data into an array over 30 seconds and then processes the it to find precise points of each peak, it then provides a reading of the heart rate variability and also a heart rate based on these calculations. There is sample count metric also displayed which shows how many peaks/pulses are taken  into account, and this gives an indication of data quality - more pulses = more better.


gfwilliams:
Awesome - thanks!

PR #614, opened 2020-12-25
Hardalarm App

jamespsteinberg:
An alarm clock where you have to count to a random number before it will turn off to make sure you wake up.


gfwilliams:
Nice - thanks!

PR #619, opened 2020-12-28
Desktop Style Launcher App

jeffmer:
Developed this for a touch screen but it works well on Bangle and gives fast access if you have lots of apps. Also IMHO looks good - see README


gfwilliams:
Looks great - thanks!

PR #620, opened 2020-12-28
Update 1

jabituyaben:
Taken out some debugging stuff not needed now and also it's a bit quicker getting a reading. I've also added options to either run this continually and log to a CSV file or just take a single reading as before.

I've added a changelog and Readme also.


gfwilliams:
Thanks!

PR #622, opened 2021-01-01
new app and slight tidy up of HRV app for app loader display

jabituyaben:
New app 'Edison's ball' based on the Steel Ball app described in the Projects forum around alerting a wearer when they begin to fall asleep. This has uses in exploring non-drug methods of combating drowsiness and fatigue and experimenting with hypnagogic states.


jabituyaben:
I've now also added another HRV metric i.e. RMSSD, which is apparently more commonly used. Readme updated to reflect change. I've also added recording of CPU temperature and movement based on accelerometer and these, combined with the HRM data are useful in analysing sleep.

gfwilliams:
Great - thanks! Just merging.

In future, if you update an app please could you add something to the changelog and bump the version number? It lets others know they can update - but I'll just add that now.

PR #624, opened 2021-01-06
HRM raw data export app

jabituyaben:
app for logging and exporting raw HR sensor data to a CSV file for upto 1 hour in user defined increments of 1 minute. I've created a companion app in Python which processes the signal and displays HRV stats as an example, linked in the readme


gfwilliams:
Also, you could add something to the app loader to allow this to be exported really easily: http://www.espruino.com/Bangle.js+Storage#reading-with-the-app-loader

jabituyaben:
> Also, you could add something to the app loader to allow the data to be exported really easily: http://www.espruino.com/Bangle.js+Storage#reading-with-the-app-loader

It is ready but I didn't see this option before! I'll implement the app loader download aspect over the next day or so - should I keep this pull request open and then drop you a note when it's done or close and open a new one when it's ready with that additional part?

gfwilliams:
Great!

Yes, just leave this PR open and drop me a note when you're ready to merge :)

jabituyaben:
@gfwilliams sorted! all good to merge once Travis has done all the tests etc

gfwilliams:
Great - thanks!

PR #625, opened 2021-01-06
0.06: Added WIDGETS.activepedom.getSteps()

hughbarney:
A getter so that other Apps can access todays step count.


gfwilliams:
Great - thanks! Just FYI - when you increase the version in ChangeLog you need to update the main apps.json as well - but I'll update that for you now.

PR #629, opened 2021-01-10
feat(app): Add "Lazy Clock" app

ollycross:
Add 'Lazy clock' app which displays an approximate string representation of the time.


gfwilliams:
Brilliant - thanks!

PR #630, opened 2021-01-11
updates to multi lock & dtlaunch

jeffmer:
Small improvements to two apps based on suggestions from @zloid.

1) make multiclock pin selected clock face so that it will return to it after running other apps

2) allow cyclic scrolling through pages it dtlaunch


gfwilliams:
Nice - thanks!

PR #633, opened 2021-01-13
devstopwatch - Persist stopwatch state

fergalwalsh:
This PR updates devstopwatch to keep all state related variables in a `state` object which is persisted to storage on `kill` and read from storage on app launch. This gives the stopwatch the illusion of continuing to run in the background while other apps are active.

Using flash storage for this seems a bit overkill but I couldn't find a way of writing to global RAM that persists between app launches. Does such a mechanism exist?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1394,7 +1394,7 @@
   "name": "Dev Stopwatch",
   "shortName":"Dev Stopwatch",
   "icon": "app.png",
-  "version":"0.01",
+  "version":"0.02",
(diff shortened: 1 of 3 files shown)


gfwilliams:
That looks great - writing to flash on `kill` is definitely the way to go. It's not going to cause a bunch of flash wear or anything.

Unfortunately there's no way to persist things in RAM - in order to keep Bangle.js running stably with apps of various quality we made the decision to tear everything down between apps, so they always start in a known state.

fergalwalsh:
Thanks for the explanation, that makes sense.

PR #634, opened 2021-01-13
banglerun v0.06 - HDOP from GPS, added README

singintime:
This PR introduces a change in the BangleRun app, which now uses the `GPS` event instead of the `GPS-raw` event to read GPS data. The change is made possible by the introduction of the HDOP (Horizontal Dilution Of Precision) value in the `GPS` event since Espruino 2v07. The value is needed by the Kalman filter to work properly, and to provide accurate tracking of position and distance.

This PR also adds a README file, containing the instructions to transpile the project's source files (written in Typescript) into code compatible with the BangleJS.


gfwilliams:
That looks great - thanks!

singintime:
Ouch... it actually doesn't look that great, now the data is read but not used at all! I'm applying the fix in another PR

PR #635, opened 2021-01-13
New Astral Clock app

jabituyaben:
This is ready to merge from my perspective.

A new astral moonphase clock app that can also display relative planet locations as well as other astronomy targets.


gfwilliams:
Great - thanks! And thanks for the really detailed README too

PR #636, opened 2021-01-18
added active pedometer face to multiclock

hughbarney:
requires activepedom widget to be installed. Calls WIDGETS.activepedom.getSteps(); to retrieve latest step count.


gfwilliams:
Thanks! Please could you bump the version number and ChangeLog for this, so people know there is a changed version?

Looking at this, I wonder if we couldn't do something like making `getFace` return `undefined` if the face isn't supported, then just not putting it in the list of multiclock faces? That's be a nice tidy way of coping with what happens if the pedometer isn't installed.

hughbarney:
I have bumped the changelog and version number.  Got a feeling this push request may now have my gpservice changes as well.

gfwilliams:
Argh - yes, it has pulled in gpservice.

I guess we don't want the GPS service stuff yet?

The GPS service stuff looks really neat though - I wonder whether it's worth splitting off? So there's one app that's just an OS grid ref widget, and then another for configuring GPS - which could be used for any app/widget?

I guess the widget itself wouldn't need the GPS configuration stuff in it then?

hughbarney:
Apologies, I did not realise that you could only have 1 pull request at a time on github.  I had planned a small push for the pedometer face and a second push for the gpsservice.  I thought I would be able to push to my clone a second time.

So the gpsservice is production ready now, I have rattled out the last few issues with the settings app which was a bit tricky to get right.  The gpsservice comprises a configurable gpsservice widget,  a settings app that runs under settings/app widget settings and the custom watch faces that are checked into multiclock.  The design is such that the gpsservice widget in a generic service and apps can call WIDGETS.gpsservice.gps_get_fix(), when they wants to retrive the last fix.  I have checked the client examples into multiclock as osref.fs and gps.fs.

I can see there is now a conflict on apps.json so I will do a refresh to get rid of that.
I could just cancel this pull request and do another with a different comment if that helps ?

hughbarney:
closing pull request will issue another for clarity.

gfwilliams:
Ok, great! Just to check: will multiclock work properly if you don't have the gpsservice installed?

You can have multiple PRs in GitHub but they work from branches - so this one is I guess off the default `master` and so any changes go onto that. If you want multiple PRs then you can have a new branch for each one.

PR #640, opened 2021-01-21
low power configurable gpsservice widget, gpsservice settings app and example client watch faces for osref

hughbarney:
low power configurable gpsservice widget, gpsservice settings app and example client watch faces for osref and gps info watch faces for multiclock. Also pedometer watch face for multiclock.


gfwilliams:
Great - thanks!

PR #641, opened 2021-01-22
added gps_get|set_settings, restore to SuperE setup on poweroff()

hughbarney:
added gps_get|set_settings, restore to SuperE setup on poweroff(), updated README
bumped version number
added change log


gfwilliams:
Neat - thanks!

PR #642, opened 2021-01-25
Add mclockplus app

skauertz:
Hello,
please review my proposal for a new clock app based on Morphing Clock.


gfwilliams:
Looks great, thanks! The screenshot and README is really handy as well.

PR #643, opened 2021-01-25
added getSteps() to wpedom widget, updated vertical face to display step count

hughbarney:


gfwilliams:
Great - thanks! Do you think you could update the ChangeLogs and versions for those two?

The 'active pedometer' also has a 'getSteps' so I guess you could check for that widget as well. It's less than ideal having to check explicitly for each widget type though.

gfwilliams:
Also this happened in a previous commit but I didn't mention it and just fixed it myself then, but as it's happened here as well, when you make a PR you should see something like:

image

If you click 'deta7ils' you'll see why it failed. In this case it looks like your editor added tabs instead of spaces. Please could you have a go at fixing that? Everyone seems to have a different tab size, so if we use tabs then the formatting of everything goes out the window and it becomes pretty unreadable.

hughbarney:
Please accept my apologies.  My first attempt to fix my .emacs / javascript mode configuration failed, fixed now.  When I do the pull request I am seeing that CI checks are taking place but no idea how long they take to run before you see the output.    I'll bump the versions and also make the apps attempt to retrieve from activepedom OR wpedom. There must be a better way to do this but cant think of one at the moment.

hughbarney:
the CI job seems to take ages to schedule, 40 minutes after I did a push it still seems to be waiting to run.

gfwilliams:
Thanks! Not sure why it's not scheduling now - I seem to recall Travis said they were changing some things and maybe they are having some difficulties

PR #644, opened 2021-01-26
BangleRun v0.07 - GPS update bugfix, added guards against NaN values

singintime:
This PR addresses two main issues in BangleRun:

- Fixing the GPS update, which I broke during the last refactor :grimacing: 
- Adding some guards to avoid divisions by 0, and thus NaN values in measurements which are integrated over time, such as total distance

Unfortunately, I'm currently living in a building entirely made from tuff and concrete, and I can't go out very much due to severe COVID lockdown restrictions. Therefore, it's almost impossible for me at the moment to verify that the GPS actually works. It would be great if somebody could double-check that this code actually works before merging it. Thanks!


apps/banglerun/src/gps.ts
```diff
@@ -34,6 +36,15 @@ function readGps(state: AppState, gps: GpsEvent): void {
...
   state.dop = gps.hdop;
+
+  state.gpsValid = state.fix > 0 && state.dop <= 5;
+
+  updateGps(state);
+  draw(state);
+
+  if (state.gpsValid && state.status === ActivityStatus.Running) {
+    updateLog(state);
+  }
```
singintime, line 47: Here's the part I forgot last time :grimacing:

apps/banglerun/src/gps.ts
```diff
@@ -80,8 +91,8 @@ function updateGps(state: AppState): void {
   const pError = dpMag + state.dop * POS_ACCURACY;
   const vError = dvMag + state.dop * VEL_ACCURACY;
 
-  const pGain = state.pError / (state.pError + pError);
-  const vGain = state.vError / (state.vError + vError);
+  const pGain = (state.pError / (state.pError + pError)) || 0;
```
singintime, line 94: These should avoid the occurrence of NaN values in the remote case of a division by 0

gfwilliams:
Thanks - that's great! I'll merge now as the old one is definitely broken and will try and give this a go soon

Shashank-Prasanna:
@singintime yay! `NaN kilometers` no more!

gfwilliams:
Ok, I was a bit late giving it a go, but it seems to be broken I'm afraid :(

It looks like in `updateGps`, initially you were expecting state.lat/lon to be in radians, but then at the end of the function you convert them to degrees. Looks like with the new GPS changes they were in degrees all the time.

I have changed this now in 4e095172dd62503dbe288ca6ad76bc35f80b7ea2 and the GPS coordinates look correct, but I haven't had time to go out properly and see if it's working as it should be for a while trail.

One thing though - are we sure that the Kalman filter really helps? I believe this is something that the GPS probably does internally anyway.

PR #646, opened 2021-01-27
Digiclock v0.02

Shashank-Prasanna:
So, I know that I said it would be coming out soon, but I got **VERY** busy. So, to make up for it, I made a *whole new* UI redesign. And, it looks much better. Yay! It would be helpful if you could go through my `apps.json` file addition because `v0.01` was causing errors probably linked to the app list. Also, heads up to @Fscked-In-The-Head if they want to update `ISOClock`.


Shashank-Prasanna:
Wait, forgot some core concepts. I Will finish those and submit another PR

gfwilliams:
Ok, thanks!

SpyGuy0215:
Yeah, FYI this is discontinued lol. I forgot to say that, my bad, I just got busy w/ school and stuff and kinda forgot abt programming for a bit.

PR #647, opened 2021-01-27
Interval trainer app

pglez82:
Hi there. This is my first app. Just playing around but I think it could be quite useful to people that likes to train using interval training. Feel free to make any comments if something is not up to standards.


gfwilliams:
Looks great - thanks for submitting this!

PR #648, opened 2021-01-28
added display id, fw ver, batt %, mem % on BTN1 in cliock, commented out reset to SuperE mode in gpsservice

hughbarney:


gfwilliams:
Nice - thanks!

PR #650, opened 2021-01-28
added hide settings for line1/line2 in activepedom widget

hughbarney:


gfwilliams:
Thanks! Nice clean addition :)

PR #652, opened 2021-01-29
Planetarium App

pglez82:
I just developed a planetarium app for Bangle.js. It shows up to 350 stars depending on the position and time (stars positions are computed in the watch). 

This is quite pushing the edge of the watch in terms of cpu. I tried to make the code as fast as possible and for me it runs ok. Maybe I have missed some optimizations... I don't know. About memory, I think it is ok as I do not load the stars information in memory but use the flash storage to read it an compute the position without storing to much data in memory.

I plan to add more constelations during the following weeks but the app is usable as it is.

Pablo


gfwilliams:
Wow, thanks! This looks awesome! Having a quick look at it, I think one of the things that might speed it up is a slight change in how you parse the file. Right now you look at each character in turn for a newline, but you could use some JS functions to do the work for you - something like:

```
var line,linestart = 0;
lineend = f.indexOf("\n");
while (lineend>=0) {
  line = f.substring(linestart,lineend);
  print(line,); // do your calculations here...
  linestart = lineend+1;
  lineend = f.indexOf("\n",linestart);
  // ...
}
```

I think that should help quite a lot

pglez82:
Thanks! I will try that solution for the next version.

PR #653, opened 2021-01-29
v0.02

pglez82:
Hi @gfwilliams . As you said, that change made the code so much faster! I think the improvement has been x2. I have added a few more stars now (up to 500). It runs great.


gfwilliams:
Great - thanks!

PR #655, opened 2021-02-02
fixed activepedom widget, only exit after lastupdate has been recorded

hughbarney:
my hide modification broke the widget, I was experiencing the widget resetting steps back to zero in the middle of the day.  This is because the draw() function updates the lastupdate variable. So move my return code to after that has been done.


gfwilliams:
Ahh - thanks! Next time please could you bump version numbers so folks are sure they got the update?

PR #658, opened 2021-02-04
add Oblique Strategies to the app loader

smcalilly:
Oblique Strategies is a virtual card deck for Brian Eno's Oblique Strategies.


gfwilliams:
Looks great - thanks!

PR #665, opened 2021-02-09
minor changes

dapgo:
delete a duplicated file and correct .img filename


gfwilliams:
Thanks! Actually I don't think you need the `.img` file either as you've got a `app-icon.js` now - but I'll remove that quickly now.

PR #666, opened 2021-02-09
gpssetup app

hughbarney:


gfwilliams:
Thanks! Looks like the CI check failed because if the icon having a semi-colon, but I'll fix that now

PR #667, opened 2021-02-11
Added Walkersclock, removed old gps and osref faces from multiclock

hughbarney:


gfwilliams:
Thanks! just merging, although the automatic checks have failed because it looks like there's a semi-colon after the icon which there shouldn't be - I'll fix that now though.

PR #673, opened 2021-02-17
Sliding Text Clock

awkirk71:
First Version of Sliding Text Clock for the Bandle.js


gfwilliams:
This looks great - thanks!

PR #674, opened 2021-02-18
testuserinput v0.03

dapgo:
App: add change of colors 
Icon: replaced E.toArrayBuffer by require("heatshrink").decompress


gfwilliams:
Thanks!

PR #675, opened 2021-02-19
v004 code improvemnt  and attemp to avoid the fail on generated .img file

dapgo:
probably the menu icon.js(require("heatshrink")...) will fail again when producing a valid .img on server. However,  app-icon.js has been generated following http://www.espruino.com/Bangle.js+App+Loader


gfwilliams:
Looks good, thanks! the menu icon looks fine. The biggest gotcha with menu icons seems to be putting `;` at the end of them, but you haven't done that.

dapgo:
> 
> 
> Looks good, thanks! the menu icon looks fine. The biggest gotcha with menu icons seems to be putting `;` at the end of them, but you haven't done that.

The instruction I was following has no reference to this  `;` 
http://www.espruino.com/Bangle.js+App+Loader

In fact seeing app-icon.js from other apps i don't see this  `;` at the end.

gfwilliams:
Exactly - what I mean is, you did it right. Lots of people add a ';' even though it isn't mentioned anywhere and then it breaks stuff

dapgo:
if  you install this app https://espruino.github.io/BangleApps/#hello
you will see that the icon is not not displayed in the launcher

gfwilliams:
It's possible you're just using a paletted image? If so it's a feature that was introduced in newer firmwares - if you try a cutting edge built it might appear....

dapgo:
I converted using the optimal 4bits because the Mac 4bits was displaying a gray color  instead of a green. I was using fw 2v08, but today i tried with cut2v08.188 with same result.

gfwilliams:
Ok, thanks! Just looked into this and it's because Espruino can only display images with up to a 2bpp palette from flash (because otherwise it has to allocate space for the palette in RAM first).

As it turns out your image only has one colour, so all you needed is 'Optimal 1bpp' - I've just updated the icon and it works fine now

PR #676, opened 2021-02-21
Added countdown timer application

Shaylin:
Hello,

This is a pull request for a simple timer application that I created. I wanted something that could better match the usability of proprietary fitness trackers I've used in the past. 

The app is not very optimal at the moment; it's quite heavy in terms of memory usage for a timer. This is just the first version to get something that works well.

I've tested this on the emulator and my own bangle.js.

The icon I'm using is under the creative commons license. I've added the required attribution link in my readme.

Here's a screenshot of the app:

screenshot


gfwilliams:
Looks great - thanks for the PR!

PR #678, opened 2021-02-24
Arrow compass, compass power status widget (indicator), added lat, lon, course displays to walkersclock

hughbarney:


gfwilliams:
Nice - thanks!

PR #679, opened 2021-02-25
Added Waypointer App, walkers waypoint navigator app,  small fix to Arrow Compass

hughbarney:
At low speeds the GPS does not provide regular course information. This means the current forms of Waypoint navigation apps will only work at higher velocities beyond 8m/s (28kmph). The Way Pointer app works out a bearing to the desired waypoint then points the compass at the direction you need to travel.   This approach has the advantage that once the bearing is established it does not matter how long it is to the next GPS fix as the compass will continue to point in the right direction.


gfwilliams:
Wow, you've been busy!

gfwilliams:
Just noticed that  `npm test`/travis gives an error:

```
ERROR: App gpsnav storage file waypoints.json is also listed as storage file for app waypointer
```

I guess this is intentional? Just checking. I think the problem right now is if you update or add `waypointer` it'll overwrite the waypoints file from `gpsnav`

If so I might change it to a `data` file and if so tweak the tester.

hughbarney:
Happy to go with whatever you suggest.  Kind of wanted to share the same waypoints.json file across Apps and reuse the same HTML interface to edit them.  In practice I suspect I wont be using gpsnav much as waypointer works much better for walking.

gfwilliams:
Ok, great - yes, it makes a lot of sense to have the same file used. I'll merge and tweak the app file

PR #680, opened 2021-02-28
gbridge: fix HRM setting

rigrig:
The settings menu didn't handle `undefined` value for `hrm` properly, so you were unable to toggle it.
Fixed by forcing the value to a boolean.

A bit weird: I know it was definitely broken for me, but it was added a couple of months back (in 9a3b203429d2270273baec6e0c8d34c21a420e4c), and looks like it has been broken all that time.
Unless something changed in the way the menu handles `undefined` values?


gfwilliams:
Weird - thanks for fixing this!

PR #681, opened 2021-03-04
refactor: real random clock support

cha0s:
I don't really like the "random" clock app. It's clunky and it's fairly trivial to put the concept of random clock into the bootloader/settings. This PR does just that, in addition to a little general refactoring to make things a little neater.

I save the clock app as ` random`, space intentional. I figured this was a good way to keep it from ever conflicting with an actual app name, but if there's a better way to do that let me know.

P.S. I didn't actually run this code yet because I haven't actually tried uploading a new bootloader app (and it seems a tiny bit scary), and I would appreciate any guidance you could give me in that regard... e.g. not bricking my watch but being able to test this

P.P.S. recommend reviewing changes in 'split' mode

### DIFF
--- a/apps/boot/bootloader.js
+++ b/apps/boot/bootloader.js
@@ -1,24 +1,42 @@
 // This runs after a 'fresh' boot
-var clockApp=(require("Storage").readJSON("setting.json",1)||{}).clock;
-if (clockApp) clockApp = require("Storage").read(clockApp);
-if (!clockApp) {
-  clockApp = require("Storage").list(/\.info$/)
-    .map(file => {
+const allClockApps = () => (
+  require("Storage").list(/\.info$/)
+    .map((file) => {
       const app = require("Storage").readJSON(file,1);
       if (app && app.type == "clock") {
         return app;
       }
     })
-    .filter(x=>x)
-    .sort((a, b) => a.sortorder - b.sortorder)[0];
-  if (clockApp)
-    clockApp = require("Storage").read(clockApp.src);
-}
-if (!clockApp) clockApp=`E.showMessage("No Clock Found");
-setWatch(() => {
-  Bangle.showLauncher();
-}, BTN2, {repeat:false,edge:"falling"});)
+    .filter((x) => x)
+);
+const clockNotFound = `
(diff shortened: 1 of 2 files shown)


cha0s:
Don't mind the noise, just a consequence of not actually running the code...

gfwilliams:
Thanks for this, however looks like your refactoring may have actually increased memory usage quite drastically, not to mention being slower (because it scans all clock apps every startup, even if one was declared).

I don't think allClockApps or currentClockApp are deleted, which means that they'll be sitting in RAM all the time any clock is running.

The original code definitely looks a bit rough, but in the bootloader in particular it is written the way it is so that it plays well with Espruino and runs quickly and doesn't use much RAM.

I'm afraid I couldn't merge this as-is.

cha0s:
No worries, I'll learn more and try again.

It isn't true that `allClockApps` gets called for every case though, only when necessary. If `clockApp` is defined, it isn't called.

I actually figured that the only reason `delete` was used was because the original code had `clockApp` declared as a top-level var. I instead allocate it on the fly in `currentClockApp`. Can the GC not be relied on to clean this up?

Learning :)

gfwilliams:
Sorry - you're totally right, I didn't spot the arrow function there.

The issue is Espruino runs the bootloader code as if it's just prepended to everything else, so *anything* you define in that code sticks around (function or var, even with `const`).

You could declare allClockApps/clockNotFound/currentClockApp inside evaluateCurrentClockApp, and then just delete `evaluateCurrentClockApp` after execution and that should do it nicely though?

If you make an empty clock app, load it, then run `trace()` you should be able to see what Espruino thinks is currently in RAM. if you want to try and track stuff down.

It's a shame there's so little memory available that we have to be really careful.

cha0s:
Vars allocated in a function will be released by GC when the function returns, right? Do we have to explicitly `delete` everything?

gfwilliams:
Yes, that's right - but `apps/boot/bootloader.js` isn't a function. You could encase everything in `(function() { ... })()` but in this case we don't do that just to make execution a bit faster

PR #684, opened 2021-03-09
Kalman filter to smooth alt and speed readings.

nujw:
After two weeks of real-world testing on the water, and cycling in bright sun light have added a couple of changes  :

- A large display mode to read the time and select a go to waypoint. Easier to see in full sun light.
- Kalman filtering on altitude and speed. Primarily to improve altitude reading stability when establishing a terrain height.


gfwilliams:
This is great - thanks!

PR #685, opened 2021-03-09
stepo:  improved time to draw, start stop timers, use long press on BTN2 to switch to launcher

hughbarney:


gfwilliams:
Thanks!

PR #686, opened 2021-03-14
new app: gbmusic: Gadgetbridge Music Controls

rigrig:
A dedicated app for Gadgetbridge music, instead of mixing it with other notifications.    
Also comes with a widget to automatically load the app when playing music. (room for improvement: it would be nice to think of some clever way to only do this if the active app is the clock)


gfwilliams:
Nice - thanks! In newer firmwares there is a `__FILE__` variable which shows the current file that's loaded - so I guess potentially you could do something like `isClock = (typeof __FILE__ == "undefined") || require("Storage").readJSON(__FILE__.split(".")[0]+".info",1).type=="clock"` 

However that code's totally untested and has no real error checking ;)

PR #687, opened 2021-03-15
Adding first version of app battleship

capybara1:
The classic game of battleship. For further information please have a look at the `apps/battleship/README.md`.

As far as I could find out there is no way to render the field in a background buffer, thus the way I repaint the entire field between the turns is a observable process. However from my personal perception it's not irritating but has a nostalgic charm.


gfwilliams:
Looks great - thanks!

> As far as I could find out there is no way to render the field in a background buffer,

You can definitely do this - for example: https://github.com/espruino/BangleApps/blob/master/apps/trex/trex.js#L1-L10

But since there's not enough RAM to store the whole screen in memory you'd have to go with a lower bit depth - as you say, if it looks fine, why not keep it - it's not like it's updating all the time.

PR #688, opened 2021-03-17
feat(app): Update "Lazy Clock" app to v0.02

ollycross:
- Fix bug with times between 12:30 and 13:00
- Tweak phrases
- Only update on threshold change, not every minute


gfwilliams:
Thanks!

PR #689, opened 2021-03-17
improved speed of arrow drawCompass() function

hughbarney:
improved speed of arrow drawCompass() function to avoid causing performance issue with buttons becoming unresponsive


hughbarney:
bit more tweaks needed on the calibrate user interface as I have changed the buffer size.

PR #691, opened 2021-03-17
Make counter work even when widgets are hidden, and make blobclk draw the time a bit sooner when LCD is powered on

jes:
The bug in "counter" was that it wasn't setting the colour when updating the screen. That means if the widgets are not drawn and the colour is set to 0, then the screen is drawn in black and you can't see anything. It also means that if the last widget to be drawn left the colour as something other than white, then the counter is not drawn in white. The fix is to set the colour to white explicitly.

I found that "blobclk" took a noticeably long time to display the time on the screen when the LCD is powered on. Moving the `redraw()` call to happen a bit earlier makes an improvement, but not a massive one. It would be better to make "blobclk" more efficient if possible, but this is still an improvement for now, in my opinion.

### DIFF
--- a/apps/blobclk/clock-blob.js
+++ b/apps/blobclk/clock-blob.js
@@ -86,9 +86,9 @@ function clearTimers() {
 }
 function startTimers() {
   g.clear();
+  redraw();
   Bangle.drawWidgets();
   intervalRef = setInterval(redraw,1000);
-  redraw();
 }
 Bangle.loadWidgets();
 startTimers();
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! Just merging - in future, if you change something please could you update the ChangeLog and version in apps.json? I'll do that for these now.

But yes, blobclk could definitely be improved upon speed-wise :)

jes:
Thanks! Yes, I will do that next time, sorry, I didn't know.

gfwilliams:
No problem at all - thanks for adding the changes!

If you're interested, all I changed was https://github.com/espruino/BangleApps/commit/d344dc127bb3545e2afb750306ee7cd3efb6e1d0

PR #697, opened 2021-03-20
Quiet mode

rigrig:
Implement #554: Quiet Mode

I had a close look at all changed apps, and I'm pretty confident this won't break any of them.   

Some options to apply during Quiet Mode:
settings

And a `qmsched` app for automated schedule:
qmsched schedule

~~Now I'm not sure if it should be called "Quiet Mode" or "DnD" (but I do know we should decide before merging this ;-)~~


gfwilliams:
Thanks! Just merging

PR #698, opened 2021-03-20
widchime: simple hour chime

rigrig:
Already respects the `quiet` setting, I figured there was no harm in already putting it in, as the default is to not be quiet anyway.

I noticed some clock faces also have this built-in, should we strip those out if this is merged?


gfwilliams:
This is a nice idea, thanks! Which clocks have the chime in them?

It may be some of them don't load widgets in which case we may have to leave the code in I guess.

rigrig:
These clocks have a chime built in
* [aclock](https://github.com/espruino/BangleApps/blob/df5fb6c052842f80807f56e38b4e81e1d4236d70/apps/aclock/clock-analog.js#L117): Analog Clock
* [beebclock](https://github.com/espruino/BangleApps/blob/df5fb6c052842f80807f56e38b4e81e1d4236d70/apps/beebclock/beebclock.js#L268): Beeb Clock
* [dotclock](https://github.com/espruino/BangleApps/blob/df5fb6c052842f80807f56e38b4e81e1d4236d70/apps/dotclock/clock-dot.js#L132): Dotclock

They all load widgets, so that shouldn't be a problem.   

Right now I'm wondering if maybe this should be a boot app instead of a widget though?

gfwilliams:
Thanks! Yes, definitely looks like it's worth taking the chime out of those then.

And yes, you're right re: widget/boot - although I guess it doesn't matter much. You could make it display a small icon, and then the widget would make sense...

rigrig:
I would've been happier if it was a boot app, so you also get chimes if the app doesn't load widgets.    
But for now I'll leave it (also because I like neither the idea of renaming an app, nor having `widchime` not actually being a widget)

gfwilliams:
:) ok - thanks!

Weiming-Hu:
Hi folks, I happen to implement a similar app using, instead of widgets, but the boot function. It's currently under testing. I sincerely invite you over for suggestions and comments.

This is my very first app with JS and in fact, my first time writing JS. So I would appreciate any suggestions on coding style, proper practice, and memory management.

The app is available [here](https://github.com/Weiming-Hu/BangleApps/tree/master/apps/hourstrike).

PR #699, opened 2021-03-21
Updated increments of brightness setting.

Tutturuuu:
This may be something of personal preference but it just seems more logical to have it go from 1 to 10 instead of 0.1 to 1.


gfwilliams:
I don't think this actually works? Have you tried it? I think the LCD would just stay at full brightness all the time.

I'm happy to have the change, but you'd need the actual value in `settings.brightness` to stay between 0.1 and 1, even if it appears to the user as between 1 and 10...

Tutturuuu:
Good point, sorry that I did not try it. Seemed simple.

PR #701, opened 2021-03-22
Made Date more visible

Tutturuuu:


gfwilliams:
Thanks! I'll do it this time, but if you make changes to apps in future please could you add a line to the ChangeLog and increase the version number in `apps.json`? Then users that have it installed will see there's an update available

Tutturuuu:
Alright, thanks!

PR #704, opened 2021-03-23
Remove hardcoded hour buzz from aclock, beebclock, dotclock

rigrig:
Because we have a separate widget (`widchime`) for that now


gfwilliams:
Great - thanks for adding the widchime note in the ChangeLogs too!

PR #705, opened 2021-03-24
fix widchime: make it actually chime the hour instead of crashing

rigrig:
Well, this is a bit embarrassing...   
Somehow the code I tested worked fine, and then I got fancy and removed semicolons before pushing, apparently without testing.   
Which means that
```js
  if (!type) return

  function chime() {
```
becomes a single statement, `chime()` isn't actually defined, and it now chimes the hour by crashing instead of buzzing.

Actually only needed two semicolons to fix, but I just added them everywhere (and will try to remember this the next time I think code w/o semicolons looks cleaner)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -585,7 +585,7 @@
   { "id": "widchime",
     "name": "Hour Chime",
     "icon": "widget.png",
-    "version":"0.01",
+    "version":"0.02",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! It's a shame the linting didn't pick that up :)

PR #707, opened 2021-03-27
kitchen combo compass, disable BTN1,BTN2 until GPS status is RUNNING and first position fix received

hughbarney:


gfwilliams:
Thanks!

PR #708, opened 2021-03-27
gbmusic: Increase text brightness, (try to) improve memory usage

rigrig:
Turns out that pure blue (#0000FF) is quite hard to read in daylight :-( So tweak the colors a bit to always throw in at least some R+G.    
Also try to improve memory usage a bit by just drawing scrolling titles twice instead of concatenating to itself, and not keeping quite as many copies of strings in memory

(I suppose inlining constants instead of [putting them all in an object](https://github.com/espruino/BangleApps/blob/6c481a118b3bf7c04ac21bf15e6a4f3ceddfdb44/apps/gbmusic/app.js#L25-L81) might also help... Maybe even just get rid of (most of) the fancy dynamic drawing and just code in a few different functions.)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3004,7 +3004,7 @@
   "name": "Gadgetbridge Music Controls",
   "shortName":"Music Controls",
   "icon": "icon.png",
-  "version":"0.01",
+  "version":"0.02",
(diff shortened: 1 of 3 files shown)


rigrig:
Whoops, seems there are some issues with font size after this.

PR #709, opened 2021-03-28
Sweep Clock Color Themes

awkirk71:
The sweep hand clock now has a number of colouring themes and saves the preferences to memory.


gfwilliams:
Nice - thanks!

PR #712, opened 2021-03-28
added graphing for Heart Rate Recorder

Andrwe:
## CHANGES

* added graphing of latest 164 measurements of chosen record file
* graphing code is based on batchart

Signed-off-by: Andrwe Lord Weber <github@andrwe.org>

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -445,7 +445,7 @@
   { "id": "heart",
     "name": "Heart Rate Recorder",
     "icon": "app.png",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! This looks great! I don't see anything that should really use the memory up too much though - so I'm a bit surprised about the large memory usage.

The big thing would be writing data into a standard JS array (rather than a types array) but it doesn't look like you do that?

PR #714, opened 2021-03-31
Added Korean locale

dorajistyle:
I added Korean locale here.


gfwilliams:
Thanks! However has this been tested?

I don't think the characters used are included in the built-in font, so I don't think any of the text will appear

PR #715, opened 2021-04-01
New app: Hour Strike

Weiming-Hu:


Weiming-Hu:
Please see [README](https://github.com/Weiming-Hu/BangleApps/tree/master/apps/hourstrike) for its features.

rigrig:
Looks neat. One suggestion though: I don't think it should `load("hourstrike.js")` to strike the hour, it would be nicer if the `notify_func` code was just inside inside `boot.js`.
Because now if you are using some other app while the hour strikes, it will exit and then load the clock.

Weiming-Hu:
Or another way to avoid the infinite loop is to only reboot the `hourstrike` app if possible. Currently, when the app is booted, it figures out how long to wait until the strike.

rigrig:
What you want is something like this:
```js
function setup() {
  // figure out how long until next hour:
  let t = ...
  setTimeout(strike_hour, t);
}
function strike_hour() {
  Bangle.buzz();
  setup(); // schedule strike for next hour
}
setup(); // schedule first strike
```

Weiming-Hu:
@rigrig Fantastic. Thank you very much for your suggestions and I have made changes accordingly. As of right now, it is not calling the `load` function anymore.

gfwilliams:
Thanks! Just merging...

PR #717, opened 2021-04-10
gbmusic: Increase text brightness, improve controls, (try to) reduce memory usage

rigrig:
Change the color selection to always mix in a bit of every color, so you never end up trying to read e.g. 100% blue. 
It also turns out that having your volume lowered to 0 every time you long-pressed BTN3 to restart the watch is kind of annoying.   
Added double/triple pressing BTN2 to navigate between songs.

And a bunch of changes to attempt to lower memory usage, which in most cases actually meant simplification of the code :-)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3034,7 +3034,7 @@
   "name": "Gadgetbridge Music Controls",
   "shortName":"Music Controls",
   "icon": "icon.png",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks!

PR #719, opened 2021-04-12
Smoother Sliding Text Clock with Colour themes

awkirk71:
- Added Colour Themes
- Sliding Text Clock is now a lot smoother
- Architecture broken up into modules to make it easy to add languages
- Language selector page added to upload page so clock doesn't have to load up all the languages


gfwilliams:
Thanks!

PR #721, opened 2021-04-15
Added "Where in the World?" App

SebTee:
The "Where in the World?" app shows where the user is on a Mercator projection world map using the coordinates from the GPS.


gfwilliams:
Nice! Thanks!

PR #724, opened 2021-04-19
batclock updated

ra101:
mainly a `apps.json` fix,
I feel like there needs to be a test case regarding that...


apps/batclock/README.md
```diff
@@ -1,13 +1,14 @@
 <div align="center">
 <h1> BatClock </h1>
 <p align="center">
-  <img src="./assets/screenshot.png" width="240px" height="240px">
+  <img src="https://github.com/espruino/BangleApps/raw/master/apps/batclock/assets/screenshot.png" height="240px">
```
gfwilliams, line 4: Please could you just use `src="screenshot.png"` here? That should be all you need...

gfwilliams:
The issue was that the app filename was wrong? Yes, it seems like `sanitycheck.js` should really have caught that

apps/batclock/README.md
ra101: Thanks, Done. Update at [here](https://github.com/espruino/BangleApps/pull/725) as well

gfwilliams:
Thanks!

ra101:
> The issue was that the app filename was wrong? Yes, it seems like `sanitycheck.js` should really have caught that

in my local, `sanitycheck.js` , stops at `moonphase`.
```
ERROR: App moonphase's moonphase.app.js is a JS file but isn't valid JS
```

gfwilliams:
That's odd - maybe try `npm install` first? I guess it's possible the sanitycheck is missing the libraries needed to parse JS.

When you go to merge, that's what the tests are based on - for instance: https://travis-ci.org/github/espruino/BangleApps/builds/767534162

PR #726, opened 2021-04-19
kicthen combo: improved stopwatch display, rewrite compass arrow to use direct screen write to avoid mem defrag

hughbarney:


gfwilliams:
Great - thanks!

PR #728, opened 2021-04-19
Adds pt_PT locale

StuntsPT:
Just adds a new pt_PT locale.
Tested it on my Bangle.js and seems to work just as expected. =-)

PS - Thank you for all your work on Bangle.js. I may not be a very active community member, but your work is highly appreciated.


gfwilliams:
Thanks! And thanks for testing - the locales especially sometimes seem ok but then don't work on the Bangle because of the limited font.

PR #729, opened 2021-04-21
[Heart] many improvements

Andrwe:
## CHANGES

* Add loading screen for viewRecord
* List average, minimum & maximum measurement in viewRecord
* Disable recording only when current recording file is erased
* Fix timezone offset
* Draw chart based on height and width of display instead of hard-coded limits
* Reduce memory usage by ~30%
* Generate scale based on defined minimum and maximum measurement
* Added background line on 50% to ease estimation of drawn values

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -460,7 +460,7 @@
   { "id": "heart",
     "name": "Heart Rate Recorder",
     "icon": "app.png",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great - thanks!

PR #731, opened 2021-04-25
Gadgetbridge Music Controls: only auto-start from clock apps, auto close after 1 hour of inactivity 

rigrig:
* only auto-start from clock apps, [(thanks for the hint)](https://github.com/espruino/BangleApps/pull/686#issuecomment-799260231)
* reformat widget code (as I was touching it anyway)
* auto close after `2*<song duration>` or 1 hour of inactivity: some (i.e. my) phones don't bother ending music updates when music simply finishes playing. (Duration seems reliable enough, but using the 1 hour fallback just in case some players don't report it.)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3043,7 +3043,7 @@
   "name": "Gadgetbridge Music Controls",
   "shortName":"Music Controls",
   "icon": "icon.png",
-  "version":"0.02",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Nice - thanks!

PR #734, opened 2021-04-29
gbmusic: setting to disable touch controls (and a small bugfix)

rigrig:
My sleeve touching the screen causing random song skips was bugging me, and I pretty much coded "completely disable touch controls" before realizing these were happening while the screen was off. So now it has three options :-p

Maybe we should just stick with a unconditional `isLDCOn` check?
But when the weather improves (i.e. no more long sleeves) blindly swiping to the next song might be nice. (At least that's the excuse I tell myself for not throwing away my shiny new code.)

Also a minor bugfix for dynamic colors, to make them actually behave as the comments claim they do.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3044,7 +3044,7 @@
   "name": "Gadgetbridge Music Controls",
   "shortName":"Music Controls",
   "icon": "icon.png",
-  "version":"0.03",
+  "version":"0.04",
   "description": "Control the music on your Gadgetbridge-connected phone",
   "tags": "tools,bluetooth,gadgetbridge,music",
   "type":"app",
--- a/apps/gbmusic/ChangeLog
+++ b/apps/gbmusic/ChangeLog
@@ -1,3 +1,4 @@
 0.01: Initial version
 0.02: Increase text brightness, improve controls, (try to) reduce memory usage
 0.03: Only auto-start if active app is a clock, auto close after 1 hour of inactivity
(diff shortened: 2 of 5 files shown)


gfwilliams:
Great - thanks! Although you're saying you get touch events when the screen is off?

I was pretty sure those should be disabled - but maybe you don't get button presses but you do get swipe/etc?

rigrig:
Both apparently: I just tested it by running:
```js
Bangle.on("touch", side => {console.log('touch:', side, 'LCD:', Bangle.isLCDOn());});
Bangle.on("swipe", dir  => {console.log('swipe:', dir,  'LCD:', Bangle.isLCDOn());});
```
and got this
```
>
{"t":"status","bat":85}
>
 ____                 _
         |_| espruino.com
 2v09 (c) 2021 G.Williams
>
swipe: -1 LCD: true
swipe: 1 LCD: true
touch: 2 LCD: true
touch: 2 LCD: true
touch: 1 LCD: true
touch: 2 LCD: true
touch: 2 LCD: true
swipe: -1 LCD: false
swipe: 1 LCD: false
touch: 2 LCD: false
touch: 1 LCD: false
> 
```

gfwilliams:
Ok, thanks - I've just pushed a change to fix that - but obviously that'll only hit in the 2v10 firmware

rigrig:
Awesome, that was fast :-)

PR #738, opened 2021-05-10
Kitchen Combo: added heart rate watch face, trip counter to stepo, reverted to direct screen write for stepo

hughbarney:
Getting close to the limits of what the memory can handle so have reverted to direct screen writes for the stepo part of Kicthen Combo.


gfwilliams:
Great - thanks!

PR #741, opened 2021-05-13
gbmusic: Setting to disable double/triple press control, remove touch controls setting, reduce fadeout flicker

rigrig:
* Setting to disable double/triple press control for http://forum.espruino.com/comments/15984222/
* Remove touch controls setting, because [firmware `2v10` will ignore touch/swipe events while the LCD is off anyway](https://github.com/espruino/Espruino/commit/917ceab9710a8ed1a8f1d50aa01dcae25059e2a5), so just having a setting to disable them seems silly.
(we still check the LCD is off before handling touch events, as probably not everybody is using `2v10` yet ;-)
* Don't clear the area while redrawing for fade out, to prevent flicker
* Some JSDoc improvements

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3044,7 +3044,7 @@
   "name": "Gadgetbridge Music Controls",
   "shortName":"Music Controls",
   "icon": "icon.png",
-  "version":"0.04",
+  "version":"0.05",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Nice - thanks for this!

PR #742, opened 2021-05-14
Dozenal Time

numerist:
Very small changes, including one to allow widgets to appear with the display


gfwilliams:
Thanks! This is quite a small change, but normally when you make changes it's an idea to change the version number (and add a line in the ChangeLog file) so users know when to update

numerist:
Thank you for your help! I’m still learning all this. You may certainly change Dozenal Time’s version number. I don’t know whether I can without unwanted consequences.

—
Paul Rapoport

numerist:
The update also added "allow_emulator":true to enable others to run the clock. Unfortunately, the update hasn't yet reached the general app loader.

PR #743, opened 2021-05-17
Sweep Clock: reduced memory footprint

awkirk71:
Was able to slash 12% from the memory footprint thanks to the RAM widget.


gfwilliams:
Great - thanks! It's really good to have these kinds of tweaks :)

PR #744, opened 2021-05-23
Font Clock: Initial Version

awkirk71:
User is able to choose between various fonts and clock faces in the upload page.


gfwilliams:
This looks amazing - thanks!

PR #745, opened 2021-05-24
multiclock - Added localisation to digi.js and timedat.js

neshanjo:
Hi there,

this is my first contribution, so I hope that I did not miss any requirements. If so, please tell me and I will fix my commit respectively.

I really like the multiclock timdat.js clock. However, it is not using the internationalisation mechanism. See also https://github.com/espruino/BangleApps/issues/136

Thus, I adapted the code - trying to change as few things as possible. I also adapted digi.js, which was quite simple. I did not touch txt.js since IMO it is alright to have this as an English skin and it would require a major effort to be adapted for other languages.

One technical remark: In timdat.js, a special date format consisting of day of week, month and day of month is used that is not available as a date pattern. Hence, I decided to build it there. I used this code together with the locales variable from [locales.js](https://github.com/espruino/BangleApps/blob/master/apps/locale/locales.js) to generate the dayFirst and withDot arrays:
```
const dayFirst = [];
const withDot = [];
for (key in locales) {
    const datePattern = locales[key].datePattern[1];
    if (datePattern.includes('.')) {
        withDot.push(locales[key].lang);
    }
    if (datePattern.indexOf('%d') < datePattern.indexOf('%m')) {
        dayFirst.push(locales[key].lang);
    }
}
console.log(withDot);
console.log(dayFirst);
```
Another possibility would be to add a third datePattern in locales.js if this is something useful in other places as well.

**Testing**: I've run the sanity check and tested the clock faces with no locale (default en_GB), de_DE and fr_FR.

Kind regards,
neshanjo

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2260,7 +2260,7 @@
   { "id": "multiclock",
     "name": "Multi Clock",
     "icon": "multiclock.png",
-    "version":"0.12",
+    "version":"0.13",
     "description": "Clock with multiple faces - Big, Analogue, Digital, Text, Time-Date.\n Switch between faces with BTN1 & BTN3",
     "readme": "README.md",
     "tags": "clock",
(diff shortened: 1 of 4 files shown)


gfwilliams:
This looks great - thank you!

PR #746, opened 2021-05-24
sliding text clock BUGFIX: turning emulator option off. 

awkirk71:
Emulator has stopped working because the clock is split into dynamic languages classes. Turning off the emulator otherwise people will think it doesn't work


gfwilliams:
Ok, thanks! At some point I'll try and get the emulator to handle apps with multiple files, but right now this doesn't work :(

PR #747, opened 2021-05-25
cliock - using locale for internationalization

neshanjo:
Hi,

I discovered that the CLI clock did not show the correct info when using the German locale, so I fixed it by using the locale mechanism, see https://www.espruino.com/Bangle.js+Locale

It should now also work with other languages. I tested with default (en_GB) and German language.

Cheers,
neshanjo


gfwilliams:
Great, thanks! This really tidies things up!

PR #751, opened 2021-05-30
T-Rex: Made x movements work and added highscore

neshanjo:
Hi there!

I really like this little game and added a highscore feature. I also enabled the move forward and backward buttons.

Highscore can be reset in app settings -> T-Rex. One little detail that I did not get to work is to show the T-Rex menu after reset, not going back (line 12 in settings.js -> what to put there to show the submenu?)

Any comments are welcome.

Kind regards,
neshanjo

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -362,13 +362,17 @@
   { "id": "trex",
     "name": "T-Rex",
     "icon": "trex.png",
-    "version":"0.02",
+    "version":"0.03",
     "description": "T-Rex game in the style of Chrome's offline game",
     "tags": "game",
     "allow_emulator":true,
     "storage": [
       {"name":"trex.app.js","url":"trex.js"},
-      {"name":"trex.img","url":"trex-icon.js","evaluate":true}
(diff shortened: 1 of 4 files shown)


gfwilliams:
That's great - thanks! Sorry for the delay, I was off last week.

It's odd about 'back' - does `'< Back'` work ok? It'd just the going back after a delay that causes a problem?

neshanjo:
I'm impressed how fast you usually answer. No need to be sorry for the delay! Also, this is not something important.

Regarding the menu structure, I created this UI flow image to illustrate the point:

TREX settings

In the current implementation, after clicking yes or no, one is directed to the settings menu with all app settings (orange path), but I'd rather have the T-REX settings menu shown again (lower part, green line).

Is this possible?

neshanjo:
Thanks! This was the missing piece. I added it to the previous commit, force-pushed to my branch and now it's ready to be merged.

gfwilliams:
Great - thanks!

PR #757, opened 2021-06-06
fix font size for firmware v2.07

fbedussi:


gfwilliams:
Just to be sure, is this fix for recent firmwares? Or does it make it work for 2v07 but then break 2v09/etc?

gfwilliams:
Any update on this? Was it just for 2v07, or is it actually for the most recent firmwares?

gfwilliams:
Just tested and it's for latest fw.

PR #758, opened 2021-06-10
barclock: use timeout instead of interval, add screenshots

rigrig:
Use `Date.getMilliseconds()` to refresh exactly when the second changes. I doubt anyone will notice a difference, but it bugged me to *know* it wasn't ticking exactly on the second ;-)   
Also add some screenshots, because everybody likes pictures.

(Also remove a duplicate `"readme": "README.md",` line from apps.json.)


gfwilliams:
Great - thanks!

PR #759, opened 2021-06-10
updated README for gpsservice app, to state its superceded by gpssetup

hughbarney:


gfwilliams:
Thanks! Is it work just removing this completely now?

PR #770, opened 2021-07-09
Add support for dismissing Gadgetbridge notifications through notify

epw:
Most of the support was already there, so this is just a few small steps:
- `notify.js` gets a `dismiss_and_hide()` exported function that can send the "notify-" event to Gadgetbridge, if installed
- `dismiss_and_hide()` is called when the Bangle.js screen is touched, instead of just `hide()`
- Gadgetbridge's `widget.js` saves "notify"-type events to a variable
- Gadgetbridge's `widget.js` sends a `DISMISS` action back to the Android app on "notify-" events, using the saved event data

Note that this is a limited implementation. For example, only the most recent notfication is saved (when a new one comes in, the old one is overridden in the variable), so it will be dismissed on the phone. A more advanced implementation might make it possible to choose which notification to dismiss.

From my own testing, this does properly count as a "swipe-away" style dismiss event in Android, so apps that react to notifications being swiped away will be triggered as expected.

### DIFF
--- a/apps/gbridge/widget.js
+++ b/apps/gbridge/widget.js
@@ -157,9 +157,9 @@
         currentNot = prettifyNotificationEvent(event);
         require("notify").show(currentNot);
         if (!(require('Storage').readJSON('setting.json',1)||{}).quiet) {
-            Bangle.buzz();
+          Bangle.buzz();
         }
-	break;
+        break;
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! This looks really neat

PR #771, opened 2021-07-11
some fixes/ improvement and a couple of clones 

FxIII:
#### Fixes:

 -  BlackJack game: ignoring buttons events on pauses

#### Improvements:

  - Desktop Launcher: reset to the clock after 2 mins of inactivity 
       - sometimes you may have the launcher mistakenly up. if so after some inactivity it comes back to the main screen
  -  App Manager: Allow negative numbers when manual-sorting
       - negative numbers when sorting can be useful with Desktop Launcher to manage secondary applications

#### Clones:

  - Magic 8 Ball Italiano: like Magic 8 Ball but in Italian :D 
        - my children love 8 ball but can barely read Italian. I have no idea how to handle it better than cloning
  - Red torch: like torch but red :D 
        - amateur astronomers will like that


gfwilliams:
Great - thanks! Sorry for the delay - I was on holiday the last 2 weeks.

With the translations - there is something planned, but for the moment I think cloning is the best solution as you say!

PR #774, opened 2021-07-22
High Contrast Clock

peeweek:
A simple clock with huge digits 10x5 pixel font, with Switchable foreground/background using BTN1.
dis
Font is inspired from WipeOut / Ikaruga games.

~~I encountered draw issues with notify/notifyfs regarding redraw upon dismissing notifications, so I added a Bangle.emit('notifyHide') upon dismissing notifications. It solves issues for both.~~ (removed as lcdPower event is already good enough)

The clock updates only necessary parts of the screen however, it redraws entirely in the following cases:
* LCD back on (sometimes the graphics buffer was cleared for some reason)
* Upon Color flipping (BTN1 pressed)
* Upon Notification Dismissed

IMG_20210722_160539-01

IMG_20210721_205615-01


gfwilliams:
Thanks! This looks nice! I just wanted to check though - what issues did you have with normal notifications though?

I can understand 'notifyfs' being a bit broken, but `notify` actually uses a separate bit of the screen buffer, so shouldn't actually touch the main screen contents at all.

peeweek:
Oh you are right, I had issues in the first place with both notify.js but I rewrote the update loop. In the end I ended up adding this double check even for the non Fullscreen but it's probably not necessary any more. I will test extensively by adding the emit() only for Fullscreen and probably revert the changes I made to the non Fullscreen.

peeweek:
After testing, It was not necessary anymore, thanks for pointing that out, now only notifyfs got an update.

gfwilliams:
That's great - thanks for testing this out so well!

Are you still sure it's needed for the fullscreen notifications? It's just looking at the code, you have this in your clock:

```
Bangle.on('lcdPower', (on) => { if(on) redraw(); });
```

And the 'hide' for fullscreen notifications has:

```
  if (Bangle.isLCDOn() || !(require('Storage').readJSON('setting.json',1)||{}).quiet) {
    // flipping the screen off then on often triggers a redraw - it may not!
    Bangle.setLCDPower(0);
    Bangle.setLCDPower(1);
  }
```

So you'd hope that the `setLCDPower` calls should trigger your `lcdPower` handler which would give it a full redraw anyway?

peeweek:
It was supposed to, however, at a rather low occurence 15% rate (especially if screen was off for a long time) the event on lcd power on was not called. I tried to log it, quite unsuccessfully. Will do more testing. (Especially as I saw a recent fix with a double initialize of the clock while exiting settings, which added side effects on my BTN1 callback)

peeweek:
After some hours testing it, it seems to work properly now. I reverted all changes I made in both notify and notifyfs.

gfwilliams:
That's great - thanks for being so thorough with this! Merging now...

If you do manage to track down some circumstances in which `lcdPower` isn't called correctly, please let me know - if I can reproduce it here I can hopefully fix it :)

peeweek:
Thanks for your guidance helping me wrap that up, I'm kinda new into bangle development but it's really cool :). I'll keep you posted it somehow my issues are coming back.

PR #779, opened 2021-07-22
Added Fahrenheit Thermometer (please disregard last pull request)

ZenMondo:
I modified the Thermometer App to display the temperature in Fahrenheit instead of Celcius for Americans and others that do not use the metric system.

(the previous pull request should have gone to my own repository as I was testing and debugging. My apologies.  It is very broken)


gfwilliams:
Thanks! There is a `locale` module which should handle temperature (I should have used that for the original temperature app) (http://www.espruino.com/Bangle.js+Locale) but it's nice to have an app that displays in Fahrenheit even in a different locale

ZenMondo:
Thank you for accepting this app modification.   The real learning experience was getting everything right to add something to the app loader, it was a bit of a learning curve, but I got there.   I plan on making more!

PR #780, opened 2021-07-29
Solar Clock: initial version

awkirk71:
Initial version of the solar clock. Using your current or chosen location shows the Sun's sky position in a day long animation.


gfwilliams:
This looks awesome - thanks!

PR #784, opened 2021-08-10
Quiet mode schedule: add widget, simplify logic

rigrig:
### Scheduling logic simplification:
I initially copied the scheduling code from the alarm app, but switching quiet mode doesn't involve loading a separate app, so there is no need to keep track of when a schedule last fired: we can just look at the times to determine which comes next.

### Widget:
Add a widget to show if quiet mode is active (which is why it also touched the settings app: to update the widget when needed)   
Doesn't add a setting to disable the widget, as I figured with a changing schedule you always want to see if quiet mode is active.
mode: Silent mode: Alarms

I think it's all good, but it's way past my bedtime, so making this a draft, and I'll give it another look-over tomorrow ;-)


rigrig:
After some sleep, still looks ok :-)

gfwilliams:
Looks great - thanks!

PR #787, opened 2021-08-12
Adjust hand lengths to be relative to tick marks

ihewitt:
Ridiculously tiny change, adjusts hand lengths to be relative to tick marks for different screen sizes.


gfwilliams:
Thanks! Just merging - in future please could you bump the version number and add a line to ChangeLog? Then users will know there is something to update. I'll do this quickly now anyway

ihewitt:
argh, wasn't paying attention! my bad. will do, thanks!

PR #789, opened 2021-08-23
Adding nixie: a watch face for both Bangle 1 and Bangle 2

yngv27:
This clock uses a layered drawing scheme to achieve a nixie tube look for the digits. Bangle 1 has better colors and images for the background, Bangle 2 makes best of use of the 8 color scheme. Bangle 1, being a slower graphic system, is quite fun to watch draw as the digits fade in and out. 

This set also introduces m_vatch.js: I hope to make this its own directory, as it provides an engine for drawing digital clocks by implementing user provided functions for drawing the background, current time, and any data. It also can support a night mode, where the user can draw a darker themed clock to use at night. It can be set up to turn on night mode automatically if the watch is placed on its side, such as on a bedside table.

There 's also some code to manage the step counter, so that it's not lost during app restarts, and it automatically saves a day's data to a history file, and resets the counter to zero on a new day.


gfwilliams:
These look great - thanks! Really good to get some Bangle.js 2 apps as well!

However please could you modify the main `apps.json` file to include the JSON for the two watch faces? I think right now you've got `g26/apps.json` and `nixie/nixie.info` which aren't needed, and won't actually get the apps in the app loader.

Info at: https://www.espruino.com/Bangle.js+App+Loader#adding-your-app

gfwilliams:
Thanks! Just tried this and had to make some small changes to make it work in the app loader: https://github.com/espruino/BangleApps/commit/409444a85ce9c73ffe69f7cb166d63f4b21643cb

But it's in now and it looks awesome!

PR #790, opened 2021-08-23
Car Crazy App

Ronin0000:
Screenshot 2021-08-23 100720


gfwilliams:
This looks great - thanks! I know it's not quite as fun, but in future please could you have the README as normal text rather than an image? You can still insert HTML formatting. It just means it'll get indexed by search engines better, and non-english speakers can easily hit 'translate' on it

Ronin0000:
Ok, Thanks @gfwilliams! I will add the normal text to the README file. Thanks for adding the game!

PR #791, opened 2021-08-24
Update README.md

Ronin0000:
Made the image into text.


gfwilliams:
That's great - thanks!

PR #797, opened 2021-09-05
make apps compatible with 2v10

ps-igel:
add LCD timeout for apps using doublebuffered LCD mode with 2v10


gfwilliams:
Great - thanks!

PR #798, opened 2021-09-05
Layout: add support for "img" with object or ArrayBuffer Graphics object

rigrig:
Rendering these already worked, because we simply pass `l.src()` to `g.drawImage()`, so we just need to determine the correct size.

I guess the minified version also needs to be updated, which minifier should I use for that? (Because it would be nice if running it again on the repo code produces the same result ;-)


gfwilliams:
Thanks! Right now I just copy the unminified version across to `.min.js`. It seems there might be an interpreter issue that gets hit when the minified+pretokenised code for the module is executed, which I still need to look at

PR #800, opened 2021-09-11
Fix swipe direction in Welcome app

qucchia:
In the welcome app, when I swipe from right to left it goes back a page instead of going forward. This PR makes it work the right way.


gfwilliams:
Thanks! Just merged this change manually as this PR also included a 'timer' app

PR #804, opened 2021-09-14
Add shortcuts app

nebbishhacker:


gfwilliams:
Awesome! Thanks for doing this! I was meaning to do something exactly like this but I never had time. It's nice using `clockupdown` as well, since it'll work on the new Bangle too.

Although I guess there's also the option in future of allowing BTN2 to be reassigned. I know sometimes people have asked if that is possible.

PR #805, opened 2021-09-14
widclkbttm (digital clock on the bottom area)

dapgo:


jes:
Hi, the build is failing here because you forgot to add a trailing comma after the last entry in `apps.json` before adding your new one.

gfwilliams:
Thanks! Yes, I think whether this works will be very app dependent :)

PR #807, opened 2021-09-14
Update weather to use setUI, for shortcuts compatibility

nebbishhacker:


gfwilliams:
Thanks! I guess the only minor issue is other stuff (like the widget clock) will assume it is inside a clock app, but maybe that's no big deal

PR #808, opened 2021-09-14
Add fd6fdetect

Smooklu:
I created a app that allows you to see how many 0xFD6F (Exposure Notifcations) beacons are near you.


gfwilliams:
Thanks! That's a nice idea!

PR #806, opened 2021-09-14
Add choozi

jes:
This is a new app for Bangle.js.

It chooses people or things at random with a fun animation. For example, you can use it to decide who goes first in a board game. There's a video demo here: https://www.youtube.com/watch?v=4cqOLNM5ei8

I hope I did everything right, please let me know what to fix if not.

I also included a fix for a typo (nu => no) in the README.

It is inspired by "Chwazi" for Android.

PR #810, opened 2021-09-15
Add README to Choozi

jes:
As suggested in #806, I've added a README.

I didn't know this was possible, else I'd have done it the first time :).


gfwilliams:
Thanks!

PR #813, opened 2021-09-18
Fixed widbatpc for Bangle 2

hughbarney:


gfwilliams:
Thanks! It'd actually be really nice to purge all the 16 bit color constants from the apps - they're just totally unreadable.

PR #816, opened 2021-09-20
fix alarm widget for bangle 2, fix widpedom width when displaying large text

hughbarney:


gfwilliams:
Thanks!

However I'm a bit concerned about the pedometer changes.

`Bangle.drawWidgets` lays out all the widgets based on their `width` values then calls `.draw` on each widget in turn. If you're setting `.width` to something else in the draw call, you're just going to end up overwriting the edge of other widgets, until the next call to drawWidgets. You probably get away with it because you're the only (or last) widget in the top left, but if anyone adds another widget there it'll cause problems.

I'll add some tweaks in for this now though which I think should fix it.

PR #817, opened 2021-09-20
Weather: refactoring and reduced ram usage

nebbishhacker:
This should reduce the memory usage of the ram widget by up to 70
JSVars, depending on the weather.


gfwilliams:
Nice - thanks!

Issue #820, opened 2021-09-22
Layout: unable to control `fillx`/`filly` of parent layouts


nebbishhacker:
If `filly` is set on a child element, it seems to automatically get set for its parent layout as well. This makes some layouts difficult to achieve, and makes nested layouts a bit harder to reason about.

My preference would for `fillx`/`filly` to always default to `0` for parent layouts regardless of what their children are doing. Failing that, I would like to at least be able to manually override it.

---

```
const Layout = require("Layout");

g.clear();
var layout = new Layout({type:"h", filly: 0, c: [
  {type: "txt", font: "50%", label: "A"},
  {type:"v", c: [
    {type: "txt", font: "10%", label: "B"},
    {filly: 1},
    {type: "txt", font: "10%", label: "C"},
  ]},
]});
layout.render();
layout.debug();
```

Expected result:

![screenshot (19)](https://user-images.githubusercontent.com/2780794/134426848-3bb4139d-f9a5-4acf-8ac4-6943d1b4a22a.png)

Actual result:

![screenshot (20)](https://user-images.githubusercontent.com/2780794/134426991-d026e6c7-2b1a-453d-a2a5-0d068f2619a9.png)

---

I ran into this while trying to replicate the layout of the weather app.

gfwilliams:
Ahh, good point! Yes, that seems fair.

The reason it got added was for the acceleration logger: https://github.com/espruino/BangleApps/blob/master/apps/accellog/app.js#L100

Basically I had a simple vertical layout but wanted one heading to go right across so the background was all filled.

In your case, I guess you'd actually need to explicitly set fill on the vertical group, but not on the horizontal?

nebbishhacker:
> In your case, I guess you'd actually need to explicitly set fill on the vertical group, but not on the horizontal?

Yes, in this case the desired end result would be for the vertical group to have a filly of 1 and the horizontal group to have a filly of 0.

Thinking about it more, I'd say the current behavior makes sense in most cases (and, indeed, matches with my intuition more often than I realized) - so all I really want is to be able to override it by explicitly setting the parent fill.

Issue #819, opened 2021-09-22
Layout: `pad` can cause text to get cut off


nebbishhacker:
```
const Layout = require("Layout");
g.clear();
var layout = new Layout({type:"v", fillx: true, c: [
  {type: "txt", font: "10%", halign:1, pad: 20, label: "0123456789"},
]});
layout.render();
layout.debug();
```

Expected result:

![screenshot (17)](https://user-images.githubusercontent.com/2780794/134425340-604b75a1-b0f2-4597-9f11-b8787d06ecd4.png)

Actual result:

![screenshot (18)](https://user-images.githubusercontent.com/2780794/134425378-7ffb5e3e-616f-41a2-8655-2feb4f8e0ebd.png)

gfwilliams:
Thanks! That's interesting. I thought pad got added to the required width, but maybe not... I'm starting to wonder if it's worth implementing a test harness now (JS + expected output). It might also serve as a useful reference.

nebbishhacker:
Here's another test case that still exhibits issues.

```
const Layout = require("Layout");

g.clear();
var layout = new Layout({type: "v", c: [
  {type: "h", pad: 20, c: [
    {type: "txt", font: "10%", label: "abcd"},
    {fillx: 1},
    {type: "txt", font: "10%", label: "1234"},
  ]},
]});
layout.render();
layout.debug();
```

Expected:

![screenshot (22)](https://user-images.githubusercontent.com/2780794/134552241-37125beb-0f0d-45b8-837e-756ad1c1e0fc.png)

Actual:

![screenshot (21)](https://user-images.githubusercontent.com/2780794/134551911-8ffcc35b-4022-42d9-a118-79e371d3c8d4.png)

PR #821, opened 2021-09-24
Layout fixes

nebbishhacker:
Fixes for #819 and #820, plus tests.


gfwilliams:
That's awesome, thanks!

PR #822, opened 2021-09-26
CSC sensor app

nxdefiant:
add support for cadence sensor


gfwilliams:
This is great - thanks!

PR #823, opened 2021-09-27
Score Tracker - Create new app 'score'

Xervon:
This app is a score tracker that should work for a ton of different sports.

Defaults are currently set for badminton but the settings should be able to be adjusted for other sports.

The app should almost be compatible with Bangle.js 2 as well. I just don't quite know how touchscreen drag is implemented to enable scrolling.

My own Bangle.js hasn't actually arrived yet, so I have only been able to test this in the emulator. Before merging I'd like to wait until me or someone else is able to test this on actual hardware. The app could also do with a bit of documentation....


Xervon:
I've added tennis scoring as well, though still no documentation. I might get to that tomorrow.

gfwilliams:
Great, thanks! Bit busy at the moment but I'm hoping to get an emulator for the new Bangle up and running soon, so we should be able to give it a shot on that then too.

Xervon:
I've added tiebreakers as a feature as well now. With this the app should be feature-complete.

Documentation is also done. So the only thing left now would be testing b2 compatibility.

Xervon:
I wanted to get these changes off of my local master branch. Please see #828.

PR #824, opened 2021-09-27
Lazy Layout: don't assume bg starts out cleared

nebbishhacker:


gfwilliams:
Thanks!

gfwilliams:
Does this mean there's no need to call `g.clear()` at all in an app now?

nebbishhacker:
Yes, as long as the outermost layoutObject fills the screen and has `bgCol` set.

PR #825, opened 2021-09-27
weather: Use Layout library

nebbishhacker:
I took a stab at updating the weather app to use the new Layout library.

Some notes:

- Experimenting with different layout variations is very convenient!
- I found a small hack to sort-of-emulate different screen sizes: `g.getWidth=()=>176;g.getHeight=()=>176;`
- The initial layout / render adds a rather noticeable delay to the app's loading time
  - This is made worse by the fact that I construct the layout with placeholder labels (which implicitly calls `layout.update()`), and then I immediately fill out the labels with actual data (if available) and call `update()` again. It would be nice if there was a way to tell the Layout constructor to skip that initial `update()` call.
- I could see memory usage becoming an issue on the Bangle 1. On my watch this change increases memory usage from 44% all the way to 74%. I know you were thinking of rewriting the Layout library in C in order to improve performance - would it be possible to optimize the memory usage as well?
- If anything other than the layout object clears the screen, the lazy rendering state needs to be cleared. Right now I'm just doing `delete layout.rects`, but I should probably add a method for that.
- If you change the length of some text that's left- or right-aligned using `halign`, and you still want it to be properly aligned, you need to call `layout.update()` again. This feels like overkill when I'm just trying to update a single line in the corner of the screen. Being able to align the text itself within the bounds of the `txt` object would help.
- Multiline text with line wrapping would be handy. If you would like, I think I could port that over from my `trui` layout library. It would require a bit of modification to the layout algorithm to do properly, though, since the wrapping text would need to know how much horizontal space is available before it can calculate its minimum height. The simplest solution would probably be to calculate `_w` and `w` for all objects in a separate pass before calculating `_h` and `h`.

screenshot (24)

screenshot (25)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -585,7 +585,7 @@
   { "id": "weather",
     "name": "Weather",
     "icon": "icon.png",
-    "version":"0.09",
+    "version":"0.10",
     "description": "Show Gadgetbridge weather report",
     "readme": "readme.md",
     "tags": "widget,outdoors",
--- a/apps/weather/ChangeLog
+++ b/apps/weather/ChangeLog
@@ -6,3 +6,4 @@
 0.07: Add theme support and unknown icon.
 0.08: Refactor and reduce widget ram usage.
 0.09: Fix crash when weather.json is absent.
+0.10: Use new Layout library
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -27,7 +27,7 @@ var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
     ]},
   ]},
   {filly: 1},
-  {type: "txt", font: "9%", id: "cond", label: "Weather line 1"},
+  {type: "txt", font: "9%", wrap: true, height: g.getHeight()*0.18, fillx: 1, id: "cond", label: "Weather condition"},
   {filly: 1},
   {type: "h", c: [
     {type: "txt", font: "6x8", pad: 2, id: "loc", label: "Toronto"},
@@ -54,7 +54,6 @@ function draw() {
   const wind = locale.speed(current.wind).match(/^(\D*\d*)(.*)$/);
   layout.wind.label = wind[1];
   layout.windUnit.label = wind[2] + " " + current.wrose.toUpperCase();
-  // TODO: split long weather conditions across multiple lines
   layout.cond.label = current.txt.charAt(0).toUpperCase()+current.txt.slice(1);
   layout.loc.label = current.loc;
   layout.updateTime.label = `${formatDuration(Date.now() - current.time)} ago`;
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -10,7 +10,7 @@ var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
   {type: "h", filly: 0, c: [
     {type: "custom", width: g.getWidth()/2, height: g.getWidth()/2, valign: -1, txt: "unknown", id: "icon",
       render: l => weather.drawIcon(l.txt, l.x+l.w/2, l.y+l.h/2, l.w/2-5)},
-    {type: "v", c: [
+    {type: "v", fillx: 1, c: [
       {type: "h", pad: 2, c: [
         {type: "txt", font: "18%", id: "temp", label: "000"},
         {type: "txt", font: "12%", valign: -1, id: "tempUnit", label: "°C"},
@@ -30,9 +30,9 @@ var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
   {type: "txt", font: "9%", wrap: true, height: g.getHeight()*0.18, fillx: 1, id: "cond", label: "Weather condition"},
   {filly: 1},
   {type: "h", c: [
-    {type: "txt", font: "6x8", pad: 2, id: "loc", label: "Toronto"},
+    {type: "txt", font: "6x8", pad: 4, id: "loc", label: "Toronto"},
     {fillx: 1},
-    {type: "txt", font: "6x8", pad: 2, id: "updateTime", label: "15 minutes ago"},
+    {type: "txt", font: "6x8", pad: 4, id: "updateTime", label: "15 minutes ago"},
   ]},
   {filly: 1},
 ]}, null, {lazy: true});
@@ -76,9 +76,9 @@ function update() {
   } else {
     layout.forgetLazyState();
     if (NRF.getSecurityStatus().connected) {
-      E.showMessage("Weather unknown\n\nIs Gadgetbridge\nweather reporting\nset up on your\nphone?");
+      E.showMessage("Weather\nunknown\n\nIs Gadgetbridge\nweather\nreporting set\nup on your\nphone?");
     } else {
-      E.showMessage("Weather unknown\n\nGadgetbridge\nnot connected");
+      E.showMessage("Weather\nunknown\n\nGadgetbridge\nnot connected");
       NRF.on("connect", update);
     }
   }
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -74,7 +74,7 @@ function update() {
   if (current) {
     draw();
   } else {
-    delete layout.rects;
+    layout.forgetLazyState();
     if (NRF.getSecurityStatus().connected) {
       E.showMessage("Weather unknown\n\nIs Gadgetbridge\nweather reporting\nset up on your\nphone?");
     } else {
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -1,96 +1,107 @@
-(() => {
-  const weather = require('weather');
-  let current = weather.get();
+const Layout = require('Layout');
+const locale = require('locale');
+const weather = require('weather');
+let current = weather.get();
 
-  function formatDuration(millis) {
-    let pluralize = (n, w) => n + " " + w + (n == 1 ? "" : "s");
-    if (millis < 60000) return "< 1 minute";
-    if (millis < 3600000) return pluralize(Math.floor(millis/60000), "minute");
-    if (millis < 86400000) return pluralize(Math.floor(millis/3600000), "hour");
-    return pluralize(Math.floor(millis/86400000), "day");
-  }
+Bangle.loadWidgets();
 
-  function draw() {
-    g.reset();
-    g.clearRect(0, 24, 239, 239);
+var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
+  {filly: 1},
+  {type: "h", filly: 0, c: [
+    {type: "custom", width: g.getWidth()/2, height: g.getWidth()/2, valign: -1, txt: "unknown", id: "icon",
+      render: l => weather.drawIcon(l.txt, l.x+l.w/2, l.y+l.h/2, l.w/2-5)},
+    {type: "v", c: [
+      {type: "h", pad: 2, c: [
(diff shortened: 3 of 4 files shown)


gfwilliams:
Hi,

Thanks! 

> The initial layout / render adds a rather noticeable delay to the app's loading time
> It would be nice if there was a way to tell the Layout constructor to skip that initial update() call.

Yes, I'm very happy to do that...

> On my watch this change increases memory usage from 44% all the way to 74%.

Wow, yes, that's not good at all. Do you have any idea where the memory usage is coming from specifically? I guess if the layout module were internal we could actually make each element an instance of `LayoutElement` and could then store all the data as binary, but that's a lot of work.

> If anything other than the layout object clears the screen, the lazy rendering state needs to be cleared.

I guess that'd be a good idea, yes... Maybe just an option to force a full redraw?

> If you change the length of some text that's left- or right-aligned using halign ... you need to call layout.update() again

Yes, that's not great. I'm pretty sure we could get a fix in for that one though.

> Multiline text with line wrapping would be handy. 

That would be really neat. What about just only allowing it if `fillx` is set, or something like that? So in the wrapped case the text doesn't really affect the size of the area, it just fills the area that is available.

gfwilliams:
Just did a quick check with your weather app:

```
>E.getSizeOf(layout)
=808
>E.getSizeOf(layout,1)
=[
  {
    name: "__proto__",
    size: 118 },
  {
    name: "l",
    size: 503 },
  {
    name: "_l",
    size: 503 },
  {
    name: "b",
    size: 2 },
  {
    name: "physBtns",
    size: 2 },  ...
  {
    name: "windUnit",
    size: 26 },
  {
    name: "cond",
    size: 23 },
  {
    name: "loc",
    size: 23 },
  {
    name: "updateTime",
    size: 26 },
  {
    name: "rects",
    size: 167 }
 ]
```

So the lazy rendering does add about 20% memory usage. However one interesting issue is that there are a lot of floating pointer numbers, and these actually hit memory usage quite hard:

```
>E.getSizeOf([ 40.5, 176.5, 198.5, 197.5 ])
=9
>E.getSizeOf([ 40, 176, 198, 197 ])
=5
```

I'll commit some layout changes in a second, but by changing `/2` to a `>>1` I've got usage from 808 down to 595

```
>E.getSizeOf(layout)
=595
```

gfwilliams:
Well, this sucks. Not sure if you have any ideas, but the tests fail with these changes. As far as I can tell, one of the `valign` tags in your all-in-one test is now in the wrong place.

Otherwise it looks ok - it's a bit of a challenge as you need to keep the floats around to ensure that when you have multiple `fill` elements, you don't end up with spare space at the bottom due to rounding.

```
--- a/modules/Layout.js
+++ b/modules/Layout.js
@@ -237,29 +237,29 @@ Layout.prototype.layout = function (l) {
   // exw,exh = extra width/height available
   switch (l.type) {
     case "h": {
-      var x = l.x + (0|l.pad);
+      var w,x = l.x + (0|l.pad);
       var fillx = l.c && l.c.reduce((a,l)=>a+(0|l.fillx),0);
       if (!fillx) { x += (l.w-l._w)/2; fillx = 1; }
       l.c.forEach(c => {
-        c.w = c._w + ((0|c.fillx)*(l.w-l._w)/fillx);
+        c.w = Math.ceil(w = c._w + ((0|c.fillx)*(l.w-l._w)/fillx));
         c.h = c.filly ? l.h - (l.pad<<1) : c._h;
-        c.x = x;
-        c.y = l.y + (0|l.pad) + (1+(0|c.valign))*(l.h-(l.pad<<1)-c.h)/2;
-        x += c.w;
+        c.x = 0|x;
+        c.y = l.y + (0|l.pad) + ((1+(0|c.valign))*(l.h-(l.pad<<1)-c.h)>>1);
+        x += w;
         if (c.c) this.layout(c);
       });
       break;
     }
     case "v": {
-      var y = l.y + (0|l.pad);
+      var h,y = l.y + (0|l.pad);
       var filly = l.c && l.c.reduce((a,l)=>a+(0|l.filly),0);
       if (!filly) { y += (l.h-l._h)/2; filly = 1; }
       l.c.forEach(c => {
         c.w = c.fillx ? l.w - (l.pad<<1) : c._w;
-        c.h = c._h + ((0|c.filly)*(l.h-l._h)/filly);
-        c.y = y;
-        c.x = l.x + (0|l.pad) + (1+(0|c.halign))*(l.w-(l.pad<<1)-c.w)/2;
-        y += c.h;
+        c.h = Math.ceil(h = c._h + ((0|c.filly)*(l.h-l._h)/filly));
+        c.y = 0|y;
+        c.x = l.x + (0|l.pad) + ((1+(0|c.halign))*(l.w-(l.pad<<1)-c.w)>>1);
+        y += h;
         if (c.c) this.layout(c);
       });
       break;

```

nebbishhacker:
Good thinking about avoiding floating point numbers. I've submitted my own take on that as #829. It's interesting to note that even when a number looks like an int, it doesn't necessarily mean it's being stored as an int...

It should be possible to store `rects` more efficiently by packing the data into strings or arraybuffers, but I don't know if it's worth getting into that right now.

>> Multiline text with line wrapping would be handy.
> 
> That would be really neat. What about just only allowing it if fillx is set, or something like that? So in the wrapped case the text doesn't really affect the size of the area, it just fills the area that is available.

The height would need to be defined too, so the requirement would be something like `(fillx || w) && (filly || h)`. That might be a good place to start, but ultimately it would be nice to be able to just set `fillx` and have it automatically figure out how much vertical space the text needs.

gfwilliams:
>  It's interesting to note that even when a number looks like an int, it doesn't necessarily mean it's being stored as an int...

That's interesting... potentially that's something I could look at optimising in Espruino.

> It should be possible to store rects more efficiently by packing the data into strings or arraybuffers

Yes... Arraybuffers aren't too efficient for small arrays (because they have a 'header' with size/type/offset) but strings would work great, especially as we know that all coordinates are below 256. I wonder if there are speed implications, but `clearRect.apply(g, rect.split("").map(x=>x.charCodeAt()))` would probably be ok.

For now, storing 'bg' in the array too works ok I think, and should save 2 memory elements per rect. I've not committed this yet as I can't find your lazy rendering test right now, but this kind of thing:

```
diff --git a/modules/Layout.js b/modules/Layout.js
index b5afbded..4b2e502e 100644
--- a/modules/Layout.js
+++ b/modules/Layout.js
@@ -162,10 +162,8 @@ function prepareLazyRender(l, rectsToClear, drawList, rects, parentBg) {
     if (c) l.c = c;
 
     if (!delete rectsToClear[hash]) {
-      rects[hash] = {
-        bg: parentBg == null ? g.theme.bg : parentBg,
-        r: [l.x,l.y,l.x+l.w-1,l.y+l.h-1]
-      };
+      var r = rects[hash] = [l.x,l.y,l.x+l.w-1,l.y+l.h-1];
+      r.bg = parentBg == null ? g.theme.bg : parentBg,
       if (drawList) {
         drawList.push(l);
         drawList = null; // Prevent children from being redundantly added to the drawList
@@ -225,7 +223,7 @@ Layout.prototype.render = function (l) {
     prepareLazyRender(l, rectsToClear, drawList, this.rects, null);
     for (var h in rectsToClear) delete this.rects[h];
     var clearList = Object.keys(rectsToClear).map(k=>rectsToClear[k]).reverse(); // Rects are cleared in reverse order so that the original bg color is restored
-    for (var r of clearList) g.setBgColor(r.bg).clearRect.apply(g, r.r);
+    for (var r of clearList) g.setBgColor(r.bg).clearRect.apply(g, r);
     drawList.forEach(render);
   } else { // non-lazy
     render(l);
```

> wrapping

Yes, totally. I think initially we make width&height required, and then maybe later it can be extended.

nebbishhacker:
> can't find your lazy rendering test right now

This? https://www.espruino.com/ide/?gist=c9e3ecbea2dfb149314c0130b7698665

gfwilliams:
Ok, great - it works, so pushed the changes now. Another annoyance is object keys >4 chars long take up 2 vars:

```
>trace({fillx:1})
#6481[r0,l1] Object {
  #6647[r1,l2] Name String [2 blocks] "fillx"= int 1
}
=undefined
>trace({fx:1})
#7958[r0,l1] Object {
  #7228[r1,l2] Name String [1 blocks] "fx"= int 1
}
```

So maybe it's worth changing fillx/halign/etc. Not sure?

nebbishhacker:
> So maybe it's worth changing fillx/halign/etc

I guess that depends on whether you'd rather prioritize ram usage or readability. If you do make that change I'd suggest doing it now, while there's still only a few apps using the library.

You could also potentially save some memory by deleting the `_w` and `_h` properties when they're no longer needed.

gfwilliams:
> If you do make that change I'd suggest doing it now, while there's still only a few apps using the library.

Yes, I know... maybe it is a step too far - unless you can think of any equally readable names that fit in 4 chars :)

> You could also potentially save some memory by deleting the _w and _h properties when they're no longer needed.

Yes, good point. Do you think it's possible though? It seems we might need them for updating the layout if the width of something changed in the future?

nebbishhacker:
> It seems we might need them for updating the layout if the width of something changed in the future?

I think they would only be needed if we want to try doing incremental updates, where only a portion of the layout is recalculated. As things stand now, though, `update()` always recalculates `_w`/`_h` for the entire layout anyways.

gfwilliams:
Good point! I guess all we'd need to do is remove them in `Layout.prototype.layout` and we're sorted.

gfwilliams:
Do you want to commit this now? I think with the recent changes the memory usage is lower and it makes sense to add this in?

nebbishhacker:
I'd still like to make a couple tweaks, but I think it makes sense to merge this in soon.

This isn't a blocker, but I'm still a bit concerned about performance.. Right with on my watch with #841 applied it takes a bit over a second between when the app starts loading and when it shows up on screen. I would say this is acceptable, but not ideal.

About a quarter second of the delay comes from having to load widgets before it can create the layout - previously I would defer loading widgets until after the app itself rendered in order to reduce the perceptual delay, but if I do that with the Layout library it doesn't leave space on top for the widgets.

gfwilliams:
Eep, yes a second is a bit much. Do you have any idea where the other delays come from?

Also, is it the same if you upload with the app loader? The app loader should minify and pretokenise all the libraries which should help a reasonable amount with performance.

> widgets

Does loading them first, then drawing them later help at all?

I have a hack for the boot code which squashes all the `.boot.js` files together into one in order to speed up boot time, and the same could be done for widgets which I guess would help a bit with the loading...

nebbishhacker:
> Do you have any idea where the other delays come from?

I believe the rest of the delay is mostly from creating/updating/rendering the layout.

> Also, is it the same if you upload with the app loader?

Seems about the same. I've been developing with minification/protokenization enabled in the IDE too.

> Does loading them first, then drawing them later help at all?

That helps a bit, but I was already doing that.

> I have a hack for the boot code which squashes all the .boot.js files together

Is that why my watch occasionally says something along the lines of "updating boot0" when it loads? I was wondering about that.

nebbishhacker:
I'm not going to have time to contribute much over the next couple weeks (getting married!) so I'll leave it to you to decide whether this is worth merging as-is, or if you want to find some way to improve the performance first.

gfwilliams:
Wow, congratulations! Hope you have a great time!

I'll merge this as-is. Performance is something I can work on in the Layout lib and it's nice to have something in the repo that actually stresses it a bit :)

PR #827, opened 2021-09-27
Pastel: handle 12hr display at 12:xx, updated description for notify default and fullscreen for clarity

hughbarney:


gfwilliams:
Thanks!

PR #829, opened 2021-09-29
Layout: avoid floating point numbers

nebbishhacker:
The previous approach involved calculating the width of each element independently, and deriving the `x` coordinate of each element by summing the previous widths. However, if the widths are rounded, summing them leads to accumulated rounding errors.

This change effectively takes the alternate approach of calculating the `x` coordinate of each element independently, and then deriving the widths from the differences between adjacent values of `x`.

As a bonus, this change corrects a couple issues in the `padding_with_fill` test.


gfwilliams:
Great - thanks!

PR #831, opened 2021-09-30
Pastel fixed for Gadgetbridge on a Bangle 2

hughbarney:


gfwilliams:
Thanks! Odd thing is the lcdPower test shouldn't be needed, since I don't believe you get the events on Bangle.js 2 - instead you get 'lock' events

PR #833, opened 2021-10-02
barclock: use theme colors, Layout library

rigrig:
Also simplifies redraw logic and adds semicolons


gfwilliams:
Wow, thanks! That's really tidied it up!

PR #835, opened 2021-10-02
Gadgetbridge Troubleshooting in README, Findphone fixed for Bangle 2

hughbarney:


gfwilliams:
Great - thanks for that!

PR #836, opened 2021-10-03
Fix translation slidingtext.locale.es.js

FranciscoTrigo:
Fixed the words for 7, SEITO is now SIETE.
20, VEINTA is now VEINTE


gfwilliams:
Awesome - thanks!

PR #837, opened 2021-10-03
settings: customize theme

rigrig:
You get to pick from these colours:
```
black: "#000", white: "#fff",
red: "#f00", green: "#0f0", blue: "#00f",
cyan: "#0ff", magenta: "#f0f", yellow: "#ff0", 
```
I guess maybe a later version can add a proper rgb color picker

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -185,7 +185,7 @@
   { "id": "setting",
     "name": "Settings",
     "icon": "settings.png",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Looks awesome, thanks! Sorry - I missed this in the barrage of emails over the weekend. Will merge now

PR #840, opened 2021-10-04
Layout: Implement text wrapping

nebbishhacker:
In the future we may want to support:
- text alignment
- text rotation
- calculation of height based on width and text contents
- line spacing


gfwilliams:
Nice - thanks!

PR #842, opened 2021-10-06
Grocery: Refactor to store grocery list in separate file

avanc:


gfwilliams:
Looks great - thanks!

PR #845, opened 2021-10-13
widbatpc - added option to fill the bar with color so the figures dont show in 2 colours

hughbarney:


gfwilliams:
Thanks! Just a note that you can use `g.dump()` on Bangle.js 2 to get a screenshot - should be a lot easier than taking a photo :)

PR #850, opened 2021-10-18
svclock - scaling for the banglejs 2 and locale support 

DeuxEtienne:


gfwilliams:
Great - thank you!

gfwilliams:
Probably:

```
switch (process.env.BOARD) {
  case "EMSCRIPTEN":
```

I guess? The real boards are named differently - the board name in the emulator is emulator specific. I'll make some changes here as it could do with having the new redraw timing code in anyway

PR #856, opened 2021-10-21
feat: set compass power off when screen is off

VSerain:


gfwilliams:
Thanks! No, all I meant was to stop the compass, which I think you have done?

PR #871, opened 2021-10-29
menusmall `wrap` option

rigrig:
Same as EspruinoDocs graphical_menu.js: https://github.com/espruino/EspruinoDocs/pull/614

I think I also got the appRect tweaks, seemd to look allright in the emulator...   
I did notice it got rid of `Bangle.drawWidgets();` though, not sure if that is ok?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4059,7 +4059,7 @@
   {
     "id": "menusmall",
     "name": "Small Menus",
-    "version": "0.01",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great - thanks! Yes, getting rid of drawWidgets should be fine - since it clears the apprect it won't clear the widget bar now so shouldn't need to redraw them

PR #873, opened 2021-11-01
menusmall: follow latest E_showMenu improvements

rigrig:
Apply https://github.com/espruino/EspruinoDocs/pull/616 / https://github.com/espruino/Espruino/pull/2085

Only small optimizations, so I figured skipping the version bump would be ok.

- remove unused variable
- remove unused parameter from `select()`
- shave a few bytes off `move()`

If a numerical item has `item.wrap` we no longer check if `item.min/max` is set, because your menu will be broken either way.


gfwilliams:
Thanks!

PR #874, opened 2021-11-01
Multiclock app update to run on Bangle and Bangle 2

jeffmer:
I have used setUI themes and scaling so that it works on both Bangle and Bangle 2 using the latest firmware - tested!  The default Background 2 colour on the Bangle seems to be white - why not a bright color - cyan or magenta?

Overall, the portability features work really well - I used "left-right" with setUI. I did not feel the need for a layout as scaling seems to solve nearly all problems.

I tried to get your floral clock twhich sets the background theme colour to run as a face but failed to find an elegant way to reset the theme over face switches - I also worried that it would affect alerts/notifications. . Is there a simple way of restoring the default theme set in settings?


gfwilliams:
Great - thanks! Good point about the bg2 colour - I'll get it set to something else.

Does `g.setTheme((require("Storage").read("setting.json",1)||{}).theme)` work to restore the theme?

jeffmer:
Yes that works if you substitute readJSON for read:). Thanks.

PR #876, opened 2021-11-03
create Settings library

rigrig:
Something like this would be useful, as mentioned in http://forum.espruino.com/conversations/369160/#16245740

This works, but some issues/bikeshedding before merging:
- Maybe remove (some of) the error checks to make it lighter. I really don't know how much that would improve Things, as I think most of this would just be executed from Flash anyway? But it seems a bit much compared to most of the other code...
- To get at global settings, you need to pass the `Bangle` object as app, we could pick some other value (or just anything falsey)
- There's probably room for some optimization before pushing this out for widespread use.
- I figured that for malformed JSON, throwing the error would be more informative than just returning the default value, as someone should probably take a look at how that happened?

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,40 +377,37 @@ that handles configuring the app.
 When the app settings are opened, this function is called with one
 argument, `back`: a callback to return to the settings menu.
 
-Usually it will save any information in `app.json` where `app` is the name
+Usually it will save any information in `myappid.json` where `myappid` is the name
 of your app - so you should change the example accordingly.
 
 Example `settings.js`
 ```js
 // make sure to enclose the function in parentheses
 (function(back) {
-  let settings = require('Storage').readJSON('app.json',1)||{};
-  function save(key, value) {
-    settings[key] = value;
-    require('Storage').write('app.json',settings);
-  }
+  function get(key, def) { return require('Settings').get('myappid', key, def); }
+  function set(key, value) { require('Settings').set('myappid', key, value); }
   const appMenu = {
     '': {'title': 'App Settings'},
     '< Back': back,
     'Monkeys': {
-      value: settings.monkeys||12,
-      onchange: (m) => {save('monkeys', m)}
+      value: get('monkeys', 12),
+      onchange: (m) => set('monkeys', m)
     }   
   };
   E.showMenu(appMenu)
 })
 ```
-In this example the app needs to add `app.settings.js` to `storage` in `apps.json`.   
-It should also add `app.json` to `data`, to make sure it is cleaned up when the app is uninstalled.
+In this example the app needs to add `myappid.settings.js` to `storage` in `apps.json`.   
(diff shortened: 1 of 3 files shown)


rigrig:
Also, the sanitycheck complains about using `<app>.settings.json` instead of `<app>.json`:
https://github.com/espruino/BangleApps/blob/master/bin/sanitycheck.js#L209-L211

`<app>.settings.json` does seem like the better option though, in case we want to use `<app>.json` for non-setting data?

gfwilliams:
I wonder whether for the path, it's almost better to just not allow '.'? Folks can always do:

```
require('Settings').get('test', 'foo').bar.baz
```

I guess I just think we should try and make 90% of things easier rather than trying a kitchen sink approach? We could always add `.` later in a compatible way, but it's hard to go backwards once people use it :)

On the settings front, I think I felt initially as we only have 28 chars of filename, `.settings.json` ate into that quite a lot - although it does make things a bit more explicit. I think changing the sanity check to avoid the warning could be a good plan though.

rigrig:
>  Folks can always do:
> `require('Settings').get('test', 'foo').bar.baz`

Not if `foo` or `bar` are missing, then you get an error. My main idea was that apps can just do  `require('Settings').get('test', 'foo.bar.baz', 'default')`, without the need to setup default values anywhere else.


Also I just thought of another interface which seems cleaner: rewrite it so you can do 
```
var appSettings = require("Settings").open("app");
var foo = appSettings.get('foo');
appSettings.set('bar', 123);

var globalSettings = require("Settings").open(); // no need for "global" value: we can check arguments.length
```

rigrig:
Updated version:
- got rid of the `path` stuff. You're right about adding it later, and it seems there aren't any apps with nested settings anyway: only global settings has them, and adding it just for that seems a bit much.
- I changed it so now you do
```js
var s = require('Settings').app(appname);
var foo = s.get('foo'), bar = s.get('bar')
s.set('foo', 1234;

var g = require('Settings').Bangle();
var v = g.get('vibrate');
```

On the one hand this seems cleaner for apps, but I guess it does eat into RAM a bit, especially if all widgets use it :-(
Maybe better to just revert it to `require('Settings').get(app,key)`?

rigrig:
Changed it back to `require('Settings').get(app, key, default)`, added a hint that apps can create their own wrapper function.

@gfwilliams: I'd be happy with the code as it is now, unless you have some suggestions on improving it? Otherwise I'll add some documentation (+changes/fixes) before marking it as ready for review.

rigrig:
I changed it to use `appid.json` after all, mainly because the README already tells people to use that (which is probably also why the sanitycheck complains about `appid.settings.json`), and it feels wrong to say "here's a handy library, but it only works for apps that *didn't* use the suggested filenames".

gfwilliams:
Thanks! Sorry for the delay looking at this again - it seems great. Merging now

PR #877, opened 2021-11-04
Fixed wrong day being displayed in nifty clocks

ff2005:
Small fix where the day of the week is display instead of day of the month.


gfwilliams:
:) thanks!

PR #878, opened 2021-11-11
apps.json fix: fwupdate is for Bangle.js 2 only

rigrig:
This app used to be tagged `b2` only, but got both devices during the great `tags`->`supports` update: https://github.com/espruino/BangleApps/commit/a94790391a9f13ab83c9658ba48080b9d5967190#diff-70e375511009c0cb767fdc64667e148359c6d6b7b8cf4f100b7eb0e853f83529L8-R10

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -7,7 +7,7 @@
     "icon": "app.png",
     "type": "RAM",
     "tags": "tools,system",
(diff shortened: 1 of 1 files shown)


gfwilliams:
Argh - thanks!

PR #880, opened 2021-11-11
BLE GATT Battery Service works on Bangle JS 2

jjok:
I've got the data going into Home Assistant already.


gfwilliams:
Great - thanks!

PR #882, opened 2021-11-16
Added "emojuino" app v0.01

jeffyactive:
First iteration of an app which lets the user select an emoji (currently implemented as text) by swiping left/right, and then tap to broadcast that emoji encapsulated in a Bluetooth Low Energy advertising packet.  The emoji can be recognised by any device in range (or by a context-aware physical space) that observes the InteroperaBLE Identifier open standard.

In future this "emojuino" app will be able to scan for peer-broadcast emojis.


gfwilliams:
Nice - thanks!

PR #883, opened 2021-11-17
Add new clock face CLI Clock Enhanced

weeurey:
This is a version of the existing clock face with some changes and enhanced for the Bangle JS2 with a bigger time label and margins adjusted to reduce screen space wastage


gfwilliams:
Thanks! Any reason for the closure, or was it accidental?

PR #886, opened 2021-11-17
Add new clock app

weeurey:
This is a version of the existing clock face with some changes and enhanced for the Bangle JS2 with a bigger time label and margins adjusted to reduce screen space wastage


gfwilliams:
Thanks! Ignore my previous comment on the closed PR :)

Is there a reason for the `app.js.png`, or is that accidental? If so I'll remove it here

PR #887, opened 2021-11-18
A Battery Widget v1.0

alainsaas:


gfwilliams:
Nice - thanks! Just merging.

It seems it's not explicitly mentioned anywhere, but please could you have a go at making the IDs of new widgets start with `wid...` - it's what we've done with others and it just makes them nice and easy to find/identify.

PR #890, opened 2021-11-18
Port pomodo to Banglejs2

adamschmalhofer:
Made the pomodo app work on Banglejs2 in addition to Banglejs1.


gfwilliams:
Nice - thanks!

PR #900, opened 2021-11-19
chargeanim Bangle.js 2 compatibility

andrewmk:
Update chargeanim to make it compatible with Bangle.js 2.


gfwilliams:
Great - thanks!

PR #902, opened 2021-11-19
[LCARS Clock] Added alarm via swipe left/right.

peerdavid:
With this PR, an alarm is added to the watch face. If the user swipes left, the alarm timer is increased by 5 min., if the user swipes right, the alarm timer is decreased. After n minutes buzz is called and the user is informed. It's very useful to not forget things, cooking, or as pomodoro timer.

Additionally, the pedometer widget is used to count the steps as this should make step counting more robust and more consistent if this widget is used.


gfwilliams:
Cool, thanks!

PR #904, opened 2021-11-19
gbmusic: Bangle.js 2 support, use Layout library

rigrig:
Bangle.js 2 has no buttons for volume up/down, but we can swipe up/down.
(Maybe a future version could add volume buttons to the top half of the screen.)

For "random" colors, only choose from the 3-bit colors, to prevent dithering.

I figured I'd also enable `allow_emulator`, as it does run in the emulator (it shows the info from the screenshots)


gfwilliams:
Great - thanks!

PR #907, opened 2021-11-20
Bugfix: Swiper Clock Launch

t0m1o1:
Fixed issue with Swiper Clock Launch causing launcher not to show due to "mode" being undefined


gfwilliams:
Great - thanks! I missed that :)

PR #908, opened 2021-11-20
Bluetooth Swipe HID Control

t0m1o1:
App to control music via HID commands using swipes on the touch screen on the Bangle.js 2. Based on the original Bluetooth Music Controls.
Swipe up/down for volume, left/right for previous and next and tap for play/pause. Button1 locks and unlocks the controls.
Enable HID in settings, pair with your phone, then use this app to control music from your watch!


gfwilliams:
Thanks!

PR #910, opened 2021-11-20
barclock works fine on the Bangle.js 2 :-)

rigrig:


rigrig:
Actually: it doesn't, somehow the time goes black sometimes...

PR #912, opened 2021-11-20
weather: Bangle.js 2 support

rigrig:
- use `g.fillPoly` when device doesn't support `g.fillPolyAA`
- use (mostly) 3-bit colors instead of dithering
- crash a bit less on incomplete weather data

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -749,12 +749,12 @@
   {
     "id": "weather",
     "name": "Weather",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Nice - thanks! It's odd that fillPolyAA isn't in there actually. That should really have still been built into Bangle.js 2 :(

PR #917, opened 2021-11-21
Updated widget id of a_battery_widget to wid_* as per Gordon's request

alainsaas:


alainsaas:
Will fix the ChangeLog file and resubmit.

PR #920, opened 2021-11-21
Update snake.js

LoginByCall:
Original control method changed to alternative control method.


gfwilliams:
Thanks! Please could you bump the version and ChangeLog though so it's more obvious that something has changed?

PR #923, opened 2021-11-21
Health, supress bleed through of Showmenus when displaying barcharts

hughbarney:
And Toucher: Debugging default = false


gfwilliams:
Thanks!

PR #924, opened 2021-11-22
Add setting to scale launcher up/down for Bangle.js2

t0m1o1:
Menu changed to use vector font and added in a scaling factor which can be changed in settings for the bangle.js2. 
Can make into a separate launcher if preferred.


gfwilliams:
Thanks! I think this might be better as a separate launcher for now though? The same launcher handles Bangle.js 1 as well, and so all Bangle.js 1 users would then get a settings screen that doesn't so anything - also I'm not sure everyone would appreciate the vector font.

t0m1o1:
Ok shall we close this request and I'll redo as a separate launcher?
Settings should only show up on Bangle.JS 2 due to settings the "supports" in apps.json.
Looking over the code although the one listing in the app loader is shared they're pretty distinct - is there a reason both are still combined into one?

gfwilliams:
> Settings should only show up on Bangle.JS 2 due to settings the "supports" in apps.json.

Ahh, missed that - that's neat - thanks!

> is there a reason both are still combined into one?

They were separate, and then someone complained it was confusing to new users so they got merged! You can't please everyone :)

Actually, thinking about this, please could you add a setting in there for font? So by default it's basically as it is right now, but then you can change to a vector font (or maybe even something else like "6x15") and can squeeze the size down. Then I reckon it's good to merge this in as the default launcher.

I think some may not like the Vector font by default (it's not quite as 'clean') but it's always good to have the option.

t0m1o1:
Haha that's definitely true!

At least in theory it shouldn't show....they will have the JSON file though. Although should remain empty I guess so take up practically nothing. 

Yeh I think you've nailed it there that's the best idea. I'll get it done.

t0m1o1:
Right thats now got a selectable font and a scaling factor for the vector font - I've made a bit of a mess with the commits though. Is this ok or should I make a new branch and put just the final changes in?

gfwilliams:
Looks great to me - thanks!

PR #926, opened 2021-11-22
messages: add telegram icon

barnybug:
Tiny change for telegram

Original below:
telegram24


gfwilliams:
Great - thanks!

gfwilliams:
Just to add - additions like this are amazingly helpful, so thank you!

PR #929, opened 2021-11-22
Update app

weeurey:
Update of a previous clock application


gfwilliams:
Thanks! Just to add - for stuff like HID, in a way a widget might be better - and then it can be used in any app or clock (rather than being specific to one clock face)

PR #933, opened 2021-11-23
[LCARS Clock] Added lcars icon to watch face.

peerdavid:
Added lcars icon and some other, minor improvements (e.g. don't change icon on HRM).


gfwilliams:
Thanks!

PR #936, opened 2021-11-23
Binwatch: re-format background images

Vingelar:
changed the format of the background images to be uploaded to the storage in a readable format (export from emulator)


gfwilliams:
Looks great - thanks!

PR #937, opened 2021-11-23
Barclock works fine on the Bangle.js 2 :-)

rigrig:
At least, it works just as well as on the Bangle.js 1, where it also breaks from fullscreen notifications.   
So fixed that, and now it really works fine on both devices.


gfwilliams:
Great - thanks!

PR #938, opened 2021-11-23
Add Bangle.js 2 support for metronome app

bengwalker:
The metronome app now can be used on the new Bangle.js 2


gfwilliams:
Great - thanks! If it supports Bangle.js 2 you'll also need to change `"supports": ["BANGLEJS"],` to `"supports": ["BANGLEJS","BANGLEJS2"],` in apps.json

gfwilliams:
Merging, and will add the line here

PR #939, opened 2021-11-24
Add Screenshots to Many Apps

Ronin0000:
I added screenshots to all the apps that worked and had a try in emulator button.


gfwilliams:
Wow, that's awesome - thanks! Just so you know I think your text editor is set to use tabs, not spaces, so things get misaligned:

image

It's an easy fix for me to do (find/replace), but if you could tweak your text editor settings then you wouldn't have the problem.

Ronin0000:
> Wow, that's awesome - thanks! Just so you know I think your text editor is set to use tabs, not spaces, so things get misaligned:
> 
> image
> 
> It's an easy fix for me to do (find/replace), but if you could tweak your text editor settings then you wouldn't have the problem.

ok thanks! That's weird it did not look like that in the editor. I will make sure to change my settings.

PR #942, opened 2021-11-24
Add new TimeCal app

madman2021:
Here is a clock I've been working on, it shows the date and time, with a 3 week calendar.

Sorry if I've missed something, this is my first time contributing to this project.


gfwilliams:
Nice -  looks good!

All I'd say is right now you're polling every second which will affect your battery life a bit. There's a way of getting called only every minute which I've been using in most new clocks that you may be able to use - there's an example at https://www.espruino.com/Bangle.js+Clock+Font

gfwilliams:
Just FYI I'm talking battery life of maybe 4 weeks down to 2, so this is not some huge issue

PR #949, opened 2021-11-26
Update Spirit Level to work on Bangle.js 2

johan-m-o:
A very small update to the Spirit Level app to make it usable on the Bangle.js 2.


gfwilliams:
Awesome - thanks!

PR #950, opened 2021-11-26
Add new Interval Timer App

jg76379:
Hello, I would like to add my new Interval Timer app to the app loader. I have tested the app on my Bangle.js 2. It may work on the Bangle.js 1 as well as I have used the layout library for all of the views, but I do not have one to test with.


gfwilliams:
Looks great, thanks! I'm a bit busy to test it now but hopefully someone will be able to see if this works on Bangle 1. I guess it might work in the emulator?

PR #951, opened 2021-11-26
Move Quiet Mode LCD options from global settings to the Quiet Mode Schedule app

rigrig:
As discussed [on the forum](http://forum.espruino.com/conversations/369394/#comment16245707)
When the app is opened for the first time, settings are migrated from `qmOptions` etc. in `setting.json` to `qmsched.json`.

Looks like I'm the first user of the `Settings` module, and already ran into missing functionality (`unset()`) :-(
I hope it's ok that I added it in this pull request.

I'm not entirely sure about how modules are handled, but it looks as if the app loader prepends the whole module to any file that uses it? So pretty much every `qmsched` file on my watch now has `Modules.addCached("Settings",...` prepended. That seems a bit worrisome. (especially since it also gets added to the boot code)


gfwilliams:
Nice - thanks! But yes, I'm afraid the Settings library can't be used in boot code or even the settings page. Even in a library is less than ideal.

I did try and document it at https://www.espruino.com/Bangle.js+App+Settings but yeah, it's a bit of a pain.

rigrig:
>  I'm afraid the Settings library can't be used in boot code or even the settings page. Even in a library is less than ideal.

Urgh, I hadn't realised modules work this way when I wrote that library (it makes sense, just hadn't thought about it) :-(   
I'm thinking it might be better to remove the module and provide a few more code examples on that page?

gfwilliams:
> I'm thinking it might be better to remove the module

Could be - I guess it's not *that* tricky to access Settings yourself. If you do remove it there's one app (I forget which) where I used it I think?

rigrig:
> I guess it's not that tricky to access Settings yourself.

Yeah, my idea for the library was
1) remove the need for boilerplate code/replace it with cleaner code
  -> this fails for all the widgets/settings/etc. where you can't use it
2) make it easy for new users to read/write settings in a failsafe way
  -> this fails because now they *also* have to worry about whether they can use the library  
 (assuming they read the tutorial instead of just looking at how other apps handle settings)

So I figure it's better to just provide a few different code snippets everybody can copy-paste for failsafe settings handling.
("If you just want to read a single value, do `...`." "To read all settings into memory, use `...`.", etc.)

gfwilliams:
Yes, I guess so. It's a pain because long-term I think a Settings library would be good as it could be built into the Bangle, but it's how we get to that point. I guess we could put a 'polyfill' in the bootloader code but I think for now probably the library isn't ideal - sorry, I hadn't really considered the implications of that when you were doing it.

rigrig:
I suppose another way to go about it would be making it upload modules as separate files instead of inlining it. But then you have to somehow clean it up when all apps that depend on it are uninstalled...
If there were more widely-used modules it might be worth the effort, but I don't think we want to plunge down the rabbit hole of dependency resolution just for the settings.

rigrig:
Ok, now updated to work without the `Settings` library.

gfwilliams:
Excellent - thanks for taking the time to sort this out!

PR #952, opened 2021-11-26
locale: Apply 12hour setting to time

### LINKED ISSUE #892: 12 hr clock setting doesn't work
I changed my locale to en_US and went to locale menu and changed clock to 12hr. Anton clock should be able to handle since it asks locale module for time, but it still shows up as 24hr time.

rigrig:
Fixes #892

I figured it would be a good idea to not read+parse the settings file every time something wants to draw the time, so we remember it after the first call.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -170,7 +170,7 @@
   {
     "id": "locale",
     "name": "Languages",
-    "version": "0.09",
+    "version": "0.10",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great - thanks! I'll add this to the locale that's built in as well

PR #956, opened 2021-11-28
Add Mandlebrot Clock app

JamesLMilner:
Adds a simple clock which is themed on the mandlebrot set.

Here's a screenshot:

screenshot_mandlebrotclock
.


gfwilliams:
Nice - thanks!

PR #957, opened 2021-11-28
Tweak styling to give a more aligned and spaced out layout

JamesLMilner:
This PR makes some minor CSS tweaks to the App Loader. Let me know if you think this feels like an improvement.

**Before:**
image

**After:**
image


gfwilliams:
Neat - thanks for this!

PR #959, opened 2021-11-28
Assorted bugfixes to the recorder apps

dratini0:
Hi,

I don't know what the etiquette is around modifying apps authored by other people. In any case, I have noticed a few bugs in the recorder apps:

1. The label for the back button and the total distance label were put into fixed positions, designed for the Bangle.js 1
2. The X axis label (there was only one!) was odd and misaligned
3. In Recorder, the speed and altitude plotting tools didn't handle missing coordinates (this doesn't affect GPS recorder, as that one doesn't log records for situations where a GPS fix is not available) (Incidentally, I think the graph library should handle `NaN`s better)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -664,7 +664,7 @@
   {
     "id": "gpsrec",
     "name": "GPS Recorder",
-    "version": "0.25",
+    "version": "0.26",
     "description": "Application that allows you to record a GPS track. Can run in background",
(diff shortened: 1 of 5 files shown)


gfwilliams:
That's great - thanks! And no worries at all about submitting fixes to other people's apps!

PR #962, opened 2021-11-29
Update degree Celsius temperature app

myxor:
What changed?

- Add degree Celsius sign after temperature value (°C)
- Made loading message more clear what actually happens ("Reading temperature")

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -3893,8 +3893,8 @@
   {
     "id": "thermom",
     "name": "Thermometer",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Just a note that you can also use the locale module for temperature: http://www.espruino.com/Bangle.js+Locale

So if someone in the USA has the locale module installed then they'll get Fahrenheit

PR #965, opened 2021-11-29
Pastel: sunrise, sunset   and MyLocation app

hughbarney:


gfwilliams:
The 'sanity test' fails on this with `ERROR: Widget mylocation has no entrypoint`

I guess 'mylocation' isn't really a widget. You could try just setting it to 'library'. I guess there's something to be said for having a really minimal `location` library which you can just say `require("location").getDefault()` to get `{lat,lon}`?

hughbarney:
will re-issue with correct checkin comments

PR #966, opened 2021-11-30
Added SensiBLE app

jeffyactive:
Initial version implements menu-display of real-time sensor data.  Subsequent versions will support BLE advertising of the sensor data.


gfwilliams:
Nice - thanks!

PR #967, opened 2021-11-30
Poweroff app

myxor:
As i am using the power off functionality from the settings menu quite often i was eager to write a (very) simple app to have this functionality availability right from the apps main menu.


gfwilliams:
Great idea - thank you!

PR #973, opened 2021-11-30
New widget: Widbars

rigrig:
Just a simple widget showing storage, RAM and battery bars.
screenshot

(Initially it also included HRM & temperature, but those don't actually seem very useful in bar form, so I commented them out.)


gfwilliams:
Nice idea - thanks!

PR #974, opened 2021-11-30
widbatpc: Fix drawing the bar when charging

rigrig:
When displaying the charging indicator, the whole widget moves to the right, now including the charge bar.
Also some minor typos in the readme.


gfwilliams:
Thanks!

PR #977, opened 2021-12-01
New app: A Speech Timer

alainsaas:


gfwilliams:
Nice idea - thanks!

In future, when naming your apps please could you avoid putting `a_` in front of them, and ideally not underscores? None of the other apps do it and it's going to mean that when someone goes to https://github.com/espruino/BangleApps/tree/master/apps they see all your apps right up the very top of the list, rather than just being able to look down to `s` for `speechtimer` :)

PR #980, opened 2021-12-01
Poweroff: icon color fix & README improvement

myxor:
- Icon is white now with a 1px black border around it so it should be visible on all background colors
- Added links to README and fixed the grammar :)


gfwilliams:
Great - thanks!

PR #982, opened 2021-12-01
remove Settings module

rigrig:
Better not to have this at all than to have a library that you can't actually use in widgets/boot/settings code.

@gfwilliams: you [mentioned](https://github.com/espruino/BangleApps/pull/951#issuecomment-981703481) you already used it in an app? I didn't actually find any usages, did I miss something?


gfwilliams:
Thanks! That's an odd one - I was sure I had something but I can't see it now either!

gfwilliams:
Just updated the Settings tutorial on the Espruino site - it'll go live later today

gfwilliams:
Sigh. I should have gone through all my emails first! Thanks for https://github.com/espruino/EspruinoDocs/pull/622 !

PR #992, opened 2021-12-03
New game app: FLOW

TheGLander:
The README/description says it all. It's a simple arcade-like game where tapping the screen moves the flow-like thing from the center of the screen to the sides and vice versa and you have to avoid white obstacles.


gfwilliams:
Great - thanks!

PR #996, opened 2021-12-04
New App: Hebrew Calendar

barakplasma:
adds a new app displaying a hebrew calendar via a menu:
hebrew calendar screenshot

Uses Rollup (as inspired by / copied from https://github.com/espruino/BangleApps/tree/master/apps/banglerun )

P.S. I didn't clean up the git history, but I don't mind squashing if that helps


gfwilliams:
Thanks!

I should add that I think in general it's probably better to have apps written directly in JS though.

BangleRun is actually a great example of this - it was contributed but then not maintained, and it's been a nightmare to keep it working correctly since none of the usual contributors know Typescript. A rewrite of it using normal JS and the `layout` and `recorder` libraries is actually on my todo list.

barakplasma:
On the next version, I'll remove typescript from hebrew_calendar, and maybe banglerun if you want

gfwilliams:
Thanks! Don't worry about this one - I just meant if you were going to do new apps in the future...

... and with BangleRun I think it needs a complete rewrite really - since it was made the layout/recorder stuff have made things a lot easier so hopefully it should turn into a pretty simple ~100 line app

PR #997, opened 2021-12-04
Fix spelling mandelbrot

JamesLMilner:
Unfortunately I misspelled Mandelbrot (from [Benoit Mandelbrot](https://en.wikipedia.org/wiki/Benoit_Mandelbrot)) in my PR :man_facepalming: 

This PR addresses that.


gfwilliams:
:) thanks for the update!

PR #1003, opened 2021-12-04
Add widget: Charging Status

T0TProduction:
This is a rather simple widget:
It shows a lightning icon whenever the watch is charging. This way one can always see the charging status at a glance, no matter which battery widget is being used. 
When the watch is not charging, the widget is not visible.
It also adds a vibration whenever charging is started.


gfwilliams:
Thanks!

PR #1005, opened 2021-12-04
new custom menu: menuwheel

rigrig:
"Wheel Menus" was the best name I could come up with, I'd be happy to take better suggestions ;-)   
(The idea being that you are looking at a wheel with menu items on it's side)

screenshot_dark screenshot_edit screenshot_light


gfwilliams:
Nice - thanks! Yes, we definitely needed a different menu.

Nice idea searching for `< Back` as well! I was trying to think of a nice backward-compatible way of doing that but I think right now it's probably the best solution

PR #1007, opened 2021-12-05
Widbata:  Battery level in the current theme

hughbarney:


hughbarney:
its also a smaller widget at 27 pixels wide.

gfwilliams:
Thanks! I thought pretty much all the widgets used the current colour theme? Or is it that this one is just black and white so matches exactly?

hughbarney:
The majority and a fixed GREEN interior.
This one os a bit smaller, black and white only and switches over with a
change in themes.

PR #1008, opened 2021-12-05
Fahrenheit Stock Weather app and bangle.js2 port of Fahrenheit Temp

elrod16:


elrod16:
I patched the weather app when I couldn't get Fahrenheit to display no matter how I configured the watch and gadget bridge, despite having imperial units for everything else. After seeing other people's screenshots now though, I think I just don't have a functioning "locale" library and I don't know why.

gfwilliams:
Hi - thanks for this, but I'm not sure including this one is a good idea.

Have you installed the Languages app? https://banglejs.com/apps/#locale

Because if you do that and you set your country, apps like `weather` should automatically show the temperature in F. I don't think having duplicate apps is a good idea - if anything we should be removing stuff like `Fahrenheit Temp` and just make it use the locale module so one app can do both.

gfwilliams:
sorry - I missed your post above - what happens if you just open the web IDE and type `require('locale').temp(150)` in the left-hand side?

elrod16:
Hi, it's all good, I know you're a busy guy. I was doing lots of tests switching between "Android" and "gadgetbridge" and "notify" vs "notifyfs" trying to figure out some notification issues and I think "locale" just got wiped or damaged along the way. Flashing back to default and manually setting everything up again corrected the issue.

PR #1009, opened 2021-12-05
Add new clock app - Weather Clock

jg76379:
Weather Clock is a clock face which displays the current weather conditions using the data from Gadgetbridge and the weather app. I have tested on my bangle.js 2.


jg76379:
Found a bug I need to fix, will submit a new PR once that is taken care of.

PR #1010, opened 2021-12-05
Adjust 93dub midnight for 12hr time

rholder:
Currently the `93dub` watch face displays midnight for the 12 hour time setting as 00:00. This PR changes it to more correctly display 12:00.

I've tested this change in the emulator to verify that the 24 hour time setting remains unchanged for various dates with hour values of midnight (00:00). Testing these same inputs for the 12 hour setting now displays values of 12:00 for 00:00.


gfwilliams:
Thanks!

PR #1011, opened 2021-12-05
Add new clock app - Weather Clock

jg76379:
Weather Clock is a clock face which displays the current weather conditions using the data from Gadgetbridge and the weather app. I have tested on my bangle.js 2.


gfwilliams:
Nice - thanks! Do you think this would work on Bangle.js 1 as well? Or should we just tag it as Bangle.js 2 only?

jg76379:
I don't have a Bangle 1 to test on so not sure. I have used the layout library which should be compatible but someone actually already opened an issue about it not working on Bangle.js 1: 
https://github.com/espruino/BangleApps/issues/1012

PR #1013, opened 2021-12-06
nl_NL improvements and added some iOS bundle-id's 'translations'

jeroenpeters1986:
I hope these improvement will be useful for others too.


gfwilliams:
Great - thanks!

jeroenpeters1986:
Oh wow, nice catch is that there is icons support now too for ios messages! Just saw that in https://github.com/espruino/BangleApps/blob/master/apps/messages/app.js#L77
Great stuff :-D

PR #1018, opened 2021-12-07
Allow Calendar app to begin on Sunday

qucchia:
Adds a setting in a Calendar app which allows it to start the week on Sunday rather than Monday.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -2392,7 +2392,7 @@
   {
     "id": "calendar",
     "name": "Calendar",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Nice - thanks! Looks great, but I think there's a typo in apps.json where you meant to say calendar.json? `"data": [{"name":"messages.json"}]`

gfwilliams:
Merging anyway - and I'll fix it

PR #1019, opened 2021-12-07
app: Scribble

enricorov:
A keyboard prototype inspired by [Tertiary Text](http://pebble.rickyayoub.com) for Pebble.


hughbarney:
This looks brilliant !!!!

gfwilliams:
Great - thanks!

gfwilliams:
Actually I just realised the automatic checks fail with this. I've made some minor tweaks here: https://github.com/espruino/BangleApps/commit/aa94abe3befe36085d4e8ab69713c312ba076cef

Please could you check I haven't broken anything? Specifically I know the current `super` handling is a bit odd so it maybe be we can't actually do what I've done, and I just need to disable those checks in the linter

enricorov:
Ya, I don't see issues with the fixed version. Thanks for checking :)

PR #1021, opened 2021-12-07
Update locales.js with en_NL and weatherClock

lunctis-viribus:
Added translations for English with Dutch units

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4721,7 +4721,7 @@
   {
     "id": "weatherClock",
     "name": "Weather Clock",
-    "version": "0.02",
+    "version": "0.021",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks - looks good, but do you think you could change the weather app version? It should go `0.01, 0.02, 0.03` not  `0.01, 0.02, 0.021`

lunctis-viribus:
> Thanks - looks good, but do you think you could change the weather app version? It should go `0.01, 0.02, 0.03` not `0.01, 0.02, 0.021`
Sure. I've changed the Weather Clock version to 0.03.

gfwilliams:
Perfect - thanks!

PR #1022, opened 2021-12-07
Integer scaling and additional configuration options for the Custom QR Code app

halemmerich:
* Integer scaling of the qr code
* Warning/Error message in case of to much data or similar conditions
* New options
  * Prevent integer scaling
  * Set backlight to max. while QR is shown
  * Do not lock or dim while showing QR
  * Hide Description
  * Replacement of default description
  * Error correction Level

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1126,7 +1126,7 @@
   {
     "id": "qrcode",
     "name": "Custom QR Code",
-    "version": "0.02",
+    "version": "0.03",
     "description": "Use this to upload a customised QR code to Bangle.js",
     "icon": "app.png",
     "tags": "qrcode",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks great - thanks!

Just a random thought but I had some folks ask about duplicating existing QR codes. I don't know if it's possible to integrate something like https://github.com/nimiq/qr-scanner which could scan the code, turn it back into a nicely aligned black and white image, and then put it on the Bangle's screen - but that'd be amazing

halemmerich:
That is actually my use case. I have scanned the QR manually and copied over the text, but integrating a scanner would be a lot less hassle. I would however scan the code to data and build a new one to keep the configuration options and prevent the scaling/alignment/probably bad image and low display resolution from interfering with the QRs scanability. I will do some experiments on the scanner lib, integration doesn't seem to be that complicated.

Issue #968, opened 2021-11-30
messages app should not pop up with old messages


mabuchner:
I've installed the messages app and the iOS integration on my Bangle.js 2.

Whenever I'm walking around in the house without my phone in the pocket, the messages app keeps popping up with old messages.

I guess, this is happening, because all notifications on my phone are treated as new notifications whenever the bluetooth connection gets re-established.

I don't mind having old messages appear in the messages app, but the messages app should not pop up, if there are no new messages.

stephenPspackman:
FYI I've seen old messages show up in Android, too—though my device is known to have had more profound mental issues.

PR #1026, opened 2021-12-08
Fix for iOS: messages app should not pop up with old messages #968

jeroenpeters1986:
Hi @gfwilliams, here are a few more changes which I hope help Bangle's platform
* Added the 'new' flag to messages from iOS integration
* Changed notification behaviour in `messages` if it's not a new message (don't buzz us then) 
* Added more common app bundle id's to the iOS integrations
* Changed nl_NL locale again because the other notation didn't fit on Bangle.js2 screen with the default Anton clock

Remarks/Questions though:
 * I don't have an Android phone and I didn't see a pointer on if Android has the preExisting indicator, so I didn't touch that
 * now iOS has more 'app detection', should we have a few more icons in messages too?
 * are you open to changing the default icon to a bell or information (just the `i` in a box) icon, as the envelope commonly indicates 'mail' instead of a notification (although the app is called 'messages')
 * maybe my coding style is a bit too verbose, please dont hesitate to change it if that is better


gfwilliams:
This looks perfect - thanks!

> I don't have an Android phone and I didn't see a pointer on if Android has the preExisting indicator, so I didn't touch that

Yes, I don't think there is one right now (or at least it's not exposed by Gadgetbridge).

> now iOS has more 'app detection', should we have a few more icons in messages too?

Yes, that'd be great. I've been getting them from icons8 and then using the Image Converter with crop + inverted + transparency

> are you open to changing the default icon to a bell or information (just the i in a box) icon, as the envelope commonly indicates 'mail' instead of a notification (although the app is called 'messages')

Yes, absolutely - that sounds like a good plan!

> maybe my coding style is a bit too verbose, please dont hesitate to change it if that is better

Generally I skip `{}` if there's just a single line in a block, but what you've done there is great. The file gets minified normally so those things will get pulled out anyway so it's not like it hurts size/performance :)

jeroenpeters1986:
Thanks for the merge and your answers! I tried using `NRF.ancsGetAppInfo(msg.appId)` but I didn't get it right..
 I will be sending more PRs soon. I like the idea that I can contribute to something that is widely used.

gfwilliams:
>  I tried using NRF.ancsGetAppInfo(msg.appId) but I didn't get it right..

It might be that it returns a promise, so you need...

```
NRF.ancsGetAppInfo(msg.appId).then(appName => {
...
});
```

But it's async - so the function will complete and lines after are executed before the 'then' part is excecuted

jeroenpeters1986:
Thanks for the pointers, I already tried exaclty that. But I am not familair enough with promises I guess. Inside the `then()` I overwrote the variable with the app name, but I think that finished after the notification 'aired'. For now I'll be making a list of apps.  I'll try to see what I think is popular and a few apps that I use myself (without a specific icon). If you have any preferences, please let me know.

PR #1029, opened 2021-12-08
Weather: Allow hiding the widget

myxor:
**New feature:** 
* Allow hiding the weather widget (can be useful when using a clock which shows the weather info).


gfwilliams:
Neat - thanks!

PR #1035, opened 2021-12-08
WIP: Settings should respect launcher font size

myxor:
**New features:**
* Settings app respects the font settings from launcher
* Default font is "12x20" - same as before these changes

**Question:**
* Do we need to make more clear that the app-settings for the launcher does also affect the settings menu? If yes, how?

**Problem:**
* The menus of app settings from other apps are not rendered respecting the font because they are just executed via `eval` (https://github.com/myxor/BangleApps/blob/3c5afeead136cec454fa9335b89809dc36772e45/apps/setting/settings.js#L650) and therefore the `'predraw': function(gfx) { gfx.setFont(font);}` line is missing in the menu config. Any idea how to solve this?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -146,7 +146,7 @@
   {
     "id": "setting",
     "name": "Settings",
-    "version": "0.35",
+    "version": "0.36",
     "description": "A menu for setting up Bangle.js",
     "icon": "settings.png",
     "tags": "tool,system",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - but I don't think that's a good idea.

If you want the menus from `E.showMenu` to be bigger, then it's `E.showMenu` that needs to change. In fact there are already 2 different menu apps that do this - one for small, one totally different. Since most apps don't mess with showMenu, changing it globally isn't a big deal.

Nothing stops you making a new app that provides configurable font sizes for menus though, in much the same way that those other apps do. I still wouldn't make it dependent on Launcher settings though - some people have their own launchers without those settings.

myxor:
After thinking about it i agree with you.
Thanks @gfwilliams .

PR #1036, opened 2021-12-08
Transmit app name as Espruino manufacturer data

jeffyactive:
Feature added to both Emojuino and SensiBLE apps.


gfwilliams:
Thanks - I wonder if this could even be a 'boot' app rather than having it in each separate application?

PR #1037, opened 2021-12-08
Create Pattern Launcher app

crazysaem:
This PR adds the initial version of the "Pattern Launcher" app which will directly launch apps from the clock screen with custom patterns.


gfwilliams:
This looks awesome - thank you!

PR #1041, opened 2021-12-09
Fix 93dub on Thursdays

rholder:
This should fix a crash for 93dub that only shows up on Thursdays. When this happens it even seems to disable navigation back to the Settings app from the button on a Banglejs 2 when set as the default clock.

I've tested this on the emulator, and it appears to do the correct thing now.


gfwilliams:
Great - thanks! I think if it errors while executing, `Bangle.setUI` never gets called so the launcher button doesn't work.

PR #1044, opened 2021-12-09
Moon Phase widget for Southern Hemisphere

Razoxane:
In the Southern Hemisphere, the Moon's phases go in a different orientation to the Northern Hemisphere.
I've just copied the existing widmp widget, and rearranged the phases accordingly.


gfwilliams:
Thanks! It's a tiny widget so I think this makes sense, but I guess for larger stuff we could have a settings menu rather than separate apps/widgets.

Razoxane:
Thanks for the feedback, and for merging!

I did think about trying to add settings functionality to support using the locale or GPS coordinates to support loading the appropriate set of orientations, but got a tad overwhelmed for my first go at a Bangle widget.
I'll have a go at it for a future PR!

PR #1045, opened 2021-12-09
Command line complete clock

myxor:
### Command line styled clock with lots of information:

It can show the following (depending on availability) information:
* Time
* Day of week
* Date
* Weather condition (requires weather app)
* Temperature (requires weather app)
* Steps (requires a step widget)
* Heart rate (when screen is on and unlocked)


gfwilliams:
Nice - thanks!

PR #1047, opened 2021-12-09
gbmusic: Fix scrolling title background color

rigrig:
We forgot to reset the background color when drawing the scrolling title.
Not very noticeable, until you get a notification and suddenly the title starts flashing in sync with the `messages` widget ;-)


gfwilliams:
Thanks!

PR #1048, opened 2021-12-09
messages: Open app when touching the widget (Bangle.js 2 only)

rigrig:
I figured that re-adding the `touch` listener every time we redraw the widget should prevent most breakage from apps doing `removeAllListeners`

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -57,7 +57,7 @@
   {
     "id": "messages",
     "name": "Messages",
-    "version": "0.10",
+    "version": "0.11",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Nice - thanks! Had you noticed an issue with apps calling removeAllListeners? I don't think it should be a huge problem now as we're trying to use setUI for most stuff

rigrig:
> Had you noticed an issue with apps calling removeAllListeners

No, that was just me wildly speculating ;-)

Issue #998, opened 2021-12-04
Typescript support


T0TProduction:
Im a fan of the simple way this repository hosts app as well as the existence of the web IDE and emulators for ease of use.

One thing I was wondering though: 
Is there any Typescript support? Obviously all distributed apps need to be JS, but in theory typing the entire BangleJS environment and creating a default export/transpile script that exports files to the correct format and structure could be a huge help to creating new apps.

The main two questions I would have are:
- Was this already considered? If so, what was the reason against it? (Or is there already something in the works that I missed?)
- Would PRs to this repo be allowed to be minified Typescript exports? Obviously the JS can be rather unrecognizable after being transpiled. 
This can be deactivated, but saves memory etc. so why should we remove that benefit of TS. 
A solution to this could be to also add the TS files to the PR, maybe in an extra folder which is used to verify the code logic in PRs, but just ignored by the app loader later on.

A probably overkill solution for the second question could also be to allow TS PRs, and then after approval a GitHub Action transpiles the TS without any human interaction - making sure the potentially unreadable code is actually created with the reviewed TS code.

stephenPspackman:
I'm not sure this is the highest priority, but it would be a lovely thing to have. Is the JS interpreter sufficiently close to standard that this would work at present?

T0TProduction:
Good point!
If the Interpreter is close enough, and If in general the wish is there to officially "support" TS, then I think this could be handled as not any official task (and therefore any specific priority), but instead get entirely crowdsourced.
Even the typings could be incrementally updated until at some point finished.

I personally would enjoy contributing to this endeavor

gfwilliams:
I think @T0TProduction may have already seen this, but there is already one Typescript app called BangleRun: https://github.com/espruino/BangleApps/tree/master/apps/banglerun

On the whole Espruino seems fine with the minified code that comes out of Typescript, and adding apps in the way BangleRun did seems like a definite possibility (with the TypeScript source but minified JS stored to get uploaded, and build instructions in a README).

**However** my concern is really with maintainability. BangleRun is a great example - it was written in thoughtful way, but the original maintainer stopped contributing after a while, and then when there are issues it falls down to the community. And that meant me, because nobody else dared poke around in it.

Actually debugging/fixing stuff like this is a nightmare. So you get a stack trace in JS? You've then got to search through the minified JS and try and find the matching bit, and then try and guess where in the Typescript it was. And then when you find it, you've got this maze of totally undocumented spaghetti code to wade through.

[...]

PR #1052, opened 2021-12-10
Add unified Typescript support

T0TProduction:
This PR builds on the discussion in https://github.com/espruino/BangleApps/issues/998

This is the first draft for a shared Typescript project setup and type collection for creating apps for Bangle.js
We *could* already merge this, as it is functional and tested, but it can also be worked on and kept open for some time before choosing to (if at all) merge this. 

Here's a list of current changes:

- added `/typescript`, which includes
   - package.json to install typescript
   - a script in the package.json to build **all** typescript apps  at once with a shared configuration
   - a readme which guides through the install and build steps
   - tsconfig, which is the TypeScript project/build config
- added `/typescript/types/globals.d.ts`
   - This is a collection of all types that bangle.js apps need. They are globally available to all TS files. 
   - this is still a very small collection of types that needs to be expanded over time. In general, every time someone needs a globally available bangle.js module, they should add the type here and automatically contribute to the completion. 
 - updated the widChargingStatus  widget to TypeScript as proof of concept
- added CI Pipeline: GitHub actions to validate both the type definitions and all typescript apps, which will run on every push. This makes code quality checks easier, as any TS app that has faulty code will fail the pipeline entirely.

Disclaimer: The app build step actively excludes the other two currently existing TS apps, as they fail with hundreds of type errors. One of the reasons for this is that they typed the globally available bangle modules as `any`, which is not allowed by the project configuration. (Typing as any defeats the entire purpose behind TypeScript.)

I'm open to feedback on everything!
I'd be especially interested in regards to the typescript configuration (anything unclear? any wishes for change?)

I added GitHub actions because it makes sense to have the type checks run in a pipeline, but I understand that adding such things is not up to me, so I'm eager to hear feedback on that as well. 

------------------------------

Things that are still to do:
- [ ] add linting. Should be easy to import the existing config and expand it with typescript/eslint to make it work with TS code as well.
- [ ] Check out other typescript projects and see If we can use those types
- [ ] Finish discussion about GitHub Actions
- [ ] Validate idea for pipeline: Check for git code difference after build step. If there is one, this means the JS Code commited was not built by the TS code committed.
- [ ] Check out inlining imports on build


T0TProduction:
I know you're probably very busy Gordon, but nonetheless, I'll remove the draft status from this PR, since it is technically ready for merging. 
Obviously, if feedback requires changes, I'll work them in, just want the status of this PR to be a bit clearer to new readers.

gfwilliams:
Hi - Thanks, this looks good - and sorry for the delay looking into this.

I really only have three comments/questions:

* `typescript/types/globals.d.ts` - you should be able to pull type info out of `http://www.espruino.com/json/espruino.json` pretty easily with a very simple script. It might be good to have a script to build the `globals.d.ts` file from that. There is also https://www.npmjs.com/package/@types/espruino (and others) and IMO it'd be better for everyone if we could just contribute back to those with updated types rather than reinventing the wheel. I'm happy to have extra type info added to the Espruino repo (https://github.com/espruino/Espruino/blob/master/scripts/common.py#L57-L89) if it's needed in order for us to be able to auto-generate good typescript files.
* I feel like this should be able to build `banglerun` - the only other typescript app in the repo. It's currently excluded from the build (hebrew_calendar got converted to JS)
* GitHub actions. This looks interesting, but we do already have Travis CI. I guess I'm not really sure why we need a second, different option *just* for typescript. It's just more complexity and maintenance overhead - and we over-use it, it's then something I have to pay for. Given we have only a handful of typescript apps, running CI on every commit for this seems a bit wasteful.

So as far as I'm concerned I'm happy to merge this if... 

* We remove the GitHub actions file for now. I think anyone developing typescript is going to have Node.js and will be able to do their own builds - maybe if a lot more developers are using typescript then we can look at re-enabling it later though.
* We add support for building `banglerun` - so every app then uses the `typescript` folder for Typescript development

T0TProduction:
Thanks for the feedback!

Concerning @types/espruino and generating types:
I didn't know this existed! Will take a closer look and see what they did ( especially differently), as well as check out generating types. 
In general I'd say we probably can't auto generate everything, but we shall see!

Concerning banglerun: 
Fair point. This will mean rewriting their code, but I guess it makes sense to do this rather sooner than later. 
This would not mean any changes to the TS config.

Concerning GitHub actions:
If we can include these in Travis, I'd be happy to migrate it there!
To my knowledge though, GitHub actions is completely free of charge for any Open Source Project. And, in addition, the runs that run on push will always run on the forks, not this repository. Nonethelese, it could be run on PR only just as Travis is.
Maintenance is non existant, because what runs there is also what needs to be run locally to build the apps. This will be maintained anyways.
If you still think we should remove it, will do, but I think it's important to enforce the typings. Any faulty TS app will break the compilation of all other TS apps, and if we merge TS without validating it, the chance is high this will happen.

gfwilliams:
> I didn't know this existed! 

I literally just googled 'Espruino Typescript' - there could be others, although https://github.com/stasberkov/espruino-ts-quickstart uses it so I guess that's a start.

> banglerun

Thanks - if you could look into that it'd be good. Out of interest, why would it need rewriting? I thought TypeScript was supposed to make maintenance like this easier. I think it's highly likely folks using Typescript will want to use multiple files (if that's the issue?).

>  GitHub actions

I had a quick look and it does seem there's some kind a quota: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions

2000 minutes/month. I get that it'll run on the forks, but it probably runs out of the credits here for any PRs. I guess the typescript stuff should be fast, but if we moved the sanitycheck.js stuff over I bet I'd be getting very close to running out of credits.

For now I'd prefer to take it out and then merging is a no-brainer. We can always look at putting it in later - I feel like when there are only a handful of contributors it's not going to be a big deal to ensure nothing gets broken.

T0TProduction:
I always only looked for bangle, not espruino. Guess that wasn't very smart of me 🤷🏻‍♂️

It needs to be rewritten to follow the rules the config enforces, that's all. Multiple files is not the issue.
A PR for that has already been opened in my fork, so that part is looking good.


Those quotas only apply to closed source. If you read the first sentence it says "GitHub Actions usage is free for both public repositories and self-hosted runners."
The quotas only apply to private repos.

T0TProduction:
Travis CI is currently failing because of the way the built banglerun app imports files now, which is not correct anymore. Will look into either inlining imports (as was done before), or switching to different syntax if that is supported.

Haven't had time to look into the other type collections yet, that will come next.

T0TProduction:
converted back to draft - once all topics are addressed I will activate it again

T0TProduction:
Successfully moved all steps happening in Travis CI to GitHub actions.
Two things to note:
- GitHub actions is a lot quicker
- Travis CI currently uses node 17 for the steps? even if it is decided to not use GitHub actions, this should be changed @gfwilliams , [as 17 is not an LTS version](https://nodejs.org/en/about/releases/).

gfwilliams:
Thanks - so instead of removing GitHub actions you've ported the Travis code to run on GitHub actions as well? I mean, it looks good, but contributors are currently used to looking on the Travis build for info and now they've got to change what they do. I guess overall it could be an improvement though.

It doesn't really solve my 'two CI solutions' issue unless you're planning to also port the whole of https://github.com/espruino/Espruino to compile in GitHub actions rather than Travis CI as well though.

In terms of BangleRun, do you consider the rewrite to be complete now? Does it actually work?

Because as far as I can see instead of one minified blob there are now a bunch of library JS files? I think the app loader will only load `app.js` and will be unable to find the other modules - however even if it does, it doesn't tree-shake or do any inter-module optimisations and it'll drastically increase the memory usage of the BangleRun app and make it slower. This seems like a big step backwards as far as BangleRun is concerned :( Is there any way to re-enable the optimisations it had?

I know what we'd said about being able to debug (so the end files not being minified) but in the case of multi-file typescript apps I think we do need to mash them together or they'll be too big/slow to run on Bangle.js

T0TProduction:
> so instead of removing GitHub actions you've ported the Travis code to run on GitHub actions as well?

Yes, because I wanted those checks for my pipeline while working on this PR. Since it's only a single file, it can easily be removed if you decide against including it in the merge. Also, I couldn't really find the TravisCI configuration in the repo, so I assumed that is done outside of the codebase. 

> It doesn't really solve my 'two CI solutions' issue unless you're planning to also port the whole of https://github.com/espruino/Espruino to compile in GitHub actions rather than Travis CI as well though.

If you're on board to switch everything to GitHub actions, I could check out the espruino repo and port the pipeline to GitHub actions there as well, sure!

> In terms of BangleRun

As noted (maybe not visible enough) I haven't tested it yet. This is something I also thought about, yes, maybe inlining all imports is the way to go. I'll check it out in more detail.


FYI I added some TODOs to the PR description, including eslint support. Currently, eslint and typescript are still not in perfect harmony within this repo, but this is something that definitely will work - just didn't do anything in that regard yet.

T0TProduction:
whoops, falsely renamed master to main on that repo, its back again

gfwilliams:
Thanks! I will get on this, just a bit busy right now.

IMO the best bet is going to be to just delete BangleRun - the Run app should be better now anyway.

gfwilliams:
Just merged! Thanks for your work on BangleRun but as being as it was obviously broken, and nobody seemed to care about it now, I've just deleted it.

T0TProduction:
Thanks for merging!
And sorry I didn't continue on the open topics, didn't have a lot of free time lately.

I still think e.g. the ability to compile many files to a single one should be created, and looking into existing types to use them here as well.
But these things can be added later on, so it's fine that this was merged now.

I might create a new issue for tracking those possible Typescript related enhancements that are currently listed in the top of this thread.

gfwilliams:
No worries - sorry it took a while. 

Thanks - the GitHub actions looks really promising, especially if it helps to stop me chewing through Travis CI credits so quick :)

T0TProduction:
Interesting, can you see if it actually "costs"/uses those credits? I understood their wording in a way that these minutes are for actions on private repos and OSS actions don't count against them, but maybe I misunderstood it.

gfwilliams:
Wow, excellent! Thanks for letting me know!

PR #1054, opened 2021-12-11
cliock: 0.16 - Stop animation when screen locked

hasmar04:
This theoretically should reduce battery usage and CPU cycles.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1755,7 +1755,7 @@
     "id": "cliock",
     "name": "Commandline-Clock",
     "shortName": "CLI-Clock",
-    "version": "0.15",
+    "version": "0.16",
     "description": "Simple CLI-Styled Clock",
     "icon": "app.png",
     "screenshots": [{"url":"screenshot_cli.png"}],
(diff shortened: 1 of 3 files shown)


gfwilliams:
I'm afraid this won't have any effect, since you're still drawing everything to the screen every second - you're just drawing the same thing.

It's a good idea though - but I think you'd have to drop the `click` interval that was created with the `var click = setInterval(updateTime, 1000);` line - you could do that with the `lock` event. There are a few examples of other clocks that do a similar thing when the LCD turns off on Bangle.js 1 so you could use that code - basically call updateTime every second when unlocked, and then every minute when locked

hasmar04:
Ah ok good to know. I could maybe even permanently reduce the update time because it doesn't display seconds anyway. I'll look into it and update this request accordingly.

gfwilliams:
Thanks, that's great! Looks like it updates every 15 secs at the moment? 

There's some code we've started using for clocks that update every minute at https://www.espruino.com/Bangle.js+Clock+Font which it might be good to look at using - it ensures the clock updates exactly when the minute changes - right now I think cliock could be as much as 15 seconds out!

PR #1056, opened 2021-12-11
Updates for impwclock BangleJs2 compatibility

lightnin:
Copied changes from wclock to handle different screen geometry of both BangleJS and BangleJS2 cases. Tested on both devices.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -538,7 +538,7 @@
     "icon": "clock-impword.png",
     "type": "clock",
     "tags": "clock",
-    "supports": ["BANGLEJS"],
+    "supports": ["BANGLEJS","BANGLEJS2"],
     "screenshots": [{"url":"bangle1-impercise-word-clock-screenshot.png"}],
     "allow_emulator": true,
     "storage": [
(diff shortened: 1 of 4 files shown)


lightnin:
Just realized there is some code in there to show the time on screen for BangleJS2, but it doesn't seem to work. Looking at the tutorial collection, it's not clear to me how to figure out how to trigger events with the touchscreen yet. If anyone else can take a look it's got a comment "Bangle2" already.

lightnin:
Ok - changed it up to use screen touch on both BangleJS and BangleJS2 to toggle display of accurate digital time. Might be better to hide the display of digital time if there was an "End of Touch" event to do that on, but didn't see a way to do that. So display state toggles, and if left on stops showing accurate time at the start of next imprecise time era.

gfwilliams:
That's great, thanks! I think you could use `Bangle.on('drag'` - it'll give you a button-down value, and you can check when that becomes 0 or 1 (you just don't want to redraw every event as you get loads when you're dragging your finger)

Issue #985, opened 2021-12-02
Astronomical step counts on GadgetBridge


aivopaas:
For unknown reasons I'm getting quite comical activity reports on GadgetBridge from Bangle 2.
Everything should be on the latest versions.
Haven't tried debugging it yet in any way so I have no idea which component exactly might be causing it (I'm pretty sure it's not my feet though).

That's yesterdays data, today I'm going to pass the 2000 km mark easily.

![SmartSelect_20211201-181316_Gadgetbridge](https://user-images.githubusercontent.com/191152/144469577-5ba016d5-c84b-42a5-a1a3-d26973860696.jpg)

gfwilliams:
Wow! Which app are you running on the Bangle? Gadgetbridge or Android?

The step reporting stuff will be in whichever ones of those are used - maybe it's not resetting the counter so is accumulating each time an update is posted

myxor:
Glad to see i am not the only one having similar step counts :)
I am using the Active Pedometer and the Android apps on the Bangle.js 1.

CountMurphy:
Same here. GadgetBridge and Active Pedometer installed here, though mine only did it once.

CountMurphy:
After installing the Android app, my counts are normal/correct

gfwilliams:
Fixed with #1058 now!

PR #1058, opened 2021-12-11
Gbridge fixes

adamschmalhofer:
Two rather single fixes for gbridge:

- Fixed inflated step count sent to Gadgetbridge http://forum.espruino.com/conversations/370644/
- Fixed getting informated about incoming calls.


johan-m-o:
The call notification bug should be fixed on the Gadgetbridge side though. This is just a bandaid that doesn't actually fix anything...

myxor:
Reference: https://github.com/espruino/BangleApps/issues/985

gfwilliams:
Thanks! Not sure how that `lastSentSteps = 0;` bug got in there, I was certain this worked fine when I did it!

Lets keep the call hack in there, but as @johan-m-o says we need to fix this in Gadgetbridge. I'm still a big confused about why it'd happening as the build done from Github works great - it's as if the release build doesn't have everything in from master.

PR #1059, opened 2021-12-11
app update: Hebrew Calendar

barakplasma:
refactored to be in javascript rather than typescript as suggested in https://github.com/espruino/BangleApps/pull/996#issuecomment-986555488


gfwilliams:
Neat - thanks! Nice idea making it a library so other apps can require `hebrewDate` if they want as well!

PR #1063, opened 2021-12-11
qmsched: Correctly size widget on load

nebbishhacker:
Fixes an issue where qmsched would cause widgets to be drawn twice on every app load.

This was especially noticeable on the Bangle 1 when it's snowing outside, since the snowflake icon renders particularly slowly :)


gfwilliams:
:) Thanks!

PR #1064, opened 2021-12-11
Custom QR Code v0.04: Adds QR code scanning

halemmerich:
Adds QR code scanning from files or browser-accessible cameras.


gfwilliams:
:o Amazing - thank you! That suddenly makes this super useful.

PR #1066, opened 2021-12-12
weatherClock Bangle.js 2 layout compatibility 

lunctis-viribus:
On Bangle.js 2 there still was a layout overlap issue for wind text.
This update fixes it. Tested for different input lengths in emulator, also for Bangle.js.


gfwilliams:
Thanks!

PR #1073, opened 2021-12-13
Pebble: Load steps from health app if pedometer widget is not available

myxor:
Pebble: Load steps from health app (if available)


gfwilliams:
I'm not sure this is a good idea - I think we should really just fix the pedometer then it'll work for all things?

myxor:
Fixing the pedometer is a good idea for sure but shouldn't the clock be able to determine the steps from the health app if no pedometer is installed?

gfwilliams:
I feel like if you want step counting it's probably fair enough that you have the pedometer installed?

The issue is that reading the daily health data for steps *every time the watch is redrawn* is going to be pretty slow and will really hurt battery life. 

Not to mention if you have health+pedometer, this PR will suddenly hurt performance for you when it would have been easy to find out steps from the pedometer

myxor:
> I feel like if you want step counting it's probably fair enough that you have the pedometer installed?

I mean, step counting is working without the pedometer. So... i am not sure.

> Not to mention if you have health+pedometer, this PR will suddenly hurt performance for you when it would have been easy to find out steps from the pedometer

So having the priority the other way around (first pedometer, then health) would be allright?

myxor:
I changed my implementation so that steps counting from health app is only fallback if pedometer widget is not installed

PR #1075, opened 2021-12-13
Add CoreTemp sensor app

ihewitt:
This is an experimental bare-bones app to start supporting a CoreTemp body temerature sensor (copy of the BTHRM app).
At the moment this simply displays the live temperature values, but also emits the values which could be logged/integrated into other tracking apps or set to alert.


gfwilliams:
Nice! I'll try and get some code in for the BT CSC sensor (cycling speed) that you could use as a reference, as it's possible to create add-ons that will allow you to log extra data (like this) alongside GPS/HRM/etc

ihewitt:
I was thinking it would be nice to merge the "collector" parts together (eg
merge also the BT HRM) and make them configurable. Will have a play.

PR #1077, opened 2021-12-13
android: Send notification id when handling notifications

nebbishhacker:
Fixes the issue mentioned in http://forum.espruino.com/conversations/370923/


gfwilliams:
I'm an idiot! Thank you so much for sorting this!

PR #1079, opened 2021-12-13
pooq Roman: autolight

stephenPspackman:
Add an option to enable the backlight on wrist-twist.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4734,7 +4734,7 @@
   { "id": "pooqroman",
     "name": "pooq Roman watch face",
     "shortName":"pooq Roman",
-    "version":"0.02",
+    "version":"0.03",
     "description": "A classic watch face with a certain dynamicity. Most amusing in 24h mode. Slide up to show more hands, down for less(!). By design does not support standard widgets, sorry!",
     "icon": "app.png",
     "type": "clock",
--- a/apps/pooqroman/ChangeLog
+++ b/apps/pooqroman/ChangeLog
@@ -1,2 +1,3 @@
-0.01: New App!
-0.02: Make internal menu time out + small fixes
+0.01: Initial check-in.
+0.02: Make internal menu time out + small fixes.
+0.03: Autolight feature.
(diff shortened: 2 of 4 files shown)


gfwilliams:
Thanks!

About the twist/face up/etc - you know this is done automatically by Bangle.js? You just set `Wave on Twist` in the LCD part of the Settings menu? If you want to enable/disable it manually for one app you can use `Bangle.setOptions`

stephenPspackman:
I have “wake on twist” on on my device. It unlocks, but it does not
illuminate, which is why I thought this option was useful. Perhaps it's a
bug?

gfwilliams:
>  Perhaps it's a bug?

It is, but it's a bug that was fixed in a firmware update several weeks ago. If you update the Bangle.js firmware it'll work fine now.

stephenPspackman:
Ah, sorry. Somehow I was unaware that there was even something that needed
to be done outside the “app store” interface. But I tried the process in
the web IDE and it said:

Sorry, the firmware update has failed.
The error was: Unable to download
https://www.espruino.com/binaries/espruino_2v10_banglejs2.zip

gfwilliams:
Best to use this for now: https://www.espruino.com/Bangle.js2#firmware-updates

The Web IDE will fail unless you manually specify a firmware because right now there is no 'official' numbered release - they are all 2v10.something .

stephenPspackman:
That method doesn't seem to work for me—the watch reboots and drops out of
DFU mode after a relatively short time (I think I saw a message saying
something something check storage? But it was up for only a fraction of a
second and I wasn't able to read it) and leaves the phone app saying it
can't find a target. The watch's about screen still reports 2v10.219. (I'm
assuming the correct target to have selected was “BANGLE.JS nnnn”? The web
page is broken on that point.)

stephenPspackman:
Ah, sorry. Tenth try was the charm. It turns out that the correct target
really *is* called DfuTarg—that's not an editing error in the
instructions—but the MAC address is one greater than usual. After several
tries it did succeed in handshaking and flashing, and now the backlight
works as expected. Sorry once again. I'll remove the unnecessary option
next time I have something to push :).

PR #1081, opened 2021-12-13
Desktop launcher: Adds settings page

halemmerich:
- Allows to selectively show/hide launchers and clocks.
- Adds setting to directly launch on touch with Bangle 2


gfwilliams:
Nice - thanks!

PR #1083, opened 2021-12-14
Red Torch - Update for Bangle 2

halemmerich:
- Use setUI instead of Bangle.watch
- Quickstart sequence does just work on Bangle 1, since Bangle 2 is missing the additional two buttons and the only remaining button is often used for functions in apps (see: https://github.com/espruino/BangleApps/pull/1083#issuecomment-993444082)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1698,14 +1698,15 @@
     "id": "rtorch",
     "name": "Red Torch",
     "shortName": "RedTorch",
-    "version": "0.01",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Does this work on Bangle.js 2 though? As far as I can see apps/rtorch/widget.js hasn't changed and that still uses BTN3 which isn't on Bangle.js 2.

halemmerich:
The widget is not installed for Bangle 2 via the supported property in apps.json. I use it via pattern launcher and did not want to change the BTN1 BTN3 sequence to only BTN1 to prevent strange interaction with other apps using BTN1.

gfwilliams:
Ahh, great - thanks!

Issue #1033, opened 2021-12-08
[Messages - Feature Request] - change the widget bar to an icon rather than flashing 'MESSAGES'


hughbarney:
Love the new Messages app.  But think it would be more elegant if the notification on the widget bar was a simple envelope icon, rather than a flashing 'MESSAGES' bit of text.  It does not need to flash or be so large.   The user will easily spot the appearance of an envelope ICON in the widget area.

gfwilliams:
I bet someone will complain it's not obvious enough if we change it, but I appreciate it doesn't look that great.

Maybe it should be a setting

jeroenpeters1986:
I was also thinking this, that it could be a setting. I was thinking of a bell tingling left and right. But I'm not even sure if I can manage that. 

Also, I have my watch/LCD to only react to the button or touch, but it seems that when I have messages it's lighting up more (but the message has already disappeared). Anyway, mightb be unrelated. I'll try to figure out the pattern. 

I will create some new icons and app-detections, I will also try to work on this

hasmar04:
This change to an icon would also mean that the alert widget doesn't interfere with the watchface theming. 
For example, the colouring and flashing affects the appearance of the cliock face, which I'm assuming is unintended.

jeroenpeters1986:
I worked on this and I have a test available, but it broken on some other codepart I think.

@gfwilliams when working with icons you pointed me to "use the Image Converter with crop + inverted + transparency". I got some new icons from icon8, but wanted to know if you have the icon-size in pixels you were using, and also, do you need to enable the compression at this point in the image convertor. I am not sure.

gfwilliams:
> the colouring and flashing affects the appearance of the cliock face, which I'm assuming is unintended.

Thanks - that'd be worth filing a bug for cliock on. `cliock`'s probably not setting up the colours properly each time it renders.

@jeroenpeters1986  I just used the '30px' versions - there are certain types of B&W icons8 icons that come in that size, and that seemed fine. I don't use compression on them as they're so small it's not really worth it :)

jeroenpeters1986:
Thanks @gfwilliams, I downloaded them as 24 and that works too now but that was just my guess. I will update them to 30px. 
Because I'm adding new icons it might be worth compressing, but I don't know if the decompression time is worth it. Do you have any insights on that?

gfwilliams:
>  I don't know if the decompression time is worth it. Do you have any insights on that?

[...]

PR #1084, opened 2021-12-14
Messages, iOS, Locale

jeroenpeters1986:
Hi, please don't mind the numerous commits, but in the end I did

**Messages app**
 - Added specific app icons
   - calendar
   - facebook
   - hangouts
   - instagram
   - gmail
   - google home
   - mail
   - messenger
   - outlook mail
   - phone
   - skype
   - slack
   - sms message (or iMessage)
   - twitter
   - telegram
   - whatsapp
   - wordfeud
 - Small layout 'fix' to not attach the icon button to the left side of the screen (give it a few pixels space)
 - Added a notification icon for when there are unread messages (instead of the blinking MESSAGES)
 - Added screenshots
_In the end I did not change the icons, as I mentioned somewhere else, because 'notifications' come in so many shapes, that the Envelope icon is just as good as a bell/bubble/phone+line icon._

**iOS app**
 - Fixed music attributes
 - Added new app-identifiers
 - Added a Readme

**Locale (nl_NL)**
 - Added translations for the Messages menu items


gfwilliams:
Nice - thanks! 

A few things though:

* Message widget: I thought in https://github.com/espruino/BangleApps/issues/1033 we'd said the messages icon would be configurable? I guess we'll go with new new one and see if anyone complains.
* Translations: It doesn't matter right now, but I wouldn't add more translations in that file. I think you've seen https://github.com/espruino/BangleApps/issues/136 but that'd be the best way to handle translations going forward.
* Message icons: this is great - but I wouldn't bother adding functions like `getFBImage` - just put them inline. Each function takes a certain amount of RAM so in a way it's better to just have a single big function.

gfwilliams:
Now it's done, please don't worry about reinstating the flashing message icon! If someone complains we'll add the settings, but no point doing it before!

> I did create functions for the icons that I used more than once

Ahh, sorry, yes, that's perfect then. I'd had a quick glance down the code and didn't see any other references. If you're using it multiple times then it definitely makes sense.

PR #1085, opened 2021-12-14
Rebble & Pebble: load steps from health app (as fallback)

myxor:
I would like to see the Rebble watch have the fallback to load the daily steps from the health app (which is installed by default).

There is already at least one person on reddit which stumbled on the problem that Rebble and Pebble watch does not deliver steps without the pedometer widget installed (which is not installed by default):

https://www.reddit.com/r/Banglejs/comments/rg1llc/step_count_on_bangle_js_2/

Contains changes from previous MR: https://github.com/espruino/BangleApps/pull/1073

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4722,7 +4722,7 @@
     "id": "pebble",
     "name": "Pebble Clock",
     "shortName": "Pebble",
-    "version": "0.04",
+    "version": "0.06",
     "description": "A pebble style clock to keep the rebellion going",
     "readme": "README.md",
     "icon": "pebble.png",
--- a/apps.json
+++ b/apps.json
@@ -4875,7 +4875,7 @@
     "id": "rebble",
     "name": "Rebble Clock",
     "shortName": "Rebble",
-    "version": "0.02",
+    "version": "0.03",
     "description": "A Pebble style clock, with configurable background, three sidebars including steps, day, date, sunrise, sunset, long live the rebellion",
     "readme": "README.md",
     "icon": "rebble.png",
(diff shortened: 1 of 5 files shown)


myxor:
@gfwilliams  is saw your commit https://github.com/espruino/Espruino/commit/c09de7bc3a206760f16fd9482c2152dd73a9f605
Do you suggest to use that new functionality for this like follows?:

```
function getSteps() {
  if (WIDGETS.wpedom !== undefined) {
    return WIDGETS.wpedom.getSteps();
  } else {
    if (Bangle.getHealthStatus) {
       const day = Bangle.getHealthStatus("day");
       if (day) return day.steps;
     }
  }
  return '????';
}
```

gfwilliams:
It's not quite that easy, since `getHealthStatus` already exists. Realistically it's probably going to be a matter of checking the version number is 2v11 or later.

BUT I really don't like the idea of having these little extra bits of code in every clock face that wants to displays steps. It's why I'm a bit hesitant about merging this.

I just don't get why we can't just ask the user to install a pedometer? Hell, it can even be put in the app JSON so a pedometer is installed automatically.

Right now, if I merge this, users without a pedometer will inexplicably find that their watch face takes a whole second to render and their battery life is seriously degraded. I don't think that's an improvement over just having '???' until they install the a pedometer widget.

myxor:
I agree that having a (maybe built-in espruino) solution for this would be better than updating all clocks, widgets, etc.

I will close this MR for now.
We could need the changes to add the pedometer widget as dependency for several clocks. I will maybe create this in the near future.

gfwilliams:
Great - thanks! Yes, I think adding the dependency is a nice easy way of handling this for now.

And when 2v11 is released we can update the pedometer to use the new Bangle.js functionality

PR #1087, opened 2021-12-14
Thermometer app working on BangleJS 2

myxor:
Make temperature value readable on smaller screens


gfwilliams:
Great - thanks! I should also add that the Bangle.js 2 has a barometer which has a more accurate temperature sensor, so it is also possible to use that if it's available.

PR #1089, opened 2021-12-15
Implemented Bluetooth Low Energy advertising of sensor data

jeffyactive:
Advertises barometric, GPS, HRM and magnetic sensor data as services using standard GATT characteristics.


gfwilliams:
Nice - thanks!

PR #1090, opened 2021-12-15
Added Latvian language to "locales.js"

marissz:
Added Latvian diacritics in "charFallbacks"
Added "lv_LV" locale
Sorted "charFallbacks" alphabetically

Also just looking at the ISO map, I guess for all lowercase diacritics we have to add uppercase ones too, right?


gfwilliams:
That's great, thanks!

>  I guess for all lowercase diacritics we have to add uppercase ones too, right?

Correct, yes

PR #1092, opened 2021-12-15
Localized "Fahrenheit Thermometer" app updated for Bangle 2

elrod16:
Title says it all. Updated to get temp on Bangle.js 2 and use ‘locale' to address issue with last PR. Latest commit adds temperature averaging to compensate for device inaccuracies. It's not even the same app anymore really. Implementation changed.


elrod16:
I'm not really sure why this is hanging in PR limbo because 

>"I don't think having duplicate apps is a good idea - if anything we should be removing stuff like Fahrenheit Temp and just make it use the locale module so one app can do both."

And that is exactly what I've done here.

gfwilliams:
It's been open 2 days - it's hardly 'PR limbo'.

>> if anything we should be removing stuff like Fahrenheit Temp and just make it use the locale module so one app can do both."

> And that is exactly what I've done here. 

Well, what you've done here is leave those 2 apps as-is, and rather than fixing them you added a THIRD app which contains the changes that really just should have been in `thermom`, but that doesn't work on Bangle.js 1 or use Bangle.js 2's more accurate thermometer. It's not really quite what I had meant.

I also can't merge it easily because something's happened with `apps.json` which is causing a merge error. So... What I've done is added the features you've added in to the original "thermom" app, and I'll close this PR.

gfwilliams:
Hi - sorry, there is no intention of hostility. I have 150+ emails every day to go through at the moment and this is my last day before christmas holiday so I just want to get things sorted asap - I don't have the luxury of writing long replies if I want to get through them all. I was delaying merging your PR because I was trying to decide what to say/do.

Sorry you got the impression every conversation you had with me was unpleasant. Do you have an example of one and what I could have done better? Also anyone else reading this - feel free to tell me if you think I haven't been friendly/welcoming.

I'll manually merge your PR now then. Users can try and decide which one of the apps they want to use.

> And that's it. Execution doesn't continue.

Do you mean Bangle.js crashes? Because `Bangle.getPressure()` should return a promise, which should then 'complete' in a few seconds, so `Bangle.getPressure().then(print)` should print something after a while?

Obviously I didn't know that wasn't working for you, but it does seem to work for other people so it'd be good to sort it out in your case rather than stopping everyone else from using the more accurate temperature sensor.

In an ideal world `E.getTemperature` would use the pressure sensor on Bangle.js 2, but it returns immediately - and to use the pressure sensor we need to wait a while for it to take a reading so the two are not compatible,

elrod16:
I appreciate the candidness of your reply and your willingness to talk this out. It shows the quality of your character. I may very well be at fault for misconstruing your intended tone, some bug reports and such I interpreted it as snippy. Could very well just be my personal biases and preconceived notions of people coloring the conversation that way too.

As far as Bangle.getPressure(), it just times out and the app doesn't move beyond it. Typing it in the console yields nothing too. That's why clear back on my first PR for this app (when I didn't know I had a locale failure) I had edited it out. It just endlessly sits on the "Loading..." screen when using that call.

gfwilliams:
Thanks! Yes, think maybe if I add short replies to things it can come across as rude. It's not my intention, it's just the sheer volume of mail I get right now means it's either that or I have to start ignoring stuff, which I don't want to do. This has only been an issue the last 2 months or so, so I'm hopeful things will calm down a bit next year.

Manually merged, so closing this now. Should be in the app loader - I changed the name to 'Localized Thermometer' so there aren't two apps of the same name.

I'll file an issue about your getPressure issue and see if I can reproduce it on a device here - it's great to know it works with sensiBLE though - it could be the difference between powering on the sensor all the time and requesting the last reading, and powering in on specifically to take a reading then powering off. It could be I'm not leaving quite long enough for every sensor to power up.

elrod16:
I replied to you in the issue you opened, but since it isn't directly relevant to that I thought I'd message you here: I know this isn't the first PR of mine where I've been trying to code around a bug I wasn't aware of. I'll try to look out for that better. As it is, with my inexperience with your platform and language, I almost always see any kind of unexpected behavior as a mistake or ignorance on part rather than a potential bug. How I said, I'll work on that, I just don't want to flood you with bug reports that result from true ignorance.

PR #1097, opened 2021-12-15
Update ptlaunch: Respect theme colors

crazysaem:
Changes:

- Respect theme colors. Fix: Do not pollute global space with internal variables ans functions in boot.js

PS: Thanks @gfwilliams for the `g.theme.fg` and `g.theme.bg` variables, but I ended up not needing them. I changed the way the circles are rendered, and now only need the fg color which is always used by default (I think).


gfwilliams:
Great - thanks!

PR #1098, opened 2021-12-15
pooq Round: tap for transient complications; bugfixes

stephenPspackman:
Add tap-to-enhance, turning on complications transiently.
Remove some dead code.
Fix a fairly serious bug arising from version skew in the orientation-dependent update rate mechanism, which was probably making a lot of unnecessary events.
Reduce the number of screen clears to eliminate unmotivated blanking.
Alter readme to elucidate that (as I now understand) lack of backlight-on-twist was a bug in the system, and this is not a permanent feature of the watchface.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4941,7 +4941,7 @@
   { "id": "pooqround",
     "name": "pooq Round watch face",
     "shortName":"pooq Round",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks!

PR #1101, opened 2021-12-16
widpedom: Save lastUpdated as a unix timestamp

### LINKED ISSUE #961: Pedometer widget step count randomly resetting
Recently while useing the Pedometer Widget I noticted that the step count seems to reset to 0. I don't know why this is happening but my guess is it is either in the way the step count is stored or how the app writes to it.

nebbishhacker:
Works around espruino/Espruino#2116 in order to fix #961

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1590,7 +1590,7 @@
   {
     "id": "widpedom",
     "name": "Pedometer widget",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks!

PR #1103, opened 2021-12-16
New clock: Circles clock

myxor:
A clock with circles for different data at the bottom in a probably familiar style.

Screenshot


gfwilliams:
Nice - thanks!

PR #1111, opened 2021-12-16
Update locales: de_DE, de_AT and es_ES

myxor:


gfwilliams:
Thanks! For US, is it intentional that you're not having clocks to use am/pm at all? For 12 hour times it means there is no way to tell if the time is morning or evening...

myxor:
I didn't change anything on the en_US locale. I only changed de_DE, de_AT and es_ES.

Or am i getting your comment wrong?

gfwilliams:
Ahh, sorry! I misread the git diff (it cut out some lines above). The lack of am/pm was for "de_AT" - and I see it's been done for some other German language stuff too

myxor:
The lack of "ampm" caused a parsing error in the web interface when opening the custom dialog for the languages app for me. Therefore I added it for de_AT.

PR #1112, opened 2021-12-16
feat: Non-dithering calendar app (clrcal)

Athemis:
I liked the existing calendar app, but imho it lack legibility on the Bangle.js 2. This apps uses non-dithering colors and additionally supports localized names for months and day of the week.


nebbishhacker:
Those sound like good changes! I wonder if it would be worth integrating these changes into the main calendar app, rather than creating a new app.

gfwilliams:
Thanks! Yes, as @nebbishhacker says - are there other changes apart from the localized names and colors? It seems like there's not much point having one barely readable calendar and one good one

Athemis:
I'm absolutely open to put my changes into the main calendar app instead. I was hesitant to do so in the first place because the current color scheme is probably fine for the Bangle 1. But if you think that it is ok to apply these changes for both Bangle 1 and 2 then I'll open a new pull request incorporating said changes into the main calendar app.
Btw is there a way to detect whether the app is running on a Bangle 1 or 2 at run time? We could then keep the original color scheme for Bangle 1 then. Otherwise I'll add a setting to change the color scheme.

myxor:
> Btw is there a way to detect whether the app is running on a Bangle 1 or 2 at run time? 

Should work with 
```
if (process.env.HWVERSION == 1) {
// Bangle.js 1
}
if (process.env.HWVERSION == 2) {
// Bangle.js 2
}
````

PR #1123, opened 2021-12-17
messages: Hide widget when all unread messages are dismissed

nebbishhacker:


gfwilliams:
Thanks! Interestingly I'd thought that the widget wouldn't have been shown while in the messages app - guess I didn't test that bit properly :)

nebbishhacker:
> I'd thought that the widget wouldn't have been shown while in the messages app

It doesn't - but the messages app doesn't need to be open for messages to be dismissed from the phone itself.

Issue #1121, opened 2021-12-17
Circleclock icon is 178x178 and takes up whole screen in dtlauncher


hughbarney:
Tried to contact the developer through github but can't figure out how to PM on github.
![download (2)](https://user-images.githubusercontent.com/3630481/146467848-487293d0-b798-4c71-9b09-e9b01d4914c8.png)

nebbishhacker:
I've noticed a couple other incorrectly sized icons (albeit not that dramatic). I wonder if it'd be worth adding an icon size check to sanitycheck.js.

myxor:
Circles clock icon is fixed via https://github.com/espruino/BangleApps/pull/1113

gfwilliams:
Thanks! I've been been through using @nebbishhacker's sanitytest additions and have fixed all the massively big or broken icons.

I haven't fixed all the ones that are too big though - I notice you've got a few too @hughbarney!

> Warning: JS icon pebble.img should be 48x48px but is instead 64x64px
> Warning: JS icon rebble.img should be 48x48px but is instead 64x64px

hughbarney:
@gfwilliams - oops, its a size setting thats difficult to remember.  It the size you choose works, then you don't question if you got the size right or not.   Maybe a tickbox on the imageInverter that said - [x] this is (48x48) an icon for a Bangle App.

PR #1113, opened 2021-12-16
Circles Clock v0.02

myxor:
- **Fix icon size**
- Add battery warn visualization
- Add battery charge visualization
- Fix name of ChangeLog file

PR #1124, opened 2021-12-17
sanitycheck: Error on incorrectly sized js icons

nebbishhacker:
Adds an error to sanitycheck.js for incorrectly sized app icons, to help catch issues like #1121. As something of a side-effect, it also ensures that app icon files follow the format described in the README.

Since there are currently 24 apps that would trigger the error, and I imagine they won't all be fixed right away, I added a `GRANDFATHERED_ICONS` list that makes them only trigger a warning instead of an error. Once the existing icons are fixed the `GRANDFATHERED_ICONS` list can be removed.

Circleclock should be fixed by #1113, so I didn't include it in the `GRANDFATHERED_ICONS` list.

These icons currently trigger the warning:

```
Warning: JS icon hebrew_calendar.img should be 48x48px but is instead 0x0px
Warning: JS icon fontclock.img should be 48x48px but is instead 40x40px
Warning: JS icon slidingtext.img should be 48x48px but is instead 40x40px
Warning: JS icon solarclock.img should be 48x48px but is instead 40x40px
Warning: JS icon sweepclock.img should be 48x48px but is instead 40x40px
Warning: JS icon matrixclock.img should be 48x48px but is instead 40x40px
Warning: JS icon speedo.img should be 48x48px but is instead 44x36px
Warning: JS icon s7clk.img should be 48x48px but is instead 50x50px
Warning: JS icon mmonday.img should be 48x48px but is instead 0x0px
Warning: JS icon bclock.img should be 48x48px but is instead 0x0px
Warning: JS icon snek.img should be 48x48px but is instead 64x64px
Warning: JS icon dane.img should be 48x48px but is instead 47x48px
Warning: JS icon fclock.img should be 48x48px but is instead 217x176px
Warning: JS icon digiclock.img should be 48x48px but is instead 254x254px
Warning: JS icon astral.img should be 48x48px but is instead 50x50px
Warning: JS icon alpinenav.img should be 48x48px but is instead 50x50px
Warning: JS icon slomoclock.img should be 48x48px but is instead 64x64px
Warning: JS icon tapelauncher.img should be 48x48px but is instead 0x0px
Warning: JS icon arrow.img should be 48x48px but is instead 50x50px
Warning: JS icon doztime.img should be 48x48px but is instead 40x40px
Warning: JS icon swiperclocklaunch.img should be 48x48px but is instead 40x40px
Warning: JS icon pebble.img should be 48x48px but is instead 64x64px
Warning: JS icon rebble.img should be 48x48px but is instead 64x64px
Warning: JS icon circlesclock.img should be 48x48px but is instead 176x176px
```

### DIFF
--- a/bin/sanitycheck.js
+++ b/bin/sanitycheck.js
@@ -60,6 +60,7 @@ const STORAGE_KEYS = ['name', 'url', 'content', 'evaluate', 'noOverwite', 'suppo
 const DATA_KEYS = ['name', 'wildcard', 'storageFile', 'url', 'content', 'evaluate'];
 const FORBIDDEN_FILE_NAME_CHARS = /[,;]/; // used as separators in appid.info
 const VALID_DUPLICATES = [ '.tfmodel', '.tfnames' ];
+const GRANDFATHERED_ICONS = ["hebrew_calendar", "fontclock", "slidingtext", "solarclock", "sweepclock", "matrixclock", "speedo", "s7clk", "mmonday", "bclock", "snek", "dane", "fclock", "digiclock", "astral", "alpinenav", "slomoclock", "tapelauncher", "arrow", "doztime", "swiperclocklaunch", "pebble", "rebble"];
 
 function globToRegex(pattern) {
   const ESCAPE = '.*+-?^${}()|[]\\';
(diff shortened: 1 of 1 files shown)


gfwilliams:
That's great - thanks! Yes, that's a really good idea.

I'm not sure what your thoughts are but I feel like having icons *smaller* than 48x48 may not be too big a deal, since if the background was transparent in that area, not storing the extra just makes things faster.

> icon digiclock.img should be 48x48px but is instead 254x254px

ouch! For that alone it's worth this PR!

nebbishhacker:
> icons smaller than 48x48 may not be too big a deal

Yeah, the slightly-undersized icons probably aren't a big deal. It should be straightforward to change the check if you would prefer to allow them.

The 0x0 icons are probably worth looking into, though - the ones I checked are actually fairly large strings, they just start with a string of zeros. I suspect they may be encoded incorrectly somehow.

gfwilliams:
Thanks! Yes I think I have fixed the 0x0 ones now.

My guess is they (and maybe some others?) were encoded as 'image object' and the base64 string was copied out of the buffer field?

SpyGuy0215:
> That's great - thanks! Yes, that's a really good idea.
> 
> I'm not sure what your thoughts are but I feel like having icons _smaller_ than 48x48 may not be too big a deal, since if the background was transparent in that area, not storing the extra just makes things faster.
> 
> > icon digiclock.img should be 48x48px but is instead 254x254px
> 
> ouch! For that alone it's worth this PR!

oops! That's on me xD

PR #1126, opened 2021-12-17
calendar: improve legibility on Bangle2 and localize strings

Athemis:
This pull request supersedes #1112 by merging its changes into the main calendar app. The non-dithering color scheme will be used by default on Bangle 2. 

A setting has been included to switch between the original color scheme and the non-dithering one.
In addition, month and day of the week will be localized if the language app is installed. Furthermore I did some minor clean-ups, mainly removing unused variables from the code.


gfwilliams:
Great - thank you!

PR #1135, opened 2021-12-17
Disable screen turn on on bluetooth status change

MindTwister:
The bluetooth widget turns the screen on on bluetooth status change. For me at least, this is undesirable since I don't sleep in the same room as my phone, but close enough that it will, from time to time, connect or disconnect, disturbing my wife's sleep.


gfwilliams:
Thanks! And sorry for the delay getting this merged over Christmas!

gfwilliams:
Also, just a note that ideally you need to also bump the version in 'apps.json' as well as the ChangeLog - but I will do that this time

PR #1141, opened 2021-12-19
[LCARS Clock] Multiple screens including graphs for health data.

peerdavid:
This PR adds multiple screens (fullscreen now). On the second screen, the step data and HRM data is shown for 1 month using the "health" app. If the app is not installed, a warning is shown to the user that this functionality is only provided, if the app is installed.


gfwilliams:
Thanks - looks awesome! And sorry for the delay getting this merged over Christmas!

PR #1142, opened 2021-12-19
new clock face

HeiLaut:
new dark full screen analog clock face (andark)


myxor:
Is the `add_to_apps.json` file requiered?

hughbarney:
any screenshots ? Would love to see what this looks like ?

HeiLaut:
> any screenshots ? Would love to see what this looks like ?

you can view the readme file, containing a screenshot here

https://github.com/HeiLaut/BangleApps/tree/master/apps/andark




> Is the `add_to_apps.json` file required?

no it isn't required

Didgeridoohan:
I tested this and it looks absolutely lovely!

But, it's going to kill the battery since it's refreshing every second even with the watch on, even though it looks like you've covered that by trying to stop updates then the LCD is off. Does the lcdPower event even work on the Bangle.js 2, since the LCD is always on (doesn't seem like it)? Maybe the isLocked command and the lock event is better suited.

Also, a suggestion: only stop the seconds updating when the LCD is off and let minutes and hours keep updating.

Reasoning: I (and probably others) have my watch set to only unlock and light up the LCD on a rather more aggressive twist than default (to save the battery draining from unnecessarily turning on the LCD). From what I could see you (intend to) stop the entire clock updating whenever the LCD is off, which means that whenever I check the time without unlocking/turning on the screen the time shown wouldn't be correct...

HeiLaut:
> I tested this and it looks absolutely lovely!
> 
> But, it's going to kill the battery since it's refreshing every second even with the screen locked and dimmed, even though it looks like you've covered that by trying to stop updates when the LCD is off. Does the lcdPower event even work on the Bangle.js 2, since the LCD is always on (doesn't seem like it)? Maybe the isLocked command and the lock event is better suited.
> 
> Also, a suggestion: only stop the seconds updating when the LCD is off and let minutes and hours keep updating.
> 
> Reasoning: I (and probably others) have my watch set to only unlock and light up the LCD on a rather more aggressive twist than default (to save the battery draining from unnecessarily turning on the LCD). From what I could see you (intend to) stop the entire clock updating whenever the LCD is off, which means that whenever I check the time without unlocking/turning on the screen the time shown wouldn't be correct...

thanks for the suggestions. I changed ist, so the watch (bangle.js2) updates only after 60 seconds, when locked and hides the second hand.

HeiLaut:
i'm new to the bangle js community and also on working with github. Don't know why checks failed, and don't know what to to next. eventually anybody can help me out?
thanks

gfwilliams:
Hi! Where it says 'Build Failed' you can click 'Details' then the highighted 'The Build' text on the next page. Looks like it's complaining about the icon. I think `"icon": "andark_icon.png",` should have been left as `"icon": "app_icon.png",` - but I'll merge and make that tweak now

gfwilliams:
I was wrong - there was just a space in `"app_icon.js "`. fixed now :)

PR #1150, opened 2021-12-21
Messages app v0.15 - fix icon width problem from v0.12

DDDanny:
widget.jus adjusted:

1. ICONS moved to toplevel
2. line break for draw function
3. dynamic clear and touch based on ICON dimensions (x,y) taken (static) from 1st! icon header information


jeroenpeters1986:
Nice catch, I only saw it after acitivating lots of widgets :-)

gfwilliams:
Thanks! But please could you keep ICONS *not* at the top level?

The problem with is being defined outside the function is it then uses RAM ALL the time (since both images are loaded). While not too big a deal on Bangle.js 2 things are quite tight on Bangle.js 1 and this could cause users problems

DDDanny:
will be replaced by new PR....

DDDanny:
OK, no usage large object-properties to save RAM.

@jeroenpeters1986: Thanks for fixing! I had some in GIT usage....

jeroenpeters1986:
No problem, I actually merged your first PR locally before the small other things I edited. When I saw Gordon’s remarks I also processed that.

gfwilliams:
Thanks!

PR #1152, opened 2021-12-21
Add Four Twenty Clock

thedod:
A clock that tells when and where it's going to be [4:20](https://en.wikipedia.org/wiki/420_%28cannabis_culture%29) next


barakplasma:
Is it possible to rewrite apps/ftclock/mkFourTwentyTz.py using Node.js instead? While python is common, since this project already relies on Node.js , I think that it would be a better choice for future maintainability.

barakplasma (changes requested):
Is it possible to rewrite apps/ftclock/mkFourTwentyTz.py using Node.js instead? While python is common, since this project already relies on Node.js , I think that it would be a better choice for future maintainability.

thedod:
> Is it possible to rewrite apps/ftclock/mkFourTwentyTz.py using Node.js instead? While python is common, since this project already relies on Node.js , I think that it would be a better choice for future maintainability.

Makes sense. It so happens that the CSV format is about to change on next release, so this could be a good opportunity to rewrite this in JS.

thedod:
> Is it possible to rewrite apps/ftclock/mkFourTwentyTz.py using Node.js instead?
Done 👍

thedod (commented):
As far as I understand, the concern raised by @barakplasma was dealt with, and this pull request is ready to merge.

gfwilliams:
Thanks! Is this definitely tested on Bangle.js 1 as well?

I'd be a little concerned that `apps/ftclock/fourTwentyTz.js` defines quite a big set of arrays that use up a lot of RAM. I'm merging now, but if you wanted to save some memory you could write it as `exports.getTimeZones = function(tz) { if (tz==0) return ... }` which would keep everything in flash memory.

thedod:
> I'd be a little concerned that `apps/ftclock/fourTwentyTz.js` defines quite a big set of arrays that use up a lot of RAM. I'm 
merging now, but if you wanted to save some memory you could write it as `exports.getTimeZones = function(tz) { if (tz==0) return ... }` which would keep everything in flash memory.

Cool. I'm on this.

> Is this definitely tested on Bangle.js 1 as well?

That was an "educated guess". Sadly, it wasn't educated enough :(
I don't have a physical Bangle.js 1 to test this on. Worse: I've just tried it on an emulator, and it doesn't seem to run [neither on bangle.js 1 nor on bangle.js 2 emulation] and claims "Module fourTwenty not found" (althogh it _does_ recognize `require("fourTwenty")` and `require("forTwentyTz")` on the REPL), so on my next PR (with the RAM-saving version of `fourTwentyTz.js`) I'll stop claiming `"allow_emulator"`, and `"BANGLEJS"` at the `apps.json` entry. Sorry for jumping to conclusions.

gfwilliams:
Thanks!

Yes, the emulator is a bit tricky - right now it only uploads the one app file, so multi-file apps don't work. It's something that could be fixed but it a bunch of work and right now is low priority :)

PR #1156, opened 2021-12-21
Update calendar weekend colors for start on Sunday

Cramill:
If week is set to start on Sunday, make Saturday (last row) and Sunday (first row) have weekend colors.


gfwilliams:
Thanks! In future, please could you provide a ChangeLog entry and update the version number in apps.json? I'll do that now though

PR #1161, opened 2021-12-22
High Contrast Clock : Added Support for messages

peeweek:
Small addition : adds support for opening messages using BTN3:
- Only enabled when app is present
- Only visible when messages are present


gfwilliams:
Nice - thanks!

PR #1165, opened 2021-12-22
Update Anton Clock with optional seconds and ISO-8601 date

dirkhillbrecht:
Hi,

I have added seconds and ISO-8601 date format (both optional) to the Anton Clock. I like that one best as it is the best readable clock app. Seconds can be configured to be shown only when display is unlocked, so power is still saved to great extent. Date and weekday presentation still obey I18N, date can optionally be switched to ISO-8601 format (YYYY-MM-DD) regardless of the locale. I use this modified Anton version for several days now and like it very much! I also tested my changes against the Bangle 1 in the emulator.

This is my first attempt to Bangle programming, so please forbear with me if I overlooked something. I tried to follow the docs to my best knowledge. Of course I will make changes if there are obstacles for inclusion of my changes to the official code repository.

I like the whole programming environment of the Bangle very much! It is extremly low-barrier which is great. I have some Arduino and lots of Java backend development experience, but Bangle and Espruino is really good at making things simple and allowing fast success when developing stuff. Very well done!

Kind regards,
Dirk Hillbrecht


MaBecker:
please create a new clock

Didgeridoohan:
Agreed... If you're going to make changes this big (your design is very different from the stock Anton clock) you should either make them an option in the app settings or make a whole new clock.

T0TProduction:
> Agreed... If you're going to make changes this big (your design is very different from the stock Anton clock) you should either make them an option in the app settings or make a whole new clock. 

They said it's all optional from settings. 

I think since these are all optional additions, it can and should be part of this watchface and not a new one. We don't need Fork-Number-21 of App-34 and spam the entire app loader with multiple versions If the same thing.

Didgeridoohan:
Fair enough... Have to admit that in my quick look I missed the settings changes and they do look like they allow the watch to remain just as it is today if one wants to. With the exception that the font used for the date and day doesn't fit on the screen.

dirkhillbrecht:
Hi everyone,

thanks for the feedback! It is always great to see that someone actually is interested in such work! I totally agree with some of the remarks and have adopted the following changes:

- Colorisation of seconds (and date) in seconds screen is configurable and disabled by default.
- Presentation of date in seconds screen is configurable, seconds are centered if no date is shown.
- Initlal settings are chosen in a way to pertain the original look (no seconds, long date format, upper-case)
- Formatting is also tweaked a bit to resemble v0.03.
- Added a README.md according to the "best practices" documentation.

I have not changed using the "Vector" font instead of the scaled "6x8" font for all non-"Anton" text. It just looks way better.

I have force-pushed my branch but retained the original commit so that my changes are a bit clearer. Note that I am not totally familiar with the usual github practices regarding rebasing, pushing and pull requests. I hope this is intuitive.

Note that you can install my "Anton clock" from https://dirkhillbrecht.github.io/BangleApps/ - either the one I issued in this pull request or an independent "Anton clock with seconds" app which contains exactly the same code but can be installed in parallel and independent of the "Anton clock".

Regards, Dirk

Didgeridoohan:
@dirkhillbrecht First of all, apologies for not doing my homework before commenting earlier.

Second, this looks like a great addition to the Anton clock.

Couple of things:
I agree that the vector font looks better, but I think it's a bit too big. The nice thing with the original 6x8 font is that the date and day is small enough that you don't really see them if just glancing at the watch. With the vector font I find my eyes being drawn to the date/day rather than the time...

With seconds enabled the colon mark before the seconds doesn't stay in place. Whenever a 1 is displayed it jumps with the changed width of the number. Should probably stay in place.

dirkhillbrecht:
@Didgeridoohan No problem, I wasn't totally clear about the configuration possibilities.

Do you really feel the Vector font being distractive? If you set the date format to "short" ("long"=false...) or ISO8601 and disable uppercase letters, it seems as unintrusive as the 6x8 font version to me. But it looks so much better and is much better readable.

Yes, the "jumping colon" is a bit unusual. But you have it in the hour-minute line, too, if the number of "1"s change (23:19 -> 23:20). After all, Anton is a proportional font. You could put the characters at absolute positions, but that would result in annoying gaps between the colon and the rest. It is just that the display is so much better than the 7-segment displays of the digital watches we all are used to. Only this makes proportial fonts possible in the first place.

Didgeridoohan:
Yes. My personal preference in this case is the 6x8 font, no matter which setup is used (although I really like the addition of ISO8601 and the lower case text). The vector font is simply too distracting (for me).

Regarding the colon between hours and minutes, that doesn't jump quite as often as the one for the seconds so that's not an issue (as I see it). I think a gap between the colon and the seconds is a lot less annoying than the jumping colon.

dirkhillbrecht:
I have now added an option to change the date and day of week between 6x8 or vector font. Default is 6x8. I have also adopted spacing. If you now just update Anton Clock without any settings, it should look exactly like version 0.03. Changes are in a seperate checkin in the pull requested branch.

Didgeridoohan:
chefskiss

andiohn:
I posted some enhancement ideas on the forum. I think this should replace the stock one, it's excellent.

dirkhillbrecht:
I have made some changes to my Anton clock enhancements:

- Seconds use a smaller version of the Anton font now. This removes the biggest flaw of my proposal: The lower widget area is now unobstructed.
- Colon in front of seconds can be disabled. This was requested several times in the discussions.
- In seconds mode, date can be shown with weekday instead of year.
- I use multi-selections in the settings now instead of only booleans. That makes configuration a bit smaller and much clearer as there are no implicit dependencies between options any more.

gfwilliams:
This looks great - thanks! I keeping the default pretty 'normal' and making it tweakable with settings works really well.

PR #1166, opened 2021-12-22
New App: Swipe back to the Clock

crazysaem:
App Description:

> Let's you swipe from left to right on any app to return back to the clock face.

Link to try it out: https://crazysaem.github.io/BangleApps/#swipe%20back%20to%20the%20clock


t0m1o1:
Nice work. Possibly worth combining with Swiper Clock Launch (or adding the function to swiping from clock to launcher to this) as that app should have this functionality but is currently broken.

crazysaem:
@t0m1o1 yeah, those were pretty much my thoughts! I actually added to the README that the "Swiper Clock Launch" app inspired me to write this app. And I also added that Quote "You can use the Swiper Clock Launch or the Pattern Launcher to further enable touch-only support of the watch."

I didn't want to add it to the "Swiper Clock Launch" app directly though, since it's a nice simple app that works nicely on it's own, and can be easily combined with this app without any issues.

t0m1o1:
I only fixed a tiny bug in it. 
I think the functionality of your app is distinct enough to exist on its own right my only concern is if the bug stopping the other one working (looking into it at the moment seems to be an issue at firmware level with the load() function) is resolved then the two apps may become incompatible.
A solution could be to add the swipe to launcher function to your app or remove the (broken) swipe from launcher function from the other one. 
I really like your implementation with a white list.

elrod16:
Yeah I personally prefer the idea of them remaining separate because it offers a higher degree of customization on a per user basis. Really, with that mindset, maybe it'd be best to just prune out that functionality from the swipe clock launcher? It'd save a tiny bit of storage and it'd make the purpose of each listing really specific to the functionality you want. Or I could totally wrong.

gfwilliams:
Thanks! Yes I think this works well as a separate app - maybe if you haven't done already you could blacklist the launcher is swipelaunch is installed, so both work together nicely?

Issue #1155, opened 2021-12-21
App crash on Bangle.js 2 related to GPS time


HilmarSt:
### Eventually GPS time is not usable on Bangle.js 2

On my Bangle.js 2 there are 2 cyclic NMEA telegrams, e.g.:
```
$GNGGA,105700.296,,,,,0,00,25.5,,,,,,*74
$GPGSV,1,1,03,01,22,177,,04,37,213,,27,43,029,,0*5F
```
The $GPGSV is long time standard but I've never seen $GNGGA before. $G?GGA contains the UTC time in the first field.

I'm not sure if this applies to the hardware of Bangle.js version 2 but there is some evidence:
> https://ava.upuaut.net/?p=768
> "Side note on the Ublox 8 series GPGGA is replaced by GNGGA by default to indicate the navigation information is from multiple GNSS sources (I.e American GPS and Russian GLONASS normally)."

The event Bangle.GPS with Call type: Bangle.on('GPS', function(fix) { ... }); 
provides a "fix" object with contains "time": Date,   // Current Time (or undefined if not known)"

It seems to me that the parser which fills fix.time does only look for "GPGGA" and always returns "undefined" on Bangle.js2, even in the GPS fix state.

This leads to a crash in the **GPS Info App**:
The function formatTime(fix.time) in gps-info.js crashes in toUTCString() if fix.time contains "undefined".
Workaround:
```
function formatTime(now) {
  if (now == undefined) { return "no GPS time available"; }
    else {
  var fd = now.toUTCString().split(" ");
  var time = fd[4].substr(0, 5);
  var date = [fd[0], fd[1], fd[2]].join(" ");
  return time + " - " + date; }
}
```

-> _Probably all Apps that rely on fix.time will not work on Bangle.js2_

HilmarSt:
Pull request created: https://github.com/espruino/BangleApps/pull/1167
The PR also contains an improvement: "The 4-dot progress indicator is replaced by the number of satellites in view."

HilmarSt:
I found out that the parser works with GGA and not GPGGA as assumed by me.
The crash will be solved with Pull Request https://github.com/espruino/BangleApps/pull/1167

PR #1167, opened 2021-12-22
GPS Info App: Added number of satellites in view and fixed crash with GPS time

HilmarSt:
1. (bugfix) Workaround for a crash in toUTCString() if fix.time has the value "undefined", see https://github.com/espruino/BangleApps/issues/1155
2. (improvement) The 4-dot progress indicator is replaced by the number of satellites in view.
Tested on Bangle.js 2 with firmware 2v11

### DIFF
--- a/apps/gpsinfo/gps-info.js
+++ b/apps/gpsinfo/gps-info.js
@@ -16,13 +16,17 @@ var lastFix = {
   time: 0,
   satellites: 0
 };
-var nofix = 0;
+var SATinView = 0;
 
 function formatTime(now) {
-  var fd = now.toUTCString().split(" ");
-  var time = fd[4].substr(0, 5);
-  var date = [fd[0], fd[1], fd[2]].join(" ");
-  return time + " - " + date;
+  if (now == undefined) {
+    return "no GPS time available";
+  } else {
+    var fd = now.toUTCString().split(" ");
+    var time = fd[4].substr(0, 5);
+    var date = [fd[0], fd[1], fd[2]].join(" ");
+    return time + " - " + date;
+  }
 }
 function getMaidenHead(param1,param2){
   var lat=-100.0;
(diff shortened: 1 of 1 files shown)


apps/gpsinfo/gps-info.js
```diff
@@ -100,12 +103,18 @@ function onGPS(fix) {
...
     layout.sat.label = fix.satellites;
-    nofix = (nofix+1) % 4;
-    layout.progress.label = ".".repeat(nofix) + " ".repeat(4-nofix);
+    layout.progress.label = "in view: " + SATinView;
   }
   layout.render();
 }
 
+function onGPSraw(nmea) {
+  if (nmea.slice(3,6) == "GSV") {
```
BartS23, line 112: On BangleJS2 there is also a message which starts with $BDGSV e.g. $BDGSV,1,1,00,0*74)
Your slice also detects this message and sets SATinView to 0.

Maybe it is better if the slice is wider:
```suggestion
  if (nmea.slice(0,7) == "$GPGSV,") {
```

HilmarSt: Thank you for the hint, I will create a version that adds the numbers of BDGSV and GPGSV.

apps/gpsinfo/gps-info.js
```diff
@@ -109,8 +109,13 @@ function onGPS(fix) {
 }
 
 function onGPSraw(nmea) {
+  var nofBD = 0;
+  var nofGP = 0;
```
BartS23, line 113: nofBD and nofGP should be global, otherwise one of them is always zero.

apps/gpsinfo/gps-info.js
```diff
@@ -100,12 +103,23 @@ function onGPS(fix) {
...
   layout.render();
 }
 
+function onGPSraw(nmea) {
+  var nofBD = 0;
+  var nofGP = 0;
+  if (nmea.slice(3,6) == "GSV") {
+    // console.log(nmea);
+    if (nmea.slice(0,7) == "$BDGSV,") nofBD = nmea.slice(11,13);
+    if (nmea.slice(0,7) == "$GPGSV,") nofGP = nmea.slice(11,13);
```
BartS23, line 117:
```suggestion
    if (nmea.slice(0,7) == "$BDGSV,") nofBD = Number(nmea.slice(11,13));
    if (nmea.slice(0,7) == "$GPGSV,") nofGP = Number(nmea.slice(11,13));
```

When I log SATinView, I sometimes get "0010" because nofBD and/or nofGP are not recognized as numbers.

apps/gpsinfo/gps-info.js
HilmarSt: Thanks, resolved with commit aa45ecd .

gfwilliams:
Thanks! In future please could you update the version in apps.json and the Changelog? I'll do this now though

PR #1169, opened 2021-12-24
Fix bug with debug info

t0m1o1:
The list didn't have min and max set and only clipped to a correct value meaning the scroll didn't work correctly at the extrema (v could go up or down beyond the bounds).


gfwilliams:
Argh - thanks for the fix!

PR #1176, opened 2021-12-25
Added vibration compatibility for Banglejs 2

untbu:
At least banglejs 2 does not support the Bangle.buzz() function (or at least mine does not). That's why I added an alternative function to vibrate for banglejs 2 to the alarm app. I moved the part of recalling buzz() down to avoid having it twice. And I changed it to let the Banglejs 2 buzz four times, but that's just my preference.

### DIFF
--- a/apps/alarm/alarm.js
+++ b/apps/alarm/alarm.js
@@ -39,20 +39,28 @@ function showAlarm(alarm) {
     require("Storage").write("alarm.json",JSON.stringify(alarms));
     load();
   });
+  function vibrate(counter) {
+    VIBRATE.write(1);
+    setTimeout(() => VIBRATE.write(0), 100);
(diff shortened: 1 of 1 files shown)


Poolitzer:
Hey, my Bangle.js 2 vibrates just fine with `.buzz()`.

HilmarSt:
Also my Bangle2 works with Bangle.buzz(100);
I'm using the "Hour Chime" widget: https://github.com/espruino/BangleApps/blob/master/apps/widchime

untbu:
Looks like this is an issue of my Bangle.js...

HilmarSt:
Just found out how to test buzz() very easily:
- go to https://www.espruino.com/ide/# and Connect the Bangle.
- on the left side (in the black window) type Bangle.buzz(200) and press return.
My Bangle2 does vibrate after the command.

PR #1178, opened 2021-12-26
Add Prettier

KTibow:
This PR adds the Prettier code formatter, by adding docs about it, adding CI to make sure the formatter is ran, and formatting the existing files.


.github/workflows/format.yaml
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  check-format:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+      - name: Check formatting
+        run: |
+          shopt -s globstar
+          npx prettier --check **/*.js
```
mabuchner, line 13: How about also formatting TypeScript files (e.g. see [here](https://github.com/espruino/BangleApps/tree/master/apps/banglerun/src)) and json files?

README.md
```diff
@@ -509,6 +509,17 @@ You can use `g.setColor(r,g,b)` OR `g.setColor(16bitnumber)` - some common 16 bi
 
 [Graphics Class](https://banglejs.com/reference#Graphics)
 
+## Code Formatting
+
+BangleApps uses [Prettier](https://prettier.io/) for code formatting, with default settings.
+
+You can use it [online](https://prettier.io/playground.html) or on your computer:
+
+1. Install Prettier with `npm install prettier --global`.
```
mabuchner, line 518: I wouldn't use a globally installed prettier version, since this forces everyone to install exactly the same version.

Instead, I would add a particular version of prettier as `devDependency` to `package.json`

```json
  "devDependencies": {
    "prettier": "2.3.0"
  },
```

Then you can run `npm install` or `npm ci` to install a local copy of prettier.

mabuchner: npm scripts in `package.json` can then be used to invoke prettier.

```json
  "scripts": {
    "check-format": "prettier --check .",
    "fix-format": "prettier --write .",
  },
```

`npm run check-format`

`npm run fix-format`

.github/workflows/format.yaml
mabuchner: Or simply let prettier take care of all supported formats?

```
npx prettier --check .
```

KTibow: Alright, I'll do that.
```suggestion
          npx prettier --check .
```

README.md
KTibow:
```suggestion
1. Install Prettier and other dependencies with `npm install` while in BangleApps.
```

.github/workflows/format.yaml
KTibow: Running Prettier has discovered a number of invalid HTML problems. I'll attempt to fix them so it can run.

KTibow:
i have no idea why this messes up travis

T0TProduction:
I'm pretty sure development needs to still be possible directly from the browser (especially within the web IDE). Adding packages that need to be installed for **all** devs won't be possible anytime soon. 
I know this because I started working on global, shared, TypeScript support and a very important requirement was to be entirely optional.
As far as I understand you want to force it onto all apps, which would entail such forced package installation.

package.json
```diff
@@ -12,6 +12,7 @@
     "start": "npx http-server -c-1"
   },
   "dependencies": {
-    "acorn": "^7.2.0"
+    "acorn": "^7.2.0",
+    "prettier": "2.3.0"
```
mabuchner: This should be a `devDependency`.

mabuchner:
> I'm pretty sure development needs to still be possible directly from the browser (especially within the web IDE). Adding packages that need to be installed for all devs won't be possible anytime soon.

Development in the browser is still possible. prettier is just a `devDependency`. It was just accidentally added as production `dependency`.

> As far as I understand you want to force it onto all apps, which would entail such forced package installation.

No, you don't have to install the NPM packages. The CI build will fail, if your code isn't formatted as expected and then you can fix those formatting issues manually or via some prettier editor plugin.

It would be nice, if the WebIDE had a "format code" button, but this merge request is already large enough and I wouldn't add this feature here.

mabuchner:
> i have no idea why this messes up travis

I have also no idea.

`BangleApps/apps/setting/settings-icon.js` was changed from

```js
require("heatshrink").decompress(atob("mEw4UA///+Nj5lCt9TH+cBqtVoALWqALTgoLUiALFgoLDqoBBAAQGCHAdRBYdFKwZECqv614ECGQQsCr2q1W1BYkVAoPqBYOrAoNUBYdXBQIAB6oLDEQgkEBYdaBYelBYt6BYetBYYvBtWq0EK1WpF4ZfBIwIFBJATCDBY6PDiuq1AEBlWqBdA7KKZZrLQZabNWZLLLcZb7LBYVV/WvAgRfCNYNRBAVVoq/FJQRECR4gnBEwQEBggLDGQg4CBag4DBaBWBBaoATA"))
```

to

```js
require("heatshrink").decompress(
  atob(
    "mEw4UA///+Nj5lCt9TH+cBqtVoALWqALTgoLUiALFgoLDqoBBAAQGCHAdRBYdFKwZECqv614ECGQQsCr2q1W1BYkVAoPqBYOrAoNUBYdXBQIAB6oLDEQgkEBYdaBYelBYt6BYetBYYvBtWq0EK1WpF4ZfBIwIFBJATCDBY6PDiuq1AEBlWqBdA7KKZZrLQZabNWZLLLcZb7LBYVV/WvAgRfCNYNRBAVVoq/FJQRECR4gnBEwQEBggLDGQg4CBag4DBaBWBBaoATA"
  )
);
```

and now acorn believes, that there is an unexpected token at (5:1).

mabuchner:
> I have also no idea.

I think, the problem is, that the sanity check wraps the JavaScript code in braces for some reason

https://github.com/espruino/BangleApps/blob/d9d5926416566762e59b953c745f989d23c9b334/bin/sanitycheck.js#L150

gfwilliams:
Hi!

> It would be nice, if the WebIDE had a "format code" button, but this merge request is already large enough and I wouldn't add this feature here.

It does! Ctrl-B. All shortcuts are listed in when you click settings in top-right.

Thanks for this, but I'm afraid I'm not sure about merging this at all. It looks like you've spent ages on this but it would have been really good to have some kind of discussion about it first.

While in theory it'd be good, we're trying to be friendly to new contributors who are generally completely new to version control and CI. For the majority I think jumping through code formatting hurdles as well would really put them off.

I think the reality is that 75% of all new commits would immediately fail the CI test, and I (or whatever maintainer) would just end up having to merge them regardless and then waste extra time trying to fix the CI checks by running the reformatter. IMO the real issue is that CI fails for formatting would then end up hiding serious failures - so we'd end up merging something that appeared just to be badly formatted, then finding out after reformatting that it was totally broken, and having to manually revert the commit.

I'm open to others' views on this, but I think it does need to be discussed with the wider community - not just those on GitHub who are already more likely to be used to this kind of thing.

Even failing that I've got some issues with the formatting rules which IMO sometimes make the code less readable. Also reformatting all the icons to add newlines is (IMO) only a negative thing. You can no longer just paste code direct from the image converter, and I can't remember offhand whether icons are minified before upload but it's possible it'll could even hurt upload speed.

Having said all that as @T0TProduction says having a tool that *could* be used if a user wanted to would only be a good thing - it's just requiring all users to do it could I feel be problematic.

mabuchner:
> Thanks for this, but I'm afraid I'm not sure about merging this at all.

Understandable, a pull request, which touches almost all files shouldn't get merged light-heartedly. Maybe there is a way to introduce Prettier more gradually.

> While in theory it'd be good, we're trying to be friendly to new contributors who are generally completely new to version control and CI. For the majority I think jumping through code formatting hurdles as well would really put them off.

I don't think, that telling people to format their code consistently, when committing to a shared code base is asking too much. Fixing the formatting can be done really easy with Prettier. You don't have to read any complicated style guide or something, just run `npx prettier --write .` and commit the changes.

Even for novice users, the WebIDE could be used to format the code.

> I think the reality is that 75% of all new commits would immediately fail the CI test,

Yes, definitely, I forget to format my code all the time. It's possible to set up git commit hooks, but that's not straight forward.

From my experience, if you don't enforce a certain style with the CI, then the formatting will always get messed up over time and then you get all those random formatting changes in pull requests. One person forgets to format the code and then another person makes a small change, but also reformats the file.

Documenting the code style and onboarding new staff also becomes a lot easier. Just tell anyone to use Prettier.

People also seem to prefer a machine telling them, that their code might need some improvements over a human doing the same.

> so we'd end up merging something that appeared just to be badly formatted, then finding out after reformatting that it was totally broken, and having to manually revert the commit.

But that's why git uses branches, so you can make sure, that the code works before merging it. Branches which fail the CI build, should not get merged.

> Even failing that I've got some issues with the formatting rules which IMO sometimes make the code less readable.

Before my company introduced code formatting tools, I was also sceptical. However, after more than two year, I have to say, that those cases happen very rarely and if they happen, you can still disable Prettier for a certain code block.

Overall the advantages outweigh the disadvantages. I can focus on the code and I don't have to think about the formatting. Merge conflicts due to formatting issues also happen a lot rarer. When doing code reviews, there is absolutely no time wasted on pointing out formatting issues. Code written by other people will also feel a lot less alien, if the formatting matches your own.

> Also reformatting all the icons to add newlines is (IMO) only a negative thing.

If the icons follow a certain name pattern, then they can easily get excluded by adding the name pattern to `.prettierignore`.

> Having said all that as @T0TProduction says having a tool that could be used if a user wanted to would only be a good thing - it's just requiring all users to do it could I feel be problematic.

An alternative could be to run Prettier over the whole code base from time to time. However, I wouldn't use this approach in my projects since it results in a messy git commit history.

T0TProduction:
I am 100% for consistent formatting in code.
And IMO a shared config would be great.

But as Gordon said, If we have additions by users that are not necessarily devs, but "just" tinkerers, adding too much around simply "uploading" code makes it much more difficult to contribute.

If the maintainers (aka. just Gordon?) are willing and able to reformat PRs it's fine to enforce it. But if that's simply out of scope, I think it just adds unwanted complexity (If forced by CI mind you, not the adding it as optional part)

Another thing to take into consideration is that while this is a shared repo, the codebase is not shared. Every app is essentially its own codebase, so a lot of the benefits of a universally shared format might not be as useful here anyways.

TLDR; 
I think adding a config that anyone *can* and maybe even should use is a great idea. This should probably have rules that are discussed in the community just like Gordon proposed.
I think the CI checks should be removed, and the changes to existing code (formatting others apps within this PR) should be removed.

mabuchner:
> I think adding a config that anyone can and maybe even should use is a great idea

Just FYI Prettier doesn't want configuration files. One is supposed to use the tool with the default configuration as much as possible.

T0TProduction:
> > I think adding a config that anyone can and maybe even should use is a great idea
> 
> Just FYI Prettier doesn't want configuration files. One is supposed to use the tool with the default configuration as much as possible.

Huh, I've never used prettier without adjusting it. But if we don't need to add a config, so be it, doesn't influence the rest tbh.

Another Idea I just had:
If/When Typescript support is added, we could force the use of prettier there. (Even with CI checks)
Since all TS apps need a Local Dev Environment anyway, adding prettier there is no big deal. Would probably be very easy to run only on .ts files as well. (In the CI)

KTibow:
I left for a bit, and a whole bunch happened.
I'm a bit confused what I should do now - GitHub says there's merge conflicts, Travis won't run, and it sounds like there's some other things I should do?
Should I try to apply these changes more manually, instead of all-at-once?

gfwilliams:
> GitHub says there's merge conflicts, Travis won't run

It's pretty much guaranteed by PRs like this that change over 900 files. Merging this is now a bit of a nightmare, but even if it is merged, it then forces pretty much any contributor or anyone with a fork to have their own merge nightmare too. 

The travis build failed because of https://github.com/espruino/BangleApps/pull/1178#issuecomment-1004609724 - @mabuchner the issue here is not that the sanity check is wrong, it's that the sanity check has detected that the icons are now broken. The icons need to not have a semi-colon on the end because the JS file is evaluated as an expression.

@KTibow thanks for doing this - but I just feel that right now this is too much, especially as a first contribution...

* this PR changes basically every app in the repo
* In many cases the code is actually changed **away** from what we suggest at http://www.espruino.com/Code+Style
* It also refactors all the markdown files away from a format that we use for the rest of Espruino
* merging would seriously inconvenience pretty much every contributor by giving them huge merge conflicts on any change they made, and I don't think the improvements here outweigh that
* This merge also destroys any commit history so we no longer know who made a specific change in an app file
* Given this change breaks the icons in the app loader, I'm a bit concerned this is pretty much untested on real devices? While formatting changes are unlikely to be an issue, it still worries me that 300 apps have been changed and probably very few of those (if any) will have been tested for regressions

I think right now it's best just to close this... While we could have a tool to allow contributors to apply formatting rules to their specific apps, right now all you have to do is hit Ctrl-B in the web IDE to do it automatically, so I wonder if a special tool is that much of a benefit.

KTibow:
> this PR changes basically every app in the repo

Commentary:
It's at least *supposed to* come out with an identical result.
Solution:
Yes, it does modify them, but it depends on your definition of "change".

> In many cases the code is actually changed away from what we suggest at http://www.espruino.com/Code+Style

Commentary:
Thanks for that link, I couldn't find it before.
Solution:
I'll address this more further down.

> It also refactors all the markdown files away from a format that we use for the rest of Espruino

Commentary:
I originally had it set up as `**/*.js`, so it was only JS files, but someone convinced me to use `.`.
Solution:
Anyway, I believe a `.prettierignore` file could fix this (if we continued to use Prettier).

> merging would seriously inconvenience pretty much every contributor by giving them huge merge conflicts on any change they made, and I don't think the improvements here outweigh that

Commentary:
I thought about that too, but any big change will cause that.
Solution:
No idea, but still, I don't think it would be that hard to reformat some code using an automatic tool.

> This merge also destroys any commit history so we no longer know who made a specific change in an app file

Commentary:
Well, it's git, so it keeps around all commits, however GitHub only shows you the latest one when you're just viewing a file.
Solution:
I could try to stack the changes onto the previous commits, or try to edit the commits to be based off of a formatted version.

> Given this change breaks the icons in the app loader, I'm a bit concerned this is pretty much untested on real devices? While formatting changes are unlikely to be an issue, it still worries me that 300 apps have been changed and probably very few of those (if any) will have been tested for regressions

Commentary:
That's just me being lazy and not uninstalling/reinstalling all my apps.
Solution:
Addressed further down.

> I think right now it's best just to close this...

Understandable.

> right now all you have to do is hit Ctrl-B in the web IDE to do it automatically

What tool is used for this? If it's made just for Espruino, has anyone bothered porting it to run on multiple files? If you use this on the app icons, will it break them?
(Couldn't find any info from a quick search)

KTibow:
> It seems to be js_beautify

I [tried it](https://github.com/KTibow/BangleApps) on the codebase, it still modified a ton. Does the web app use some special config? Or should I just drop this all (which I can)?

gfwilliams:
If you look at the code in the EspruinoWebIDE repo there may be some config, but...

> Or should I just drop this all

Sorry, but I think you should drop it, yes.

There are loads of issues with Bangle.js software and your help with those would be hugely appreciated, but really I've haven't heard of anyone that has had any issues with code formatting. At most the formatting is a minor inconvenience to one or two people, and those people probably don't like the default formatting style we use anyway.

But if I merge these changes it's going to cause a lot of inconvenience. There are almost 600 forks, many by users who aren't properly familiar with Git and are even just using the GitHub online editor - for them it'll be a total mess. I really think that right now the small improvement to readability doesn't outweigh the trouble this will cause many of the contributors.

KTibow:
Okay. I'll see if there's anything else interesting that would help.

PR #1180, opened 2021-12-27
Update locales.js for Italian language

fparri:
I am providing small fixes for the Italian localization.

- using : to separate hours from minutes, instead of .
- in datePattern: i changed to the more usual way of representing dates in Italian.
- in day: there was an extra space character before sabato, showing in certain apps. I removed that.


fparri:
I closed this because I have found another bug which needs to be fixed. I'm opening a new commit.

PR #1181, opened 2021-12-27
Added grid positioning and swipe controls to calculator app

untbu:
added grid positioning and for Bangle.js 2 swipe controls to switch between numbers, operators and specials. 
On Bangle.js 2 only one page of numbers, operators and specials is displayed. Swiping left displays the operators, swiping right displays the specials and swiping up or down displays the numbers. After an operator is entered, it also switches to the numbers.


gfwilliams:
Nice, thanks - that looks much better!

Just FYI it's an idea to change the version on apps.json as well as adding the ChangeLog line but I'll do that now

PR #1182, opened 2021-12-27
Tea Timer - a simple timer

ThFGG:
A simple timer. You can easyly set up the time. The initial time is 2:30

On the first screen, you can
- tap to get help
- swipe up/down to change the timer by +/- one minute
- swipe left/right to change the time by +/- 15 seconds
- press Btn1 to start

Press Btn1 again to stop the timer
- when time is up, your Bangle will buzz for 15 seconds
- and it will count up to 60 seconds and stop after that


gfwilliams:
Nice - thanks!

PR #1183, opened 2021-12-27
New App: Touchtimer - Quickly create a timer with a number pad input

crazysaem:
App Description:

> Quickly and easily create a timer with touch-only input. The time can be easily set with a number pad.

Link to try it out: https://crazysaem.github.io/BangleApps/#touch%20timer


gfwilliams:
Thanks! Looks like a great addition!

PR #1185, opened 2021-12-28
Custom QR Code: Change brightness on touch

halemmerich:
In some conditions (and with some smartphone cameras) a brightness change can help the qr scan to work better.


gfwilliams:
Nice - thanks!

PR #1188, opened 2021-12-29
Add Pro Menu App

Ronin0000:
Pro Menu is a app for Bangle.js 1 that replaces Bangle.js 1's built in menu function.


gfwilliams:
Great - thanks!

PR #1189, opened 2021-12-29
Tidy up CoreTemp app

ihewitt:
I've tidied up the coretemp app and made it a bit more friendly/pretty and to be able to run in the background.
Added skin temp display, an activity widget and a settings page.
also added this sensor into the recorder app as an experiment.

I'd like to reorganise this all into a generic "sensors" environment for activating/deactivating all sensors in one place (e.g. CSC, BTHR etc), with hooks for other apps to be able to interrogate for sensor availability/state. But I made a complete mess on first attempt so will start again.


gfwilliams:
Thanks! I'm not sure if you spotted but you can define sensors in such a way that the `Recorder` app can pick them up and then write their data into the file it records:

https://github.com/espruino/BangleApps/blob/master/apps/recorder/widget.js#L102-L113


I've been meaning to do it with CSC and the BT chest monitor but haven't got around to it yet.

I guess they could even be used in other apps too

ihewitt:
Cool I'll take a look (once I've got some free time, cleared the work inbox and completed the weekends race) I like the idea of a more generic "sensors" that could be queried similar to having the global WIDGETS instance but I made a complete mess on attempt #1. :grin:

PR #1190, opened 2021-12-29
Time and Life app (draft)

isaacphi:
Adds "Time and Life" app.

Displays the time and then decays according to the rules of Conway's Game of Life.


gfwilliams:
Looks nice - it's marked as draft, but are you happy with merging?

isaacphi:
> Looks nice - it's marked as draft, but are you happy with merging?

@gfwilliams  I think there's something wrong still. It works when I run the code on my bangle directly from the web IDE but it gets stuck on the loading screen when I try uploading it from https://isaacphi.github.io/BangleApps/#time%20and%20life. There could be something wrong with my configuration. Are there any logs I can look at or specific debugging steps to follow? Are you able to run it?

gfwilliams:
Just a note on this - looks like the error given is:

```
Uncaught Error: 'E.InlineC' calls should have been replaced by the Espruino tools before upload
 at line 49 col 2 in timeandlife.app.js
```

It looks like the BangleApps loader isn't compiling the inline C code - which is kind of odd because I thought it should have been able to.

gfwilliams:
Just merged but I made some changes - I manually pulled in the compiled version of your code, but also you were drawing 484 8x8 images when you could have just used `g.drawImage` so I tweaked that and it should be much faster

isaacphi:
@gfwilliams I appreciate it, thanks!

PR #1192, opened 2021-12-31
Add Sleep Widget

kyleplo:
# Sleep Widget
A simple widget for the Bangle.js 2 that completely clears the screen when the watch is locked. I have not done extensive testing, but this should save battery life.

## Features
- Completely clears screen and stops running app when watch locks
- Returns to clock when woken
- Widgets continue running even when locked (for apps that use widgets for timers and notifications)

## Controls/Usage
- Sleeps when watch would normally lock (do not press anything for this to happen)
- Once asleep, use any of the controls you would normally use to unlock the watch to wake it

## Known Issues
- Only works in apps that show widgets
- May react badly to widgets that use the full screen for notifications

## Creator
[kyleplo](https://kyleplo.com)

See my website for contact information


kyleplo:
I'm closing this because I'm working on a better version that doesn't use a widget.

PR #1193, opened 2021-12-31
Fixed Italian localization

fparri:
I am providing small fixes for the Italian localization.

1. using : to separate hours from minutes, instead of .
2. datePattern: changed to the more usual way of representing dates in Italian.
3. datePattern: fixed so that clocks such as Anton Clock work fine
4. day: removed an extra space.


gfwilliams:
Thanks!

PR #1199, opened 2022-01-02
Presentor App release 3.0

7kasper:
WIP, but already submitting for possible comments and reference.

TODO:
- [ ] Fix Bangle.js 1 working.
- [ ] Fix layout.
- [ ] Better code structure.


gfwilliams:
Looks good - should I merge anyway? Then it's easier for folks to try and comment.

I do notice you have SpecialReport - do you think any of the existing report types might work? If they're in boot0 it'll likely make pairing a bit more reliable as the Bangle will then always have HID enabled: https://github.com/espruino/BangleApps/blob/master/apps/boot/bootupdate.js#L22

7kasper:
Hi Gordon!
Yeah I guess that's fine too! I can just push bug fixes with new PRs.
As for the HID I don't know.
I need a mouse in the HID but couldn't get the default mouse HID to work. Well It did work on Android and Linux but not on windows 10 interestingly enough. So I tried a lot of different stuff. In the end I only got HID mouse to work if I advertise that I am a keyboard mouse combination. Wierd but ok... 😄 So I wrote the Special HID interface that can send keystrokes and mousemovements at the same time.

Maybe we can incorporate it in Bangle or a library instead? Advertise a standard HID that advertises a mouse, keyboard etc and then add all the functions in the library? [So basically similar to how the keyboard is now but then one that does both a keyboard and a more feature-complete mouse]. If you want I can take a look at that.

gfwilliams:
Thanks! That sounds like a good plan - adding it as another option for HID (or maybe even just replacing the two keyboard+mouse with just your version?). Definitely having a library or something to allow apps to use it nicely would be great.

For now what's in there looks great though

PR #1203, opened 2022-01-04
Pastel: Added weather icons, other small changes that have been soak tested over christmas

hughbarney:
added weather icons to pastel
fixed 2 icons to 48x48
estings put  Turn Off  into  Utils mmenu

fixed issue with dub93 clock when in light theme


gfwilliams:
Great - thanks!

Issue #1049, opened 2021-12-10
Weather conditions not matching an icon on non-english languages


myxor:
When using Gadgetbridge and the Weather notification app on android (as suggested on https://codeberg.org/Freeyourgadget/Gadgetbridge/wiki/Weather) to push weather conditions to the Bangle.js the weather object 'weather.json') looks like this in German:

```
{ "hide": true, "expiry": 14400000,
  "weather": { "temp": 276, "hum": 74,
    "txt": "Ein paar Wolken",
    "wind": 7, "wdir": 113,
    "loc": " -- ",
    "time": 1639131366761.5966796875,
    "wrose": "se"
   }
 }
```
You can see that the textual representation of the condition is in German: "Ein paar Wolken" ~ "Some clouds".

The weather app does string comparison with the english words of weather conditions to find the matching icon:
https://github.com/espruino/BangleApps/blob/master/apps/weather/lib.js#L207

```
if (condition.includes("thunderstorm")) return drawThunderstorm;
if (condition.includes("freezing")||condition.includes("snow")||
  condition.includes("sleet")) {
  return drawSnow;
}
[...]
```

Any idea how we can get the condition matching working in multiple languages?

myxor:
Probably this is a problem which should be fixed upstream in the weather app. There is already an issue in the repository but development seems to have stopped: https://github.com/gelin/weather-notification/issues/56

nebbishhacker:
Matching on the weather text was always a bit of a hack - ideally GadgetBridge would include some sort of language-independent condition code or icon name in the weather object.

gelin/weather-notification#56 appears to be more about being able to control the language returned by OpenWeatherMap, which I guess would allow you to avoid the issue by always setting the language to english.

See also #541

myxor:
I created an issue on the Gadgetbridge git: https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/2504

myxor:
Looks like we will get the weather conditions code in the next release of GadgetBridge: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2511

thanks @rigrig

myxor:
Recently released version 0.64.0 contains change
* Bangle.js: send weather condition code to device

I will have a look at updating weather app code later. If someone else wants to do it feel free.

Didgeridoohan:
[...]

PR #1205, opened 2022-01-04
Weather app v0.14: use weather condition code for icon selection

myxor:
If weather code is available. use weather condition code for icon selection else use textual representation as before.

See
#1049 (comment)


gfwilliams:
Brilliant - thanks!

PR #1208, opened 2022-01-04
Updates ios, messages, widbatpc

jeroenpeters1986:
APP ios: 0.08: Added more app identifiers, added 'cannot display' in case a message goes empty because of replacements
APP messages 0.17: Widget icon width
WIDGET widbatpc: 0.15: Added option to always display the icon when charging (useful if 'hide if charge greater than' is enabled)


gfwilliams:
Great - thanks!

PR #1209, opened 2022-01-04
Ported Grocery to Bangle.js2

adamschmalhofer:
Well, it didn't need any changes actually (other then in apps.json). I didn't bump the version number or add a Changelog entry as that doesn't seem to add value in this case.


gfwilliams:
Nice! thanks!

PR #1210, opened 2022-01-04
Ported cprassist to Bangle.js2

adamschmalhofer:
Ported cprassist to Bangle.js2 by converting the fixed positions to calculated ones by the screen size.


gfwilliams:
Thanks!

PR #1214, opened 2022-01-05
weather: fix widget icon

rigrig:
Only the widget was broken, not the app.

Fixes https://github.com/espruino/BangleApps/issues/1049#issuecomment-1005560918


myxor:
Thanks @rigrig for fixing this!

PR #1217, opened 2022-01-05
New App: recentapp

storm64:
This app is a tiny boot script to access the recent/active app info data as object with: `require(\"recentapp\")`


gfwilliams:
Hi - I'm really not sure this is a good idea. It seems the reason for this is the recent regression which causes `__FILE__` not to be set?

I'm literally just getting a fix in for that now, so at that point you'll just be able to check `__FILE__` from your apps/widgets/boot code.

storm64:
> Hi - I'm really not sure this is a good idea. It seems the reason for this is the recent regression which causes `__FILE__` not to be set?
> 
> I'm literally just getting a fix in for that now, so at that point you'll just be able to check `__FILE__` from your apps/widgets/boot code.

Hi,

thanks for the fast reply. I'm not sure if I fully understand what you mean because I didn't know about a `__FILE__` variable or module for now.
If it is possible to get the name of a loaded file from within `__FILE__`, this would be perfect and make this app unnecessary.

Could you tell me the firmware version this fix would be in?

storm64:
Just installed 2v11.7 and it works great.

gfwilliams:
Perfect - thanks for letting me know!

PR #1218, opened 2022-01-05
Puzzle15

dirkhillbrecht:
A 15 puzzle implementation with 3x3, 4x4, 5x5 playing fields, drag gesture interface and animatied stone movement, rebased onto the current master version of the official loader app.


gfwilliams:

Issue #1020, opened 2021-12-07
Split apps.json


qucchia:
It seems a little counter intuitive to me that every app has its own folder for code, image, README, and yet their actual data is outside in a file called `apps.json`. This file is already quite large, and will probably not scale well.

My suggestion is to split this file into many files called `app.json` (or another sensible name) inside each of the apps' folders. We'd have to automate this process as there are a lot of apps, but it wouldn't be too difficult. Then at runtime these can all be put together into one file automatically.

elrod16:
I would second this because I've already ran into trouble with trying to edit apps.json and have had it exhaust my browser's RAM.

T0TProduction:
In theory, we should not need the `apps.json` at all.
The app loader could look at all the folders it has, scrape them for their `app.json`, and give IDs based on the path (which afaik is forced to be the same as the ID currently anyways)

elrod16:
The only issue I could see would be metadata like app category and which bangle it is for.

jeroenpeters1986:
I like this idea too, and started to think about it, but at the moment I wouldn't know how to loop over the subdirectories with native JS. Currently in `core/js/index.js`, `apps.json` is explicitly given as a file to check and is handled with XMLHttpRequest etc.

T0TProduction:
> The only issue I could see would be metadata like app category and which bangle it is for.

That information should be included in their respective `app.json`s, right?

elrod16:
I'm referring to the proposal of eliminating any app.json

T0TProduction:
> I'm referring to the proposal of eliminating any app.json

That's not what I meant to propose, sorry for the confusion 😅

gfwilliams:
This has actually been asked several times.

> ran into trouble with trying to edit apps.json and have had it exhaust my browser's RAM.

It's a 160k file, so not huge at all - so I so wonder if your browser needs looking at :)

> The app loader could look at all the folders it has, scrape them for their app.json

Have you considered how slow that might be, not to mention the load on the server? Someone goes to the app loader on a mobile network and then it has to do over 300 HTTP requests before it can even display the names of every app?

Also it's not possible to do a directory search when hosted on GitHub pages, so you'd have to have some specific code to use the GitHub API. If you're on a webserver where directory listing is disabled (like many) running the app loader would not be an option at all.

There's a bit of detail on it in this issue: https://github.com/espruino/BangleApps/issues/46

So really I think there are two solutions here:

[...]

Issue #46, opened 2019-11-15
[Suggestion] Code organization, folder per app


ra100:
Would it make sense to have a folder per app, for better navigation in the source?
example:
- `apps/animals/animals.js`
- `apps/animals/animals.json`

gfwilliams:
Thanks - yes, that could make sense. To be honest even with that it's going to get pretty hard to navigate, so I'm open to other ideas as well.

While having a single `apps.json` is a bit of a pain, there's something really handy about being able to test using GitHub pages on your own fork though.

ra100:
So for `apps.json`, would it make sense, to have that info actually in json of every app, and then just generating apps.json by simple script?
And going even further, what if every app had it's own repo and be linked to BangleApps just by apps.json? But that would mean, that apps could change without PR review to this repo, which sounds a bit unsecure.
Maybe just for now, some simple structure would be enough, and later, when it gets out of hands, to think more about it :)

gfwilliams:
> and then just generating apps.json by simple script?

Except that you could no longer just test it out using GitHub pages :(

While running a script seems fine to us, for a non-Node.js developer, installing Node.js then downloading and learning command-line git is quite a big learning curve.

Another option is to use something like https://github.com/gfwilliams/ThinGallery (so basically just scrape the website using client-side code), but that's slow.

> But that would mean, that apps could change without PR review to this repo, which sounds a bit unsecure.

Yeah, I mean if you don't want to do a PR you can always host your own repo. I think there's definitely some possibility of allowing proper third party stuff later, but especially for beginners I think having one repo with some level of vetting is a good idea.

ra100:
Yes, I didn't think about non-Node.js users. What about leaving it like it is, you'll just edit `apps.json` manually, for normal people :) and for lazy ones have js script, that would update it based on changes in apps. Buuut, right now, as I'm reading this, it looks like too much overhead. You'll have always the "check" in PR, where you'll see that apps.json was (or was not) updated.

Now I'm thinking how to combine good developer experience (with nice toolset, automated things and so on), but also keep it as simple as possible to add your own app.

ra100:
So how would moving to app folders look like.
- Have folder based on app id? Eg: `mclock` for Morphing clock, or `clock-morphing`?
- Should folder be specified in `apps.json`? `{"name":"+mclock","url":"clock-morphing/clock-morphing.json"},` or if it's the id, then create the path automatically, or add new property like `rootDir` or something with better name?

[...]

PR #1221, opened 2022-01-05
Let apps.json be generated by jekyll

adamschmalhofer:
Reading

#1020
#46

My understanding is that the requirement for a split apps.json solution
is:

 - work seamlessly with github pages (no manually calling a script)

 - single (or at most a few) apps.json for the browser to load for the
    store so the store page is fast

As the github pages already use jekyll to build the store page this can
be solved by a minimal template. This is a proof of concept of
that. If the concept is accepted: in a next step, I will create the
split files for more that just the two here. In addition we will need to 
update the documentation


rigrig:
I'd also like a javascript merge script for local development/testing using `npm start`.   
But I think that should be relatively easy using `npm watch`.

adamschmalhofer:
> I'd also like a javascript merge script for local development/testing using `npm start`. But I think that should be relatively easy using `npm watch`.

Or you could use `jekyll serve`

rigrig:
> Or you could use `jekyll serve`

Oh, cool. I didn't know jekyll could run standalone, figured it was just some preprocessor language for GitHub pages.    
...too bad it doesn't just work :-(
```
$ jekyll serve
Configuration file: /home/richard/src/BangleApps/_config.yml
jekyll 3.8.6 | Error:  The jekyll-theme-minimal theme could not be found.
...
```

So it looks like it requires you to also figure out/setup jekyll/ruby, instead of just sticking with npm/javascript which we already use for everything else.

But I guess this works fine for people using GitHub pages, and if we go this way I'd be happy to look into making it work locally with just npm.


Some bikeshedding: I would prefer the per-app files being named `app.json`, just makes more sense to me to combine several `app.json`s into one big `apps.json`.

gfwilliams:
Thanks! So actually everything can be handled with the `{%- assign apps = site.static_files | where: "name", "metadata.json" -%} ... ` code in `apps.json`?

I'll have to think about this as it's quite a big change, but it definitely seems to tidy things up a lot.

I'm in two minds about `metadata.json` - while I like `app.json` it's a bit confusing for widgets/etc, so probably keeping with what's here is better

BartS23:
I like @adamschmalhofer's solution very much.

I had another idea:

It would also be solvable via an additional branch ( e.g. gh-pages).
With every commit the apps.json is rebuilt via actions and merged with the commit into the branch gh-pages.

You can find an example here: https://github.com/BartS23/BangleApps/blob/master/.github/workflows/PreparePages.yml


Both solutions also have the possibility to leave the existing apps.json unchanged and only use the app.json/metadata.json for new apps.

gfwilliams:
Thanks for keeping this up to date! 

I just posted a question about this on the forum to see what those outside of GitHub think: http://forum.espruino.com/conversations/372018/

adamschmalhofer:
Updated. Looking forward to it being merged \o/.

gfwilliams:
Excellent - thanks!

gfwilliams:
@adamschmalhofer is it possible that when you ran your script, you pulled apps.json from `banglejs.com/apps` (which would have been out of date)? I'm seeing a few cases where apps didn't have metadata.json files or they are out of date

Issue #1137, opened 2021-12-18
Update battery level while charging.


pzo:
Using "Battery Level Widget (with percentage) (v0.14)", only while charging, it would be nice if the battery level percentage was updated. 
Now, you have to press the button to see at which level the battery is.

gfwilliams:
fixed with last update afaik

PR #1222, opened 2022-01-05
Update GPS Info App and "Battery Level Widget (with percentage)"

HilmarSt:
GPS Info App: Resolve one FIFO_FULL case and exit App with button press.

Battery Level Widget (with percentage): Increase screen update rate when charging - as requested by issue https://github.com/espruino/BangleApps/issues/1137

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -936,7 +936,7 @@
     "id": "widbatpc",
     "name": "Battery Level Widget (with percentage)",
     "shortName": "Battery Widget",
-    "version": "0.15",
+    "version": "0.16",
     "description": "Show the current battery level and charging status in the top right of the clock, with charge percentage",
     "icon": "widget.png",
     "type": "widget",
--- a/apps.json
+++ b/apps.json
@@ -1501,7 +1501,7 @@
   {
     "id": "gpsinfo",
     "name": "GPS Info",
-    "version": "0.06",
+    "version": "0.07",
     "description": "An application that displays information about altitude, lat/lon, satellites and time",
     "icon": "gps-info.png",
     "type": "app",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks - I'm interested in the FIFO_FULL issue... Was it this code that you found helped?

```
  Bangle.setGPSPower(0, "app");
  Bangle.removeListener('GPS-raw', onGPSraw);
  Bangle.removeListener('GPS', onGPS);
  load();
```

I've have hoped that `load()` would have effectively done the `removeListener` already - although we do generally leave GPS on while switching apps so that if the next app needs GPS you don't have to wait for it to get a lock again.

HilmarSt:
@gfwilliams 
> I'm interested in the FIFO_FULL issue... Was it this code that you found helped?

This is what had helped, because GPS-raw spills out many characters within a second:
```
Bangle.removeListener('GPS-raw', onGPSraw);
```

> I've have hoped that `load()` would have effectively done the `removeListener` already

Sure, but as I can't prove if a listener is still active, I've done it just for safety. I will remove the removeListener() functions with the next update.

> although we do generally leave GPS on while switching apps so that if the next app needs GPS you don't have to wait for it to get a lock again.

Really? It draws a big current while active and I've found (with the help of "GPS Widget (v0.03) Tiny widget to show the power on/off status of the GPS") that the GPS power is switched off (may be in load()?) when you return to the clock - at least in my Bangle.js 2 . I've also done it for safety and will remove it with the next update.

gfwilliams:
Thanks!

> I will remove the removeListener() functions with the next update.

Don't worry about it - it doesn't matter either way. It's just if doing that really helps then it's a sign that I may be able to do something in the firmware to help.

> Really? I've found ... that the GPS power is switched off when you return to the clock

GPS is left on, then 0.5 sec after the the next app loads it checks to see if it's still needed or not and changes state accordingly. As far as the next app is concerned, if it didn't ask for it, GPS is off.

But if you do explicitly turn GPS off in an app after that initial 0.5 sec then that state change is immediate :)

PR #1226, opened 2022-01-06
Update GPS Info app: Leave GPS power switched on on exit (will switch off after 0.5 seconds anyway)

HilmarSt:


myxor:
May i ask why this change is required?

gfwilliams:
More info at https://github.com/espruino/BangleApps/pull/1222 - thanks @HilmarSt !

PR #1230, opened 2022-01-06
Add Ruuvi Watch

emarc:
Add initial version of Ruuvi Watch, an app that allows you to check the status of RuuviTag bluetooth sensors.


gfwilliams:
Great idea - thanks!

PR #1231, opened 2022-01-06
creating new application

psbest:
mmind application first update


gfwilliams:
Thanks! Looks good, however it seems the icon (in mmind.icon.js) is 176x176 pixels when it should be 48x48. When it's bigger it won't render right in the launcher on the Bangle.

Please could you have a go at updating that?

psbest:
Thank you for the tip, changed the pixel size. Was hoping this would also resolve the app not showing up in the menu. But it still is not showing after upload using the app loader.  I can see the info, img and js file on the bangle storage, also start it through the webIDE works fine. Any other hints?

psbest:
update committed

PR #1234, opened 2022-01-07
Round dev

stephenPspackman:
Font fixes.
Remove auto-illumination, since the system once again does this.
Add screen auto-rotation while charging.
Add a little haptic feedback on touchscreen operations.


gfwilliams:
Nice - thanks!

PR #1237, opened 2022-01-07
Circlesclock v0.04

myxor:
* make configurable what to show in each circle
* add step distance & weather
* allow switching visibility of widgets
* increase size of circles and font inside for better readability (suggestion from http://forum.espruino.com/comments/16337868/)
* fix some graphical glitches

Changes reviewed by forum user Fteacher: http://forum.espruino.com/comments/16340930/


gfwilliams:
Looks great - thanks!

PR #1238, opened 2022-01-07
0.04 Fix alarm not activating

qucchia:
Just one keystroke was needed to fix the bug where Q Alarm did not activate sometimes!


gfwilliams:
Thanks!

PR #1241, opened 2022-01-07
[LCARS clock] Fixed bug that clock is unresponsive if gadgetbridge connects.

peerdavid:
During tests I found, that the clock became unresponsive, whenever gadgetbridge connects. This happens, because the LCD powers on and the clock is drawn again. Then, the line `require("health").readDay(new Date(), h=>steps+=h.steps);` never finishes. Not sure why but its clearly reproducible.

As a hack I the clock only partially when lcd is powered on, avoiding to access health. Not sure if this is a bug of this clock or a more general bug.


gfwilliams:
Thanks! The `require("health").readDay` not returning sounds like a bug - I'm pretty sure that doesn't normally happen so I'm not sure what's going on.

but... Who is calling `require("health").readDay`? Because I only see that code being called in the health app right now?

peerdavid:
Hi,

I'm using the `readDay` of the health module in the lcars clock. I (1) use it to show daily steps rather and (2) I provide a feature to show the step of the current day in form of a graph directly on the watch face:

```javascript
/*
 * Step counter via widget
 */
function getSteps() {
  var steps = 0;
  try {
    health = require("health");
  } catch(ex) {
    return steps;
  }

  health.readDay(new Date(), h=>steps+=h.steps);
  return steps;
}

[...]

try {
    health = require("health");
  } catch(ex) {
    g.drawString("MODULE HEALTH", 20, 110);
    g.drawString("REQUIRED.", 20, 130);
    return;
  }

health.readDay(new Date(), h=>{
      data[h.hr]+=h.bpm;
      if (h.bpm) cnt[h.hr]++;
    });
    require("graph").drawBar(g, data, {
      axes : true,
      minx: 1,
      gridx : 4,
      gridy : 100,
      width : 140,
      height : 50,
      x: 5,
      y: 25
    });
```

gfwilliams:
Thanks! Sorry for not getting this merged sooner. I'm still a bit confused by readDay failing. I can't seem to find a way to reproduce it here - but if you can figure out some way to get this to break when run from the IDE, please let me know and I'll see if I can fix it.

PR #1243, opened 2022-01-07
Ffcniftya clock 0.02

DDDanny:
added option to show ISO weeknumber (default is on)


DDDanny:
Hmmm, how happenend this core update ?!
https://github.com/espruino/BangleApps/pull/1243/commits/1ef8a0bfb0b52107cac461c7f009a7e74a6efa36

gfwilliams:
Thanks! For screenshots, did you know that when you're connected with the Web IDE you can just write `g.dump()` and you'll get a nice image of the clock displayed?

gfwilliams:
I have no idea how the core got automatically updated :( I think it's ok though as it should only be changing to what is already there on master

PR #1245, opened 2022-01-07
menuwheel: override Bangle.setUI to clear up touch handler

rigrig:
Because apps/libraries expect setUI to clean up touch handlers.

(I spend hours wondering why my `E.showScroller` was broken, turns out it was because I opened a menu :-()


gfwilliams:
Nice - thanks!

PR #1246, opened 2022-01-07
messages improvements

rigrig:
- Use theme colors and app-specific icon colors
- Spread message action buttons out
- Back button now goes back to list of messages

Still draft, because:
Haven't had a proper look at B1 yet, maybe it's better to stick with custom colors instead of theme there?
I'm not entirely sure if `fg/fg2`+`bg/bg2` are supposed to be used in this way (as alternating rows)    

Also: is it me, or is the Instagram icon broken?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -77,7 +77,7 @@
   {
     "id": "messages",
     "name": "Messages",
-    "version": "0.17",
+    "version": "0.18",
     "description": "App to display notifications from iOS and Gadgetbridge",
     "icon": "app.png",
     "type": "app",
--- a/apps/messages/ChangeLog
+++ b/apps/messages/ChangeLog
@@ -24,3 +24,4 @@
 0.15: Don't buzz when Quiet Mode is active
 0.16: Fix text wrapping so it fits the screen even if title is big (fix #1147)
 0.17: Fix: Get dynamic dimensions of notify icon, fixed notification font
+0.18: Use app-specific icon colors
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks! Just fixed the instagram icon.

So, I'm not sure about the theme colors - does anyone have any thoughts on this? It was kind of intentional that the colour of messages looked different to other apps, but maybe it is better just to keep with theme colours.

Moving between bg2/bg for the list seems ok, but I feel like it's probably a step backwards in looks from where it is now?

Also I think for the navigation - I know you can't test this right now - it but looks like you should be using `bg2` or `bgH` not `bg`, because as-is you'll have got rid of the 'accented' background on some bits - so it'll all be plain white/black.

rigrig:
I'm not too sure either, my thoughts were mostly "what's the point of having theme colours, if apps won't use them?".

Hand picked colours would (probably) look a lot better, but I think that when I realised the need to pick 4 sets (`[dark/light]`x`[Bangle 1/2]`), laziness tipped the scales towards the "just use theme" side ;-)

For navigation: I just realized we probably want to do a whole bunch of stuff with colour:
- highlight unread messages
- highlight active item for B1 navigation
- alternate rows

So that's probably not going to work with just theme colours anyway.

gfwilliams:
Hmm - yes, it's a shame I didn't expose the functionality but it'd be nice to be able to blend colors. For example to go halfway between the bgH+bg for the lists (or something like that). There are hacks we could use for that though...

Please could you take out the background color changes for now? And then I can merge icon colors and message button sizes and we can figure out a nice way to do this properly later

rigrig:
> Please could you take out the background color changes for now?

Sure. (I keep hoping someday I'll learn to properly separate commits ;-)

gfwilliams:
Great - thanks!

PR #1247, opened 2022-01-08
Limelight - simple analogue clock with configurable fonts, support for full or normal screen

hughbarney:
new (configurable fonts) Analgue clock based of based on the work of @Andreas_Rozek (Simple_Clock)
typo flipper to Flipper
use queueDraw() method in simplest clock


gfwilliams:
Thanks!

PR #1250, opened 2022-01-08
Fixed sv_SE formatting

benjarn:
Long date does not work well for Bangle.js2
This aligns with default locale en_GB.


gfwilliams:
Thanks!

PR #1253, opened 2022-01-09
Workout HRM 0.09

adamschmalhofer:
- Ported to Bangle.js2
- Home returns to clock, instead of menu
- Add settings


gfwilliams:
Great - thanks for these ports!

PR #1256, opened 2022-01-09
Anton-Clock can optionally show ISO-8601 calendar weeknumber

DDDanny:
The ISO-8601 calendar weeknumber added as or behind weekday.
The "display with seconds " mode was untouched.
No "core" update this time ;)
PS: copied GIST for ISO-Week # is als "MIT-license"


gfwilliams:
Thanks!

PR #1260, opened 2022-01-10
New widget: Heart rate alarm

halemmerich:
This invisible widget vibrates whenever the heart rate gets close to the upper limit or goes over or under the configured limits.


gfwilliams:
Looks great - thank you!

Issue #1259, opened 2022-01-10
Anton clock seems to have dow string hardcoded.


pzo:
cfr app.js  / function draw()  :
...
    var dowwumStr = require("locale").dow(date);
      dowwumStr = "thursday";
...

gfwilliams:
@DDDanny any idea what's going on here? This is obviously not ideal for the default clock...

DDDanny:
Yeah, really sorry - isn't it thursday oO
I discovered and fixed this morning via Pull Request #1262
and missed to check for issues...

gfwilliams:
Thanks! Sorry, missed your PR when I was responding here :)

PR #1262, opened 2022-01-10
Antonclk clock 0.05 show weeknumber *bugfix*

DDDanny:
I fixed the typo and format of the new option in readme.md
Furthermore I removed my testdata "weekday name" hardcoded to "thursday" - would be a mess if it is always thursday ;)
and tagged the string "week" that it has to be added in locale-app

### DIFF
--- a/apps/antonclk/app.js
+++ b/apps/antonclk/app.js
@@ -188,9 +188,8 @@ function draw() {
     g.drawString(dateStr, x, y);
     if (weekDay || calWeek) {
       var dowwumStr = require("locale").dow(date);
-      dowwumStr = "thursday";
       if (calWeek)
-        dowwumStr = (weekDay ? dowwumStr.substr(0,Math.min(dowwumStr.length,6)) + (dowwumStr.length>=6 ? "." : "") : "week ") + "#" + ISO8601calWeek(date);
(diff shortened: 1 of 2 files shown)


apps/antonclk/app.js
```diff
@@ -188,9 +188,8 @@ function draw() {
     g.drawString(dateStr, x, y);
     if (weekDay || calWeek) {
       var dowwumStr = require("locale").dow(date);
-      dowwumStr = "thursday";
       if (calWeek)
-        dowwumStr = (weekDay ? dowwumStr.substr(0,Math.min(dowwumStr.length,6)) + (dowwumStr.length>=6 ? "." : "") : "week ") + "#" + ISO8601calWeek(date);
+        dowwumStr = (weekDay ? dowwumStr.substr(0,Math.min(dowwumStr.length,6)) + (dowwumStr.length>=6 ? "." : "") : "week ") + "#" + ISO8601calWeek(date); //TODO: locale for "week"
```
BartS23, line 192: Maybe it is better that you use the default short weekday names require("locale").dow(date**, 1**) if you have no space.
```
var dowwumStr;
if  (calWeek && weekDay) {
  dowwumStr = require("locale").dow(date, 1) + " #" + ISO8601calWeek(date);
} else if (calWeek) {
  dowwumStr = "week #" + ISO8601calWeek(date);
} else {
  dowwumStr = require("locale").dow(date);
}
```

DDDanny:
Fixes #1259

apps/antonclk/app.js
DDDanny: Geat Idea!  I'll consider on next patch

gfwilliams:
Thanks - sorry - missed this!

PR #1264, opened 2022-01-11
Circlesclock: Show correct percentage values in circles & show humidity as weather circle progress

myxor:
Small bugfix release:

- Show correct percentage values in circles
- Show humidity as weather circle progress


gfwilliams:
thanks!

PR #1268, opened 2022-01-11
Devstopwatch app 0.04 - fixes #1157

### LINKED ISSUE #1157: Dev Stopwatch app does not work
@Ronin0000 Hey! Thanks for the Dev Stopwatch app! App really handy and helpful. But at the moment it doesn't work on Bangle.js v1 :(

And how about time counting down endlessly in minutes? For example, with a limit of 24 hours?

DDDanny:
* *bugfix* stopwatch broken with v0.03 setUI - fixes #1157
* realigned quick n dirty screen positions
* help adjusted to fit bangle1 & bangle2 screen-size with top/lower widgets
* fixed bangle2  foreground color for header: chrono and last lap highlight 3bit color (blue)
* added screenshot for bangle2 and a small README for both


gfwilliams:
looks great - thanks!

PR #1269, opened 2022-01-12
[LCARS clock] Adds humidity to datapoints.

peerdavid:
This PR mainly adds the feature to show humidity if selected by the user. The data is loaded from gadgetbridge. It also contains some other improvements and refactorings.


gfwilliams:
Thanks!

PR #1270, opened 2022-01-12
BanglExercise

myxor:
# BanglExercise

Can automatically track exercises while wearing the Bangle.js watch.

Currently only push ups and curls are supported.

## Disclaimer

This app is experimental but it seems to work quiet reliable for me.
It could be and is likely that the threshold values for detecting exercises do not work for everyone.
Therefore it would be great if we could improve this app together :-)

## Usage

Select the exercise type you want to practice and go for it!
Press stop to end your exercise.


gfwilliams:
Great! I guess star jumps should be pretty doable as well.

PR #1279, opened 2022-01-12
RAM efficient version of `fourTwentyTz.js`

thedod:
As suggested by @gfwilliams here:
https://github.com/espruino/BangleApps/pull/1152#issuecomment-1009739110

Also make `apps.json` entry more accurate, and play nice with dark theme.


gfwilliams:
Thanks! Looks perfect!

PR #1280, opened 2022-01-12
Select which GNSS system to use

benjarn:
Fewer GNSS systems could decrease time to fix.
This patch adds a checkbox for each choice.

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1536,7 +1536,7 @@
   {
     "id": "assistedgps",
     "name": "Assisted GPS Update (AGPS)",
-    "version": "0.02",
+    "version": "0.03",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - there's actually a GPS Setup app for Bangle.js 1 - and I think it might be a good idea to 'port' that (create a new app for Bangle.js 2, but still call it 'GPS Setup') that provides this functionality?

HilmarSt:
@benjarn 
Maybe you remove the possibility to disable "GPS", so that nobody is able to remove all systems. I'm not sure if anybody wants to use GNSS without GPS.
My proposal:
- GPS only
- GPS+BDS
- GPS+GLONASS
- GPS+BDS+GLONASS

BTW: Does any of the Bangle.js 1/2 GPS hardwares support GALILEO or some kind of SBAS?

PR #1281, opened 2022-01-12
Select which GNSS system to use

benjarn:
Fewer GNSS systems could decrease time to fix.
This patch adds a checkbox for each choice.


benjarn:
Someone might want e.g GPS+GLONASS, not only GPS

gfwilliams:
Thanks! As mentioned in the closed PR, a `GPS Setup` app would be good, but as being as we set the mode anyway in AGPS I guess it makes sense here

PR #1283, opened 2022-01-12
qmsched: theme switching

### LINKED ISSUE #1266: Quiet mode: schedule theme?
What do you think about scheduling the theme via quiet mode? 
I am thinking about switching to dark theme during the night times.

Pinging @rigrig  because he wrote the quiet mode schedule app/widget.

rigrig:
Fixes #1266
**Firmware:** tested with `2v11.23` (the Cutting Edge build):
- It uses `wrap` in the menu, which was added with `2v11`
- It uses `__FILE__` to determine whether to reload the current app, which I think was broken in earlier `2v11` versions?

Switches between the default "Dark BW" and "Light BW" themes, so custom theme settings will be lost.   
- If the current app is a clock: `load()` to apply the new theme   
- If the current app is this one: apply the changed theme directly
- Any other app: we wait for the user to switch apps, which will apply the new theme.

Also changes time selection to 5-minute intervals, as scrolling through minutes felt tedious.


gfwilliams:
Looks great - thanks! For checking if it's a clock, it is possible to use `Bangle.CLOCK` though. https://github.com/espruino/BangleApps/search?q=Bangle.CLOCK&type=code

All clocks *should* set that (usually via `Bangle.setUI`) so it's a fast, easy way to figure out if it's a clock (and won't fall foul of the 2v11 regression)

PR #1284, opened 2022-01-12
Coretemp - Move recording code into recorder plugin

halemmerich:
Code was in generic recorder app.


halemmerich:
Botched the rebase ;)

PR #1285, opened 2022-01-12
Recorder improvements

halemmerich:
* Allow recording barometer data
* Move over recording for CoreTemp to the coretemp app
* Always record HRM if data is available
* Record empty string for HRM if no data available instead of 0


halemmerich:
I have only moved the code from the recorder app and tested by emitting the events on the console. For deeper going changes i would rather have somebody with hardware to test do those.

PR #1286, opened 2022-01-12
BTHRM - Prevent mixed readings from both sensors

halemmerich:
* Prevent readings from internal sensor mixing into BT values
* Mark events with src property
* Show actual source of event in app


gfwilliams:
Great - thanks!

PR #1288, opened 2022-01-13
Anton clock 0.06 - fixes #1271

### LINKED ISSUE #1271: [Anton clock] week number is not displayed
The week number is not displayed even though configured as "on" because there is a difference between the settings and the app code : 
setting.js : uses settings.weekNum
app.js  : uses settings.calWeek

DDDanny:
& some code rework
+ moved settings-helper method def to load settings method
+ settingsfile as const
+ isBangle1 via HWVERSION==1
cal Week is buffered until date or timezone changes

fixes #1271

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4279,7 +4279,7 @@
   {
     "id": "antonclk",
     "name": "Anton Clock",
-    "version": "0.05",
+    "version": "0.06",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks!

PR #1290, opened 2022-01-13
qmsched: use Bangle.CLOCK instead of __FILE__

rigrig:
Thanks for pointing out `Bangle.CLOCK`, I know I've seen it before, but totally forgot about it.   
Also just checks for `global.setAppQuietMode` instead of `__FILE__`, so that should work with borked `__FILE__` as well.

I figured this wouldn't need a version bump, because it hasn't made it into the official store yet.


gfwilliams:
Thanks! Looks great - and yes, don't worry about the version bump

PR #1291, opened 2022-01-13
Add colour configuration to vector clock

StuntsPT:
Vector clock was too monochrome for my taste.
I decided to try and change the weekday font colour to yellow.
Well, one thing led to another, and I ended up configuring a possible different colour for each "section".
Then I also added a configuration file, and a settings menu to configure it.
It ended up looking good enough to "go public".
I hope I didn't bloat the app too much, as I'm not very experienced with JS, and is my first foray into programming on such a memory starved device.
Any feedback is more than welcome, so I guess you can consider this PR more like a "request for comments". =-)

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -4200,7 +4200,7 @@
   {
     "id": "vectorclock",
     "name": "Vector Clock",
-    "version": "0.03",
+    "version": "0.04",
     "description": "A digital clock that uses the built-in vector font.",
     "icon": "app.png",
     "type": "clock",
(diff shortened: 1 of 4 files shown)


gfwilliams:
Hi! Just checking but have you tried this on different themes? It seems that instead of `0xFFFF` you should probably use `g.theme.fg` as otherwise when running on a light theme (Bangle.js 2 default) by default everything will be white now (so invisible)

StuntsPT:
@gfwilliams Thank you for your feedback. I have changed the default colour to `g.theme.fg` as you recommended. In the same "spirit" I have also added he colour "black" at the end of the configuration spectrum (and also "magenta" since I was at it).

gfwilliams:
Thanks!

PR #1299, opened 2022-01-14
Add Crow Clock app

jameycribbs:
This PR adds a new clock app called Crow Clock, as simple clock based mostly on the code from Bold Clock.  It shows an image of MST3K's Crow T. Robot as its face.


DDDanny:
> mostly on the code from Bold Clock

1. Would'nt it make sense to extend Bold Clock to show a picture and then 
* dunno if it is possible to upload a clock like custom apps (so with the picture) - i know conversion is required....
* in this repro require the bold clock and just provide and set ur png
* or at elast use a 1:1 hardcopy with set ur png

2. Furthmore just be to be safe: is apps/crowclk/crow_clock.png clear of copyrights and MIT licensed? (beacuse we run under MIT license here)
If the user can select the image during upload app to clock or select a picture from clock storage then we get als rid of this topic, because the enduser would be responsible.

3. And one hint I got that I would like to share with you:
You can use ´´´javascript: g.dump();´´´ on the lift side ">" within webIDE and you receive a 1:1 screen dump - so no need to capture it manually -  it will have exactly 176x176:
dump

jameycribbs:
Thanks, Gordon!

Thanks for the feedback, DDDanny.  For your third point, I'm not sure I understand it (I'm pretty new to this stuff), but I will tinker with it and see if I can figure it out.  :)

PR #1303, opened 2022-01-15
wid_edit: Widget Editor

rigrig:
Widget Editor, so we don't need to edit code to move our widgets around.

Every feature seems like a good idea while happily coding away, but "disabling" widgets is tricky. I *think* it works fine with most widgets, but they are still loaded.   
And apps that try to call `WIDGETS[<myappid>].someCustomFunction()` probably break. 
Maybe better to strip this out, and if people want to get rid of a widget we adjust that app? (Maybe someday we can come up with a standard setting for this, so apps don't all need to build their own "show widget?" setting)

Also, maybe we should "officially" get rid of bottom widgets? [`Bangle.appRect`](https://github.com/espruino/Espruino/blob/35e46ef6a12ab3dd72e639439ba5f1d409253958/libs/banglejs/jswrap_bangle.c#L4884-L4916) doesn't looks like it supports it for one...   
The only bottom widget is [`wdclkbttm`](https://github.com/espruino/BangleApps/tree/master/apps/widclkbttm), I haven't tried that particular one, but it seems Things (like menus) Break with bottom widgets.

With all the "disabled"/bottom row breakage, I figured this should be a separate app.
But if we strip out disabling, and only allow choosing between "top left"/"top right", maybe this could go directly into the settings app?


gfwilliams:
Thanks! So... 

> Also, maybe we should "officially" get rid of bottom widgets?

I think so, yes... I mean with `appRect` we could kind of support them in apps (and I should add that to the firmware), but I'm not sure anyone really seems to care about them.

> but "disabling" widgets is tricky.

I think for disabling we should just ask the user to uninstall. But maybe you could switch to hiding? This seems quite successful and is already used a few places: https://github.com/espruino/BangleApps/blob/9b4d2e140b8876428fb6fff4c3a5d1b060ec0dd3/apps/widviz/widget.js#L10-L11

rigrig:
> I think for disabling we should just ask the user to uninstall. But maybe you could switch to hiding?

It tries to hide them now, and could probably try a bit harder (and properly label it "hide" instead of "disable"), but "just ask the user to uninstall" seems like the better way. Also because otherwise you end up with a whole bunch of widget code loaded for no good reason, and it would mean much cleaner code for this app.
Same goes for only supporting top-row widgets: that way the boot code doesn't need to mess with `appRect`.

I'll update it to just support `tl`/`tr` and `sortorder`.

rigrig:
Updated to just support tl/tr and sortorder.
Also removed "bottom bar" references from the general README.

gfwilliams:
Thanks!

PR #1304, opened 2022-01-15
Option to keep or not the textual translations

diesphink:
As most apps aren't translated, I found odd the mixed language on apps, so I thought it would be useful to have a locale with formatting but no translation. this way the apps language are all in english, but the format for numbers, dates, etc are in my locale.

As forum post: http://forum.espruino.com/conversations/371915/#comment16353541

### DIFF
--- a/apps/locale/locale.html
+++ b/apps/locale/locale.html
@@ -11,7 +11,7 @@
     </select>
     </div>
     <div class="form-group">
-      <input id="translations" type="checkbox" checked /> <label for="translations">Mark this option if you want translations for commont text like "Yes", "No", "On", "Off".</label>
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Just replied in the forum post above about this. I totally agree with it, *but* longer term I'll be removing them anyway so I'm not sure if it'll be confusing for folks to put the setting in and then just take it out later

gfwilliams:
Just a note that translations are now working in https://github.com/espruino/BangleApps/issues/1311

So I think once everyone seems ok with that, the next step will be to pull the translations out of `locale` completely.

gfwilliams:
Translations are in - but actually I'll pull this in now, and change the default to false

PR #1306, opened 2022-01-15
Fix a couple of issues with TouchMenu

kyleplo:
- Menu event listeners are now properly cleared when going directly from the menu to the launcher
- Menu now fills all available screen space, respecting widgets if there are any (using `Bangle.appRect`)


gfwilliams:
Neat - thanks!

PR #1307, opened 2022-01-15
Unify launcher for bangle.js 2 and 1

### LINKED ISSUE #1177: Why does the launcher have a different version for Bangle 1?
The default launcher has a version for Bangle 1, and Bangle 2. The version for Bangle 2 seems to be more maintained, and has config options.  

However, when I edit the app to use the Bangle 2 version, it works just fine for the Bangle 1. Would it be possible to remove the Bangle 1 version and always use the Bangle 2 version?

KTibow:
## Changes
- Delete separate launcher code for bangle.js 1.
- Slightly modify bangle.js 2 code to work on both devices.
## Issues affected by this
Fixes #1177
(This means bangle.js 1 users will have a way to change the size of the text, and there's less code to maintain.)

- [x] Tested on Bangle.js 1
- [x] Tested on Bangle.js 2 (by @adamschmalhofer)


adamschmalhofer:
Tested on Bangle.js 2 and works well.

PR #1309, opened 2022-01-17
Run: set format of time to h:mm:ss, pace to mm:ss, settings to disable GPS/HRM

hughbarney:


gfwilliams:
Great - thanks! I guess at some point mentioning settings in the README might be handy too though

PR #1310, opened 2022-01-18
1. GPS Info App: Show satellites "in view" separated by GNS-system; 2. Spirit Level App: Now also visible on Bangle.js 2

HilmarSt:
Does anybody ever seen a GLONAS satellite?

### DIFF
--- a/apps.json
+++ b/apps.json
@@ -1541,7 +1541,7 @@
   {
     "id": "gpsinfo",
     "name": "GPS Info",
-    "version": "0.09",
+    "version": "0.10",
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks! With the spirit level I guess the colour issue is that it doesn't support themes?

If it used `g.theme.fg` and not `1,1,1` as the colour I bet that'd fix it properly without HW-specific defines. I'll do that now though

HilmarSt:
Sure you're right with the themes, but it's not running in the emulator and I don't have a Bangle.js 1

PR #1312, opened 2022-01-18
Let the parrot party on Bangle.js 2 too!

HilmarSt:


gfwilliams:
Is that really all it needed?

gfwilliams:
Just made a few small tweaks to it too

PR #1320, opened 2022-01-19
provide metadata for "configurable_clock"

rozek:


rozek:
Thanks - that was fast!

PR #1321, opened 2022-01-19
Let git ignore apps.json

myxor:
Let git ignore apps.json so that nobody tries to commit it


gfwilliams:
Good plan - thanks!

adamschmalhofer:
as apps.json is already in git adding it to .gitignore won't do anything.

PR #1323, opened 2022-01-19
Jekyll apps.json changes

adamschmalhofer:
- ported bin/firemwaremaker*.js to metadata.json
- Offer an easy way to use a restricted apps.json


gfwilliams:
Thanks!

PR #1324, opened 2022-01-19
Support for localhost appstore

rigrig:
Depends on https://github.com/espruino/EspruinoAppLoaderCore/pull/10

- Adds a script to generate `apps.local.json` from all metadata
- Makes the loader use `apps.local.json` when serving from localhost
- Adds npm `local` script to watch for changed metadata, while serving from localhost


rigrig:
I got a merge conflict for `.gitignore`, and decided to undo https://github.com/espruino/BangleApps/commit/a83e38425acfffb3e83fbac0aac38d98520e9441, because that won't work: git doesn't ignore files it already know about.

gfwilliams:
Hi - does the Node.js code do anything that `https://github.com/espruino/BangleApps/blob/master/bin/create_apps_json.sh` didn't?

I like the `npm start` idea but it just seems a shame to have more code duplication

rigrig:
Well, it errors out for invalid JSON and results in a consistently sorted+formatted file. But mostly I created this, then rebased onto master without noticing `create_apps_json.sh` was added already.   
I'll get rid of the js code and adjust `create_apps_json.sh` to take an (optional) filename argument instead. I agree that we don't want code duplication, and being able to generate `apps.json` without needing `npm` is nicer.

rigrig:
Updated to use `create_apps_json.sh`.

- Added optional filename argument
- Changed to not add `null` at the end, at least `language_scan.js` chocked on that.
- I found a way to make git ignore modified `apps.json`: [`git update-index --skip-worktree`](https://compiledsuccessfully.dev/git-skip-worktree), and modified the script to do so if you call it without an explicit filename.

gfwilliams:
Looks great - thanks for changing that!

PR #1327, opened 2022-01-20
Days left App now also works on Bangle.js 2

HilmarSt:


gfwilliams:
Great - thanks for this! Do you think the `predraw` is actually needed?

I saw this pattern used on another app with the same bug, and `predraw` was never called on old firmwares which is why it worked and then broke later (when `predraw` got added). So in the end I could just remove the whole thing

HilmarSt:
> Do you think the `predraw` is actually needed?

I'm new to both, Bangle.js 2 AND Javascript and have no idea - but it still works if removed ;-)

@gfwilliams 
I've seen that the day counter widget was not updated when I unlocked, only when changing into the settings menu. What would be the best way to update the counter, at least when unlocking?

PR #1329, opened 2022-01-20
File Manager For Bangle.js v2

Ronin0000:
Made it so you could get file manager for bangle.js 2.

### DIFF
--- a/apps/fileman/ChangeLog
+++ b/apps/fileman/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02: Improve handling of large amounts of files (fix #579)
 0.03: Update RegExp use (Was using backreference instead of character code)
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi - I see a lot of extra commits and non-filemanager specific changes here - is this actually ready to merge?

In the apps that don't have ChangeLogs, adding one is great, but please could you set the ChangeLog version to the app version, rather than resetting the app version to `0.01`? Otherwise folks that already have the app installed will be really confused when they find they have 0.03 installed but the latest is 0.01!

T0TProduction:
Correct me if I'm wrong, but I'm pretty sure activating emulator does not need a new app version since it's just something that changes how the app is displayed in the app loader.

Ronin0000:
I am going to close this. Sorry this was not a good idea.

PR #1330, opened 2022-01-21
Pastel: remove dependancy on widpedom

hughbarney:


gfwilliams:
Nice - thanks! The Pebble one could do this too I guess?

gfwilliams:
Only thing to note is that all the firmwares of Bangle.js 2 will execute that `getHealth` line without errors, just the pre-2v11 ones will give it for that 10 minute window, and 2v11 itself won't ever reset it :(

Soon there'll be 2v12 though and that'll fix it

hughbarney:
> Nice - thanks! The Pebble one could do this too I guess?

Yes it could.  Was dipping my toe in the water first, to see how this went.

PR #1331, opened 2022-01-21
Add Timer Clock app

untbu:
It's a simple clock based on the Anton Clock. It includes apps for stopwatches, timers and alarms. Two stopwatches or timers are displayed above the time.
On Bangle.js 2 the times are modifiable by swiping over the digits, on Bangle.js 1 right/left switches between the digits, up/down changes the value and the middle button switches between controlling the buttons and changing the times.
The apps can be started by swiping over the time/date or using the Launcher entries. The launcher entries are sorted to the back by default because they aren't really needed.
The font size and type is configurable for clock and apps.


gfwilliams:
Wow, this looks great.

Just thinking aloud here - but does it make sense to try and make these independent apps/widgets, so you can use some of the work you've done here with other watch faces?

untbu:
I moved some functions used by all parts to a library file. I will move them back. Is there an option to automatically install the other apps if the clock app is installed?

untbu:
Just overlooked that you already merged it. And maybe this way is better because of storage usage and the effort to change controls or something like this.

gfwilliams:
> Is there an option to automatically install the other apps if the clock app is installed?

Yes, there's a `dependencies` entry: https://github.com/espruino/BangleApps/blob/master/README.md#metadatajson-format

But for now this is fine, it's just especially if the apps are copied from existing apps (or if in future others copy your apps) it ends up being a bit of a nightmare if a bug is found and then it needs fixing in loads of places.

PR #1333, opened 2022-01-21
New Clock for BJS 2 - Not Analog

peerdavid:
An analog watch face for people (like me) that can not read analog watch faces, but like the design of analog watch faces. It looks like an analog clock, but its not! It shows the time digital - check the 4 numbers on the watch face :)

The white hand shows always the battery level, the colored hand data depending on the state: the steps or the minutes left for a timer that can be set within the watch face. Other data that is shown: Lock status, Charging state, Gps On.

screenshot_1

screenshot_3


gfwilliams:
Thanks!

PR #1334, opened 2022-01-21
Just File Manager For Bangle.js 2

Ronin0000:
Hi, Sorry about the previous pull request I realized that was not a good idea. This is just a pull request for making File Manager compatible with Bangle.js 2.


gfwilliams:
Thanks!

PR #1336, opened 2022-01-22
Add Light Switch Widget

storm64:
Add a new widget. Description:
A fast way to switch LCD backlight on/off, change the brightness and show the lock status. All in one widget.


gfwilliams:
Wow, nice - thanks!

Just to say I think unless there's some reason I'm missing, I think it's going to end up a lot faster and more efficient to just define all the images you need in `drawIcon` (as you have already done for some) rather than having them in the images.json file...

storm64:
The Idea behind a separate file was to make the variety of images simply customizable through a _custom.html_-menu in the app launcher and maybe add an option to upload a personal png-file completely without coding.

gfwilliams:
Ahh, right - thanks!

PR #1340, opened 2022-01-24
[NotAnalog clock] Improved readability of steps

peerdavid:
360° are now 12k steps such that its easier to read the current steps.


gfwilliams:
Thanks!

PR #1342, opened 2022-01-24
Add gpsrawtime

BartS23:
I am not sure if it is too much code for the boot.js.
Should I put it in a separate file?

Could someone review the code?

See also http://forum.espruino.com/conversations/372114/?offset=50#comment16367783


gfwilliams:
Hi - thanks for this, but I'm not sure all that code is really needed, and adding 'GPS-raw' could cause FIFO_FULL errors... 

Looks like you're decoding the `RMC` nmea data to get the GPS time? But Espruino already does this and puts it in the `time` field of the GPS fix: https://github.com/espruino/Espruino/blob/master/libs/misc/nmea.c#L53-L61

So unless I'm missing something, all you need to do is `Bangle.on('GPS', fix => print(fix.time))` and you'll be able to grab the latest time? Also RMC is on by default, so there's no need to send the `PCAS03`

There is actually already a widget which does this as far as I can see: https://github.com/espruino/BangleApps/blob/master/apps/gpsautotime/widget.js

But right now it's not enabled for Bangle.js 2 (although it should work). Do you think it's worth just enabling that? Although I know it doesn't have logging.

BartS23:
My intention was to get the time without a fix.

The app enables gps, wait for the for the first RMC message and disable gps.
If within 10 seconds no RMC message is received gps will also disabled.

I'll remove PCAS03.

gfwilliams:
> My intention was to get the time without a fix.

Yes, that should be fine to do with the `GPS` event? The GPS event fires whenever there's data from the a GPS - it doesn't wait until there's a fix (or even a time - so the first few events `time` may be `undefined`).

Using `GPS-raw` every time any app starts will really slow things down - I don't see why you need this when the data is already being parsed internally by Bangle.js

BartS23:
Yes, you are right. I didn't see it, sorry.

I will take a look at gpsautotime. 
Right now it's waiting for a fix, but that should be able to be changed.

gfwilliams:
Thanks!

> Right now it's waiting for a fix

I didn't spot that! Wow, that really limits its use - but yeah, that should be an easy tweak

PR #1343, opened 2022-01-24
sleepphasealarm: Compatibility to Bangle.js 2 and new firmware

storm64:
Make app compatible with Bangle.js 2:
- check hardware
- _drawString(s, x, y)_: removed _x_, always centered
- _drawString(s, y)_: _g.reset()_ before _g.clearRect(...)_
- adjust heights for Bangle.js 2 for each _drawString(s, y)_ call
- _drawApp()_: on Bangle.js 2's always on display save energy by drawing time only at every next minute
- _drawTime()_: hide seconds on Bangle.js 2 (only refreshed every minute)
- _drawApp()_: add _"Alarm at "_ to distinct from actual time (no difference in seconds)
- execute _load()_ on Bangle.js 2's BTN

Make app compatible with _E.showMessage(...)_ on new firmware:
- adjust actual time height to be drawn above the alarm time from _E.showMessage(...)_

### DIFF
--- a/apps/sleepphasealarm/ChangeLog
+++ b/apps/sleepphasealarm/ChangeLog
@@ -1,3 +1,3 @@
 0.01: New App!
 0.02: Respect Quiet Mode
-0.03: Add compatibility for Bangle.js 2, added "Alarm at " for the alarm time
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great - thanks!

PR #1345, opened 2022-01-25
Simplest: renamed some files

hughbarney:


gfwilliams:
Just to say generally I advise everyone just to stick with `app.js`/etc so this is a move away from that.

IMO it just makes stuff tidier - if I need to look at or edit a file it's nice and easy just to type `apps/theapp/app.js` rather than the duplication of `apps/theapp/theapp.app.js`

PR #1346, opened 2022-01-26
Improved proximity identification settings specific to Bangle.js

jeffyactive:
Twice as responsive and far more sensitive at slight expense to battery life.


gfwilliams:
Great - thanks!

PR #1350, opened 2022-01-26
Add Waypointer Moto

jes:
This is a fork of the "Way Pointer" app which makes it suitable for motorcycle navigation. The most significant changes are:

1.) Use GPS course instead of compass bearing, as this is more accurate (the original app doesn't use it because the GPS course is either unavailable or inaccurate at walking speeds).
2.) UI changes: hide the compass bearing as it's not relevant; bigger writing; higher-resolution arrow.
3.) Use the Espruino menu system instead of the custom waypoint selector.
4.) Add OpenStreetMap integration to the waypoint editor on the BangleApps loader.
5.) Add an STL for a 3d-printable "artificial wrist" that mounts on the handlebars so that the watch can be mounted on the bike.

Arguably the OpenStreetMap integration should also be upstreamed to the other apps that use `waypoints.json`, but I'll leave it at this for now.

Please let me know if there is anything I should change.


gfwilliams:
Wow, this looks awesome - thanks!

Yes, I guess longer term a generic waypoint editor would be really neat. Maybe we could have a `waypoints` app that had the editor as part of it, then other apps could depend on that app?

PR #1351, opened 2022-01-26
Make lapcounter behave better on `Bangle.js 1`

thedod:
* 1 bit icon on opening screen (appears in fg color)
* Say "button 1" instead of "button"


gfwilliams:
Thanks!

PR #1353, opened 2022-01-26
bthrm - Better stability and auto reconnect

halemmerich:
* Automatically reconnect BT sensor
* App buzzes if no BTHRM events for more than 3 seconds


gfwilliams:
Great - thanks!

PR #1354, opened 2022-01-26
health - Adds 3 minute setting

halemmerich:
Allow to take HRM every 3 minutes while waiting 1 minute on data from HRM


gfwilliams:
Thanks - but how does the 3 minute option work? The only change I see is that it will keep searching for HRM for less time, not that it'll try every 3 minutes?

gfwilliams:
Looks good - thanks!

PR #1355, opened 2022-01-26
Adds app for recording HRM and accelerometer events

halemmerich:
This is an experiment for creating data on HRM vs Accelerometer. Since it is only temporarily useful (while fine tuning the HRM algorithm), maybe merging into the main repo is not relevant?


gfwilliams:
Thanks! This looks like a good thing to merge in to me... Even in the future the more data that can be gathered for testing the better.

Only things I'd say are:

* I'd remove the `eval(...boot0...)` bit - this happens automatically even when you do `reset()` (you can check what's in RAM by running `dump()`) so you don't want to load it twice as that could cause all kinds of problems.
* It seems to fail the self-test at the moment because the `metadata.json` name is long and there's no 'shortname' for the  launcher - I can fix that though.

halemmerich:
Travis wants an entrypoint, what does that mean?

myxor:
@halemmerich 
This comes from https://github.com/espruino/BangleApps/blob/a0ca017af78679d42de298d9e43022adcc61fc63/bin/sanitycheck.js#L266
Looks like something is wrong with the `storage` object. Maybe try to change it into `storage:[]` as assistedgpshas it: https://github.com/espruino/BangleApps/blob/a0ca017af78679d42de298d9e43022adcc61fc63/apps/assistedgps/metadata.json#L12
You have a space in it.

halemmerich:
It just needed an app.type != "app" to prevent the mentioned line to be expecting an entrypoint (*.app.js file in storage array).

gfwilliams:
Looks great - type=ram looks perfect - thanks!

PR #1358, opened 2022-01-27
Avoid erasing bottom widgets

thedod:
Turns out bottom widgets (specifically, digital clock widget) were disappearing during maze generation. Still not clear _why_, but calling `Bangle.drawWidgets()` from inside the `Maze` object constructor solves this.

Also: say "Btn1" in order to be less confusing on `bangle.js 1` (before I ask people to test this app on that platform).

### DIFF
--- a/apps/acmaze/ChangeLog
+++ b/apps/acmaze/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New App!
 0.02: Faster maze generation
+0.03: Avoid clearing bottom widgets
--- a/apps/acmaze/app.js
+++ b/apps/acmaze/app.js
@@ -11,13 +11,10 @@ function Maze(n) {
   this.margin = Math.floor((g.getHeight()-this.total_length)/2);
   this.ball_x = 0;
   this.ball_y = 0;
-  this.clearScreen = function() {
-    g.clearRect(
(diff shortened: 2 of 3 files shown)


gfwilliams:
I'll merge, but I really don't think this is a good idea - you're making redraws much slower by redrawing *all* widgets every time.

If you need to clear the screen it's better with just a `g.clearRect(Bangle.appRect)` which'll give you the area that's available (having said that, bottom widgets are kind of deprecated and aren't well supported right now). I'll make sure appRect updates for them in the 2v12  release.

thedod:
> you're making redraws much slower by redrawing _all_ widgets every time.
There's no "every time". This happens _once_ when we start generating the maze.
I'm really puzzled by what clears bottom widgets. I've removed any calls to `clearRect()` I could (in fact, I'm surprised that even without the call I used to make on line 20, the maze area seems to get cleared by _something_).
I was suspecting that the calls to `E.showMenu()` (without parameters) were doing this, but I've done some tests on the ide, and `E.showMenu()` doesn't seem to remove the bottom digital clock widget. I wish I knew what _was_ causing this, but at least the "voodoo" solution isn't too inefficient IMHO.

PR #1359, opened 2022-01-27
[NotAnalog clock] Performance improvements

peerdavid:
This PR includes mainly performance improvements which should also extend the battery life if the analog clock is used.


gfwilliams:
Thanks!

PR #1360, opened 2022-01-27
Add "Chime the Time" to Vector Clock

HilmarSt:
Chimes the time with a up/down swipe

### DIFF
--- a/apps/vectorclock/ChangeLog
+++ b/apps/vectorclock/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Use Bangle.setUI for button/launcher handling
 0.03: Bangle.js 2 support
 0.04: Adds costumizable colours and the respective settings menu
+0.05: "Chime the time" (buzz or beep) with up/down swipe added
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks! However does this provide any different functionality from just installing the "Hour Chime" widget?

I feel like this kind of thing is much better implemented as a widget/boot code so it can apply to all clock faces, and as there's one already maybe it's better to just improve that?

HilmarSt:
> does this provide any different functionality from just installing the "Hour Chime" widget?

Yes, it will vibrate the hour and minute count of the current ime - I've taken this functionality from the "Vibrate Clock" App. That means you "feel" the time without having to look at the display.

> is much better implemented as a widget/boot code

I've also thought about it, but I fear interferences with other clock faces or widgets because of the general "reservation" of swipe up/down.
I'd like to add the functionality to the great "Light Switch Widget" but I'm not yet experienced enough to risk this.

Maybe you accept this PR as a beta test.

gfwilliams:
Ok, thanks - I'll merge this then.

But I still think the widget/boot code is preferable longer term. Maybe it will conflict with one or two clocks but in that case the user can just uninstall the widget - but the benefit is for the vast majority of clocks you can add the functionality if you want to.

Issue #1363, opened 2022-01-28
Rename 'About' tab to 'More...'


gfwilliams:
http://forum.espruino.com/conversations/372207/#comment16372415

Ronin0000:
Should this Issue be closed?

gfwilliams:
Thanks - yes :)

PR #1364, opened 2022-01-28
Rename 'About' to 'More...'

jes:
#1363


gfwilliams:
Great - thanks! Just have to change all the docs now!

jes:
Ah, hadn't thought of that! I've had a first pass in the mentioned PR on EspruinoDocs, it's possible there may be more that I'm missing.

PR #1368, opened 2022-01-28
fixed Font-Name in Morse App

Benjamin-6848:
fixed Font-Name in Morse App


gfwilliams:
Thanks!

Benjamin-6848:
Thank you for merging so fast! 

I did a similiar thing for MatrixClock - fixing some Color-Issues and the Screen-Update-Order of MatrixClock.

PR #1369, opened 2022-01-30
`acmaze` and `ftclock` are `bangle.js 1` compat

thedod:
Both apps were tested on a `bangle.js 1` by forum user Mi: https://forum.espruino.com/comments/16376253/ (thanks for testing).


gfwilliams:
Great - thanks!

PR #1370, opened 2022-01-30
New Ruuvi Watch version, redesigned and refactored

emarc:
A bit of re-design and refactor before making things configurable.
Removed flicker. Added image to docs and fixed formatting.


gfwilliams:
Thanks!

PR #1372, opened 2022-01-30
use the correct key to check for translations

alessandrococco:
I'm working on the Italian translation and I noticed that `language_scan.js` prints *Missing translation* for existing translations.

This commit fix the issue.


gfwilliams:
Great - thanks!

PR #1374, opened 2022-01-31
ContourClock update 

xxDUxx:
Version 0.2
-settings page
-new fonts
-inverse digits


gfwilliams:
Wow, that's great! In future please can we keep the version number in the `#x##` format? All the other apps do so it's nice to keep things kind of the same.

PR #1375, opened 2022-01-31
MatrixClock: Fixed Color-Issues and Screen-Update-Order

Benjamin-6848:
I fixed 2 Problems in MatrixClock: 
1. When trying to clear the old Text, so it could get updated, previously it got cleared by using "g.theme.fg" as color which is white. But it should be "g.theme.bg" which is black. 
2. Previously the Date got "deleted" by the falling Shards, this problem should be fixed.


gfwilliams:
Excellent - thanks! When you push updates, please could you also add a note in the app's changelog and bump the version? I'll do it this time though

Benjamin-6848:
> Excellent - thanks! When you push updates, please could you also add a note in the app's changelog and bump the version? I'll do it this time though

Sorry, i am currently getting started with GitHub - i never used it before. 
I will watch out in future for things like that. 
Thank you for your advice and your help.

PR #1376, opened 2022-01-31
Circlesclock v0.09

myxor:
# Changes

* Support to show temperature, air pressure or altitude from internal pressure sensor
* Fix sunprogress calculation during night
*  Refactor settings menu
* The color of each circle can be configured. The following colors are available:
  * Basic colors (red, green, blue, yellow, magenta, cyan, black, white)
  * Color depending on value (green -> red, red -> green)
 * Good HRM value will not be overwritten so fast anymore

Changes were beta tested by @halemmerich and some idea are from him. Thx for that :)

### DIFF
--- a/apps/circlesclock/ChangeLog
+++ b/apps/circlesclock/ChangeLog
@@ -17,3 +17,4 @@
       Fix sunprogress calculation during night
       Refactor settings menu
       Colors of circles can be configured
+      Color depending on value (green -> red, red -> green) option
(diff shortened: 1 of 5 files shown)


gfwilliams:
Great! You think this is ready for merging?

myxor:
It is ready now :)

gfwilliams:
Thanks!

Issue #1311, opened 2022-01-18
App Translations / Internationalisation


gfwilliams:
This is a replacement for https://github.com/espruino/BangleApps/issues/136 - without the discussion about how to implement it...

Historically, we used the `locale` module for translations (http://www.espruino.com/Bangle.js+Locale) but this only provided extremely basic translation (yes/no/etc). While more translations could have been added, these would use up memory and would make apps run slowly.

**We are now moving towards translating apps automatically in the app loader during the upload process**

It works pretty simply:

* In the `app.js` file, write any text in English and put `/*LANG*/` in front of any text string that should be translated
* Upon upload, the app is scanned for these strings, and they are replaced if they are found in the translation file in the `lang` folder, for instance https://github.com/espruino/BangleApps/blob/master/lang/de_DE.json

Current status:

* It works. 'Translations' is added to the settings in the About page of the App Loader
* About the only app with `/*LANG*/` entries in at the moment is Settings - **we need more apps with the LANG tag added**
* Translations for different languages are very lacking - run `bin/language_scan.js` to get an idea. **we need more translations!** Maybe we want to try doing translations automatically so we have some placeholders?
* Translations are not 'smart'. While they are aware of punctuation at the beginning and end they're not aware of capitalisation - perhaps if this worked better we could get away with less translations required.

Any help that can be provided is hugely welcome! This has been requested a lot so I've done what I can, but my language of non-english languages is seriously lacking.

HilmarSt:
sorry for this newbie question, but how do I "run `bin/language_scan.js`" on a windows PC?

gfwilliams:
You'd have to install node.js (https://nodejs.org/en/) but once you do that, I believe you get a shell window you can open. You can go to the directory and run `node bin/language_scan.js`

alessandrococco:
Hi @gfwilliams, I have a doubt about the key of a translation.

`language_scan.js` returns `This will remove everything` but the phrase is `This will remove everything!` (as in `E.showPrompt(/*LANG*/'This will remove everything!', ...`) What should I use as a key? `This will remove everything` or `This will remove everything!`? (I think the latter)

gfwilliams:
I think you need to use `This will remove everything` since the actual translation uses the keys without any punctuation.

At some point it could probably be improved but right now that's how it works: 

[...]

PR #1377, opened 2022-01-31
Italian translations

alessandrococco:


gfwilliams:
This looks awesome - thanks! I noticed:

```
    "Delete all messages": "Cancella tutti i messaggi",
    "Delete All messages": "Cancella tutti i messaggi",
    "Delete All Messages": "Cancella tutti i messaggi",
```

Was this for testing, or does it actually seem to be required? I wonder whether it's worth me changing the translation code so it does something like check for the exact match (as now) and failing that use one that matches, just with different capitalisation

alessandrococco:
Actually it is the output of the language_scan tool. E.g: 

    /*LANG*/"Delete all messages" : () => {
      E.showPrompt(/*LANG*/"Are you sure?", {title:/*LANG*/"Delete All Messages"}).then(isYes => {

(Probably `Delete All messages` is a test,  I can no longer find the source)

---

Maybe this is a silly question: is there a simple way to test the translations on my bangle?

gfwilliams:
> Maybe this is a silly question: is there a simple way to test the translations on my bangle?

Yes, absolutely! If you've got the changes in your `master` branch (or I think you can set up GitHub pages to serve off a different branch) you can go to https://alessandrococco.github.io/BangleApps/ then 'more...' then 'Translations' and choose italian.

Any apps you upload from then-on that have the /*LANG*/ comments in them (Settings is the obvious one) will have your translations

alessandrococco:
Damn, Italian is so verbose..!

"Wake on FaceUp" &rarr; "Attiva se girato a faccia in su" / "Risveglia se girato a faccia in su" 😭 

Anyway, what's the difference between *Wake on FaceUp* and *Wake on Twist*?  They look the same to me 🤔

Rarder44:
ciao, 
( vi prego per dare un help nelle traduzioni non fatemi scrivere in inglese 😭 ) 

di solito ho visto questo tipo di "notazione"
"Wake on FaceUp" → "Attiva a faccia in su"

credo che la differenza tra "FaceUp" e "Twisted" sia 
"FaceUp"→ "faccia in su"
"Twisted"→ "scosso, mosso, fare il movimento rapido del polso che muova l'orologio di un certo gradiente"
( ok, forse l'ultima è un po' lunga ma la differenza credo che sia quella )

alessandrococco:
> ciao, ( vi prego, non fatemi scrivere in inglese anche per dare un aiuto nelle traduzioni 😭 ) (please do not let me write in English also to help with translations 😭)
> 
> di solito ho visto questo tipo di "notazione" "Wake on FaceUp" → "Attiva a faccia in su"
> 
> credo che la differenza tra "FaceUp" e "Twisted" sia "FaceUp"→ "faccia in su" "Twisted"→ "scosso, mosso, fare il movimento rapido del polso che muova l'orologio di un certo gradiente" ( ok, forse l'ultima è un po' lunga ma la differenza credo che sia quella )

Ciao! The problem is that the length of a line is limited. "Wake on FaceUp" is already too long:

image

"Wake on Twist" should be "Attiva ruotando il polso" but it is too long.  Maybe "Attiva col polso" or "Att. con polso"? And "Wake on FaceUp" &rarr; "Att. faccia su"

---

@gfwilliams  To save space would it make sense to replace On/Off with Y/N (Italian: S/N)? E.g: `Wake on Twist Y`

Rarder44:
the solution could be scrolling text....
I think "Att." may not be understood....

gfwilliams:
> To save space would it make sense to replace On/Off with Y/N

Yes, however I have plans to make some big changes to `showMenu` which would allow it to have some kind of checkbox/radio button.

While I'm not sure if there'd be scrolling text, menu items will be higher (to make them more finger-sized) and so should be able to display 2 lines of text then.

... so I wouldn't worry for now. Long titles is a problem with the standard English translations too and it's something I do plan to sort out pretty soon

alessandrococco:
> While I'm not sure if there'd be scrolling text, menu items will be higher (to make them more finger-sized) and so should be able to display 2 lines of text then.
> 
> ... so I wouldn't worry for now. Long titles is a problem with the standard English translations too and it's something I do plan to sort out pretty soon

Great, good to know!

---

**Can an app name be translated?** E.g. In the launcher an Italian user should find "Impostazioni" instead of "Settings".

Not all app names should be translated: Nifty-B Clock should remain Nifty-B Clock but all generic names should/could be translated (eg.: Compass &rarr; Bussola, Weather &rarr; Meteo, Calculator &rarr; Calcolatrice, etc.)

Maybe we could add a new field in app's metadata.js, e.g. "shouldTranslateAppName": true/false. If true, search for translated name (default to original name if not found). Or we could add the translated names inside the metadata.js:

```
{
  "id": "setting",
  "name": "Settings",
  "translatedName": {
     "it_IT": "Impostazioni",
     "de_DE": "Einstellungen",
     ...
   }
   ...
}
```

What do you think @gfwilliams?

gfwilliams:
Good idea! I guess all we really need is for the app loader to run the name of the app through the translation process it uses for other things?

Something like `Nifty-B Clock` wouldn't be translated anyway unless there was a specific entry in the translations for it, so I think we'd be safe. One option is to have a `launcher` entry in the translation JSON so you could have a separate translation (The `Run` app may have a different meaning from `Run` that is mentioned in other apps)

alessandrococco:
In my opinion a `launcher` entry in the translations file is the better solution 👍

gfwilliams:
Do you see any reason why I shouldn't merge this now? Seems like what you have is a huge improvement over normal

alessandrococco:
> Do you see any reason why I shouldn't merge this now? Seems like what you have is a huge improvement over normal

Two reasons:

1. The length of some of the translations
2. The `\n` problem I reported in #1311 (partially fixed now?)

I'll review again the translations because there are few things I want to fix but if _1._ is not a real problem tonight I'll mark the PR as ready ;-)

gfwilliams:
Thanks! So yeah, I think the "\n" should be fine now.

In terms of translation length, I'd be tempted to leave them long for now and hopefully a firmware update soonish will sort out the menus so it's not an issue.

alessandrococco:
Done!

I also commited a small improvement in the language_scan tool.

alessandrococco:
Is there any Italian who wants to proofread the translations before merging? :D

PR #1379, opened 2022-02-01
android/gbridge: Include charging state in battery updates to phone

rigrig:
This seems like a useful thing to include for GadgetBridge: not only do you get the nice lightning bolt icon, but it also means the app knows to dismiss the "low battery" warning when the watch is connected to a charger.

Before you merge:
- `charging`:`true/false` seemed like the cleanest option, but I have no idea how frugal we need to be with bluetooth-bytes, does it really matter? Because then I guess `pow`:`1/0` (or `chrg` or even `c`...) would be better.   
  Asking because this will also need some changes in GadgetBridge and EspruinoDocs, so I'd like to get it right instead of amending 3 PRs later on ;-)
- I noticed we could also include `volt` (from `NRF.getBattery()`?), not sure if that has any added value over `bat`, but I'd be happy to work it in while I'm messing with this code anyway.

### DIFF
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -5,3 +5,4 @@
 0.04: Android icon now goes to settings page with 'find phone'
 0.05: Fix handling of message actions
 0.06: Option to keep messages after a disconnect (default false) (fix #1186)
+0.07: Include charging state in battery updates to phone
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -50,8 +50,9 @@
   };
 
   // Battery monitor
-  function sendBattery() { gbSend({ t: "status", bat: E.getBattery() }); }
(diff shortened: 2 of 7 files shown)


gfwilliams:
Thanks! While size doesn't matter *so* much, ideally we want the total under 50 bytes so it fits in one packet. All the other events tend to use quite short field names, so I'd say maybe do the same. Maybe make it `chg` instead of `charging` and make it an integer (so just 0/1 rather than true/false)?

Also I really wouldn't bother with the voltage. I have no idea what we'd do with it in Gadgetbridge and its meaning would vary depending on whether you had a Bangle.js 1 or 2

rigrig:
Thanks, `chg`:`1/0` does look more like the other fields, I'll use that.

> I really wouldn't bother with the voltage

Ok. (The unused code will bug me a bit, but I'll just pretend that "it's optional" makes everything OK ;-)

rigrig:
Updated: now sends `chg`:`0/1`

gfwilliams:
Thanks!

PR #1381, opened 2022-02-02
Custom QR - Load widgets in background

halemmerich:


gfwilliams:
Thanks! I know you don't call 'draw', but the widgets are still there and I believe will try and show themselves. Eg if you plug the watch in to charge (or after a few minutes) the battery widget will try and redraw.

What were you trying to do with this? Was there some widget that did something, but that wasn't working?

halemmerich:
Yes, I sometimes forget closing the QR code and then widgets don't do their background work. For example the recorder stops. You are right, just calling loadWidgets will cause problems with widgets drawing themselves, I did not test long enough for that to happen.

gfwilliams:
Thanks! Which widgets do you have a problem with? It's possible we should tweak those to use `.boot.js` so they run all the time?

halemmerich:
At least my "invisible" hralarm widget has that problem. That is probably a classic case for code running with *.boot.js, it doesn't even have an icon it could draw. It is more of a theoretical concern, because I forget to close the QR display and suspect I would miss something 😅.

PR #1382, opened 2022-02-02
Recorder - Extract number from file name for overwrite prompt

halemmerich:


gfwilliams:
Thanks! I've been meaning to add another option there for 'new file' as well :)

Issue #1015, opened 2021-12-06
[Bangle.JS 2] [Feature request] Add a setting to remove clocks from the app menu?


Aracnoss:
Since you can't actually (or at least, I've not been able to) properly set the clock from the clocks in the app menu. I end up just using the app manager to drop them to the bottom of the app menu, it'd be nice if there was an option to just hide clocks entirely.

I suggest an option since some people / devs, might want to temporarily view their clocks without setting it.

More of a QoL feature, but if possible it'd certainly be nice.

zloid:
By the way, this functionality is available in the Desktop Launcher app.

alessandrococco:
@Aracnoss Done ;-)

PR #1386, opened 2022-02-03
Launcher - New option to hide clocks from the app list

alessandrococco:
Closes https://github.com/espruino/BangleApps/issues/1015

I use Desktop Launcher as a model, I am still learning :-)


gfwilliams:
Looks good - my only concern would be that you need `if (settings['showClocks'] == undefined) settings['showClocks'] = true;` in app.js too, otherwise by default the clocks will be hidden.

alessandrococco:
Done!

gfwilliams:
Great - thanks!

PR #1387, opened 2022-02-03
Recorder - Allow creation of new file instead of overwriting

halemmerich:


halemmerich:
Integrated in #1388 to prevent changelog chaos

PR #1388, opened 2022-02-03
BTHRM - Allow reading more characteristics if the sensor supports them

halemmerich:
BTHRM
* Allow reading additional data if available: HRM battery, position and RR
* Better caching of scanned BT device properties
* New setting for not starting the BTHRM together with HRM
* Save some RAM by not defining functions if disabled in settings
* Always emit BTHRM event
* Cleanup promises code and allow to configure custom additional waiting times to work around bugs
* Disconnect cleanly on exit

Recorder
* Log src property of HRM-event
* Allow creating new files instead of overwriting


gfwilliams:
Nice - thanks!

PR #1389, opened 2022-02-03
BT HRV - Adds app for taking HRV measurements

halemmerich:
New app, depends on the additional data from #1388.


gfwilliams:
Looks great - thanks!

PR #1391, opened 2022-02-04
Torch For Bangle.js 2

Ronin0000:
Should the red torch app be deleted because now you can set the color of the torch in this app?


gfwilliams:
Thanks! Yes, I guess we could remove the red torch...

Please could you try and just change the existing app.js rather than adding a new one for Bangle.js 2 though? It seems they are identical apart from the very last few lines.

You could just add:

```
// Any button turns off
setWatch(()=>load(), BTN1);
if (global.BTN2) setWatch(()=>load(), BTN2);
if (global.BTN3) setWatch(()=>load(), BTN3);
```

gfwilliams:
Great - thanks!

PR #1398, opened 2022-02-06
feat: add barometer

fbedussi:


fbedussi:
good idea, it will be in v0.2

Issue #1294, opened 2022-01-13
messages app: Feature request - variable font size


lightnin:
Font size of messages app is a bit small for my old eyes. 

Please consider:

1. Variable font size: When text is < n characters, use larger font size that makes most of screen real estate. > n use a minimum size (that is larger than the one now used) and scroll.
 or
2. Setting to control font size of messages.
and/ or
3. Autoscroll lines of larger text.

It seems like an option to get rid  of the buttons on the bottom in favor of a gesture, in order to retain more usable screen real estate for message text, could be another way to make messages more readable.

Thanks for making this great platform! It's really fantastic.

DDDanny:
I support this request.  👍 

I thought in the morning about to use bangle2 touchscreen in showMessage function:
* use the swipe up/down (maybe on the right screen side like clockmode) to imitate something like a scroll/paginate through larger messages
* and furthermore to use also swipes and/or double taps to replace the buttins confirm, abort, back - which also will give the opportunity to have a larger textarea or to support buttom left/right widgets. (Haven't seen one yet)

Any feedback which swipe/tap/drag would fit the best?
I can imagine: 
* back -> swipe left
* confirm -> swipe right (like a lot of "delete" actions in smartphones)
* double tap to abort

During such an extension it should be "easy" to implement small, medium, large font size

DDDanny:
prototype swiping (will allow a larger textarea and prepares to set font large for body) without buttons but a global variable and one swipe handler and one duble tap handler

Testmessages - Mock
```javascript
var MESSAGES = [{t:"notify",id:"mail", new: true, title:"This is long title. Very Long. Ipse lorum actum ractum dactum benedictum benedactum showMessage should be scroll :-p", body:"This is long BODY. Very Long. Ipse lorum actum ractum dactum benedictum benedactum showMessage should be scrollalble - yes, scroll and use the touch screen and replace back button with swipe left, so we can read until here with font medium :-p", src:"Mail", subject: "SUBJECT", positive: true, negative: true},
{t:"notify",id:"phone", new: true, title:"short title.", body:"This is short  BODY.", src:"Mail", subject: "SUBJECT"}];
```

```javascript
var currentShowMsgId;
function showMessage(msgid) {
```
....
```javascript
  // Normal text message display
  currentShowMsgId=msgid;
```
No Buttons: so commented out
```
  /*
  var buttons = [
...
if (msg.positive) {
...
if (msg.negative) {
    }});
...
  }*/
```
[...]

PR #1400, opened 2022-02-06
Messages: Use large font for message body when message < 2 lines 

lightnin:
In: https://github.com/espruino/BangleApps/issues/1294#issuecomment-1012921003, @gfwilliams said: 
"Right now the message title font does change dynamically depending on how much text there is, so doing the same for the body text is probably no big deal."

This mirrors the implementation used for the message title in the message body. 

I can see there are other issues open related to controlling font size globally. But I think the main need for most users will be in the messages app, where the ability to choose a larger font for body text would be useful even in the short term, IMHO. Perhaps getting rid of the buttons in favor of labels to view next / previous via swipe direction would leave more screen real estate for larger text as well. The buttons do not work for me most of the time.
As always - thanks for making this great ecosystem!

### DIFF
--- a/apps/messages/app.js
+++ b/apps/messages/app.js
@@ -270,11 +270,11 @@ function showMessage(msgid) {
       checkMessages({clockIfNoMsg:1,clockIfAllRead:1,showMsgIfUnread:1});
     }});
   }
-  // If body of message is only one line, use large font.
+  // If body of message is only two lines long w/ large font, use large font.
   var body=msg.body, bodyFont = fontLarge, lines;
   if (body) {
     var w = g.getWidth()-48;
-    if (g.setFont(bodyFont).stringWidth(body) > w)
+    if (g.setFont(bodyFont).stringWidth(body) > w * 2)
       bodyFont = fontMedium;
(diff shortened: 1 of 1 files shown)


lightnin:
I realized after the first submission that two lines of large text will fit nicely on Bangle.js 2's screen. So now it checks to see if the message body will fit on two lines of text w/ large font, and only decreases font size if message is larger than that.

gfwilliams:

And yes, swipe was suggested as a way to remove the buttons. I've been meaning to think about that as I think we need something to show users that swiping does something, but otherwise it makes a lot of sense (and button for back)

gfwilliams:
Sorry - one more thing - when you make changes, please can you add a line to the changelog and bump the version number in `metadata.json` - I'll do that now though

lightnin:
> Sorry - one more thing - when you make changes, please can you add a line to the changelog and bump the version number in `metadata.json` - I'll do that now though
Yes! I would have done but wasn't sure if the patch would likely get merged in to a release with other patches. I will in future.  Thanks!

PR #1402, opened 2022-02-06
ContourClock: Fixed bug in settings menu / four more fonts

xxDUxx:
Now all listeners are removed when exiting the settings menu.


gfwilliams:
Thanks! Just merging, *but* this does fail the build (see the red cross by your commits) because it seems you've got tabs used in some files instead of spaces. I'll fix that this end though.

PR #1403, opened 2022-02-06
bthrv - Always write interval data on kill and reset if at least the first measurement is available

halemmerich:
This works in most situations, but the kill handler set on line 98 just doesn't trigger. That causes lines without \n in the csv file when currently taking measurements.

### DIFF
--- a/apps/bthrv/ChangeLog
+++ b/apps/bthrv/ChangeLog
@@ -1,11 +1,2 @@
 0.01: New App!
-0.02: Make overriding the HRM event optional
-      Emit BTHRM event for external sensor
-      Add recorder app plugin
-0.03: Prevent readings from internal sensor mixing into BT values
-      Mark events with src property
-      Show actual source of event in app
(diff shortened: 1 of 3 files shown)


gfwilliams:
Odd about the kill handler not triggering - how are you changing apps?

What's the reason to writing data as soon as possible though? From the point of view of Bangle.js it's definitely more efficient to write to Storage less often, but with more data, so writing whole lines would probably be better

gfwilliams:
Ok, sorting the kill event would definitely be the best solution then...

So if you just paste `E.on('kill', ()=>print("Kill event"))` into the terminal and change app, do you see `Kill event` being printed to the console? Just tried it here and it seems pretty reliable.

halemmerich:
Yes, I had tried that, worked without fail. I suspect some of my other code prevents it from working.

gfwilliams:
Excellent - thanks!

PR #1404, opened 2022-02-06
HRM: Don't force screen to be on for Bangle 2.

dashavoo:
The heart rate monitor app forces the backlight to be on and the screen not to lock.
There is no need on the Bangle 2, as the screen is readable with the backlight disabled.


gfwilliams:
Thanks!

PR #1406, opened 2022-02-07
[LCARS clock] Version 0.15

peerdavid:
Using the wpedom widget to count steps.


gfwilliams:
Thanks! In 2v12 you should be able to do `Bangle.getHealthStatus("day").steps` if you want to get the step data quickly in a way independent of the widgets - it's something I'll be updating widpedom to do at some point soon (when I'm happy enough people are on 2v12 :) )

Issue #1366, opened 2022-01-28
health tracking app: graphic glitch in the widgets bar


alessandrococco:
I just started the HRM (3 min) and the app shows a loooong bar nevertheless there's no data yet.

![image](https://user-images.githubusercontent.com/2331798/151560350-ef4096eb-4ef1-4643-957d-fef5850579f0.png)

After returning to the menu the top of the bar remains in the widgets bar:

![image](https://user-images.githubusercontent.com/2331798/151560778-8cbf5296-d1e4-4f97-8eb5-99f8062c5e05.png)

Android 11, latest stable firmware, all apps are up to date.

alessandrococco:
I took a look at the code - Would it make sense to remove the `drawWidgets` before drawing the bar chart?

https://github.com/espruino/BangleApps/blob/2c9e164a060d92fc751c551922bf34b01f1a5b4d/apps/health/app.js#L104 

I'd like to try to make the change:
- remove the drawWidgets() calls
- update the code to go fullscreen (steps charts also)

What do you think @gfwilliams? :-)

gfwilliams:
It's a bit harder than just removing the drawWidgets call I'm afraid since even if you clear the screen any event that'd cause a widget to redraw (eg battery going down slightly) would put it back on the screen.

Looks like the graph drawing just needs fixing to not draw past the top of the screen - Maybe it should use Bangle.appRect to get the area it can draw in

Ronin0000:
Do you just want the bar not to be drawn at all because there is no data or do you want it to draw a long bar in the middle, but not drawing into the widget area?

Ronin0000:
I have made a pull request for this.

alessandrococco:
> Do you just want the bar not to be drawn at all because there is no data or do you want it to draw a long bar in the middle, but not drawing into the widget area?

Two things:

1. If there is no data for a day/hour then the app should draw nothing for that day/hour
1. The app should not draw into the widget area

Thank you :)

PR #1408, opened 2022-02-08
Don't draw a long bar when there is no data supplied (health app)

Ronin0000:
#1366


gfwilliams:
Is this tested? It seems to me that it now it doesn't draw *anything* at all?

PR #1409, opened 2022-02-08
Update fr_FR locales.js: dow not part of date?

myxor:
I think the day of week should not be part of the date in fr_FR as it is not in most of the other locales.

Fteacher suggested this in the forum: http://forum.espruino.com/comments/16389770/ as well.

**Note**: i am **not** a french speaker so i am not 100% sure about this change but in my opinion this should be changed.


gfwilliams:
Thanks - yes, that sounds like a good plan

alessandrococco:
Maybe we should support both long and short date?

PR #1413, opened 2022-02-09
GPS widget: remove out of date 'requires firmware' messages

myxor:
GPS widget: remove out of date 'requires firmware' messages


gfwilliams:
wow, very out of date! thanks!

PR #1414, opened 2022-02-09
Auto Translated all languages, Updated Language scan script

copoer:
... This will need some QA :sweat_smile:

PS. I don't speak all these languages lol


apps/circlesclock/app.js
```diff
@@ -227,7 +227,7 @@ function isCircleEnabled(type) {
 
 function getCircleColor(type) {
   const pos = getCirclePosition(type);
-  const color = settings["circle" + (pos + 1) + "color"];
+  const color = settings["circle" + (pos + 1) + /*LANG*/"color"];
```
myxor, line 230: This is a key in the array - this should NOT be translated

apps/clicompleteclk/app.js
```diff
@@ -35,7 +35,7 @@ function setting(key) {
     'battery': true,
     'batteryLvl': 30,
     'weather': true,
-    'steps': true,
+    /*LANG*/'steps': true,
```
myxor, line 38: This should not be translated

apps/clicompleteclk/app.js
```diff
@@ -45,7 +45,7 @@ function setting(key) {
 let showBattery = setting('battery');
 let batteryWarnLevel = setting('batteryLvl');
 let showWeather = setting('weather');
-let showSteps = setting('steps');
+let showSteps = setting(/*LANG*/'steps');
```
myxor, line 48: This should not be translated

gfwilliams:
Hi! I really like the idea, but there are a few issues with this PR IMO:

* As @myxor noted a bunch of stuff has `/*LANG*/` applied (eg settings) where it shouldn't be and would break those apps
* There's stuff like `status ? "on" : /*LANG*/"off";` where `on` should be translated too
* I even see cases with `/*LANG*//*LANG*/"string"` which is obviously broken
* Japanese/Chinese translations won't work because the default character set on Bangle.js is ISO Latin (http://www.espruino.com/Fonts#character-sets) - it can't handle Unicode. At some point it'd be possible to render these un-displayable characters into bitmaps and place them in the string, but realistically I think that's something to handle a bit down the road.

Please could you do a PR (or update this one) with *just* the updated language_scan tool with the translate option for now?

* I think the actual `/*LANG*/` additions need a lot of manual vetting before we can add them. Merging something like that now would instantly break half the apps!
* In terms of the translations, I think maybe we should have a 'template' translation file that includes all the stuff we actually want translated - then we can ensure stuff like `"BTNs 1:startlap 2:exit 3:reset"` doesn't get put in there, and we can use that as a basis to feed into the auto-translator?

copoer:
> In terms of the translations, I think maybe we should have a 'template' translation file that includes all the stuff we actually want translated - then we can ensure stuff like "BTNs 1:startlap 2:exit 3:reset" doesn't get put in there, and we can use that as a basis to feed into the auto-translator?

Since we will only translate the things with `/*LANG*/` manually place at the front, we can just auto translate everything so down the road there is a large library of translations that could automatically be implemented

gfwilliams:
Looks great, thanks! Just merging! I'll add something to disable the choice for chinese/japanese, but will keep the translations in there.

Romanization is an interesting thought, although given the lack of space on the display maybe it wouldn't really be sensible.

If you fancy another challenge, it shouldn't be too bad to detect non-roman words, render them to a canvas, convert them and add them to the string with:

```
str += "\0"+imageconverter.canvastoString(canvas, {mode:"1bit",output:"string"});
```

PR #1419, opened 2022-02-10
[LCARS] Stability improvement and possibility to show wind

peerdavid:
Inow included a try catch around the draw function to ensure that - also if something goes wrong - at least the next minute is queued again. Then after one minute draw is executed again. This PR also adds "wind" as additional datapoint which can be configured by uses in the settings.


gfwilliams:
Thanks! I guess if you wanted to avoid the try..catch you could queue the redraw at the start of the function?

peerdavid:
Ah yea thats much better - will be changed in the next version :)

Issue #277, opened 2020-04-10
New app idea - Sleep tracker 


brainfart-be:
Hi !
I use to wear a fitbit and i liked the sleep tracker feature.

I'm going to try to build one myself.
I think i will need some help to build that so don't hesitate to reply bellow if you are interested!

Well time to learn how a sleep tracker work :)

Purple-Tentacle:
I am interested as well.
Currently I am using SleepAsAndroid on my phone. This has the ability to connect to a Google Wear Watch to track movement. 

I do not think that this is possible with Bangle.js. 

But whatI would like
- Start and end of sleep
- Rating of sleep quality (manual, 5 stars)
- Movement detectio 
- Local logging of data
- In the long run: Support for Gadgetbridge, to store and show data on the phone

brainfart-be:
Since i have no clue of how any of this work, i was planning to wear my fitbit and my bangle together.

The bangle app would simply record heart rate and accelerometer (or any other usefull sensor)

Then build a webpage to extract those data from the bangle and try to get the same result as the fitbit.

I obviously have to do some research on this topic

Purple-Tentacle:
Maybe some of the code I used in Active Pedometer could be used in a sleep tracker app?
Many movements are currently counted as steps. We could use use to detect movement.
We have

- Minimum and maximum time between movements
- Threshold of how much movement has to happen before we go into 'movement' mode
- Timer to reset from 'movement' to 'still' mode

Didgeridoohan:
Gonna post this here, just for the record:
https://en.wikipedia.org/wiki/Actigraphy

nxdefiant:
Hi,

I wonder if a "Sleep Phase Alarm clock" is part of the idea here. I have an Axbo which can wake me up as much as 30 minutes before my wake up time. It does that by estimating by sleep phase from an accelerometer at my wrist.

But I agree that we definitely need some real data first. Any progress on this?

brainfart-be:
Hi,

My initial goal was just to display a graph describing my sleep (like my fitbit does, i should post a screenshot of the graph)

I still didn't start anything yet, i kind of took a break from "code"

As a started we could simply "clone" the heart rate monitor app and add other stuff to track like accelerometer and temperature

Purple-Tentacle:
I am currently coding a simple app based on my Active Pedometer app. 
It will simply detect when a certain number of "steps" have been registered in a certain time. 

[...]

PR #1422, opened 2022-02-11
sleeplog: Add Sleep Log App

storm64:
This app logs and displays the four following states:
_unknown, not worn, awake, sleeping_
It derived from the SleepPhaseAlarm and uses the accelerometer to estimate sleep and wake states with the principle of Estimation of Stationary Sleep-segments ([ESS](https://ubicomp.eti.uni-siegen.de/home/datasets/ichi14/index.html.en)) and the internal temperature to decide _sleeping_ or _not worn_ when the watch is resting.

For now this app does not send the status to Gadgetbridge but I wish to add this feature.  
I'm not sure, but it seems as I read in issue [#277](https://github.com/espruino/BangleApps/issues/277#issuecomment-907078912) there is some work on Gadgetbridge to do beforehand?


gfwilliams:

Just out of interest, have you been able to compare with the `movement` field from the `Bangle.on('health')` event that occurs every 10 minutes?

It is also possible to request the current health data. The reason I ask is the movement calculation there is built in, so if that works then we could implement the sleep tracking with basically zero impact on battery life.

storm64:
@gfwilliams
Your welcome, its so rewarding and pleasant to contribute to this huge project. :grin:

For now I haven't thought about the built in calculation. But you have a point there.
I just viewed the output of `Bangle.getHealthStatus().movement`  
and `Bangle.on('health', data => print(Date(), data.movement))`:  
In my opinion it is not possible to use the ESS calculation on this values, but it should be possible to do a slightly more inaccurate sleep prediction.

The idea I have in mind is to add an energy saving option where the health listener value is checked against a separate threshold. I'm not sure how big the impact on accuracy might be but I don't think it will be that much.

PR #1428, opened 2022-02-11
Fix bug and update red7 game.

eyecreate:
I left a big bug in the game where game wouldn't end in some situations. Fix it now.


gfwilliams:
Thanks!

PR #1430, opened 2022-02-11
Moon Phase Widget: Replace the 8 phases by a more exact drawing

HilmarSt:
see forum.espruino.com/conversations/371985


gfwilliams:
Thanks!

PR #1431, opened 2022-02-11
Waypointer Moto: Add routes, support for Bangle.js 2, better editor UI

jes:
"Routes" are where you preconfigure several waypoints in order and it automatically
jumps from one to the next as you approach. There is a flaw in that if you
accidentally miss one waypoint on your route, there's no way to make it go on to
the next one without going all the way back. We probably want some way to manually
skip back/forth in the route.

Bangle.js 2 support is provided via hardcoded layout coordinates. This should
probably use the Layout library instead, but it'll probably do for now.

I've also replaced the table-based waypoint editor UI with an entirely map-based
replacement which I find much more intuitive.


gfwilliams:
This looks great, thanks! I bet the sunlight readable screen on Bangle.js 2 really helps

jes:
Thanks!

PR #1432, opened 2022-02-11
boot - Sort *.boot.js files

halemmerich:
Sorts the *.boot.js files during bootupdate.js and allows an optional priority with the following scheme: 
`name.<optional_priority>.boot.js`

The resulting order would be like this:
```
xyz.0.boot.js
xyz.3.boot.js
abc.boot.js
xyz.boot.js
```


gfwilliams:
Looks great - thanks!

PR #1434, opened 2022-02-11
dtlaunch - Optimize line breaks for Bangle 2

halemmerich:
Allows multiple words in one line and splits camel case app names to allow wrapping those.


gfwilliams:
Great! Not sure if it helps, but have you seen `g.wrapString` exists now?

PR #1438, opened 2022-02-12
[Hebrew Calendar] big refactor to add features

barakplasma:
Did a big refactor to add features via a customization page.

Caveat:
One problem I encountered was running out of memory at runtime if I tried to load more than 3 months of calendar data into the custom app. It can't filter that large a calendar. I also tried to heatshrink the calendar data by month, which made it smaller, but didn't succeed in getting around the lack of memory. I might try again in the future.

HebrewCalendar-Screenshot


gfwilliams:
Is that bug fixed now, or should I close this PR until you're ready?

barakplasma:
closed for now, I'll reopen when it's ready

PR #1440, opened 2022-02-12
Recorder webinterface: Catch some errors

myxor:
I had a broken recorder track which causes errors downloading the GPX file.
I was not able to see the errors except in the web console.
So here are changes which make the errors visible if they happen.


gfwilliams:
Thanks! Are the errors anything that can be recovered from?

myxor:
This not 100% clear to me yet. I still have the files and will have a deeper look into it asap.

PR #1441, opened 2022-02-13
[LCARS] Optional fullscreen mode

peerdavid:
Before this release, LCARS was fullscreen only. The drawback is that some really cool widgets are not shown and probably some things are reimplemented. Now users can *enable* fullscreen mode in the settings and widgets are shown. Per default `fullscreen=false` - the user can then decide to explicitly enable fullscreen mode and hide widgets. 

Here is the difference between fullscreen enabled / disabled:

screenshot_1

screenshot_3


alessandrococco:
Nice!

Can I suggest to draw a separation between the widgets and the clock?

A quick Paint cut&paste to give the idea:

image

What do you think?

peerdavid:
I agree it looks better, but then have problems showing the second screen correctly... that was the main reason why I decided to design it this way...

peerdavid:
Ok @alessandrococco - I just pushed a new commit which includes your suggestion - below is a screenshot of my bangle with fullscreen disabled. For the second screen, no top line is shown, because then the graph also nicely fits :D Thanks again for your suggestion I like the new design much more :)

screenshot_3

gfwilliams:

PR #1442, opened 2022-02-13
gbridge: Move event listener from widget to boot code

rigrig:
This probably should've been `boot` code from the start

Apart form the reorganizing it also runs for any app (not just clocks): it still won't autoload the app for non-clock apps, but it does prevent music from showing up in the Messages app if the song changes while you aren't looking at a clock.


gfwilliams:
Thanks - looks good!

PR #1446, opened 2022-02-13
BTHRM - Allow settings modes

halemmerich:
There are 4 new modes: Off,Default,Both,Custom
These all set a valid subset of the settings.


gfwilliams:
Thanks!

PR #1447, opened 2022-02-13
Support naming, Farenheit, and alerts.

emarc:
Add ability to rename tags.
Add Sauna, Fridge & Freezer alert. Support °F based on locale.

Prepare for Bangle 2 w/ swipe support (maybe – need to read up on that and try the emulator).


gfwilliams:
Thanks!

Issue #1423, opened 2022-02-11
[sleeplog] Log file not found: Uncaught Error


myxor:
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

Hello @storm64 
your sleeplog app looks promising but it crashes when no "sleeplog.log" exists in the filesystem:

```
Uncaught Error: Expecting a string, got undefined
 at line 93 col 59 in sleeplog.boot.js
...(storage.read(this.logfile)));
                              ^
in function "log" called from line 74 col 45 in sleeplog.boot.js
          this.log(Math.floor(Date.now()), 2);
                                            ^
in function "calc" called from line 164 col 102 in .boot0
...idth) global.sleeplog.calc();
                              ^
in function called from system
```

## How to reproduce:
1. install sleeplog
2. connect watch to console and see errors in log

storm64:
Thanks for the fast fix @myxor. I added the `atob()` conversion to save storage, but forgot to test it without a logfile. :see_no_evil:

PR #1448, opened 2022-02-14
sleeplog: Add power saving mode, fix #1445 

### LINKED ISSUE #1445: [sleeplog] Having sleeplog enabled hinders execution of kill event handlers
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

For reproduction install sleeplog and enable it. Then run the following code from IDE:
```
E.on('kill',()=>print("killed"));
```
Cause a reload, "killed" will not be printed.
Then deinstall sleeplog and run the code again.
A reload prints "killed"

storm64:
* Add power saving mode using build in movement detection
* Move all read/write log actions into lib.js with issue #1423 in mind
* Prevent removing other kill listeners, issue #1445
* Add changing menu entries depending on power saving mode, implemented workaround to espruino/Espruino#2149


gfwilliams:
Great - thanks!

PR #1450, opened 2022-02-14
Add Neon X & IO X Clock

bundyo:
This is a Bangle.js version of the Pebble's Neon X and Neon IO X watchfaces by Sam Jerichow. They are rather similar, so I combined them in a single clock that can be toggled from Settings.

The originals are here:
https://apps.rebble.io/en_US/application/5572db913b1d5162a3000029
https://apps.rebble.io/en_US/application/5572dd5d3b1d51c76f00002f


gfwilliams:
Looks great, thanks!

Issue #1437, opened 2022-02-12
Problem with "Bangle.CLOCK"


HilmarSt:
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

In some apps / widgets <code>**Bangle.CLOCK**</code> is used to detect if a clock is running, e.g. in "Digital clock widget":
<code>if (!**Bangle.CLOCK**) WIDGETS["wdclk"]={...</code>

This seems not to work any more.

gfwilliams:
Which clock faces have you tested on? Because I just tried here and it works fine for me.

If the clock calls  `Bangle.setUI("clock` *after* loading widgets then this will break, so that could be it?

HilmarSt:
I'm using the Vector Clock, and you're right - Bangle.setUI("clock") was executed _after_ Bangle.loadWidgets() .
I'll create a PR to fix this.

gfwilliams:
Great - thanks!

PR #1451, opened 2022-02-14
Vector Clock: Fix problem with "Bangle.CLOCK"

HilmarSt:
https://github.com/espruino/BangleApps/issues/1437


gfwilliams:
Thanks!

PR #1452, opened 2022-02-14
Circlesclock v0.10

myxor:
# Changes

* Use roboto font for time, date and day of week and center align them
* Move default settings to external file


gfwilliams:
Thanks!

PR #1454, opened 2022-02-15
Rolex watch face

DC959:
A Rolex style watch face.

Hasn't been tested on a physical watch as I am not purchasing one till the end of the week but works well on the emulator.

Please Ignore the first 7 commits as I forgot to make a sub folder, the last 3 commits are the changes that have been made


gfwilliams:
This looks great - thanks! Just tried it on a real watch and the only issue seems to be if the watch has a light theme then the widgets draw wrong. I'll put in a quick fix for that (temporarily changing the theme to dark).

One thing I did notice is your images seem to be 8 bit. You could probably get away with using 1 (or 2) bit greyscale images which would save a bunch of RAM, and the benefit is the Bangle would automatically be able to render them and flip colours so you could actually have the watch face in a light theme if you wanted

PR #1458, opened 2022-02-15
Fixed theme honoring and made images 2 bit

DC959:
This update should fix theme honoring so the watch face now changes to match the theme set on the watch.
The images have also been updated to be better compatible with 2 bit and also now set to 2 bit to save ram.
Also fixed the date text as that stopped displaying in the last update.


gfwilliams:
Excellent - thanks!

However just a note about the ChangeLog - looks like you created a new file called `Changelog` rather than updating `ChangeLog` - i'll fix it here

DC959:
Ah sorry I didn’t have a local file for ChangeLog so decided I’d create one as the last one I just created through GitHub, apologies for the mistake

PR #1459, opened 2022-02-16
sleeplog: Fix #1445, display loading info, add icons to display service states

### LINKED ISSUE #1445: [sleeplog] Having sleeplog enabled hinders execution of kill event handlers

storm64:
Fix logfile disabling:
- fix logfile correction in `setEnabled(...)` to make disable logging possible in lib.js
- simplify logfile checks in lib.js
- fix error on reading a non string logfile value in settings.js


gfwilliams:
Thanks!

PR #1461, opened 2022-02-16
[NeonX & IO X] Added optional fullscreen mode.

peerdavid:
First of all, thanks @bundyo for your awesome NeonX watch face - I really really like it. I thought it would be cool to have the option to show it in fullscreen mode - therefore, I implemented this and opened a PR. Per default, the fullscreen mode is disabled such that the user must enable it in the settings. Here is an example for both, Neon X & IO X:

download

download

I hope it's fine to add this functionality directly to this watch face as I'm not the creator of it?

------------
To be still able to show some information such as battery, steps etc. relatively easily and fast, I added additional "Info". Instead of adding it directly to the src of the clock, I decided to export this functionality into a new app as it can be useful in general to check data, versions, states etc.:

screenshot_1
screenshot_2
screenshot_3

Now in combination with the PatternLauncher, I can use the NeonX watch face in fullscreen mode which looks pretty cool and still show all other important data such as battery or steps with a single swipe :)

I hope its ok to have both apps in one PR.

David


bundyo:
No problem from my side, great and thanks! :slightly_smiling_face: 

I guess Gordon is the one to decide, as I don't have the rights to merge.

gfwilliams:
Looks great - thanks!

PR #1470, opened 2022-02-17
Anton Clock v0.07

jeroenpeters1986:
Aligned default settings with app.js, otherwise the initial displayed settings will be confusing to users. I was puzzled by it everytime I had to reset my watch and wanted to change settings.


gfwilliams:
Thanks!

Issue #1154, opened 2021-12-21
[TimeCal - Feature request] - Configure if week starts on Monday or Sunday


m-p-3:
I'm used to my calendars starting on Sunday, and would like the ability to configure TimeCal as such. Thanks!

DDDanny:
Okay, I'll update the clock for u like this:

0.02: Feature Request #1154 and some findings...
      -> get rendered time from locale
      -> *BATT SAFE* only update once a minute instead of once a second
      -> *RAM optimized* clean code, corrected minute update (timout, no intervall)
      -> locale: weekday name (first two characters) from locale
      -> added settings to render cal view begin day (-1: today, 0:sunday, 1:monday [default])

I personally don't use it:
* Are you able to test the patched clock?
* Is it ok for you that I provide it via my github custom app loader?

Of course I will pre-dev check & test it!

DDDanny:
> Of course I will pre-dev check & test it!

I changed a lot, so added multiple customisations of colors and highlights. I still work on some UI(automated) tests.

Preview: 
![grafik](https://user-images.githubusercontent.com/31635744/149848571-03a464a2-8f89-4621-8b6c-0879738e71ea.png)
the widgets area is here for testing reasons below the clock, but it will dynamically scale it's width. - calendar border is configurable and also sunday color and today's coler

m-p-3:
Hi, I checked on https://dddanny.github.io/BangleApps/index.html but TimeCal is still at v0.01 there, would you kindly point me in the right direction? I'll gladly test it on my end :)

DDDanny:
I still struggled with my (generic automated UI) testscript. 
The mentioned location is correct. 
I pushed and merged the branch now to my master so you can test in espruino WebIDE and/or with your bangle.

m-p-3:
Loaded on my Bangle JS2 running firmware 2v11 Rel

It looks like this
![TimeCal_v0 02](https://user-images.githubusercontent.com/3731432/150895318-1b087783-e53e-4d4a-906a-6a0591a7dfe7.jpg)

Also, when going in `Settings > Apps > TimeCal`, it gives me the message

```
TimeCal:
Invalid
settings!

BTN1 to go back
```

DDDanny:
Sorry for the late reply !!!
1. I missed the settings app ;) This is working now!
2. Something with the bottom widget recognition is wrong -> I forced +24 now
3. I tested in webIDE and installed successfully - maybe some settings configurations will not work together as expected
[...]

PR #1475, opened 2022-02-18
Timecal clock 0.05 - settings fixed

DDDanny:
fixes settings (defaults) of #1154
I have invalidated the .test.js by rename it to ._js


gfwilliams:
Thanks!

PR #1476, opened 2022-02-18
[NeonX & IO X] Added optional settings to show the lock status.

peerdavid:
I added optional settings to switch colors whenever the bangle is unlocked. Per default, those settings are off to keep the standard behavior and design. This especially helps in full-screen mode to easily detect the lock state while not changing the cool design of this watch face.

Also, I found that the watch face is updated every 10 seconds - wouldn't it be enough to do it every 60 seconds in order to save some battery? What do you think about this PR @bundyo?

### DIFF
--- a/apps/neonx/ChangeLog
+++ b/apps/neonx/ChangeLog
@@ -1,2 +1,3 @@
 0.01: Initial release
 0.02: Optional fullscreen mode
+0.03: Optional show lock status via color
--- a/apps/neonx/README.md
+++ b/apps/neonx/README.md
@@ -22,3 +22,8 @@ Shows the current date as DD MM on touch and reverts back to time after 5 second
 ### Fullscreen
 Shows the watchface in fullscreen mode.
 Note: In fullscreen mode, widgets are hidden, but still loaded.
+
+### Show lock status
+In fullscreen mode it can be useful to detect, whether the BangleJs is locked or not.
+If these settings are enabled, the first digit is shown red if the BangleJs is locked
+and purple otherwise.
--- a/apps/neonx/README.md
+++ b/apps/neonx/README.md
@@ -24,6 +24,4 @@ Shows the watchface in fullscreen mode.
 Note: In fullscreen mode, widgets are hidden, but still loaded.
 
 ### Show lock status
-In fullscreen mode it can be useful to detect, whether the BangleJs is locked or not.
-If these settings are enabled, the first digit is shown red if the BangleJs is locked
-and purple otherwise.
+If enabled, color changes when unlocked to detect the lock state easily.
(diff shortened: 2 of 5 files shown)


bundyo (approved):

bundyo:
There is one possible place that would probably benefit from faster refreshes - when you sync the time from external connection. It probably would get updated on the next refresh. I wonder why there is no `Bangle.timeChange` event (with maybe a `resolution` argument) that clocks can just subscribe to, instead of refreshing at regular intervals?

peerdavid:
I totally agree with you - a sync from an external connection would take in the worst case 60 sec. instead of 10... Also not sure how much the battery consumption improves - honestly I'm happy with both things simply let me know and I will change it accordingly.

I'm currently running my own device on 60 sec. to check if there are any other drawbacks - currently everything looks fine for me...

I have also seen that the original antonclk updates each 60 sec. in case only minutes are shown, and whenever seconds are shown its updated each second:

```python
function updateState() {
  if (Bangle.isLCDOn()) {
    if ((secondsMode === "Unlocked" && !Bangle.isLocked()) || secondsMode === "Always") {
      secondsScreen = true;
      queueMillis = 1000;
    } else {
      secondsScreen = false;
      queueMillis = 60000;
    }
    draw(); // draw immediately, queue redraw
  } else { // stop draw timer
    drawTimeout = undefined;
  }
}
```

peerdavid:
Ok @bundyo I made some tests and learned the following:  With setInterval we call draw() each n seconds, independent when the minute really changes. So with an interval of 60 we are in the worst case 60 seconds late. With 10 seconds we are still 10 seconds late in the worst case. On the other hand, we can "setTimeout" based on DateTime.now i.e. we schedule the next draw exactly when the minute changes. I implemented this in the current version to test it also on my device, but it looks good until now and also I used this for lcars and not analog... let me know what you think :)

Also, @gfwilliams I hope its not wrong what I write here...

bundyo:
Seems okay to me.

gfwilliams:
>  Maybe @gfwilliams can tell if that's okay?

Thanks - this looks good. I think realistically the `setTimeout(..., 60000 - (Date.now() % 60000));` code here is the way to do. It's what I'm suggesting for new clocks. Numerals is older (I think?) so doesn't use it (but probably should)

In an idea world, yes, we'd have some kind of time changed event, but I think for now this code is a good compromise. And if in future the event does get added, we'll be able to search for all instances of `60000 - (Date.now() % 60000)` and update them accordingly.

bundyo:
I've tested the 60 seconds code the last two days - it draws around 4-5% of battery per 24 hours - only clock, nothing else enabled, not connected to phone.

PR #1488, opened 2022-02-19
messages: add Home Assistant logo

rigrig:
Based on https://github.com/home-assistant/assets/blob/master/logo/logo-small.svg, this is the best I could do: ha-1bit


gfwilliams:
Thanks!

PR #1489, opened 2022-02-19
Simplify gbmusic

rigrig:
- simplify touch controls: just touch for play/pause, swipe for next/previous
- remove date+time

Discussed here: http://forum.espruino.com/conversations/373020, although it only got 1 reaction.


gfwilliams:
Thanks!

PR #1491, opened 2022-02-19
[messages] Fix music text clipping

### LINKED ISSUE #1478: [messages] Music information overlapping back button
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### Description

When there is music with a particularly long album title, the text can clip over the back button.

### Expected behavior

I would expect the text to either end (probably with an ellipses) or side scroll.

### Steps to reproduce
1. Start playing music on phone that has a long album title
2. Open music view on messages app on Bangle
3. Back button has text clipping over it

nh-99:
This fixes a bug (fixes #1478) in the music control in the messages app that would cause the text to clip the back button or off the screen.

I tried out two versions of this fix and they're contained in the commit history. Initially I tested a version that just used an ellipses when the text was too large to fit. This didn't work very well in practice, as you couldn't see much of the music information in a lot of cases. I ended up going with scrolling text instead due to this, however that makes the display refresh every second which isn't ideal.

Let me know what you think, I'm willing to make more changes to this implementation.


gfwilliams:
Thanks! This looks like a good plan, but re-initialising the layout all the time is going to be really slow - it's possible just to update the labels: http://www.espruino.com/Bangle.js+Layout#updating-the-screen

I'll have a go at tweaking this and merging in now though

gfwilliams:
Just merged with some tweaks - looks great!

gfwilliams:
Just merged with some tweaks - looks great!

nh-99:
Thanks for that doc link! I figured there was probably a way to do a partial refresh, but I didn't dig into it. I'll keep that in mind for future applications.

PR #1494, opened 2022-02-21
No idea what I am doing

Catking473:
: (


Catking473:
Accidentally added a request on the wrong BangleApps repo. I was attempting
to do it on my copy.
I am going to post this app though with different file names, once it is
finished.

PR #1495, opened 2022-02-21
[Neonx] Ensure that widgets are hidden in fullscreen mode

peerdavid:
I have seen some cases, where widgets are shown although the fullscreen mode was active. I assume that it happened because widgets were not hidden whenever a "redraw" happened. Now widgets are hidden within the draw function to ensure that they are never shown...

Unfortunately, I was not fully able to reproduce when this bug happened, but always when I got messages from gadgetbridge and the message widget started blinking...


gfwilliams:
Great - thanks!

PR #1496, opened 2022-02-22
New app - smclock

smekras:


gfwilliams:
Do you think this is ready to merge in? I know there were a few changes recently.

gfwilliams:
is this ready to merge?

smekras:
Might as well. Plenty to update, but it will need some work, so... better if there is an initial version up.

gfwilliams:
Thanks!

Issue #1460, opened 2022-02-16
[messages] file write optimisation


gfwilliams:
Right now, we write a new JSON file every time a message is added or removed. If you're constantly getting notifications this could make Bangle.js run slow pretty quickly.

I think it'd make things a lot faster if we just used StorageFile instead, and we *only* append to it. Either a new message, or something saying a previous message should be removed. We could then check if the StorageFile was getting too big and just delete it and create a new one without the junk in it.

Not too hard, and it should help a lot, at the expense of a bit of a delay when reading it

jeroenpeters1986:
I was reading this and I thought I got it, but I think I don't. 
Are you also planning on changing the format, like a row per message?
I am a bit lost on only appending and something saying a previous message should be removed. Arent you then drawing messages and later retracting them?

rigrig:
Sounds like a good idea.   

If we're going to change message storage, maybe we should special-case music, as it seems those lead to a lot of pointless storage activity? i.e. keep `musicstate`/`musicinfo` in memory, and only write+read them to flash when launching the app?

gfwilliams:
That's a good point - actually it might be worth just special-casing if the messages app is open - storing the messages in RAM and then only writing them on exit. Things like navigation will also really thrash the storage otherwise

rigrig:
If we're messing with it anyway, it might be a good idea to decouple the library from the app.

In [my (WIP) message app rewrite](https://github.com/rigrig/BangleApps/blob/new-messages/apps/messages/lib.js) I changed it to do `Bangle.emit("message", type, event);`.
I don't think we want different message apps, but maybe clocks can do something nice with it, and it seems cleaner in any case.

gfwilliams:
Sorry for taking so long to check https://github.com/espruino/BangleApps/pull/1498 out properly... I still think this is a good plan - I just wanted to sit down and properly run through what's happening in the PR and haven't felt like I had time yet.

> maybe we could just store messages as a regular file each?

Yes, that would definitely be a big improvement over the current situation.

... however what really slows the Bangles down at the moment is having lots of 'trash' files in Storage. Your solution would use a lot less storage (and RAM) but would still generate one new file per message (however deleting a message wouldn't create a file so I guess it's twice as good as now ;)

[...]

PR #1498, opened 2022-02-22
Save messages in append-StorageFile

rigrig:
Implements #1460
* No special case for music (yet), but I think that would probably be best combined with autostart-for-music
* Right now it writes a fresh StorageFile when exiting the app, and [randomly](https://github.com/espruino/BangleApps/pull/1498/files#diff-73da835ccb0c190b9545584c57154f5d990da6450c16586486edd30a7afa8d8bR154) when "modify"/"remove" lines are written and the app is not open. Not sure if that is the best way to go and/or if these are reasonable values.

### DIFF
--- a/apps/messages/ChangeLog
+++ b/apps/messages/ChangeLog
@@ -74,3 +74,7 @@
 0.52: Fix require("messages").buzz() regression
       Fix background color in messages list after one unread message is shown
 0.53: Always use "message" events, even for the default app/widget
+
+
+
+0.XX: Save messages in a StorageFile
--- a/apps/messages/ChangeLog
+++ b/apps/messages/ChangeLog
@@ -73,3 +73,4 @@
       Move WIDGETS.messages.buzz() to require("messages").buzz()
 0.52: Fix require("messages").buzz() regression
       Fix background color in messages list after one unread message is shown
+0.53: Always use "message" events, even for the default app/widget
--- a/apps/messages/app.js
+++ b/apps/messages/app.js
@@ -62,6 +62,11 @@ var onMessagesModified = function(msg) {
   }
   showMessage(msg&&msg.id);
 };
+Bangle.on("message", (type, msg) => {
+  msg.handled = true;
+  onMessagesModified(msg);
+});
+
 function saveMessages() {
   require("Storage").writeJSON("messages.json",MESSAGES)
 }
--- a/apps/messages/app.js
+++ b/apps/messages/app.js
@@ -48,9 +48,47 @@ we should start a timeout for settings.unreadTimeout to return
 to the clock. */
 var unreadTimeout;
 /// List of all our messages
-var MESSAGES = require("messages").getMessages();
-if (!Array.isArray(MESSAGES)) MESSAGES=[];
+var MESSAGES, modified = [];
+function saveMessages() {
+  // only write modified messages
+  require("messages").write(MESSAGES.filter(m => modified.includes(m.id)));
+}
+function set(msg, key, val) {
+  if (msg[key]!==val) {
+    msg[key] = val;
+    if (msg.id && !modified.includes(msg.id)) modified.push(msg.id);
+  }
+}
+function markRead(msg) {
+  if (msg.new) set(msg, "new", false);
+}
+function remove(msg) {
+  if (!MESSAGES.some(m => m.id===msg.id)) return; // already gone
+  MESSAGES = MESSAGES.filter(m => m.id!==msg.id);
+  modified = modified.filter(id => id!==msg.id);
+  require("messages").remove(msg);
+}
+try {
+  MESSAGES = require("messages").getMessages();
+  // Write them back to storage when we're done
+  E.on("kill", saveMessages);
+} catch (e) {
+  g.clear();
+  E.showPrompt(/*LANG*/"Message file corrupt, erase all messages?", {title:/*LANG*/"Delete All Messages"}).then(isYes => {
+    if (isYes) {    // OK: erase message file and reload this app
+      require("messages").clearAll();
+      load("messages.app.js");
+    } else {
+      load();// well, this app won't work... let's go back to the clock
+    }
+  });
+}
+// used by lib.js to inform app of updates
 var onMessagesModified = function(msg) {
+  if (msg.t==="remove") {
+    remove(msg);
+    msg = undefined;
+  }
   // TODO: if new, show this new one
   if (msg && msg.id!=="music" && msg.new && active!="map" &&
       !((require('Storage').readJSON('setting.json', 1) || {}).quiet)) {
@@ -60,16 +98,13 @@ var onMessagesModified = function(msg) {
     if (msg.state && msg.state!="play") openMusic = false; // no longer playing music to go back to
     if (active!="music") return; // don't open music over other screens
   }
+  if (msg && msg.id && !modified.includes(msg.id)) modified.push(msg.id);
   showMessage(msg&&msg.id);
 };
 Bangle.on("message", (type, msg) => {
   msg.handled = true;
   onMessagesModified(msg);
 });
-
-function saveMessages() {
-  require("Storage").writeJSON("messages.json",MESSAGES)
-}
 
 function showMapMessage(msg) {
   active = "map";
@@ -101,8 +136,7 @@ function showMapMessage(msg) {
   g.reset().clearRect(Bangle.appRect);
   layout.render();
   function back() { // mark as not new and return to menu
-    msg.new = false;
-    saveMessages();
+    markRead(msg);
     layout = undefined;
     checkMessages({clockIfNoMsg:1,clockIfAllRead:1,showMsgIfUnread:1,openMusic:0});
   }
@@ -134,8 +168,7 @@ function showMusicMessage(msg) {
     updateLabelsInterval = undefined;
     openMusic = false;
     var wasNew = msg.new;
-    msg.new = false;
-    saveMessages();
+    markRead(msg);
     layout = undefined;
     if (wasNew) checkMessages({clockIfNoMsg:1,clockIfAllRead:1,showMsgIfUnread:0,openMusic:0});
     else checkMessages({clockIfNoMsg:0,clockIfAllRead:0,showMsgIfUnread:0,openMusic:0});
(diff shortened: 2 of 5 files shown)


gfwilliams:
Thanks! There's a lot there so I'll have to go through this later. It looks good but it's a surprising amount of extra code needed.

I think it's probably worth not rewriting it in the messages app (as I imagine that'll be opened for most notifications anyway), but maybe reallocate if the size would grow above 40k (which is the point where storagefile would create a new file)

rigrig:
> it's a surprising amount of extra code needed.

Well, "needed"...

The two main bulks are
- `exports.load()`: it keeps a list of ids while reading messages, so removing/modifying only needs to look at that list, instead of filtering the whole array every time.
- `exports.haveNew()`: basically the same as `load()`, except it only keeps a list of new ids. This might be over-optimized, we could just replace it with `exports.load().some(e=>e.new)`. (But I was happily coding away, so I figured I'd optimize a bit for the widget while I had the thing in my mind)

> maybe reallocate if the size would grow above 40k

I thought about looking at the size, but [the docs ](https://banglejs.com/reference#l_StorageFile_getLength) says "This requires Espruino to read the file from scratch, which is not a fast operation.". So I figured checking that every time we append a message would kind of defeat the purpose. Unless most of the delay comes from parsing JSON?


Also I wasn't sure if every event definitely has an `id` set, if so It would probably simplify some things.

rigrig:
I rebased it onto master, and got rid of separate `exports.haveNew()` code: we don't check for new messages often enough that it needs special optimized parsing.

gfwilliams:
Thanks! I will get on this, I'm just pretty busy at the moment

rigrig:
Sure, there's no rush, and I definitely want you take a good look at changes like this before merging.

rigrig:
Still TODO: look into the [numbers](https://github.com/espruino/Espruino/blob/9327f504c7e0bcdedb29ff47d3cc3bc4122eec9d/src/jswrap_storage.c#L32), come up with a sensible value to compare to `StorageFile.getLength` to determine whether to `compact` the messages file (and maybe make it smart enough to check if compacting will help at all)

rigrig:
Rebased it onto master, and got rid of `compact()`: it already rewrites the entire file every time we close the app anyway, so I figured that keeps it small enough for now.

But we can probably do better:
1. Instead of saving an updated `MESSAGES`, just append changes to the file, and only rewrite it when it gets too large
2. Make it even smarter: keep track of changes, and only write messages that need updating when leaving the app. (Alternatively: compare the in-memory `MESSAGES` with the saved file.)

gfwilliams:
Looks great - thanks! I'll have a play with this branch on mine and see how it's working :)

rigrig:
If you've got some time now that's great, but note that I really meant the **untested** bit for my last changes: I think it *should* work, but haven't gotten round to actually testing it, at all.

rigrig:
I tested it a bit in the emulator and fixed a few bugs, seems to work OK now:   
* I uploaded `messages/lib.js` as `messages`, loaded `android/boot.js` into RAM, and pasted `gbridge/sample_messages.js` into the REPL. After that I checked `require('messages').load()`, and it looked good.
* After pasting in enough copies of the sample messages for `messages.jsonl` to grow over 40000bytes, it erased the file and started a fresh one without losing any messages.
* Doing 
  ```
    M = require('messages').load();
    /* modify M: delete/modify messages */
    require('messages').write(M);
  ```
  only added lines for the deleted/modified messages.

gfwilliams:
Ok, great! Maybe it's best to hold off on merging this just a little more then until you've been running with it on the watch itself for at least a few days.

Just thinking aloud here but the actual file size is ~40930 bytes I think? So if you do a check at 40000 and then you add another 930 bytes of messages, a new file would get created anyway. I think it's unlikely but I guess we could drop the check to 38000, in which case it would be hugely unlikely that any one message would push it into a second file.

I'd also be interested to see if you notice any slowness when exiting the messages app. While I like the idea of the `.write` function, I guess in a lot of cases (marking read/deleting/etc) it'd be pretty easy to just write the extra line directly into the messages file without having to parse the entire thing. It also avoids having to read the whole file on exit even if nothing has changed, and potentially opens the way for not actually ever having to load all messages into RAM (we could just parse the file and read the single message out as and when it was needed).

... but maybe that's one for later when we've got this merged ;)

rigrig:
>  if you do a check at 40000 and then you add another 930 bytes of messages, a new file would get created anyway. 
> we could drop the check to 38000

Good point, I'll adjust that.

> in a lot of cases (marking read/deleting/etc) it'd be pretty easy to just write the extra line directly into the messages file

Sure, but in most cases you probably open the app, open a message (marking it as read), and then delete it, so you end up writing a copy of the message except with `new=false`, and then delete it anyway.   
But we could do this:
* write deletions to the file right away
* keep track of changed messages, only write those on exit   

That way we don't need to parse the file on exit.

gfwilliams:
Could do, but probably not worth the extra effort just yet. Let's set your this PR goes.

Taking of which I used this branch and put in on my bangle and something is very wrong - short buzz and no notifications at all. But I'll have to look into it on Monday

rigrig:
Turns out it wasn't that much effort, so I added it before reading your comment.   
I guess somewhere along the line something broke during some merge/rebase :-(    
I just rebased onto master, figured the easiest way forward was to just fix merge conflicts in the rebase as best as I can, and now try to stamp out the bugs. (Merging in half a dozen options every now and then probably wasn't great for the code)

rigrig:
I gave it a try on my Bangle.js 1, but didn't have time to properly debug stuff yet.
Apart from other issues, I just noticed my `messages.jsonl` now contains exactly this line:
```
{"t":"modify","track":"\u0000\u0012\u0015\u0001\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u000F\u0000\u0003\u00F0\u0000\u00FF\u0000?\u00F0\u000F\u00F8\u0003\u00F8\u0000\u00F8\u00008\u0000\b\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000 Body Machine | Priest","dur":-1,"c":-1,"n":-1,"id":"music","title":"Music","new":true,"state":"play"}
```
That's from Firefox playing media being forwarded to my phone, so the "Body Machine | Priest" bit is fine, but all the `\u0000` and the fact that there is exactly one line worries me.

gfwilliams:
Ahh, that'd do it! Although if you're not in a clock, do message icons still appear? That didn't work for me either.

Do you not get out of memory errors with the current latest messages app? I've noticed the newish message icon code appears to keep the entire messages list in RAM, which is pretty wasteful.

How many messages did you have? It doesn't look to be like your code should use *that* much extra memory.

rigrig:
>  if you're not in a clock, do message icons still appear? That didn't work for me either.

Whoops, only tested it from the clock.

> Do you not get out of memory errors with the current latest messages app?
 
I don't know, I've been using my [own version](https://github.com/rigrig/BangleApps/tree/new-messages). I didn't get round to merging in either these changes *or* the icon update yet though.   
But it happened with just a few messages, so I'll at least look at it a bit harder.   
(Also: it looks like I broke `music` messages...)

rigrig:
Just tried the messages app from `master`, and getting `LOW_MEMORY` there as well, so I'm going to assume it isn't because of these changes :wink: 
(Looks like drawing the scroller with all messages is what pushes it over the edge, loading a single message works fine)

gfwilliams:
Interesting - thanks! Might have to have a look at that - although I guess for most users you probably have 5 messages max

rigrig:
I didn't mean that the scroller uses excessive memory, just that creating it was the tipping point when the memory ran out. (I didn't test it on an empty watch, so there were already widgets eating up memory)
Anyway, it didn't run out of memory since the widget was [fixed](3be786841e1f65006ef744f7d77daa3a93989aa0), and [this PR](https://github.com/espruino/BangleApps/pull/2065) should help a bit more.

Apart from that, it has worked fine for me the last couple of days.

rigrig:
Whoops, fixed another small bug, and rebased onto master.

rigrig:
TODO: I just noticed that [`7x7dotsclock`](https://github.com/espruino/BangleApps/blob/387e037a6111af84ff3625f2c7091376a5582d91/apps/7x7dotsclock/7x7dotsclock.app.js#L272) and [`hcclock`](https://github.com/espruino/BangleApps/blob/387e037a6111af84ff3625f2c7091376a5582d91/apps/hcclock/hcclock.app.js#L247) load `messages.json` directly, so those will need fixing.

gfwilliams:
Thanks - I'll file an issue for that as there should really be a way to just query the messages library/widget

rigrig:
Strange, I *know* I tested filling up the storage file before, and it worked fine. But today I finally filled up my watch "naturally",and it broke :-(

Fixed now:   
Turns out you can't do this (simplified a bit):
```js
    const file = require("Storage").open("messages.jsonl", "a");
    if (file.getLength()>38000) {
      file.erase();
      file.write(JSON.stringify(e)+"\n"); // throws "Can't write in this mode"
```
So I changed it to
```js
    let file = require("Storage").open("messages.jsonl", "a");
    if (file.getLength()>38000) {
      file.erase();
      file = require("Storage").open("messages.jsonl", "a");
      file.write(JSON.stringify(e)+"\n");
```
and now it works again.

(I'm not surprised that writing to an erased file throws, just that it used to work before, maybe I broke it during some merge/rebase...)

gfwilliams:
Ahh, interesting, thanks! It's good it fails though - that could have led to storage corruption before.

I'm pretty sure you don't actually need an erase. You can just use `w` and it'll erase as part of creating the new file:

```
file = require("Storage").open("messages.jsonl", "w");
```

rigrig:
Rebased it onto `master`.   
I notice this PR has accumulated quite a bit of formatting cruft, I'll see about cleaning that up.

rigrig:
Rebased onto `master` (i.e. message events).

rigrig:
Set it back to draft, as it will need a bit of work(+more testing) to apply onto #2205 

Some good news: now that the `messages` library can be extended, I am using this code as daily driver for my `smessages` custom messages app, so the chance of me running into any lurking bugs is higher :-)

gfwilliams:
> I am using this code as daily driver for my smessages custom messages app, so the chance of me running into any lurking bugs is higher :-)

You mean you're using append-storagefile?

Thanks! and sorry It's taken so long for me to test/merge this one in - it just feels like quite a big one :)

rigrig:
> You mean you're using append-storagefile?

I have been for quite a while now, just not this exact code:   
My own (still WIP) fork already uses the storage file. The difference is that now the [message events PR](https://github.com/espruino/BangleApps/pull/2146) landed, it uses `require("message")` (the version in this PR) instead of duplicating a bunch of code.

> Thanks! and sorry It's taken so long for me to test/merge this one in - it just feels like quite a big one :)

I totally understand, it's a big change and we definitely don't want the messages app to break.   
And waiting a bit doesn't really bother me: I can just load this code onto my watch anyway ;-)

rigrig:
Putting this on hold for a bit, at least until we have a stable `require("messages")` API.

gfwilliams:
I guess we do now have a more stable API - and I think we could in fact just create a separate `messages_sf` app now that would allow both the old and new to coexist?

However, do you think there's a need for this now? I've made a few changes to storage in the last 6 months that has made slowdowns from lots of file writes basically disappear completely - even completely uncompacted it should be basically as fast as right after a compact.

By contrast I'm a bit concerned that reading the messages line by line from a StorageFile may actually end up being slower than the current solution.

With your recent changes to the library it seems we have the option to *not* save messages too, and we already fast-load the `messagesgui`, so with a little more work I think we could probably avoid saving the messages to Storage in 90% of cases (eg. when getting a message when on the clock, dismissing it, and going back to the clock).

Issue #1479, opened 2022-02-18
run 0.6 - Cadence never changes from 0


hughbarney:
### Affected hardware version

Bangle 2

### Your firmware version

2v12.32

### The bug

Run 0.6 setup with 
Box1 : Time, Box2: BPM, Box3:  STEPS, Box 4 Cadence.

The cadence never updated from 0 and stay at zero throughout the test walk.
Screenshot attached.
![download (2)](https://user-images.githubusercontent.com/3630481/154743851-123299c5-1b59-456f-8fc3-9c10bb38459f.png)

copoer:
Here is a potential fix I am testing:
https://github.com/espruino/BangleApps/pull/1502

copoer:
It seems to fix it. :+1:

gfwilliams:
Thanks for the fix @copoer !

PR #1502, opened 2022-02-23
Fixed stepHistory

copoer:
It seems to fix it. +1

https://github.com/espruino/BangleApps/issues/1479


gfwilliams:
Thanks! Looks good!

The odd thing is I'm certain I had cadence working. Maybe I just forgot to commit something

PR #1504, opened 2022-02-23
[Contour Clock] Fix Locale

copoer:
Fixed contour clock to use locale formatting


gfwilliams:
Thanks, but this looks really dodgy - you're taking a string with locale and then guessing at the formatting (which might not even contain a ':') and then converting that string back to a number and using divide and modulo on it.

What are you trying to achieve here? Is it that you just want the 12/24 hour option to work?

copoer:
> What are you trying to achieve here? Is it that you just want the 12/24 hour option to work?
Yes

I might do this a different way if the locale library has a clear way.

copoer:
Maybe I could pass the processed locale time back into a new date object?
Like this?
```
var date = new Date(require('locale').time(new Date(),1));
```

gfwilliams:
Ahh, ok. So this will definitely fail for `wae_CH` by the look of it.

I would actually just use the previous code, but with this function instead of `date.getHours()`:

```
var is12;

function getHours(d) {
  var h = d.getHours();
  if (is12===undefined) is12 = (require('Storage').readJSON('setting.json',1)||{})["12hour"];
  if (!is12) return h;
  return (h%12==0) ? 12 : h%12;
}
```

It's what the locale library uses internally, but is sadly not exposed. I don't think it's safe to use the locale string and then try and extract the 2 numbers as I bet at some point someone will add extra characters at the beginning or end of the format string for some language, or will do something that otherwise breaks it :(

gfwilliams:
> Maybe I could pass the processed locale time back into a new date object?

Nice idea, but the problem is there's no guarantee that the time string produced by locale is actually parsable by Date...

gfwilliams:
Thanks!

PR #1505, opened 2022-02-23
Messages: Threema icon

myxor:
I added the threema icon but i wasn't sure about the size of the icon.

Should it be 32x32px or 48x48px? The code suggests it to be 48x48px but it looked to big for me so i made it 32x32px.

--
As this is just a minor change i did not increase the version and changelog.


gfwilliams:
You can check existing ones with `g.imageMetrics`:

```
>g.imageMetrics(atob("GBiBAAAAAAAAAABAAAHvAAHvAADvAAAPAB/PMB/veD/veB/mcAAAABzH8B3v+B3v+B3n8AHgAAHuAAHvAAHvAADGAAAAAAAAAAAAAA=="))
={ width: 24, height: 24, bpp: 1, transparent: true }
```

Looks like they are 24px - you can see from the size of the text string that this one's a bit big :)

myxor:
Thank you @gfwilliams. I changed the threema icon to 24x24px with 1bit color.

gfwilliams:
Great - thanks!

PR #1506, opened 2022-02-23
Initial commit of widcw

avanc:
widcw is a widget to display the current calendar week.


gfwilliams:
Thanks! Are those SVG files just there as source images for the icon?

avanc:
> Thanks! Are those SVG files just there as source images for the icon?

Yes, I want to make a fancier icon, but don't have a good idea yet.
But I just saw that I committed the similar file under two names. I will fix this when I do an update.

gfwilliams:
Thanks!

PR #1507, opened 2022-02-23
Daisy: added settings menu

hughbarney:


gfwilliams:
build failed with `ERROR: App daisy file daisy.settings.js doesn't exist`

Looks like `{"name":"daisy.settings.js","url":"daisy.settings.js"}` should be `{"name":"daisy.settings.js","url":"settings.js"}`?

hughbarney:
pushed fix.

gfwilliams:
Thanks!

PR #1509, opened 2022-02-24
messages: Remove left-over debug statement

rigrig:
Not sure if this warrants a version bump, but I suspect unexpected output might confuse the App Loader?


gfwilliams:
Oops - thanks!

PR #1510, opened 2022-02-25
Messages: icon for alarms

myxor:
Alarm clock reports as "alarmclockreceiver" on my side but i guess it is "alarm" on some other devices


gfwilliams:
Nice - thanks!

PR #1512, opened 2022-02-25
New widget - widadjust

malaire:
This widget adjusts clock continually in the background to counter clock drift.


gfwilliams:
Thanks!

PR #1513, opened 2022-02-25
[WIP] A new timer widget which exposes functions for other apps and clocks.

peerdavid:
Now the library is added to the qalarm app. A smpltmr app is added which is simply an alternative UI to demonstrate how this library can be used. Clearly, the qalarm is still compatible with bjs1 and bjs2 while the smpltmr is specially designed for bjs2.

ToDo:
- [ ] Clarify if the current solution is ok
- [ ] Update lcars app
- [ ] Update notanalog app


gfwilliams:
This looks like a great idea - just some questions though:

* Looks like `widtmr` is set as working for Bangle.js 1, but I see a touchscreen event handler there that uses coordinates (when the Bangle.js 1 only has a 2 zone touchscreen). Does it actually work on Bangle.js 1?
* The default 'Alarm' app does timers as well, and they even work when you're in an app that doesn't load widgets. Is it actually worth trying to base this on that? - even if there's a separate app with a different UI for setting them at least it could use the same basic code for handling timers, so we don't get in the situation where you can have 5 independent timers running all in their own apps and you have to cancel them all from different places.
* Same with the Chrono widget really - I think even if it doesn't make sense to use the default app, please could you modify the original widget rather than forking please? It hasn't been touched in 6+ months so I think it makes sense to just improve that...

thanks!

peerdavid:
Thanks a lot for your great suggestions! I marked this PR as [WIP] until I made all changes :)

> Looks like widtmr is set as working for Bangle.js 1, but I see a touchscreen event handler there that uses coordinates (when the Bangle.js 1 only has a 2 zone touchscreen). Does it actually work on Bangle.js 1?

Very good hint thanks a lot - I have to check this again and will do more tests with the BJS1 emulator and adapt the UI accordingly.

> The default 'Alarm' app does timers as well, and they even work when you're in an app that doesn't load widgets. Is it actually worth trying to base this on that? - even if there's a separate app with a different UI for setting them at least it could use the same basic code for handling timers, so we don't get in the situation where you can have 5 independent timers running all in their own apps and you have to cancel them all from different places.
Same with the Chrono widget really - I think even if it doesn't make sense to use the default app, please could you modify the original widget rather than forking please? It hasn't been touched in 6+ months so I think it makes sense to just improve that...

Totally fine for me. The reason why I created a fork is, that I have no idea if I'm really allowed to change the app of others... that's the reason why I forked it and created a new one. Otherwise, I also prefer to change the current widget.  So I will remove the "widtmr" widget and implement it in the chrono widget. Also I will try to adapt the chrono widget such that it is based on the default Alarm app which sounds a great idea to me too. I simply was not aware that I can do this :)

-- I only need some days to implement and test those things as I'm a bit busy this week :)

David

gfwilliams:
That sounds excellent - thank you!

About modifying other people's apps - generally I wouldn't worry too much. If you check the commit history and it's being worked on a lot recently maybe check with the author, but especially if it's something that was committed and then never worked on for 6 months I wouldn't worry :)

For the default alarm app - maybe have a `timer` library provided by the `Alarm` app that it creates that provides the `setTimer/etc` functionality you were providing before (I imagine at some point there will be an `Alarm` library too). I think we do similar things for the messages/health apps, but if you need a hand with it or some advice please let me know!

alessandrococco:
> I marked this PR as [WIP] until I made all changes :)

There's a better way: https://github.blog/2019-02-14-introducing-draft-pull-requests/ ;-)

peerdavid:
Thanks a lot @gfwilliams

peerdavid:
Hi @gfwilliams I just finished the first version and wanted to ask you about a short review of what I have down so far. If you think its ok I will update the lcars and notanalog app too.

First of all, I had a problem with the timer of the default alarm app on my device - but the qalarm app worked fine so I added a lib there. The library exposes functions to easily create, update or delete a timer.  Each app must select a unique ID if a timer is created which ensures that only timers that are created by this app are changed. I created the "Simple Timer" app which shows how this library can be used. But its quite easy:

With the following line I create and start a 1h 10min. and 5 sec. timer:
```javascript
qalarm.editTimer("my_unique_id", 1, 10, 5);
```

With the following line I stop/disable the timer:
```javascript
qalarm.deleteTimer("my_unique_id");
```

And if I want to know how many minutes are left I simply call
```javascript
qalarm.getTimerMin("my_unique_id");
```

alessandrococco:
> First of all, I had a problem with the timer of the default alarm app on my device - but the **qalarm** app worked fine so I added a lib there.

Could your alarm library become a standard module (like `layout`)?

gfwilliams:
Hi - thanks! What issues did you have with the default alarm/timer app? It'd be good to get those sorted if there's a problem there.

Just a minor note on the library - I'd use `exports.xyz = function() { ... }` rather than defining named functions separately and referencing them as it saves some RAM, and I wouldn't call `Bangle.loadWidgets` from the library, as when you require the library the widgets may already be loaded.

My main concern really is making this part of qalarm rather than the default alarm. We're just having the exact same problem in that a user installs the 'default' alarm, installs your clock and now there are two separate alarms. 

Worth nothing that QAlarm will throw up `BTN2 not defined` errors and the not buzz when the hard alarm goes off by the look of it - I'm not sure if that breaks other things. It's just stupid having two separate (but mostly identical) alarm apps that are broken in different ways. I think probably the best bet here is if someone (probably me I guess) just merges the two alarm apps together and removes the Q Alarm app.

*Or* we have a new 'Core Alarm' library that provides the `boot.js` and `lib.js` and then the Alarm apps sit on top of that. Looking at them, I have a feeling that when merged there's not going to be a massive difference in apps though.

The question really is whether if I merge the apps, someone will come along with a `Q Alarm v2` that's just another fork. If that's the case I guess we need to go with the 'Core Alarm' approach

peerdavid:
Thanks for your answer. I can easily remove the laodWidgets and also change the exports as suggested. 
But I had problems with the original alarm app. I simply uploaded the app on my bangle and created a one minute timer.
I waited for 5 minutes but nothing happened:

qalarm


 Could be that something is wrong with my bangle but it seems that the timer
in this app is not working at all - so decided to not create an api for it as I don't want to fix this app too. 
On the other hand, the qalarm app worked immediately and therefore I simply created an api to it.
 
Honestly, I don't want to invest more time into this little thing - it burned already too much time and I want to spend 
my time a bit different :) So I will close this PR now and keep my fork as it is - its totally fine for
me to run this beside the original sotre - then the original store can be clean and I add my features there :)

Still what I wonder a bit - which is also important if I want to push something in future to the original store 
- why you ever released the qalarm app if there should be only a single alarm app? And also if something already exists,
(lets say a GPS recorder) I should not push another one right?

Thanks a lot,
David

gfwilliams:
Ok, thanks. I'll look into that. I just did a 1 minute timer and it was fine, but I did exit the app after. I tried without a reload and it failed, so maybe timers don't get reloaded properly when things change inside the app.

I notice you edited the post, but I just wanted to try and explain:

I'm sorry, I know stuff like this is frustrating. It's bad for everyone involved.

You've done loads of work on Bangle.js apps - and crucially you create them and *then* you maintain and improve them, which many folks don't - and it really is hugely appreciated.

But basically I'm stuck in the middle. I have a bunch of contributions, and I want to try and accept everything I can and be friendly to new contributors, but also to make Bangle.js easier to use and less confusing. Whatever I do, someone's going to be unhappy with the outcome - it's just whether it's a contributor that's very unhappy, or thousands of slightly confused Bangle.js users.

I shouldn't have merged the 'Q Alarm' or 'Hard Alarm' apps in the first place, but I don't actually like saying no to people and have been insanely busy - it seemed kind of harmless at the time. But we've now got `hardalarm`, `alarm` and `qalarm` - all of which have issues. Initially you'd come up with your own, and I know someone else forked `qalarm` just so they could make it work with Gadgetbridge. So then we end up with 5 separate alarm apps which all do basically the same thing and in some cases look identical, but none of which work together - it's got stupid now, and I need to do something.

Q Alarm is a bit of a sore point for me as it's had a few bugs, some of which have made all apps installed at the same time unreliable. If I hadn't spent the hours tracking down issues related to Q Alarm and Hard alarm on the forum, or maintaining them when I made changes to the menu, I could have used that time to make the original alarm app actually work well and integrate with Gadgetbridge.

But you're totally right - please just have fun with Bangle.js.

**It's a problem I need to sort out**, and I'll try and simplify things in the coming weeks. I'll ensure there's an API and hopefully when it's done, if you want to, it'll be pretty easy to tweak your apps to use it.

peerdavid:
Thank you so much for the clarification - I now really understand your goal which totally makes sense for the future of this project! I will definitely update all my apps if there is an API available and open a PR afterwards. Also, I totally understand that you have to maintain and keep the code clean, otherwise, it will become unmaintainable and probably also unusable (as you said having 5 alarm apps installed etc.) soon which nobody wants...

Still I think its a nice solution to split the store through the fork for now - then I can play around with it and do some quick and dirty things - simply write some apps that I want. Whenever I implemented a good solution or something that can be useful also for the whole community, I will open a PR containing only a clean solution. I guess that's the best workflow.

Thanks again for your great and hard work.

David

PR #1514, opened 2022-02-25
Fix Daisy Clock

dukky:
The new `daisy` clock was missing a line in `metadata.json` to make is actually usable as a clock, as it stands you can only open it as an app from the launcher but not set it as a clock in the settings.


gfwilliams:
Thanks!

PR #1515, opened 2022-02-26
Fix typo in Time and Life shortName

tjsilver:
The shortName for the Time and Life app was misspelled in the metadata, so it showed in the menu as "Time and Lfie". This PR fixes that.


gfwilliams:
Thanks! Please can you avoid committing other whitespace refactoring in future though?

Generally all the other files look like this one used to, and it doesn't really make it more readable.

PR #1525, opened 2022-02-28
RUN App - README.md: Make links to github absolute

BartS23:
otherwise they will not work in the apploader

### DIFF
--- a/apps/run/README.md
+++ b/apps/run/README.md
@@ -46,8 +46,8 @@ record GPS/HRM/etc data every time you start a run?
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - but I wonder whether this should be fixed properly in the app loader? I've had other PRs for *removing* absolute URLs in other (non-app) areas of documentation and it feels like the way to go.

I guess just using `../../` would work?

BartS23:
../../modules/README.md becomes https://banglejs.com/apps/modules/README.md in the apploader.

It then displays the readme in raw, it is better than 404 but not perfect in my opinion.

I'll adjust the PR.

PR #1530, opened 2022-03-01
added g.clear() for fix the trasparent widget bar if there are no widgets - and new setting

Rarder44:
###issue
if there are no widgets, the widget bar remains transparent and does not draw the background


gfwilliams:
Thanks! Please can you update the version and Changelog too?

Rarder44:
sorry, you're right! I forgot!

gfwilliams:
Great - thanks!

PR #1541, opened 2022-03-04
Ratchet Launcher: Initial version

jsphpl:
> Ratchet Launcher is a fork of the default Launcher with modified user interaction. Instead of free scrolling, apps are selected by swiping up and down, but in discrete "ticks", just like in the settings menus.

This is my first contribution to bangle.js, so i am especially happy for any feedback on coding style, espruino-specific JS best practices, etc.

Is it okay to list `"supports": ["BANGLEJS2","BANGLEJS"]` without having tested it on Bangle.js v1? I am only assuming, it *should* work, but have on idea if it actually *does*.

Link for quick testing: https://jsphpl.github.io/BangleApps/?q=ratchet


gfwilliams:
Thanks! I'd say it's always best to try it on the emulator if you don't have a Bangle.js 1.

I just gave it ago and you're using fonts that Bangle.js 1 doesn't have - but you could fix it with:

```
var font = g.getFonts().includes("6x15") ? "6x15" : "6x8:2";
var largeFont = g.getFonts().includes("12x20") ? "12x20" : "6x8:3";
```

Please could you put those lines in, and then I'll merge and we'll know it works on both :)

jsphpl:
Do i need to increment the version number? It hasn't been merged yet, so i left it at 0.01...

gfwilliams:
Great, thanks! And no need to bump the version as I hadn't merged yet. Merging now!

PR #1545, opened 2022-03-06
Add optional notification of milestones reached to exstats and use it in run app

GrandVizierOlaf:
I started this as just a test to try and get my feet wet with app development, but then it expanded a bit and I thought it could be useful so I wanted to open a PR for discussion, but understand if it needs some rework.

Basically, the `exstats` module can now be optionally instructed to emit `notify` events when `dist`, `step`, or `time` reach milestones set by the user (like `dist` going up 1000 m or `step` going up 30 seconds). I set up the `run` app to make use of this and it will allow the user to customize a different `Bangle.buzz()` pattern per notify event type. I think this could also be useful for pedometers and eventually apps like `wohrm` which want to know when `hrm` has gone outside of given bounds. I thought that it could be helpful to have that consolidated in the `exstats` module rather than expecting each app to perform similar basic actions separately.


apps/recorder/app.js
```diff
@@ -305,7 +305,7 @@ function viewTrack(filename, info) {
     g.drawString(require("locale").distance(dist),g.getWidth() / 2, g.getHeight() - 20);
     g.setFont("6x8",2);
     g.setFontAlign(0,0,3);
-    g.drawString("Back",g.getWidth() - 10, g.getHeight() - 40);
+    g.drawString("Back",g.getWidth() - 10, g.getHeight() - 40); // TODO: this position depends on Bangle1 vs 2
```
myxor, line 308: Is this TODO still required?

GrandVizierOlaf: Yes, but it is just something I ran into with the `recorder` app while messing around and is unrelated to the `exstats` and `run` changes. When plotting a map, altitude, or speed of an existing track, the back button is displayed in the bottom right of the screen where it would be on a banglejs1.

gfwilliams: Thanks! That's a good point. I just fixed this in an update, so it might be worth removing this line from the PR

gfwilliams:
Thanks! I'm not too sure whether we should start loading loads of extra features into the exstats library (things like this *could* be in a separate `exstats-notify` library I guess), but this doesn't seem to use too many extra resources, and integrates quite nicely.

Merging now :)

Issue #1521, opened 2022-02-27
Feature Request: Barometer - Pressure change alerts


fparri:
Would that be possibile to add Air Pressure change alerts?

https://www.thoughtco.com/how-to-read-a-barometer-3444043

gfwilliams:
You mean in the barometer app itself, or just as something that pops up in normal use?

It's definitely something that'd be pretty easy to do, but with it on your wrist the pressure will change a lot just due to changes in height as you walk up and down stairs, so I guess you'd have to be careful that didn't trigger anything.

fparri:
I am testing a Huawei Watch GT Runner: it offers the possibility to enable or disable alerts for pressure drops, measuring the drop in hPA over the last 3 hours, with the following options:

2 hPa/3 h
2.5 hPa/3 h
3 hPa/3 h
3.5 hPa/3 h
4 hPa/3 h
4.5 hPa/3 h
5 hPa/3 h
5.5 hPa/3 h
6 hPa/3 h

I attach some pics

![01](https://user-images.githubusercontent.com/1544681/156017067-c10b6870-bfeb-4d81-9446-0839c0e6c03f.jpg)
![02](https://user-images.githubusercontent.com/1544681/156017071-0a8d4169-e1ad-471b-9fab-78a3adee1a57.jpg)
![03](https://user-images.githubusercontent.com/1544681/156017076-9e8fe056-6949-4cfb-a9f0-3ec344e338f2.jpg)

gfwilliams:
Looks like a good idea - it could be handled as a widget pretty easily.

Honestly though, I've spent the entire day today *just* answering messages, so for me personally I don't have much time to implement something like this

myxor:
I am working on a widget which implements this. Will create a merge request in the next days.

fparri:
Just read and just seen your pull request. Looking forward to trying it! :)
Thanks a lot!

myxor:
@fparri v0.02 is available since yesterday. Please give it a try and i am happy to receive feedback from you :)

fparri:
> @fparri v0.02 is available since yesterday. Please give it a try and i am happy to receive feedback from you :)

Avsolutely. I'm going to update the app and try it tomorrow. Thanks a lot :))))))

fparri:
@myxor I tried it yesterday. It worked perfectly. The only difference I noticed with the same feature on my Huawei GT Runner is that the GT Runner probably compensates pressure variation due to a change in altitude.
[...]

PR #1546, opened 2022-03-07
Barometer Alarm Widget

myxor:
This is an initial implementation of  #1521

@gfwilliams sometimes the barometer reports obviously wrong values (e.g. 900 when all previous values are around 1002). Any idea how i can avoid this?
I tried to solve this by calculating averages over 3 values but this does not help if one of the three values are so far away from the other two values.


gfwilliams:
Looks great! I'm happy to merge as-is if you want?

In terms of wrong values, I'm not really sure why they'd be off, but I'd suggest median filtering...

So maybe each time, turn the barometer on, take 5 readings, then sort them and pick the middle one?

myxor:
I am now using median calculation with 5 values as you suggested.

Ready to merge.

gfwilliams:
great - thanks!

PR #1547, opened 2022-03-07
qmsched: support new bangle.js 2 menu, add /*LANG*/ hints

rigrig:
Whoops, this app was quite broken with the new menu, because of how it abused `format`.


gfwilliams:
Ahh - thanks :)

PR #1552, opened 2022-03-09
Barometer Alarm Widget: Try not to warn multiple times for the same exceedance

myxor:
This is required because the widget warns on every run of the check for threshold exceedances before. 
I don't think this is what the user wants ;)

I also removed the screenshot because i was not able to capture a screenshot of the notification via `g.dump()`  - i think that's a bug on the notify module.


gfwilliams:
Thanks! You think this is ready to merge?

And yes, the notify module makes a fake `g` instance so you don't get any apps in the background trying to overwrite what's on the screen, so what you see in `g` is pretty much nothingness

myxor:
Thanks for the explanation of the notify module.

I will test this MR for one more day before i will make it ready to merge.

gfwilliams:
Great - thanks!

PR #1554, opened 2022-03-09
bthrm - Allow scanning for sensors in settings

halemmerich:
Set a specific sensor to use to prevent chaos when more than one is in range. Without a set sensor the old "connect to first found" behaviour is kept.


gfwilliams:
Excellent - thanks!

PR #1556, opened 2022-03-10
Update metadata.json

fparri:
I only changed the name of the app so that it shows the initials in uppercase.


gfwilliams:
Thanks!

PR #1557, opened 2022-03-10
messages: Setting to auto-open music

rigrig:
Apart from automatically launching the app for music, also switch back to music after handling incoming "real" messages.


gfwilliams:
Great - thanks!

PR #1565, opened 2022-03-13
TodoList v0.01

francoislanc:
This is a simple Todo List application.

The content is loaded from a JSON file.
You can mark a task as completed.


gfwilliams:
Nice - thanks!

PR #1567, opened 2022-03-13
powermanager - Allow forcing monotonic battery voltage/percentage

halemmerich:


apps/powermanager/boot.js
```diff
@@ -26,4 +26,26 @@
     Bangle.on("charging",handleCharging);
     handleCharging(Bangle.isCharging());
   }
+  
+  if (settings.forceMonoPercentage){
+    var p = (E.getBattery()+E.getBattery()+E.getBattery()+E.getBattery())/4;
+    var op = E.getBattery;
+    E.getBattery = function() {
+      var current = (op()+op()+op()+op())/4;
```
rigrig, line 34: This should be rounded: `Math.round((op()+op()+op()+op())/4)`, as apps/widgets expect an integer percentage.
(Is `E.getBattery` really so unstable that we need this in the first place?)

foostuff: > (Is `E.getBattery` really so unstable that we need this in the first place?)

In my experience with my bangle2, E.getBattery() can return a range of ±3. (max-min=6). Averaging 4 samples brings it down to ~±1. 

I sometimes play with even higher counts:
```
function bat(c) {
var c = (typeof c !== 'undefined') ?  c : 100, bat=[]
while (bat.length < c) bat.push(E.getBattery());
return{count:c,avg:bat.reduce((a, b) => a + b, 0) / c,min:Math.min.apply(Math, bat),max:Math.max.apply(Math, bat)}
}
```
It could be more stable with the latest firmware or due to the higher charge, because it currently has a range of ±1.5 on my bangle.

gfwilliams:
Thanks! As @rigrig says, please could you add a `Math.round(...)` call in there? It's possible that other things (eg widgets that just print the percentage) will get broken by having non-integer values

apps/powermanager/boot.js
halemmerich: I suspect the newer firmware will amplify that behaviour since the mapping follows the Li-Ion discharge-curve more closely and maps a smaller part of the voltage values to more percentage values.

gfwilliams:
Thanks!

PR #1568, opened 2022-03-13
boot: fix alphabetic sorting of *.boot.js files

rigrig:
No version bump: assuming that where it really matters we already set a priority, which worked fine.


gfwilliams:
Thanks!

PR #1570, opened 2022-03-14
Update recorder to make widget work if app has not been run before.

GrandVizierOlaf:
The recorder widget had some assumptions about what would be in `recorder.json` and would break if the app had not been run to initialize it properly. This is what changed:

1. Properly attach `app-settings.json` as `recorder.json` and make sure it includes the updated log name schema
2. Fix `getTrackNumber` so it can handle there not existing a matching track file
3. If the widget is enabled and there is no file, set a default

By itself, fix 1 would have been enough, but the other two are nice guards.


gfwilliams:
Excellent - thanks for this!

Issue #1566, opened 2022-03-13
[Question] Howto install SleepPhaseAlarm?


EnerGeh:
I'm unable to install the SleepPhaseAlarm App on my Bangle.JS 2. Is there a hidden feature to do so? I have only the like button (heart) but no Upload Button.
I appreciate any help 🤗
![Screenshot_20220313-201208_Bromite](https://user-images.githubusercontent.com/28715981/158075280-9385caad-b648-445a-93bc-0eed29ef4abd.png)

gfwilliams:
That's an odd one. If you run the App Loader on another computer, do you have the same issue? I just tried here and the upload button appears for me

EnerGeh:
Fair point!
But I just discovered, that the content inside that surrounding frame is horizontally scrollable 🤦
All other apps I installed are with a upload button that doesn't need scrolling. 😩

nxdefiant:
Reproducible when Browser is set to mobile layout with width less then ~500px

gfwilliams:
Thanks for tracking this down @nxdefiant!

PR #1571, opened 2022-03-14
main.css: Force line break for smaller resolutions

nxdefiant:
See #1566


gfwilliams:
Great - thanks for this!

PR #1572, opened 2022-03-15
New watchface based on the Seiko 5actus

DC959:


gfwilliams:
Thanks! Looks good. However please could you rename the directory from `seiko_5actus` to `seiko-5actus`? The automatic checks fail at the moment: https://github.com/espruino/BangleApps/pull/1572/checks

Longer term it might also be worth using smaller images for the hands? Looks like they are 176x176 right now when maybe you only need 176x32 or something? It'd save memory *and* make it render faster.

DC959:
Sorry for some reason seiko-5actus.js wasn't working no matter what I did, kept giving me no app source found when trying to launch the clock, bug maybe? I triple checked the metadata file and it seemed fine, I ended up changing it to app.js and all is fine, everything seems to be working correctly now

gfwilliams:
Great - thanks! Merging now.

In terms of the alignment, I think the images are centered when you rotate them - so if you just crop out the area of unused space *on both sides* so the hand is still centered in the image you should be fine. You'd only have issues if you (for instance) cropped the shorter side of the watch hand down so the middle of the watch hand was no longer centered in the image.

PR #1573, opened 2022-03-15
Included sun clock and daysleft widget

bernschneider:
Included new sun clock app showing date/time, sunset/sunrise, H = current sun height/noon sun height, Az = sun azimuth.

Please only merge the apps/sunclock folder!


HilmarSt:
Do I see it right that some of the default locations in mylocation.app.js are removed?
Please tell the reason for it, I've explicitly added one location in the southern hemisphere that you plan to remove.
The moon phase widget needs this location to also work in the southern hemisphere.

HilmarSt:
I don't know if it's possible to split a pull request.
If I had to revert the changes that I don't want to have in the PR, I would simply commit the original file that's still in the master.

bernschneider:
Done: I committed the original file.

HilmarSt:
Looks good for mylocation.app.js - it's not in "Files changed" any more, thank you.

Why do you want to add a new widdaysleft widget? It looks nearly identical to the existing apps/daysl/widget.js with only setTimeout() added.
Wouldn't it be better to just add this function to the existing "Days left"widget?

bernschneider:
Hi Hilmar,

the existing Days Left widget prints the number of days left in the current month. My widget takes a list of dates and shows the number of days left until the first date in the daysleft.json files, which lies in the future. (daysleft.json includes a list of increasing dates.)

I do see a way to join both apps, if we provide a configuration functionality in the Apps Loader,. There the users could decide, if they count the days until the end of month or want to provide a list of custom dates.

Can you point me to a documentation of the Apps Loader apps configuration setup?

Thanks, Peter

gfwilliams:
Just merged - thanks! If you can try and only submit PRs for the files you want to change in future it'd be great though, as It's a bit of a pain for me to manually remove all the other files

bernschneider:
Thank you, Gordon!
I will do.
Best regards,
Peter

HilmarSt:
> the existing Days Left widget prints the number of days left in the current month.

I think this is not true, I've used this widget before to remind me on a birthday: You can enter a date in the settings menu for the app until the days are being count.

> I do see a way to join both apps, ... the users could decide, if they count the days until the end of month or want to provide a list of custom dates.

I think the easiest way would be to add just one item, like "One Date or Date List", in the settings menu <code>function showMenu()</code>of the app:
https://github.com/espruino/BangleApps/blob/master/apps/daysl/app.js

bernschneider:
Hi Hilmar,

I was talking about the days left in month widget, and didn’t yet know about the daysleft app you are linking to - sorry!

What you think about this behavior:

1. App: 
If there is no daysleft.json, the app creates an empty list, otherwise it reads the date list from daysleft.json.

The app´s menu now allows definition of a new date (as currently in your app). This date is appended to the list and the list is saved.

2. Widget:
The widget reads daysleft.json and looks for the first date in the date list, which lies in the future. It computes the days left until this date and shows it in the widget area at the top (as currently in my widget).

Best regards,
Peter
-
Peter Bernschneider
0151-675 675 70
***@***.***

PR #1580, opened 2022-03-15
Improved dozenal watch face

numerist:
Novice here; haven't found the change log yet to update it. Changes:
-- extraneous comments and code removed
-- display improved
-- now supports Adjust Clock widget, if installed

### DIFF
--- a/apps/doztime2/app.js
+++ b/apps/doztime2/app.js
@@ -206,7 +206,7 @@ function modeTime()
 {
 	timeActiveUntil = new Date();
 	timeActiveUntil.setDate(timeActiveUntil.getDate());
-	timeActiveUntil.setSeconds(timeActiveUntil.getSeconds()+15);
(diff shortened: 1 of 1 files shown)


numerist:
Not understanding the conflicts; just trying to update apps/doztime2/app.js

numerist:
It's app-bangle2.js that needs the changes.

gfwilliams:
I think the issue may be that I don't think `apps/doztime2` exists in the main repository. I'm pretty sure last time you submitted, you'd made that a copy of `apps/doztime`, so we decided to merge the two together so there was one app, but that just had one file for Bangle.js 1 and one for 2.

It might be worth trying to get your repository to be exactly the same as the main one (even if that means removing and re-adding it from GitHub) and I think things will be a lot easier then. I'll try and make your changes here though.

gfwilliams:
Just merged manually

PR #1586, opened 2022-03-17
Seiko-5actus: shrunk hand images to save memory

DC959:
Your response in my last pull request made me realize how simple it would be to keep the hands centered and shrink the image size down so there wasn't a bunch of extra transparency around each image.

I've shrunk the minute and hour down so that the point I need centered stays centered but removes all the whitespace around the images. The second hand isn't able to be changed due to already being at the max size it can be.

Hopefully this improves memory utilization a little

### DIFF
--- a/apps/seiko-5actus/ChangeLog
+++ b/apps/seiko-5actus/ChangeLog
@@ -1,1 +1,2 @@
 0.01: Initial Release
+0.02: Shrink hand images to save memory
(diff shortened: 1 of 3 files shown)


gfwilliams:
Excellent, thanks! The compressed base64 doesn't get much smaller I know, but once it's decompressed it'll make a big difference to RAM

PR #1588, opened 2022-03-18
Fix for when user decides to not delete a track in the recorder app

GrandVizierOlaf:
This fixes an incorrect variable reference when backing out of the menu to delete a track.


gfwilliams:
Great - thanks!

Issue #1539, opened 2022-03-03
Desktop Launcher Does not show image of white images.


Ronin0000:
### Affected hardware version

Bangle 1

### Your firmware version

2v12

### The bug

The Desktop launcher does not show white images, when not selected like the Black Jack Game Image. I can post a screenshot.

Issue #1532, opened 2022-03-01
Crow Clock Does Not Look Good In Dark Mode


Ronin0000:
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

When I open clow clock in dark mode it looks like this.
![image](https://user-images.githubusercontent.com/89286474/156267144-1d34936c-0f20-41f2-9ac0-2f13670a114c.png)
And it should look like this:
![image](https://user-images.githubusercontent.com/89286474/156267173-d0acf403-ae2c-4e61-8fc3-e220ed1343e1.png)

PR #1591, opened 2022-03-18
Fix Crow Clock For Dark Mode and Fix Desktop Launcher Bangle.js 1 not showing white images.

Ronin0000:
#1532 
#1539


gfwilliams:
Great - thanks for sorting this!

PR #1595, opened 2022-03-20
add BangleJS 1 compatibity to the terminal clock app

Stiralbios:
Tested only in the emulator like I don't own the first BangleJS


Ronin0000:
I justed tested this on a actual bangle.js 1 and it works great!

gfwilliams:
Great, thanks! And thanks for testing @Ronin0000

PR #1596, opened 2022-03-20
Cycling App: Initial version

jsphpl:
From the README: This is a fork of the CSCSensor app using the layout library and separate module for CSC functionality. It also drops persistence of total distance on the Bangle, as this information is also persisted on the sensor itself. Further, it allows configuration of display units (metric/imperial) independent of chosen locale. Finally, multiple sensors can be used and wheel circumference can be configured for each sensor individually.

Decided to fork instead of changing the original version because of several significant changes:
- internal calculation based on SI units
- completely changed layout
- dropped crank / cadence support (for now)
- could not test on actual Bangle.js 1 hardware with real Bluetooth (therefore dropped support in metadata.json)

Question:
1. The BLESCS implementation is currently just uploaded as a separate file. Should it be maintained as a "module"?
2. Should i squash commits?

Link for quick testing: https://jsphpl.github.io/BangleApps/?q=cycling

CC @berkenbu @nxdefiant


gfwilliams:
Thanks! Merging now! Generally I'd try and avoid duplication as much as possible though.

> The BLESCS implementation is currently just uploaded as a separate file. Should it be maintained as a "module"?

Well, ideally I'd make the bluetooth comms a separate app and then have a 'dependency' on that app - for this *and* the original CSC app.

The reason is we could then have a 'settings' page where things like which CSC sensor to connect to could be supplied.

*Also* it could then provide a 'btcsc.recorder.js` file which would allow you to log CSC data alongside GPS, etc with the `recorder` app, which would be really neat.

> Should i squash commits?

No, they can be handy for debugging :)

PR #1603, opened 2022-03-21
touchtimer: ADD ability to repeat the timer

bhubbb:
This is an update to the touch timer app so that you are able to repeat the last run timer and not have to create a new one each time.


gfwilliams:
Thanks!

PR #1610, opened 2022-03-23
touchtimer: ADD 5 second count down buzzer

bhubbb:
- This is a configurable option (Default is off) for the timer to buzz
  each second when there are 5 seconds left.


gfwilliams:
Neat - thanks!

PR #1611, opened 2022-03-23
ClockCal: Configurable launch gestures

foostuff:
also removed a bug with the small calendar not coloring the current day.

### DIFF
--- a/apps/clockcal/ChangeLog
+++ b/apps/clockcal/ChangeLog
@@ -1,2 +1,3 @@
 0.01: Initial upload
 0.2: Added scrollable calendar and swipe gestures
+0.3: Configurable drag gestures
--- a/apps/clockcal/README.md
+++ b/apps/clockcal/README.md
@@ -9,25 +9,24 @@ I know that it seems redundant because there already **is** a *time&cal*-app, bu
 |![unlocked screen](screenshot2.png)|unlocked: smaller clock, but with seconds|
 |![big calendar](screenshot3.png)|swipe up for big calendar, (up down to scroll, left/right to exit)|
 
-
-
-
 ## Configurable Features
 - Number of calendar rows (weeks)
 - Buzz on connect/disconnect (I know, this should be an extra widget, but for now, it is included)
-- Clock Mode (24h/12h). Doesn't have an am/pm indicator. It's only there because it was easy.
+- Clock Mode (24h/12h). (No am/pm indicator)
 - First day of the week
-- Red Saturday
-- Red Sunday
-- Swipes (to disable all gestures)
-- Swipes: music (swipe down)
-- Spipes: messages (swipe right)
+- Red Saturday/Sunday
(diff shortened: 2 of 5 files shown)


gfwilliams:
Cool, thanks! Have you seen pattern launcher? It seems like this adds similar swipe functionality but to any clock app...

foostuff:
> Cool, thanks! Have you seen pattern launcher? It seems like this adds similar swipe functionality but to any clock app...

I appreciate the feedback. Pattern Launcher sure is great for launching apps. :) But there are two main things I wanted to be "extra quick" without the loading times of switching apps: Glance at the calendar and (soon) a quick way to control music. Thats why I opted for small modules that can't be started with the pattern launcher. (As far as I know.)
If the loading times could drastically improve, I'd put them in extra apps. But even the tiniest app takes a good second of loading time. Is there anything I could do to "instantly" load an app? Maybe even something stupid like putting everything in the boot code?

gfwilliams:
Ahh, sorry, I hadn't spotted from this PR that this was loading from modules.

> Is there anything I could do to "instantly" load an app?

Well, compacting storage will help if you seem to have abnormally bad load times, but otherwise I think mainly there are some things internally in Espruino that could be sped up a lot. For instance we can use the internal flash for storing JS, which should give a pretty big speed boost. Even the connection speed to the external flash can be boosted a lot.

Unfortunately there are a lot of conflicting things taking my time right now, and the loading speed is quite far down the list :(

To load from the clock quickly yourself, it may be possible to `eval(storage.readFile(...))` but you need to make sure you've torn down the clock stuff, and even then you have issues like what happens if the app doesn't want widgets but the clock has loaded them

foostuff:
> Ahh, sorry, I hadn't spotted from this PR that this was loading from modules.

I don't know if my use of "modules" is the right term :) [it's a js function to draw stuff.]

> Well, compacting storage will help if you seem to have abnormally bad load times

Unfortunately a 10-line-music-app seems to take "the same" time as the larger message-app. I guess unloading the watch is also part of the loading time.

> To load from the clock quickly yourself, it may be possible to `eval(storage.readFile(...))` ...

Thats a great idea. Not for this situation, but I'm sure it could come in handy for very special use cases.

Thanks for taking the time to reply/react to pull requests and forum posts! This watch, the github-based appstore, the overwhelmingly good documentation, the constructive community and your input ... what a game changer :)

gfwilliams:
Thanks! I'm glad you're enjoying it :)

>  I guess unloading the watch is also part of the loading time.

A little, but not much I think. I think the vast majority of the time is split between loading the boot code, app, and widgets...

If you run `t=getTime();Bangle.loadWidgets();print(getTime()-t);` you can see how long something (like loading widgets) takes to run. In my case it's 0.25s for widgets alone. I just tried and having the internal flash enabled drops that to 0.06s, so I think it's definitely something I'll look at enabling for later builds.

PR #1613, opened 2022-03-23
New app: 2047pp game

berkenbu:
A '2048' style game for the Bangle. Looks like someone beat me to the punchline with "Game1024" yesterday, but perhaps this one is different enough to warrant inclusion in the app loader (it does support both Bangle HW versions, on version 1 'BTN1' and 'BTN3' are used in lieu of swiping up/down).


gfwilliams:
Thanks! Looks great :)

Yes, amazing you both submitted within a few days of each other!

PR #1616, opened 2022-03-24
waypointer

nxdefiant:
Bangle.js 2 compatibility

recalculated layout to rescale on BangleJs2.

ping @hughbarney - if you want to take a look


gfwilliams:
Thanks, that's great!

PR #1619, opened 2022-03-25
Fix run notifications bugs and improve accuracy

GrandVizierOlaf:
This fixes a few issues with the `run` app, mostly ones I introduced with the notifications, but not exclusively. It also removes an unnecessary variable assignment in the `recorder` app.

* Don't try and enable/disable recording if the `recorder` widget is unavailable
* Don't configure notifications for stats which do not have a box enabled
* Look up the `vibTimes` rather than the `vibPatterns` to determine the current notification configuration
* Use the previous notification state to determine the next state, rather than the current stat (decreases notification drift)
* Reset GPS state in `reset` to prevent distance miscalculation when stopping one run and starting another

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -17,3 +17,4 @@
 0.11: Fix KML and GPX export when there is no GPS data
 0.12: Fix 'Back' label positioning on track/graph display, make translateable
 0.13: Fix for when widget is used before app
+0.14: Remove unneeded variable assignment
--- a/apps/recorder/metadata.json
+++ b/apps/recorder/metadata.json
@@ -2,7 +2,7 @@
   "id": "recorder",
   "name": "Recorder",
   "shortName": "Recorder",
-  "version": "0.13",
+  "version": "0.14",
   "description": "Record GPS position, heart rate and more in the background, then download to your PC.",
   "icon": "app.png",
   "tags": "tool,outdoors,gps,widget",
--- a/apps/recorder/widget.js
+++ b/apps/recorder/widget.js
@@ -248,7 +248,7 @@
       }
       var buttons={Yes:"yes",No:"no"};
       if (newFileName) buttons["New"] = "new";
-      var prompt = E.showPrompt("Overwrite\nLog " + settings.file.match(/\d+/)[0] + "?",{title:"Recorder",buttons:buttons}).then(selection=>{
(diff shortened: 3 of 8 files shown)


gfwilliams:
Thanks! Does the GPS state thing really help to fix a problem though?

It seems that because we measure distance as that between the current and last period, setting it to {} may actually have the effect of skipping out one second's worth of distance?

GrandVizierOlaf:
> Thanks! Does the GPS state thing really help to fix a problem though?
> 
> It seems that because we measure distance as that between the current and last period, setting it to {} may actually have the effect of skipping out one second's worth of distance?

If you stop a run and start another it would have the old GPS state and if you had moved at all it could cause the initial distance calculation of the second run to be wrong by the distance you had traveled between stopping and starting.

gfwilliams:
Do you actually see this happening though? Or is it just what you think might be an issue?

Because I believe GPS is on all the time so those should update every second regardless of whether you have started the run or not.

GrandVizierOlaf:
I see what you're saying and it might have then been a coincidence because I was having trouble with the initial distance on a second run and then it went away after making that change. Perhaps I had a better GPS fix by that point, though it had been around 30 minutes of consistent testing by then. I had assumed that when `run` told the `recorder` widget to `setRecording(false)` it would stop GPS updates to `exstats` too, but see now that wouldn't be the case because `exstats` is doing its own `setGPSPower`.

If the GPS is always on with exstats then does that mean leaving the watch on the `run` app screen will run the battery out faster than not having the app open, even when not running? The `run` app is my primary use of the watch and I leave it open a lot, even between runs.

gfwilliams:
It looks like the GPS 'fix' likely caused this issue? https://github.com/espruino/BangleApps/pull/1630

> If the GPS is always on with exstats then does that mean leaving the watch on the run app screen will run the battery out faster than not having the app open, even when not running?

Yes, that's the case. But it's important the GPS is turned on as soon as possible, because it can take a while to get a lock. Last thing you want it to start recording and then have it miss the first 10 minutes of your run

GrandVizierOlaf:
@gfwilliams I went back and tested and retested and could not duplicate the issue that caused me to introduce this breaking change, instead confirming the GPS fix is updating nearly once per second as you said, regardless of whether a run is active. Would you like me to revert it and/or the change from #1630 or leave things as-is?

gfwilliams:
Thanks for checking into this! I've just removed the `={}` bits - I guess #1630 is probably a good thing to have in anyway as *maybe* there's a case where you start the app and then start the run within 1 sec of boot - and you could hit the issue in that case :)

PR #1623, opened 2022-03-26
Magic the Gathering Watch Face

brantpeery:
New watchface that is MTG themed


gfwilliams:
Thanks! However I just checked the result of the test: https://github.com/espruino/BangleApps/runs/5717439702?check_suite_focus=true

And there's an issue with the app icon: [apps/mtgwatchface/app-icon.js](https://github.com/espruino/BangleApps/pull/1623/files#diff-69d6c50d0d4e3ac43529e6897665861c35c7bd74ce6ad7ed6bc11f27c372a8b4)

Basically you need to choose the 'image string' option in the image converter, not the 'image object' that you put in...

Also you don't need the ` "supports": [.. ]` lines for each storage file - because you're only supporting Bangle.js 2.

PR #1626, opened 2022-03-26
Quick Launch app

frigis1:
This app lets users tap or swipe up/down/left/right on their clock face to quickly launch up to five apps of their choice.


gfwilliams:
Nice - thanks!

PR #1627, opened 2022-03-27
Cosmetic update to Authenticator web interface

andrewgoz:
No version change - visual update to web interface only.


gfwilliams:
Thanks!

PR #1628, opened 2022-03-27
Fix run bug where recorder did not stop.

GrandVizierOlaf:
I introduced a bug in `run` version 0.11 after misreading an if statement and `recorder` would no longer stop at the end of a run. This fixes that.


gfwilliams:
I'm pretty sure this breaks the code. 

What if you don't have recorder installed? You'll get an exception from `WIDGETS["recorder"].setRecording` when stopping recording

BartS23:
No, it does not break the code because it is already covered in line 52
https://github.com/espruino/BangleApps/blob/master/apps/run/app.js#L52

gfwilliams:
Thanks! Sorry, didn't spot that :)

PR #1629, opened 2022-03-27
[RUN] Fix exception on Dist/Step/Time pattern selection in settings

BartS23:
Fix additional entry in new menu and exception in old menu
New Menu
Exception


gfwilliams:
Thanks!

PR #1630, opened 2022-03-27
[RUN] Fix.speed could be NaN

BartS23:
If you start the run before the first GPS fix, Speed may be NaN when the first fix occurs.


gfwilliams:
Thanks!

gfwilliams:
Looks like this was probably made worse by this PR from friday (https://github.com/espruino/BangleApps/pull/1619) since GPS always seems to report a value for speed even when there's no fix. To trigger otherwise I guess you'd have to start the app and click the button within 1 second

PR #1634, opened 2022-03-29
Improved Messages and Bluetooth widget

KungPhoo:
Messages:
* Option to unlock + lighten up the screen when a new message arrives.

Bluetooth:
* Forked widget 'widbt_notify'
* Vibrate when the connection is lost.
* Show short message about the lost connection

Clock:
* glbasic - a clock that is a copy/paste merge of the Pebble watchface and a 3-week calendar.


apps/messages/lib.js
```diff
@@ -57,15 +57,23 @@ exports.pushMessage = function(event) {
   // otherwise load messages/show widget
   var loadMessages = Bangle.CLOCK || event.important;
   // first, buzz
-  var quiet = (require('Storage').readJSON('setting.json',1)||{}).quiet;
-  if (!quiet && loadMessages && global.WIDGETS && WIDGETS.messages)
+  var quiet       = (require('Storage').readJSON('setting.json',1)||{}).quiet;
+  var unlockWatch = (require('Storage').readJSON('setting.json',1)||{}).unlockWatch;
```
gfwilliams, line 61: You're reading setting.json here but 'unlockWatch' seems to be written to `messages.settings.json` ?

gfwilliams:
Thanks for renaming the widget. I think I just noticed a bug in this - commented above.

Also, did you intend to include the `glbasic` clock app too? or should I take that out when merging this PR?

KungPhoo:
Yes. Please include the glbasic clock. It's a copy&paste production of the Pebble clock and a calendar widget.

gfwilliams:
Thanks - but looks like you're now using `messages.setting.json` when you should be using `messages.settings.json`

It does make me a little concerned that this isn't actually tested?

KungPhoo:
The settings bug is fixed and I put the bluetooth widget with vibration into a separate widget 'widbt_notify'.
Not sure if this is all right. I'm very new to 'github' and pull requests. This is the first project I actively support, so pardon me.

KungPhoo:
I tested as good as I can, but I didn't notice, that the options were not saved/restored in the first place. Is it enough to do a hard reset (hold button for 7 seconds) to see if the setting was written?
I tested toggling the setting, and it seemed to work good for me.
Pardon me. I didn't mean to cause you extra work.

gfwilliams:
No worries - and thanks for the contribution and fixes, this looks great!

'Messages' could do with the version number and changelog bumping, but I'll do that when I merge now

Actually just switching between apps reloads everything so that should be enough to check if a setting is saved :)

PR #1636, opened 2022-03-29
touchtimer: FIX countdown buzz to be in last 5 seconds

bhubbb:
- The countdown buzz was happening on each minute and not only on the
  last 5 seconds


gfwilliams:
Thanks!

PR #1640, opened 2022-03-31
Update metadata.json

numerist:
The word "digital" is added to distinguish this watch face from a coming similar analog face.

### DIFF
--- a/apps/doztime/metadata.json
+++ b/apps/doztime/metadata.json
@@ -1,9 +1,9 @@
 {
-  "id": "doztime",
-  "name": "Dozenal Time",
-  "shortName": "Dozenal Time",
+  "id": "dozdigtime",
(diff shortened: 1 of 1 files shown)


gfwilliams:
Hi - I'm afraid you can't just change the app ID to `"id": "dozdigtime",` while the app is still in the `doztime` folder.

Maybe just leave the ID as `doztime` and change the name/description?

Also the idea of `shortName` is it's shorter than `name` - or at least short enough to display in the watch's launcher. If they're going to be the same value you could just remove it, but I'd be concerned that `"Dozenal Digital Time"` wouldn't fit on the screen?

gfwilliams:
Ok, just done - and definitely make the Analog clock a whole new app, not just files under the existing one

PR #1647, opened 2022-04-02
Antonclk fix calweek substring

DDDanny:
bugfixed calendar week shown 3 digits (`substring` -> `slice`) furthermore replaced deprecated `substr` by `slice`

### DIFF
--- a/apps/antonclk/ChangeLog
+++ b/apps/antonclk/ChangeLog
@@ -9,3 +9,4 @@
       when weekday name and calendar weeknumber are on then display is <weekday short> #<calweek>
       week is buffered until date or timezone changes
 0.07: align default settings with app.js (otherwise the initial displayed settings will be confusing to users)
+0.08: fixed calendar weeknumber not shortened to two digits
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thank! But is there a reason for replacing substr with slice? As far as I can tell they both work, and `substr` is what tends to be used in other Espruino apps.

DDDanny:
Thanks for merge.
Reason for replacement:
I just got the deprecated warning from my VS Code and found it mentioned at

[Mozilla deprecated and obsolete features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Deprecated_and_obsolete_features#string_methods)

> String.prototype.substr()
> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

[Mozilla dev substr](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr)

PR #1651, opened 2022-04-03
[exstats/run] Include maxbpm as a stat that can be tracked

GrandVizierOlaf:
This has `exstats` keep track of the maximum BPM reached while active and emit a `changed` event every time it goes up. `run` makes this one of the options available to put in a box.


gfwilliams:
Great - thanks!

PR #1652, opened 2022-04-03
v0.09 added settings menu, removed symbols button, added highscore reset, clockmode optional

chiefdaft:
Including some bugfixes


gfwilliams:
Thanks!

PR #1653, opened 2022-04-04
Add README.md to metadata.json if exists

BartS23:
I have found several apps that have a README.md but do not mention it in the metadata.json.

So it is not shown in the apploader. This PR fixes the problem.


gfwilliams:
Great - thanks for sorting this out!

PR #1658, opened 2022-04-05
Added norwegian locale

hberland:
Added norwegian locale to the list, and also listed our three special characters æ, ø and å in the fallback list. They are all in the codebook, but I'm unsure if they will be rendered correctly with the current code (the two last days of the weeks contains the ø).


gfwilliams:
Thanks! Looks great - some of the built in fonts do have support for those characters, like 6x8:

PR #1659, opened 2022-04-06
New clock - Fuzzy Text Clock

sir-indy:


gfwilliams:
Looks great - thanks!

Have you seen the internationalisation support? https://github.com/espruino/BangleApps/issues/1311

It'd be possible just to include the JSON in your app, then add `/*LANG*/` before each string. Then the translations can be put in https://github.com/espruino/BangleApps/tree/master/lang and the language is adjusted based on the user's current language preference

sir-indy:
Thank you Gordon!

No, I hadn't seen the internationalisation support, that would be an ideal place to put it! I'll have a go at modifying the clock and lang files to use it.

To quote one of my favourite films (Fifth Element) 'I only speak two languages: English and bad English', so I can't guarantee the translations I've got are actually correct. They seem right but might not be how a native speaker would put it.

PR #1661, opened 2022-04-06
New and updated apps using the new "sched" library

peerdavid:
Hi, 

first of all sry for this huge PR -- as we discussed in my las PR I simply continued with development in my own fork. Now, as the really cool sched lib is available (thanks again for this cool and useful library), I think this fork is ready to merge it upstream. Therefore, this PR contains new as well as updated apps.

**New:**
- 90s Clock - The fullscreen mode is optional.

screenshot

- BlackWhiteClock -- If sched lib is installed, it provides a timer. If the sched lib is not installed, it still works, only the timer is disabled. The fullscreen mode is optional as well as the lock icon.

screenshot

Also the theme is considered and steps etc. can be shown (tab right/left). :
screenshot_2

- SimpleTimer - A simple GUI for a single timer using the sched lib. Especially useful for cooking etc as only a few tabs are needed to set the minutes. Nice also in combination with the pattern launcher :) Note: For this app the sched lib is necessary, therefore its included as a reference in the metadata.

screenshot

**Updated:**
- lcars -- I replaced the custom timer with sched lib. If the sched lib is not installed, this functionality is disabled. The app still works.
- NotAnalog -- Same as for lcars.

Best, David


gfwilliams:
Excellent - thanks! Just checking - did `"dependencies": {"scheduler":"type"},` work for you? I notice you used `sched:app`.

The difference is that if in the future someone *were* to make their own `sched` fork (i know, i know) then at least the app would be able to use that one and wouldn't try and auto-install `sched`

peerdavid:
So what I tried is that I uninstalled the sched lib and installed the simpltmr app. Then first the sched lib is installed and afterwards the app itself.

If I understand you correctly it is possible to simply say "hey I need some kind of sched, independent of which app provides this functionality" -- is this right? If so I think that's cool but I'm not 100% sure how I can change the dependency accordingly...

Thanks a lot :)

peerdavid:
Ah sry I got it -- changed the dependency to sched:type...

gfwilliams:
I think it needs to be 'scheduler' - but I'll tweak that here :)

peerdavid:
Ah great thanks a lot :) So the interface is "scheduler" and the app that implements it is "sched" right?

gfwilliams:
Yep. It's a bit confusing. But basically if you set `type:"xyz"` in an app's metadata.json you can  do `"dependencies": {"xyz":"type"}` in another app. And that's what we've done here (i hope!)

PR #1662, opened 2022-04-06
Rework word lookup algorithm to use bisection, get rid of index file.

berkenbu:
Speed up word look ups by using bisection search of alphabetically sorted word list. This also removes the need for an index file.


gfwilliams:
Thanks!

PR #1663, opened 2022-04-06
lightswitch: masking (espruino/Espruino#2151) + oversize

storm64:
@gfwilliams Utilizing the issue espruino/Espruino#2151 to stop the corresponding event handlers works very nice on my own watch for the last two month, but I don't know if it is "good manners".

Update ChangeLog and metadata.json
Update README.md and settings.json
 - add oversize setting + description
Update settings.js
 - add oversize setting
 - correct default value of touchOn according to value in README.md
Update widget.js
 - add oversize setting
 - correct default value of touchOn according to value in README.md
 - add masking touch and drag input by adding own event listeners first and messing up the handler on a widget related event, using espruino/Espruino#2151


gfwilliams:
Hi, I'll merge this but...

> Utilizing the issue https://github.com/espruino/Espruino/issues/2151 to stop the corresponding event handlers works very nice

I'm pretty iffy about this, as it really is taking advantage of a bug. When it finally gets fixed, that functionality in your app will stop working. Not sure if there's a way I could implement the feature nicely in Espruino (doing something based on the return value?).

PR #1665, opened 2022-04-07
locales: Fixed day order for no_NB

zerodogg:
The data had monday as the first day, while the BangleJS seems to assume the first day is sunday. My Bangle claimed today (7. april) was friday (it's thursday) after installing the no_NB locale.

As an aside, what kind of format is used for the locale code? The POSIX locale code for "Norwegian Bokmål" is nn_NO, while the locale code for "Norwegian Nynorsk" is nn_NO, whereas this file is using no_NB.


gfwilliams:
Thanks! Yes, ` no_NB` is wrong - I've just fixed it :)

PR #1666, opened 2022-04-07
Added Norwegian Nynorsk translation

zerodogg:
Norway has two written languages, Bokmål and Nynorsk. This adds the Nynorsk variant.

I was uncertain which language code to use.

nn_NO is the POSIX identifier, used in Linux et.al. so this pull request uses that, however as mentioned in https://github.com/espruino/BangleApps/pull/1665 the language code that has been used for Norwegian Bokmål is "no_NB" (ie. the other way around, country_LANGUAGE rather than language_COUNTRY). Browsers similarly use "nn-NO" for nynorsk, and "nb-NO" for bokmål.

### DIFF
--- a/apps/fuzzyw/fuzzy_strings.json
+++ b/apps/fuzzyw/fuzzy_strings.json
@@ -114,6 +114,29 @@
     ],
     "text_scale":3.5
   },
+  "nn_NO":{
+    "hours":[
+      "tolv", "eitt", "to", "tre", "fire", "fem",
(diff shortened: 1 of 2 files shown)


gfwilliams:
I think nn_NO is right.

>  the language code that has been used for Norwegian Bokmål is "no_NB"

This looks wrong. We assume language_COUNTRY for putting the flag by the items in the drop-down list, so I guess that need changing

gfwilliams:
Thanks - just merging, and I'll tweak the Bokmål one

PR #1669, opened 2022-04-08
sched: Fix `getTimeToAlarm` for timers

storm64:
Only set `last` for alarms to prevent a wrong output of `getTimeToAlarm` for a timer already used at same day.


gfwilliams:
Thanks!

However am I right in thinking that because 'last' is reset to 0 in the `alarm` app whenever you edit a timer, this doesn't actually affect how timers work - it's more an issue if you've got a custom app using them that doesn't change `last`?

PR #1670, opened 2022-04-08
Swedish abbreviations for speed

johandalabacka:
The Swedish abbreviation for speed is km/tim (even if km/h often is used) but NOT kmh


gfwilliams:
Thanks!

PR #1673, opened 2022-04-08
Let language_scan optionally add stubbed translations

zerodogg:
This adds an environment variable that, when set, will add stubbed translations for any strings not currently present in the files, so that it can be updated without using the auto-translation feature.

### DIFF
--- a/bin/language_scan.js
+++ b/bin/language_scan.js
@@ -3,6 +3,9 @@
 outputs a list of strings that have been found.
 
 See https://github.com/espruino/BangleApps/issues/1311
+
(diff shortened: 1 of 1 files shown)


zerodogg:
Updated the pull request, now also includes a simple command-line parameter parser. The old ENV method for DEEPL still works, but the env variable for auto-updating stubbed translations has been removed in favour of --refresh. It'll respond to --help so the script becomes a bit easier to figure out.

gfwilliams:
Looks excellent - thanks!

PR #1674, opened 2022-04-08
New app "OpenWind"

berkenbu:
App that connects to an OpenWind bluetooth sailing (I guess one could use it as an expensive weather station, too) instrument.


gfwilliams:
This looks great - thanks!

PR #1675, opened 2022-04-08
sched: `getTimeToAlarm` check next dow if t<now

storm64:
Fix undefined output of `getTimeToAlarm` on an alarm that is already in the past to check for the next dow instead of the actual dow.
 - if alarm.t < currentTime, (time.getDay() + 1) % 7


gfwilliams:
Thanks!

PR #1677, opened 2022-04-09
Noteify app

frigis1:
Write notes using an onscreen keyboard and use them as custom messages for alarms or timers.


gfwilliams:
Looks great - thanks!

PR #1678, opened 2022-04-09
New app. Golf View

jdekarske:
This app aims to fill the same functionality as golf specific smartwatches. It uses OpenStreetMap to fetch course geometry and measure real time distance to golf holes.


gfwilliams:
Wow, this looks awesome! I'm not a golf player but I love the idea of dragging all the course info off openstreetmap. There must be other sports that could do stuff like this too

PR #1679, opened 2022-04-09
sleepphasealarm

nxdefiant:
- Read alarms from new scheduling library
- account for higher acceleration sensor noise on Bangle.js 2

### DIFF
--- a/apps/sleepphasealarm/app.js
+++ b/apps/sleepphasealarm/app.js
@@ -125,9 +125,9 @@ if (nextAlarm !== undefined) {
 
   // minimum alert 30 minutes early
   minAlarm.setTime(nextAlarm.getTime() - (30*60*1000));
-  setInterval(function() {
+  Bangle.on('accel', (accelData) => { // 12.5Hz
     const now = new Date();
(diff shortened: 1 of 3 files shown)


apps/sleepphasealarm/app.js
```diff
@@ -38,13 +38,20 @@ function calc_ess(val) {
   }
 }
 
+// copied from alarm app
+// time in ms -> { hrs, mins }
+function decodeTime(t) {
```
nxdefiant, line 43: I have seen this function in alarm/app.js and
sched/sched.js. Should probably be moved to
library

nxdefiant:
Yes, extra commit in this PR is intentional

apps/sleepphasealarm/app.js
gfwilliams: Thanks - yes, that's a good idea. Do you want to do a PR for it?

gfwilliams:
Thanks for switching to `sched`!

apps/sleepphasealarm/app.js
nxdefiant: I can try that later

PR #1681, opened 2022-04-09
Update apps/kbswipe/lib.js (Swipe keyboard)

thyttan:
@gfwilliams

Change the way the display is updated as to keep user input trace intact until they lift their finger from the screen. 

Now only the marker is updated every second as opposed to the whole screen. To achieve this the marker is reimplemented outside the wrapString-method with g.fillRect() and g.clearRect() as well as a newly introduced function findMarker().


gfwilliams:
Thanks!

PR #1683, opened 2022-04-10
GBalarm - sync GB alarms with sched.json

frigis1:
Is there a need for an app that syncs Gadgetbridge alarms, or is it already being developed?


gfwilliams:
Thanks! This looks like a good idea - but honestly I'd just move it into the `Android Integration` app itself. Do you see a problem with that? I know you do have the separate app for viewing Gadgetbridge alarms, but do you think that is needed or could the default `Alarm` one be used?

frigis1:
Adding it to Android Integration makes sense since it handles Gadgetbridge related stuff. Should I add it to Android.boot.js? As for the app for viewing alarms, do you mean Noteify? I copied and pasted the default alarm apps simply for the convenience of editing or removing alarms created by Noteify without leaving the app. Would it be better to let the default alarm app handle that instead?

PR #1684, opened 2022-04-10
[dtlaunch, b2] Change swipe direction to UpDown

thyttan:
@jeffmer

On bangle 2 change swipe directions to move between pages from **_LeftRight_** to **_UpDown_**, as to match the page indicators. Implemented like 'natural scrolling'.

Also added function **_showClock()_** which is called on **_'left-to-right' swipe_**.  Here I imagine the clock to be to the left of the launcher, if you do that then this also follows the natural scrolling principle.


jeffmer:
Looks good to me - originally up down was not available so this change is a definite usability improvement.

gfwilliams:
Thanks!

PR #1685, opened 2022-04-10
[BW Clock] Minor improvements

peerdavid:
I made some minor improvements including a shorter name (its better to read in the settings in my opinion) as well as a better positioning of the date in fullscreen mode.


gfwilliams:
Thanks!

PR #1688, opened 2022-04-10
Tagging bootloader messages, and all "Loading..." messages

zerodogg:
Incidentally, where does the "Loading..." message that is displayed when pressing btn1 on bangle2 when the clock is displayed come from?

### DIFF
--- a/apps/accellog/app.js
+++ b/apps/accellog/app.js
@@ -29,7 +29,7 @@ function showMenu() {
 }
 
 function viewLog(n) {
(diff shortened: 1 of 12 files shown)


gfwilliams:
Thanks! The `Loading...` message comes from inside the Bangle.js firmware I'm afraid. It's a good point though - some were asking about disabling it a while back but maybe it needs to be able to be localised (or even customised as an image)...

PR #1699, opened 2022-04-13
Attempt to add font scaling for text label to layout lib

KTibow:
This just attempts to allow `scale` to be specified for a text label, which is passed as the second parameter to `g.setFont`. See diff.

### DIFF
--- a/modules/Layout.js
+++ b/modules/Layout.js
@@ -30,7 +30,8 @@ layoutObject has:
 
 * A `type` field of:
   * `undefined` - blank, can be used for padding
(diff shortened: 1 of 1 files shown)


gfwilliams:
Hi - thanks, but this is actually not required now. If you check out http://www.espruino.com/Reference#l_Graphics_setFont you can include the scale inside the font name itself.

```
g.setFont("Vector:12") - vector font 12px high
g.setFont("4x6:2") - 4x6 bitmap font, doubled in size
g.setFont("Vector12") - vector font 12px high
g.setFont("4x6",2) - 4x6 bitmap font, doubled in size
```

KTibow:
Ah, nice!

PR #1700, opened 2022-04-13
Choozi Support Bangle.js 2

Ronin0000:


gfwilliams:
Thanks! Out of interest, how much changed between Bangle.js 1/2? Is it something where it might make sense to make just one sourcefile for both?

PR #1701, opened 2022-04-13
Allow quiet mode to optionally override message auto-open

zerodogg:
The messages app will auto-open messages (but not buzz) even when the watch is in quiet mode. This adds a setting that disables this behaviour.


gfwilliams:
Nice idea - thanks! Please could you just bump the version number and add something to the ChangeLog?

zerodogg:
Done

gfwilliams:
Thanks!

PR #1702, opened 2022-04-13
[New app] [Messages Music] Pushes message to open Messages music controls.

thyttan:
Hacky app that uses Messages app and it's library to push a message that triggers the music controls from within Messages itself.

Making the music controls accessible this way lets one start a music stream on the phone in some situations even though the message app didn't receive a music message from gadgetbridge to begin with. (I think.)


gfwilliams:
Thanks! Nice idea!

Issue #1577, opened 2022-03-15
[Feature Request] Change LCARS clock theme colors


Ronin0000:
[Feature Request] Because the LCARS clock has three main theme colors it would be cool if you could change those colors in the settings of the app.

peerdavid:
Would be cool to keep the original colors as default. Also currently parts of the watch-face are drawn using a background image so you have to redraw those round corners etc. manually...

PR #1703, opened 2022-04-13
LCARS clock custom theming 

Ronin0000:
#1577 
@peerdavid 
EDIT: Fixed the bug


peerdavid:
Hello,

I just tested this and its not properly working on my bangle, so I'm not sure if its a problem on my side or a bug: I updated the app from the source repo as well as the settings of lcars. I changed the color to red and got the following:
screenshot

It seems that the image is not rendered correctly. This is my settings file:

`{"alarm":-1,"dataRow1":"Battery","dataRow2":"Steps","dataRow3":"Temp","speed":"kph","fullscreen":false,"themeColor1BG":"#f00","themeColor2BG":"#FF00DC","themeColor3BG":"#0094FF"}`

Best, 
David

peerdavid:
Update: If I select green, I get the following: 
screenshot

peerdavid:
At least on my bangle its not working correctly as the color of the images (left) is different than the lines (horizontal). Also sometimes no images are shown. So I would suggest to fix this first.

Ronin0000:
Hi, sorry about this I thought I had tested it. I will close this for now and open it when I have fixed it.

Ronin0000:
Looks like it had to do with how I defined some of the colors in the settings page I did not realize that all the hex coeds had to be 6 digits long and you can see that the red green and blue colors are not the same as the other ones.
image

Ronin0000:
Now looks like it is working 
image

PR #1705, opened 2022-04-14
[Pebble Clock] Support for internationalization / Get steps from built-in step counter

### LINKED ISSUE #1697: [pebble] request: using built-in step counter #1696
Would that be possibile to have this app use the built-in step counter (Bangle.getHealthStatus("day").steps) and not the widget?

In this way we could do withouth the pedometer widget and save some memory (especially useful on Bangle JS 1).

alessandrococco:
Closes #1697


gfwilliams:
Thanks!

PR #1707, opened 2022-04-15
[Scheduler] New functions & Settings page

alessandrococco:
_edit: This first message is obsolete_

Unlock the watch at buzz so the user can interact with the prompt immediately (it also reduces the use of the button :-D)

PS: This could be an option but I don't know where to put it:

1. in a new system setting (like the quiet mode)
2. in the Scheduler-owned settings (sched/settings.js)
3. Other?

### DIFF
--- a/apps/sched/README.md
+++ b/apps/sched/README.md
@@ -53,21 +53,27 @@ use too much RAM.
 It can be used as follows:
 
 ```
-// add/update an existing alarm
+// Get a new alarm with default values
+let alarm = require("sched").newDefaultAlarm();
+
+// Get a new timer with default values
+let timer = require("sched").newDefaultTimer();
+
+// Add/update an existing alarm
 require("sched").setAlarm("mytimer", {
   msg : "Wake up",
-  timer : 10*60*1000, // 10 Minutes
+  timer : 10 * 60 * 1000 // 10 minutes
 });
 // Ensure the widget and alarm timer updates to schedule the new alarm properly
 require("sched").reload();
 
 // Get the time to the next alarm for us
-var timeToNext = require("sched").getTimeToAlarm(require("sched").getAlarm("mytimer"));
-// timeToNext===undefined if no alarm or alarm disabled
+let timeToNext = require("sched").getTimeToAlarm(require("sched").getAlarm("mytimer"));
+// timeToNext === undefined if no alarm or alarm disabled
 
-// delete an alarm
+// Delete an alarm
 require("sched").setAlarm("mytimer", undefined);
-// reload after deleting...
+// Reload after deleting
 require("sched").reload();
 
 // Or add an alarm that runs your own code - in this case
@@ -76,12 +82,15 @@ require("sched").reload();
 require("sched").setAlarm("customrunner", {
   appid : "myapp",
   js : "load('setting.app.js')",
-  timer : 1*60*1000, // 1 Minute
+  timer : 1 * 60 * 1000 // 1 minute
 });
 
 // If you have been specifying `appid` you can also find any alarms that
 // your app has created with the following:
-require("sched").getAlarms().filter(a=>a.appid=="myapp");
+require("sched").getAlarms().filter(a => a.appid == "myapp");
+
+// Get the scheduler settings
+let settings = require("sched").getSettings();
 ```
 
 If your app requires alarms, you can specify that the alarms app needs to
--- a/apps/sched/README.md
+++ b/apps/sched/README.md
@@ -8,8 +8,17 @@ Other apps can use this to provide alarm functionality.
 App
 ---
 
-The Alarm app allows you to add/modify any running timers.
+The **Alarms & Timers** app allows you to add/modify any running alarms and timers.
 
+Global Settings
+---------------
+
+- `Unlock at Buzz` - If `Yes` the alarm/timer will unlock the watch
+- `Default Auto Snooze` - Default _Auto Snooze_ value for newly created alarms (_Alarms_ only)
+- `Default Snooze` - Default _Snooze_ value for newly created alarms/timers
+- `Buzz Count` - The number of buzzes before the watch goes silent
+- `Buzz Interval` - The interval between one buzz and the next
+- `Default Alarm/Timer Pattern` - Default vibration pattern for newly created alarms/timers
 
 Internals / Library
 -------------------
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -18,3 +18,4 @@
 0.17: Moving alarm internals to 'sched' library
 0.18: Cope with >1 identical alarm at once (#1667)
 0.19: Ensure rescheduled alarms that already fired have 'last' reset
+0.20: Use the new 'sched' factories to initialize new alarms/timers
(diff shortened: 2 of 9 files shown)


alessandrococco:
I thought about it for a moment. I'd create a Settings menu for Scheduler. It will contain some global scheduler options as 

- Unlock at buzz Yes/No
- Auto-snooze minutes (default is 10 minutes, [hard-coded](https://github.com/espruino/BangleApps/blob/3eed3749c19a4f0bb30fac14e8e0427b3bc1121c/apps/sched/sched.js#L41))

I'd also add an option for

- The buzz count (default is 10, [hard-coded](https://github.com/espruino/BangleApps/blob/3eed3749c19a4f0bb30fac14e8e0427b3bc1121c/apps/sched/sched.js#L33))
- The buzz interval (default is 3s, [hard-coded](https://github.com/espruino/BangleApps/blob/3eed3749c19a4f0bb30fac14e8e0427b3bc1121c/apps/sched/sched.js#L58))

storm64:
Adding this options for more customization would be realy nice.
I also thought about a settings menu for sched and would like to add some more options in the future, e.g.:
 - Save on Back (and replace the `Save   >` with `Cancel   >`)
 - Change default value of `Vibrate   ..`
 - Change default value of `Auto snooze   No`

alessandrococco:
> Adding this options for more customization would be realy nice.

I'm working on it!

image

image

> I also thought about a settings menu for sched and would like to add some more options in the future, e.g.:
> 
> * En-/Disable Save on Back (and replace the `Save   >` with `Cancel   >`)
> * Change default value of `Vibrate   ..`
> * Change default value of `Auto snooze   No`

Interesting! Do you mind if I work on it? :-)

alessandrococco:
Quick update:

image image

I'm also updating the Alarms & Timers app to read the new settings.

alessandrococco:
Done! I'm not 100% sure about the menu labels. Are *Unlock at Buzz*/*No. Buzz*/etc. understandable? Would you suggest better labels?

storm64:
> Interesting! Do you mind if I work on it? :-)

Feel free, I like what you already did. ;)  
Would you like to implement the `- En-/Disable Save on Back (and replace the Save > with Cancel >)`, too?

> Done! I'm not 100% sure about the menu labels. Are _Unlock at Buzz_/_No. Buzz_/etc. understandable? Would you suggest better labels?

Looks realy nice.  
1. In my opinion the internal `buzzCount` might be more catchy:  
_No. Buzz_ -> _Buzz Count_

2. I would like to highlight that _Auto Snooze_ and the _Patterns_ are changing the default values:  
_Auto Snooze_ -> _Default Auto Snooze_  
_Alarms Pattern_ -> _Default Alarm Pattern_  (reads more fluently without the s's)
_Timers Pattern_ -> _Default Timer Pattern_

---

One big concern that I haven't thought about in the first place is, that we combine the settings of two different apps:
 - __sched settings:__
   - _Unlock at Buzz_
   - _Snooze_
   - _No. Buzz / Buzz Count_
   - _Buzz Interval_
 - __alarm settings:__
   - _Auto Snooze / Default Auto Snooze_
   - _Alarms Pattern / Default Alarm Pattern_
   - _Timers Pattern / Default Timer Pattern_
   - (_Save on Back_)

On the one hand I like to have all settings in one place and even on a forked version of the alarm app this settings could come in handy. But on the other hand it doesn't feel right.

---

alessandrococco:
Thank you!

> One big concern that I haven't thought about in the first place is, that we combine the settings of two different apps [...]
> On the one hand I like to have all settings in one place and even on a forked version of the alarm app this settings could come in handy. But on the other hand it doesn't feel right.

Yes, you're right. The "problem" here is that "sched" is a library, not an app. An alarm/timer could be created from the default alarm app but also from other apps and every app should be able to read/update the alarms/timers created from others. IMHO we should avoid to duplicate the same settings: there is no need to have a "Default Auto Snooze" implemented in 10 different alarm/timer apps :-D

Probably I need to make the *This is a default for all new alarms/timers* part clearer so I like your suggestion to add a "*Default*" prefix to the labels.

alessandrococco:
> Would you like to implement the `- En-/Disable Save on Back (and replace the Save > with Cancel >)`, too?
> 

Sorry @storm64 , I missed this part. Could you elaborate? I don't think I understand this "Save on back" thing 🤔

alessandrococco:
> Maybe the 'default' alarm/timer state (eg the object at the start of editAlarm) should be a function in the library, so anything implementing an alarm gets the same default values without them having to be hard-coded?

Good idea. I'll export 4 new functions:

- newDefaultAlarm (or newAlarmDefault?)
- newDefaultTimer
- get/setSettings

gfwilliams:
Great - thanks! Just in general I think it's good to try and keep the exported function count for libraries as low as possible (as once you use a library all the functions in it gets loaded and it uses more RAM) but this all looks good.

Do you think it's ready to merge?

alessandrococco:
Thanks!

> Do you think it's ready to merge?

Almost ready! Before the merge I want to rebase & clean up my commits and add some documentations about the new functions :-)

alessandrococco:
> When I change the config (mostly en-/disable) of an alarm or timer, I don't want to navigate to the `Save >`-button. I would prefer to just return with the _back_-icon in the top left corner and have the changes saved. If enabled the "Save on Back"-option should switch the behavior of the menu entries `Back >` <-> `Save >` and rename the `Save >` to `Cancel >`.

Understood! It would be useful to me too: I have often configured a timer without saving it 😅

This setting belongs to the alarm app only. I'll open a new PR!

storm64:
> - newDefaultAlarm (or newAlarmDefault?)

I wouldn't change it, the default as prefix groups alarm and timer together.


> This setting belongs to the alarm app only. I'll open a new PR!

Good point.


> > Do you think it's ready to merge?
>
> Almost ready! Before the merge I want to rebase & clean up my commits and add some documentation about the new functions.

I merged your branch into my own fork for testing and realy like the settings and the QOL improvement.

alessandrococco:
@gfwilliams: Is saving all the settings after each selection a best practice? Or should I save the settings only once?

apps/sched/README.md
```diff
@@ -53,7 +62,13 @@ use too much RAM.
 It can be used as follows:
 
 ```
-// add/update an existing alarm
+// Get a new alarm with default values
+let alarm = require("sched").newDefaultAlarm();
+
+// Get a new timer with default values
+let timer = require("sched").newDefaultAlarm();
```
storm64, line 69: You should change .newDefaultAlarm() to .newDefaultTimer() ;)

gfwilliams:
> Is saving all the settings after each selection a best practice? 

I think that's fine... If nothing has changed Espruino is smart enough not to rewrite the file, and someone may break out with a long press and be surprised their change didn't take effect otherwise

alessandrococco:
> Do you think it's ready to merge?

Yes! 😄

gfwilliams:
Great!

alessandrococco:
@gfwilliams In the alarm app code there is this comment:

image

Out of curiosity what does it mean?

gfwilliams:
There's an issue with object/array indices that include a `\0` character (which is used for inline images). It's *possible* the comment is out of date, because when I looked into it basically everything after the `\0` got ignored:

```
a = {};
a["\0foo"] = 1;
a["\0bar"] = 2;
a == { "\0foo": 2 } // !!!
```

So actually the real workaround was to just put the images at the end of the string rather than the beginning

alessandrococco:
@storm64  I just implemented the "Save on Back", you can find it in [this branch](https://github.com/alessandrococco/BangleApps/tree/alarms-save-on-back) - Feel free to do some tests, a feedback is appreciated!

image

PR #1708, opened 2022-04-16
lang: added Croatian locale

wespen:
Added Croatian translation and locale.  
Let me know if I missed something.


gfwilliams:
This looks great - thanks! I think you need to add the new file to `lang/index.json` but I'll do that now anyway :)

PR #1709, opened 2022-04-18
Add HeartZone app

Uberi:
I wanted something similar to what Workout HRM does, but with these features:

* Larger text, more contrast, and color-coding for better readability while exercising.
* Configurable buzz interval, instead of at every heart rate reading (which was too distracting).
* Pause for a rest and resume afterwards without having to restart the heart rate sensor (which takes several seconds each time to stabilize).
* Configure the minimum heart rate confidence threshold (bad readings cause buzzes that have to be ignored).

To do this I needed to write a separate app. This app is HeartZone.

You can see it live over at: https://anthony-zhang.me/BangleApps/?q=heartzone

P.S., the Bangle.js development experience has been one of the smoothest I've ever seen - congrats on making such a great set of tools!


gfwilliams:
Looks great, thanks! Glad development worked so well for you!

You could even add calls to the `recorder` app (if installed) when you start/stop I guess (a bit like `run` does) so you could record your HRM (and maybe GPS/etc) over time in case you wanted to replay it.

PR #1713, opened 2022-04-18
widcal: fix memory leak

rigrig:
Repeated redrawing would create multiple redraw-timeouts


gfwilliams:
Thanks for tracking this down!

PR #1715, opened 2022-04-19
Another pedometer widget

sir-indy:
Yes, I know, another one. But this one is mine, it's as narrow as I can make it, existing step counters took up half the widget bar once they got up to 5 digits. This one counts in steps of 0.1k.

I've also added a step goal into the health app. Not sure if this is the best way to do this, or if I should add a separate settings page for this widget, but I saw it listed as a TODO so I did it.


sir-indy:
Thank you for the feedback!

Not removed, but moved. That was one of the things I wasn't sure about, having the setting for the step goal and the `Heart Rt` together in a settings menu, or separately under `menuStepCount `and `menuHrm`. To avoid confusion I'll move 'Heart Rt' back to the settings menu, and put Daily Step Goal in there too.

Good point with the `loadSettings` function, I like your idea to avoid that. I should have spotted that myself, I spent a long time chasing a bug from having `settings` instead of `wsSettings`, then that being overwritten by something elses `settings` when I  went into certain menus.

I'll make the changes and try again.

sir-indy:
Changes made as suggested, but do not merge, something really funky is going on with the settings.

Reading out the settings after changing things, it seems to be picking up a different file (not health.json?), containing `"scroll":-24`, and adding the changed setting to that. Then, change the other setting, and it forgets about the previous setting. Question is, what file? I can't find any talking about scroll -24? I'll keep investigating...
```
{ "scroll": -24, "hrm": 1 }
{ "scroll": -24, "hrm": 0 }
{ "scroll": -24, "stepGoal": 10000 }
{ "scroll": -24, "stepGoal": 10000 }
{ "scroll": -24, "stepGoal": 9800 }
{ "scroll": -24, "hrm": 1 }
{ "scroll": -24, "stepGoal": 10000 }
```

gfwilliams:
> Not removed, but moved

Ahh sorry, not sure how I didn't spot that!

>  it seems to be picking up a different file

That's odd. 'scroll' will be from the variable returned by `E.showMenu` or `E.showScroller` I guess, but I don't know how that's happening.

Does it happen if you have minification turned off?

sir-indy:
I don't think I had minification turned on. It still happened when I loaded the file in from the ide, with a small mod to print the settings out whenever they were changed, which is how I got the list above.

I still don't know why it happens, but the last change stopped it happening: `var s=getSettings();` -> `var healthSettings=getSettings();`
I've noticed the same thing has happened with the settings for my mosaic clock, 'scroll -24' has turned up there too, and I used `s` as a reference to the settings object.

Does `s` have some special relevance to `E.showMenu` or `E.showScroller`?

gfwilliams:
So I guess this is ready to merge now?

sir-indy:
Mystery solved! Excellent, thank you!
Yes, I'm happy if you're happy.

gfwilliams:
Great!

PR #1719, opened 2022-04-20
Allow timeouts to run on the message list screen

zerodogg:
Currently, when a new message is delivered, the messages app starts a timeout timer that returns to the clock after a while. If the message being displayed is removed without user interaction (for instance by the user dismissing a notification on their phone), the app returns to the message list and stops the timer. The result is that if a message is received and then before the timeout is dismissed on the phone, the device will permanently display the message list until the user manually returns to the clock.

This change lets the timer keep on running as long as the user doesn't interact with the screen, restoring the clock as usual.

One possible improvement to this change (that I haven't attempted to write) could be to clear the timeout if the screen gets unlocked by the user while on the messages screen.

### DIFF
--- a/apps/messages/ChangeLog
+++ b/apps/messages/ChangeLog
@@ -45,3 +45,4 @@
 0.30: Add new Icons (Youtube, Twitch, MS TODO, Teams, Snapchat, Signal, Post & DHL, Nina, Lieferando, Kalender, Discord, Corona Warn, Bibel)
 0.31: Option to disable icon flashing
 0.32: Added an option to allow quiet mode to override message auto-open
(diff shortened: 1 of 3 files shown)


gfwilliams:
Ok, thanks - we'll see if anyone complains :) I guess a timeout of 60 secs if you leave the watch of that screen seems fine - although as you say even if you're scrolling up and down it, you'll still end up reloading the clock after 60 sec

zerodogg:
Yep, if that ends up being an issue I won't mind adding something that checks locked/unlock status before returning (that ought to solve that usecase for most people I think) or just adding a setting for it, so just @ me in the issue if it comes up.

PR #1720, opened 2022-04-20
[LCARS] Bugfixes and other improvements

peerdavid:
The alarm control (of the sched lib) was broken after the last update. It is fixed now. Also `Bangle.getHealthStatus("day").steps` is used as fallback if no widget is installed. I still use the steps of widgets to ensure that the behavior does not change with this update.


gfwilliams:
Thanks!

PR #1721, opened 2022-04-20
Update .gitignore

alessandrococco:
I run the App Loader locally and I noticed that the `_site` folder wasn't in .gitignore. This simple PR fix that!


gfwilliams:
Thanks! Actually now I personally just host the folder directly - the app loader will now manually scan every app folder is `apps.json/apps.local.json` wasn't created - it's not too slow when hosting off localhost :)

PR #1722, opened 2022-04-20
Refactor some methods to scheduling library

nxdefiant:
As communicated in #1679, moved a few functions to scheduler lib.

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -19,3 +19,4 @@
 0.18: Cope with >1 identical alarm at once (#1667)
 0.19: Ensure rescheduled alarms that already fired have 'last' reset
 0.20: Use the new 'sched' factories to initialize new alarms/timers
+0.21: Refactor some methods to scheduling library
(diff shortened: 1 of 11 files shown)


nxdefiant:
@frigis1 You might want to update noteify app if this PR gets ok.

alessandrococco:
Aaaarg, I also just pushed an alarms app refactor  🤣

frigis1:
@nxdefiant Thanks for the heads up! So if I understand correctly, instead of copying the decode/encodetime functions from the default alarm, I can use the sched library to do do that?

nxdefiant:
Yes

gfwilliams:
Thanks - looks good - but...

What's with the [css/main.css](https://github.com/espruino/BangleApps/pull/1722/files#diff-f131fc9dca7d0116bd2258c95e127905bb0db3b922c83db9c7648b32770fc532) change? `word-break: break-word;` would appear to basically be the default, and it seems to word break already for me just fine?

I'm just iffy about pulling in CSS changes like this after the last one broke app layout.

nxdefiant:
Sorry, the css should not have been commited (git commit -a accident).

gfwilliams:
Thanks!

PR #1726, opened 2022-04-21
[fuzzyw] Moving translations for Fuzzy Work Clock

sir-indy:
As Gordon suggested in #1659 , moved the translations from a custom file into the standard lang files, as per #1311 
Removed the settings as they weren't really needed anymore.


gfwilliams:
Excellent - thanks! It's also great to have text translations for numbers in a bunch of different languages now :)

PR #1728, opened 2022-04-21
[Sched] Correct decodeTime(t) rounding

storm64:
Correct `decodeTime(t)` to return a more likely expected time:
- before:  
  ```
  >require("sched").decodeTime(60*60*1000-1)
  ={ hrs: 0, mins: 60 }
  ```
- after:   
  ```
  >require("sched").decodeTime(60*60*1000-1)
  ={ hrs: 1, mins: 0 }
  ```


gfwilliams:
Thanks!

PR #1730, opened 2022-04-21
dtlaunch, make left to right swipe exit optional

hughbarney:


gfwilliams:
Thanks for the flipper tweaks too!

PR #1731, opened 2022-04-22
[hebrew calendar] major renovation

barakplasma:
This PR contains the (now debugged) major overhaul of the Hebrew Calendar app.

One of the biggest changes is the new customizer, which will more accurately calculate the hebrew calendar date change based on the location of the user. The customizer has an opt-in geolocation api usage now. The customizer embeds the upcoming 3 months of calendar data into the watch app (since I couldn't figure out how to get it to use a data file that large otherwise)

The UI is brand new and improved as well. If someone else wanted to make their own phone-less calendar app, this app would be a good place to start.

feat: use customizer to more accurately generate calendar data file
feat: brand new UI to display upcoming/current calendar events
feat: add holidays, shabbat torah readings, and time display to clock


gfwilliams:
Great - thanks!

PR #1734, opened 2022-04-22
[Health Tracking] Add LANG placeholders, minor tweaks

alessandrococco:
This PR add the `LANG` placeholders to **Health** app.

---

Is there a reason this app settings are in the app instead of the global Settings menu?

### DIFF
--- a/apps/health/app.js
+++ b/apps/health/app.js
@@ -1,42 +1,54 @@
 function getSettings() {
-  return require("Storage").readJSON("health.json",1)||{};
+  return Object.assign({
+    hrm: 0,
+    stepGoal: 10000
+  }, require("Storage").readJSON("health.json", true) || {});
 }
 
-function setSettings(healthSettings) {
-  require("Storage").writeJSON("health.json",healthSettings);
+function setSettings(settings) {
+  require("Storage").writeJSON("health.json", settings);
 }
 
 function menuMain() {
   swipe_enabled = false;
   clearButton();
   E.showMenu({
-    "":{title:"Health Tracking"},
-    "< Back":()=>load(),
-    "Step Counting":()=>menuStepCount(),
-    "Movement":()=>menuMovement(),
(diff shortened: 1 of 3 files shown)


apps/health/app.js
```diff
@@ -1,42 +1,54 @@
...
-    "":{title:"Health Tracking"},
-    "< Back":()=>menuMain(),
-    "Heart Rt":{
-      value : 0|healthSettings.hrm,
-      min : 0, max : 3,
-      format : v=>["Off","3 mins","10 mins","Always"][v],
-      onchange : v => { healthSettings.hrm=v;setSettings(healthSettings); }
+    "": { title:/*LANG*/"Health Tracking" },
+    /*LANG*/"< Back": () => menuMain(),
+    /*LANG*/"Heart Rt": {
```
alessandrococco, line 33: I do not really like *Heart Rt* 🤔

gfwilliams:
Thanks! By the way, you know Espruino doesn't actually handle `let` correctly right now? Usually it's not a big deal but sometimes minifiers minify everything down to overlapping variable names, and then it breaks the minified code. So randomly refactoring var->let can sometimes cause things to break.

Good point about the settings. I think having an app is needed (so you can actually see what's been recorded), but it'd definitely make sense to pull the settings out into a `health.settings.js` file and then load that when needed.

alessandrococco:
> By the way, you know Espruino doesn't actually handle let correctly right now? Usually it's not a big deal but sometimes minifiers minify everything down to overlapping variable names, and then it breaks the minified code. So randomly refactoring var->let can sometimes cause things to break.

Aaargh, I didn't know it! :-/

> Good point about the settings. I think having an app is needed (so you can actually see what's been recorded), but it'd definitely make sense to pull the settings out into a health.settings.js file and then load that when needed.

Ok, I'll send a PR abouth this.

PR #1736, opened 2022-04-23
[New app] [Dragboard] A draging/swiping keyboard

thyttan:
A keyboard app/library where characters are selected by draging/swiping along one of two rows for numbers or letters.


gfwilliams:
Great - thanks!

Issue #1737, opened 2022-04-23
[Swipe keyboard] Marker positioning doesn't account for widget field.


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v13

### Describe the bug
When widget field is present on the display, the text and marker vertical positions doesn't match.

### Expected behaviour
The marker should be positioned correctly to the right of the base of the last letter.

### Steps to reproduce
1. Install [Swiping keyboard](https://banglejs.com/apps/?q=swipe%20keyboard).
2. Install [Noteify](https://banglejs.com/apps/?q=noteify).
3. Load Noteify -> 'New note'
4. Make sure the widget field is showing/takes up screen real estate.
5. Swipe to write a letter.
6. See that the height of the marker doesn't match the bottom of the text input.

thyttan:
Now I've made a PR regarding this issue, #1738.

PR #1738, opened 2022-04-23
[Swipe keyboard] Consider widget field for marker positioning

thyttan:
This solves issue #1737 regarding incorrect vertical positioning of the marker.


gfwilliams:
Thanks!

PR #1741, opened 2022-04-24
New app - Tap Keyboard

frigis1:
An onscreen keyboard that was made for Noteify, now a separate app.


gfwilliams:
Looks like the pre-commit checks have failed for this: https://github.com/espruino/BangleApps/runs/6145338826?check_suite_focus=true

You can see it flagged on https://github.com/espruino/BangleApps/pull/1741/files

In general I'd strongly discourage 'placeholder' getters/setters anyway, as they do tend to use more memory and slow down execution.

frigis1:
Thanks for the tip! I guess I don't fully understand what getter/setters are supposed to be used for, and it turns out I didn't need them at all here.

gfwilliams:
Great - thanks!

PR #1746, opened 2022-04-25
[AppLoader] New categories

alessandrococco:
This PR adds 3 new categories to the app loader:

image

Not sure about the label for "Apps Without Installations" (aka: apps with `type: RAM`) - does anyone have a better name?

---

Companion PR: https://github.com/espruino/EspruinoAppLoaderCore/pull/20

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,11 +75,14 @@
         <div class="filter-nav">
           <label class="chip active" filterid="">Default</label>
           <label class="chip" filterid="clock">Clocks</label>
+          <label class="chip" filterid="launch">Launchers</label>
           <label class="chip" filterid="game">Games</label>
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! I'll merge for now, but while more search options are good I'm a little worried that especially on mobile devices it's probably not ideal having multiple lines of tags there.

Especially something like Keyboards - where there's I think just a single app in the app store that uses keyboards, it probably doesn't make sense for that to be immediately visible to every user?

alessandrococco:
> Especially something like Keyboards - where there's I think just a single app in the app store that uses keyboards, it probably doesn't make sense for that to be immediately visible to every user?

Yes, you're probably right. We could hide the `Keyboards` chip for the moment.

PR #1747, opened 2022-04-25
[Health Tracking] Move settings

alessandrococco:
As suggested in https://github.com/espruino/BangleApps/pull/1734#issuecomment-1108413622.

### DIFF
--- a/apps/health/ChangeLog
+++ b/apps/health/ChangeLog
@@ -12,3 +12,4 @@
 0.11: Pre-minified boot&lib - folds constants and saves RAM
 0.12: Add setting for Daily Step Goal
 0.13: Add support for internationalization
+0.14: Move settings
--- a/apps/health/app.js
+++ b/apps/health/app.js
@@ -1,14 +1,3 @@
-function getSettings() {
-  return Object.assign({
-    hrm: 0,
-    stepGoal: 10000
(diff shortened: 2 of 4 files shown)


gfwilliams:
You sure this works? It seems like the `});` at the end may break it?

alessandrococco:
> You sure this works? It seems like the `});` at the end may break it?

Yes, I run it on my Bangle. The `});` at line 43 closes the `(function (back) {`. Maybe it feels strange due to the `setSettings` function after the menu 😄

gfwilliams:
Ok, thanks - in future can you try and avoid it though - because it's supposed to be an expression executed by `eval` so shouldn't really have the `;` on it

I'll merge now, but I'll also add a 'settings' link to the app menu to folks can still find the settings

alessandrococco:
> but I'll also add a 'settings' link to the app menu to folks can still find the settings

Should the 'settings' link in the app be a standard practice? IMHO it's confusing if some apps link to theirs settings and some do not.

gfwilliams:
I feel like where apps have a menu and it makes sense, they probably should link to settings as otherwise users may not be aware things *can* be changed. For clocks/etc where there's no obvious way to go to the settings menu I think they shouldn't.

PR #1748, opened 2022-04-25
[Alarms & Timers] "Save on Back", bulk actions

alessandrococco:
This PR adds ~~an option to set~~ the "Save on Back" (https://github.com/alessandrococco/BangleApps/issues/5) and three new menu items:

- `Enable All` (iff at least one alarm/timer is OFF)
- `Disable All` (iff at least one alarm/timer is ON) 
- `Delete All` (iff at least one alarm/timer exists)

~~Moreover it removes the custom boolean formatter since the new menu system handles the booleans better.~~

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -22,3 +22,5 @@
 0.21: Fix time reset after a day of week change (#1676)
 0.22: Refactor some methods to scheduling library
 0.23: Fix regression with Days of Week (#1735)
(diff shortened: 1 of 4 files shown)


alessandrococco:
> I think all round your change is an improvement, so we should probably just have it by default.

No problem! I'll update the PR later today.

alessandrococco:
> Actually, would it be possible for you to change this PR so that there was no menu option, and your 'save on back' behaviour was the default?

Done!

gfwilliams:
Excellent, thanks!

PR #1749, opened 2022-04-25
Fix some version numbers

Ronin0000:
Fixed 3 version numbers


gfwilliams:
Great - thanks! Maybe we should make the sanity checker enforce the `x.yz` version format?

PR #1750, opened 2022-04-25
after 2v14 release: [Scheduler] UI improvements

alessandrococco:
This PR removes the custom boolean formatter since the new menu system handles the booleans better.

image

### DIFF
--- a/apps/sched/settings.js
+++ b/apps/sched/settings.js
@@ -4,11 +4,10 @@
   E.showMenu({
     "": { "title": /*LANG*/"Scheduler" },
 
-    /*LANG*/"< Back": () => back(),
+    "< Back": () => back(),
 
     /*LANG*/"Unlock at Buzz": {
       value: settings.unlockAtBuzz,
-      format: v => v ? /*LANG*/"Yes" : /*LANG*/"No",
       onchange: v => {
         settings.unlockAtBuzz = v;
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! The only thing to say here is that I think pre-2v13 firmwares will output true/false instead of the checkbox.

I'm not sure how much we care about that (how many people are going to be on older firmware still?)

alessandrococco:
I would not worry about it, this is not a breaking change (the app will still works for those users).

gfwilliams:
Good point :) Although I just checked and on Bangle.js 1 2v13 we still display those menu items as true/false.

I just changed it, but it'll mean that until 2v14 Bangle.js 1 users get a rubbish experience.

I'm tempted to maybe hold off on this until the 2v14 release and then do a bit of a blitz and remove Yes/No and On/Off from pretty much all apps in BangleApps

gfwilliams:
Thanks! I guess we could look at doing this for other apps too...

alessandrococco:
> Thanks! I guess we could look at doing this for other apps too...

I have a local branch with many apps updated. I think I'll send a PR in the weekend.

gfwilliams:
That'd be great, thanks! I'm away next week so won't get it merged then, but I will on my return :)

PR #1753, opened 2022-04-26
[locale] dow/dowAll

alessandrococco:
This PR adds a new function `dowAll` to the `locale` module. It returns *all* the days of the week in 3 versions: long, short, first char only.

The PR also updates the `dow` function to add the _first char only_ version.

### DIFF
--- a/apps/locale/ChangeLog
+++ b/apps/locale/ChangeLog
@@ -19,3 +19,4 @@
 0.16: Remove global variables that used RAM
       Add second 'dp' argument for decimal places in distance/speed/temp (fix #1523)
 0.17: Fix regression where long month names were 'undefined' (fix #1641)
+0.18: Add function dowAll(), update function dow()
--- a/apps/locale/README.md
+++ b/apps/locale/README.md
@@ -1,25 +1,34 @@
-# Languages (locale)
+Languages (locale)
+==================
 
 Country-specific app internationalisation.
 
 This is not an app, but instead it is a library that can be used by
-other applications or widgets to display messages.
+other applications or widgets to provide locale-friendly
 
-## Usage
+- Dates
+- Time (12h/24h)
+- Days of the Week
+- Months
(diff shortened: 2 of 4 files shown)


gfwilliams:
What does `dowAll` do? Is it just like calling `dow` for all 7 days? eg:

```
days = [];
for (var i=0;i<7;i++) days.push(require("locale").dow({getDay:()=>i}));
```

I'd be a bit concerned about changing this unless absolutely required because this isn't the only implementation. The `locale` module is also built-in to the firmware itself (for when nobody has installed locale), so any app you make that uses `dowAll` is going to have to check to see if the `dowAll` method actually exists, at least until the firmware is updates *and* everyone's on 2v14.

It seems like for something that's basically a utility function it's maybe not worth it? What we *could* do is add a module to the `modules` folder that provided these kind of utility functions (like the ability to get the month from just the month number, and so on?)

That way it's also not using up memory unless an app specifically wants it

alessandrococco:
Hi @gfwilliams,

> It seems like for something that's basically a utility function it's maybe not worth it? 

The idea is to show the localized dow in the list of the configured alarms (some discussion [here](https://github.com/alessandrococco/BangleApps/issues/5)), eg.: 

```
Alarm 7:00
_MTWTF_ (en) --> _LMMGV_ (it)

Alarm 8:30
S_____S (en) --> D______S (it)
```

Sorry, I didn't know the implications of adding a new function to this module (but I'm learning 😄) No problem, I'll use your snippet instead.

gfwilliams:
Ok, just done - you should now be able to do `require("date_utils").getDOWs(2)` and it'll give you an array of single-letter day names. (check out https://github.com/espruino/BangleApps/blob/master/modules/README.md for usage though)

Feel free to add other stuff to that lib if you need it :)

alessandrococco:
Great, thank you!!

PR #1755, opened 2022-04-26
settings: Add some Bangle 1 colours to theme customizer

rigrig:
Orange, purple and grey seemed distinct enough to be nice additions

(Just in time for [King's Day](https://en.wikipedia.org/wiki/Koningsdag) ;-)


gfwilliams:
Thanks!

PR #1758, opened 2022-04-27
Fix backup when event listeners are active

BartS23:
I noticed that the backup script stopped running for me. 
Further analysis showed that the event `Bangle.on('accel', ...)` was the cause.

Excerpt from the Debug.Log:
```
<BLE> Sending "\nfor (var i=0;i<s.length;i+=384) Bluetooth.print(btoa(s.substr(i,384)));\nBluetooth.print(\"ÿ\");\n})()\n"
puck.js:406 <BLE> Received "Execution Interrupted during event processing.\r\n"
puck.js:406 <BLE> Sent
comms.js:435 <COMMS> readTextBlock read started...
puck.js:406 <BLE> Received "416\r\nKKooKXujKCFCYW5nbGUuaXNMb2NrZWQpqztCYW5nbGUub24oImxvY2siLKoob24pe1dJREdFVFNbImxvY2siXS53aWR0aD1CYW5nbGUuaXNMb2NrZWQoKT8xNjo"
puck.js:406 <BLE> Received "wO0JhbmdsZS5kcmF3V2lkZ2V0cygpO30pO1dJREdFVFNbImxvY2siXT17YXJlYToidGwiLHNvcnRvcmRlcjoxMCx3aWR0aDpCYW5nbGUuaXNMb2NrZWQoKT8xNjowLGR"
puck.js:406 <BLE> Received "yYXc6qih3KXujKEJhbmdsZS5pc0xvY2tlZCgpKWcucmVzZXQoKS5kcmF3SW1hZ2UoYXRvYigiRGhBQkgrRC93d01NRERBd3dNZi92Ly80ZitIL2gvOC8vUC96Ly8vZi9"
puck.js:406 <BLE> Received "nPT0iKSx3LngrMSx3LnkrNCk7fX07fSkoKQ==ÿ"
ui.js:86 <TOAST>[error] Backup failed, InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.
```
the line `puck.js:406 <BLE> Received "Execution Interrupted during event processing.\r\n"` causes an error in the calculation of the file length.

You can reproduce it if you install the `sleeplog` app and move your wrist a little while backing up.

Removing all listeners solves the problem for me.

Perhaps there is also a more elegant way?

### DIFF
--- a/backup.js
+++ b/backup.js
@@ -9,6 +9,7 @@ function bangleDownload() {
   console.log("Listing normal files...");
   Comms.reset()
   .then(() => Comms.showMessage("Backing up..."))
+  .then(() => Comms.write("\x10Bangle.removeAllListeners()\n"))
   .then(() => Comms.listFiles({sf:false}))
   .then(f => {
     normalFiles = f;


gfwilliams:
Thanks!  I guess the issue here is that while the app loader calls `reset()` it's still loading the `.boot0` code, which ideally wouldn't be doing anything too drastic - but which in reality can be doing all kinds of crazy stuff depending on what's installed.

As an aside, `sleeplog` should really be using the custom made 'health' event's movement field for this which I think should have the same effect, but without causing any real battery drain (or this kind of issue).

I think this is a problem for all kinds of things, so I think there's a better fix. Realistically we need to at minimum do this in `Comms.reset()`, but I think the best option is to modify the Bangle.js firmware such that `reset()` can be called in such a way that it really does reset the bangle without loading any extra code.

Filing an issue for this now

gfwilliams:
Please can you try the development app loader and see if it works any better now?

BartS23:
Your fix works perfectly.

Thanks

storm64:
>As an aside, sleeplog should really be using the custom made 'health' event's movement field for this which I think should have the same effect, but without causing any real battery drain (or this kind of issue).

@gfwilliams  I'm working on it right now.
In my opinion there is no need for complex ESS-calculations in sleeplog to work properly.

PR #1763, opened 2022-04-28
[alarm] Fix redrawing entry inside its submenu

storm64:
1. Fix redrawing selected Alarm/Timer entry inside its edit submenu through opening the menu after a short timeout.
2. Apply redrawing fix to dow submenu, too.
3. Removing the obsolete check-mark images.
4. Removing format on boolean values to use new check-marks.

---

5. Missed to ignore the _.gitignore_ file, should I remove the entry or can it stay?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ tests/Layout/bin/tmp.*
 tests/Layout/testresult.bmp
 apps.local.json
 _site
+._sync_*.db*
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@ tests/Layout/bin/tmp.*
 tests/Layout/testresult.bmp
 apps.local.json
 _site
-._sync_*.db*
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -24,3 +24,4 @@
 0.23: Fix regression with Days of Week (#1735)
 0.24: Automatically save the alarm/timer when the user returns to the main menu using the back arrow
       Add "Enable All", "Disable All" and "Remove All" actions
+0.25: Fix redrawing selected Alarm/Timer entry inside edit submenu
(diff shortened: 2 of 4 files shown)


alessandrococco:
About 1 and 2: Could you explain the issue? Every submenu should be called throught the setTimeout?

About 3 and 4: see Gordon's comments in #1750.

storm64:
1+2) On all menu entries displaying a value at the right-hand side (all except the entries with an >), the entry is redrawn after `onchange` is called to display the possibly changed value. See more: espruino/Espruino#2149

3+4) I see, will revert this changes asap.

nxdefiant:
If I may hijack this: Currently there seems to be no way to exit the vibration pattern menu, do you have an eye on this?

storm64:
> If I may hijack this: Currently there seems to be no way to exit the vibration pattern menu, do you have an eye on this?

This is handled inside [modules/buzz_menu.js](https://github.com/espruino/BangleApps/blob/master/modules/buzz_menu.js), will add a separate PR for this.

nxdefiant:
+1

gfwilliams:
So is this fixing an issue on Bangle.js 1 or 2?

And is it still an issue on recent firmwares? I haven't seen any problems, but I seem to recall that on some old firmwares we used the `rising` edge of button presses rather than `falling` and I can see that could cause issues.

storm64:
> So is this fixing an issue on Bangle.js 1 or 2?

Bangle.js 2

>And is it still an issue on recent firmwares? I haven't seen any problems, but I seem to recall that on some old firmwares we used the rising edge of button presses rather than falling and I can see that could cause issues.

The problem occurred on fw 2v13.15 and I remembered this issue espruino/Espruino#2149.

gfwilliams:
Thanks for the explanation. I'm really surprised I didn't hit this during testing - I wonder why not...

storm64:
> If I may hijack this: Currently there seems to be no way to exit the vibration pattern menu, do you have an eye on this?

@nxdefiant: Fix is merged: espruino/Espruino#2188

Issue #1682, opened 2022-04-10
[Touch/Swipe keyboard] Swipe keyboard always installs when "textinput":"type" dependency is used, interferes with Touch keyboard


frigis1:
### Affected hardware version

Bangle 2

### Your firmware version

2v13

### Describe the bug
I'm attempting to make use of the keyboard libraries in my app [https://frigis1.github.io/BangleApps/?id=noteify](url). When I try to load my app, the app loader always installs Swipe Keyboard, even though I've already installed Touch Keyboard. When I then uninstall Swipe Keyboard, Touch Keyboard no longer works, and will only work if I delete and re-install Touch Keyboard.

### Expected behavior
When "textinput":"type" is added to dependencies, the app loader will not load Swipe Keyboard if a keyboard is already installed. Also, when multiple keyboards are installed, perhaps there can be a way to choose between the two keyboards?

### Steps to reproduce
1. Install Touch Keyboard.
2. Install an app that uses "textinput":"type" dependency (e.g. [.](https://frigis1.github.io/BangleApps/?id=noteify). Swipe Keyboard will also install.
3. Uninstall Swipe Keyboard, and then attempt to use keyboard input in app. There will only be a black screen.
4. Delete and re-install Touch Keyboard, then again use keyboard input in app.

thyttan:
I just verified the following: Keyboard apps all write their lib.js (app code) to the file 'textinput'. If installing a second one, 'textinput' is overwritten. If one of the keyboard apps is deleted then so is 'textinput'.

If one wanted to handle this differently than above a backup of the first installed 'textinput' could be made before it's overwritten which is restored when the second installed keyboard is uninstalled.

thyttan:
I cannot reproduce this bug using [frigis1's app loader](https://frigis1.github.io/BangleApps/?q=keyb)
For me this happens when trying to reproduce:
1. Install Touch keyboard: Touch keyboard is installed.
2. Install for e.g. Noteify: Scheduer is installed (dependency). Noteify is installed.
3. Uninstall Swipe keyboard and attempt to use keyboard in app: No Swipe keyboard to uninstall. Text input works.
4. Delete and re-install Touch Keyboard: Not necessary in this case.

Bangle.js2,  FW 2v13

frigis1:
Thank you for your feedback! I factory reset my watch, updated the firmware, and as thyttan said everything works as expected. If nobody minds, I'll edit the original post as installing multiple keyboards is still an issue.

[...]

PR #1764, opened 2022-04-28
[Quick Launch] Update 0.02

frigis1:
Moved configuration from launcher to Settings->Apps.


gfwilliams:
Thanks! Just wondering about `"type": "launch",` - you're still expecting there to be another launcher on the system as well? Eg this isn't expected to erase the default launcher?

There's an issue open at https://github.com/espruino/BangleApps/issues/1682 and when that gets fixed, it might mean that installing this tries to replace the default launcher with it.

Please could you use `'`"type": "bootloader"'`?

> // if this is 'bootloader' then it's code that is run at boot time, but is not in a menu

frigis1:
Ah, I see. Updated metadata. I wanted to hide it from the default launcher but I wasn't sure what app type I was supposed to specify. Is there a list of app types I can look at?

gfwilliams:
Thanks! Yes, there's a list at: https://github.com/espruino/BangleApps#metadatajson-format

If you think I can add some better docs (or links to that somewhere) please let me know though - or even better do a PR ;)

PR #1765, opened 2022-04-28
Bradbury Watch

nomusictuesdays:
Bradbury Watch, a watchface based on the Seiko Voice Note. Displays time in hours/minutes/seconds, as well as date, day of week, battery percentage, and step count. Touch screen toggles the visibility of widgets.


gfwilliams:
Nice - thanks!

nomusictuesdays:
> Nice - thanks!

nomusictuesdays:
Hi, I'm still trying to figure things out here. I'm just wondering what the full process is for the app to appear in the official app loader, if there's anything else I need to do. I don't mean to pester you, I just don't really understand the process very well.

gfwilliams:
No, it's just when I manually push it up - I don't always do it because sometimes there might be something merged that I feel could do with some testing before it goes live. Also generally pushing new changes on Friday and then not looking at email over the weekend always seems like a recipe for disaster :)

I'll try and do it today though.

PR #1769, opened 2022-04-28
sleepphasealarm: Add logging

nxdefiant:
- add logging and plot data with chart.js. The plot is not useful at all, but does pretty pictures.
- switch to Layout library
- add ETA (time until alarm) to display


apps/sleepphasealarm/interface.html
```diff
@@ -0,0 +1,108 @@
...
+        <option selected disabled>No day</option>
+      </select>
+    </div>
+    <div class="chart-container">
+      <canvas id="sleepChart"></canvas>
+    </div>
+
+    <script src="../../core/lib/interface.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chart.js@3.7.1/dist/chart.min.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chartjs-adapter-date-fns@2.0.0/dist/chartjs-adapter-date-fns.bundle.min.js"></script>
```
nxdefiant, line 19: fetching files from CDN instead of downloading
them. Not very privacy friendly, but eg. openstmap 
also fetched from cdn.

gfwilliams: Thanks - I think that is probably fine. I guess we could maybe start hosting the files locally, but it feels like for `interface` files this is fine - it's not like they get loaded every time you run the app loader

PR #1770, opened 2022-04-28
Bug fix for OpenWind, new scientific calculator app

berkenbu:
Fixed small bug in true wind computation in openwind app.
New scientific calculator app (Bangle2 only, uses touch screen).


berkenbu:
Fair point, I'll add some screenshots. At least this time I remembered to
enable the emulator 😉

PR #1771, opened 2022-04-29
stopwatch: Do not register as watch

nxdefiant:
manually start clock on button, avoids conflict with quicklaunch which
otherwise believes that a clock is running and reacts on touch events

@hughbarney Are you ok with that?

### DIFF
--- a/apps/stopwatch/ChangeLog
+++ b/apps/stopwatch/ChangeLog
@@ -1,2 +1,3 @@
 0.01: first release
 0.02: Adjust for touch events outside of screen g dimensions
+0.03: Do not register as watch, manually start clock on button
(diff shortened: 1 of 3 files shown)


gfwilliams:
It's probably better to use `setUI("updown", dir=>{if (!dir) load()});` - while not a big deal here, if this were ever ported to Bangle.js 1 you might want BTN2 to go back, rather than BTN1?

nxdefiant:
Also the first version had a `process.env.HWVERSION == 2 ? BTN : BTN3` switch, I removed it when I realized that stopwatch is BJ2 only.

gfwilliams:
> updown would also act on touch click (tap), right?

Ahh, yes - good point! Or:

```
Bangle.setUI({mode:"custom",back:()=>load()});
```

And then you get the 'back' icon - but yes, I'll just merge this as-is.

PR #1774, opened 2022-04-29
Add screenshots to SciCalc app

berkenbu:
Per advisement, I added some screenshots to the README file. Did not bump the version since the app code remains unchanged.


gfwilliams:
Thanks! You can also add the screenshots to `metadata.json` so they appear when scrolling in the app loader - I'll do that now though

berkenbu:
Ah, thanks. In my (weak) defense, I read most of your (excellent, btw)
documentation on the app loader, etc. almost 2 years ago when I got my
Bangle 1 and haven't really kept up with a lot of the changes....

PR #1777, opened 2022-04-29
Update boot.js

boski1987:
edit uncodeRemap to map polish characters to characters supported by Bangle.


gfwilliams:
Great - thank you!

PR #1778, opened 2022-04-29
ActivityReminder: Obey system quiet mode & extend allowed dismiss delay

myxor:
- Obey system quiet mode
- Extend allowed dismiss delay

### DIFF
--- a/apps/activityreminder/ChangeLog
+++ b/apps/activityreminder/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New App!
 0.02: Fix the settings bug and some tweaking
 0.03: Do not alarm while charging
+0.04: Obey system quiet mode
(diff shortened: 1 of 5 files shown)


Stiralbios:
Thanks,

It's pretty cool for the quiet mode. And I just saw you did another pull request with cool features too ;).
However for the "Dismiss delay" with the current implementation it can pose some problems if it's higher than the "Max inactivity".
I plan to rework this part to prevent such bug, be able to add a longer delay "Pause" (for meetings and such) and reduce the battery usage which is quite high right now.

myxor:
Thanks for your feedback @Stiralbios. I added a notice to the README. I hope this is fine for now.

gfwilliams:
Thanks!

PR #1781, opened 2022-05-01
[Touch keyboard] Introduce settings for changing layout and functionality

thyttan:
@gfwilliams 

Updated keyboard can be loaded here: https://thyttan.github.io/BangleApps/?q=touch%20keyboard

I considered making a new app instead but decided it felt better to use settings for this app as to not clutter the app loader. 

A pro for doing a new app could be some small performance improvement, I guess. Any thoughts?

### DIFF
--- a/apps/kbtouch/ChangeLog
+++ b/apps/kbtouch/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New App!
+0.02: Introduced settings to customize the layout and functionality of the keyboard.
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks - just tweaking this app seems great to me. Maybe if someone really has speed issues and the app can't be improved there can be a new one made later on, but for now I think 4 separate keyboard apps is enough when there's only one app (`noteify`) that actually uses them :)

PR #1782, opened 2022-05-02
Changes to gpsautotime / sleepphasealarm

nxdefiant:
gpsautotime: Add setting to hide the widget
sleepphasealarm: Add check for correct day of week


nxdefiant:
Can be split into two PRs if requested.

gfwilliams:
Thanks, looks good!

Issue #1780, opened 2022-04-30
Alarm: Day of week settings should not always start on sunday


nxdefiant:
Right now the day of week settings in the alarm app always has Sunday as first day of the week. Dependent on the locale it should switch to Monday as first day. Problem is: This seems to be not yet provided by the locale or am I missing something?

So the alternatives are:
1. Add week-starts-on-monday to locale
2. Add week-starts-on-monday to alarm settings
Any opinions on this?

alessandrococco:
Hi @nxdefiant,

in the last days I made various changes to the alarms app and the "Week starts on Sunday" always bothered me so I started to thinking about it. This morning I made some updates to the `date_utils` module to handle the start of the week (not pushed yet) as I'd want to add this _qol_ improvement to #1773)

The module works

![image](https://user-images.githubusercontent.com/2331798/166106349-9afe011e-f1e8-445c-90ba-47a916a7546d.png)

now I need to integrate this to the alarms app (maybe tomorrow).

---

At the moment we have nothing that says "the week start on Monday/Sunday". I'd like to add a global option (maybe in `System -> Locale`?) Moreover we could include a default in each locale but the user can change it using the Settings app.

nxdefiant:
Great, I see you got it covered. Good for us not living in USA :)

alessandrococco:
In the sched library the days of week are stored as binary:

```
  dow : 0b1111111, // Binary encoding for days of the week to run alarm on
    //  SUN = 1
    //  MON = 2
    //  TUE = 4
    //  WED = 8
    //  THU = 16
    //  FRI = 32
    //  SAT = 64
```

Question: if the "Week starts on Monday" system setting is `true` should I consider MON = 1, TUE = 2 ... SUN = 64? Or should I keep as is?

nxdefiant:
I would keep the API unchanged and just change it in the user interface.

alessandrococco:
Some updates!

Improved `System / Locale` menu (renamed to `Time Options` or `Time Settings`)

![image](https://user-images.githubusercontent.com/2331798/166138511-7d2fa6b9-719e-4756-a1c0-1d576f4c315c.png) ![image](https://user-images.githubusercontent.com/2331798/166138521-d5068bee-0cb8-4870-a6fe-491ca375287a.png)

`Alarms & Timers` app:

- Start week on Monday:

[...]

PR #1773, opened 2022-04-29
[Alarms & Timers] Support for Monday as first day of week

alessandrococco:
_PR updated, see comments below_

~~This PR improves the UI of the default alarms app.~~

~~- Show localized abbreviations for days of week (current version always uses english abbreviations)~~
~~- An alarm on repeat Monday to Friday is labeled "repeat on workdays"~~
~~- An alarm on repeat Saturday and Sunday is labeled "repeat on weekends"~~

~~Some screenshots:~~

image 

image 

Any feedback/suggestions?

---

Closes #1780

PR #1784, opened 2022-05-02
New module "locale_utils"

alessandrococco:
This PR is preparatory to #1780.

I made some changes to the [`date_utils`](https://github.com/espruino/BangleApps/pull/1753#issuecomment-1109569187) module and I renamed it to `locale_utils` (it seems more appropriate to me as it is a container of _utility_ functions that use the _locale_ module).

This PR does not delete `date_utils.js` because the actual **Alarms & Timers** (and PR #1773) depend on it.

### DIFF
--- a/modules/locale_utils.js
+++ b/modules/locale_utils.js
@@ -1,0 +1,60 @@
+// module "locale_utils"
+//
+// Utility functions that use the "locale" module so can produce text in the
+// currently selected language.
+//
+// Some functions have a "firstDayOfWeek" parameter. Most used values are
+// - 0/undefined --> Sunday
+// - 1           --> Monday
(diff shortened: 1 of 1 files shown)


gfwilliams:
Can we just keep `date_utils` please? I really don't see this as a big deal.

Also, it's much better to keep these libraries small so apps that need then aren't pulling in a bunch of extra stuff. As such I could imagine that someone may want to do stuff with distance/temperature/etc and it'd be better to have `distance_utils`/etc than to lump them all into `locale_utils`

alessandrococco:
> Can we just keep date_utils please? I really don't see this as a big deal.

Sure, no problem!

PR #1785, opened 2022-05-02
waypointer

nxdefiant:
silently fallback to built in compass heading instead of throwing an Exception.


gfwilliams:
Looks good - thanks!

gfwilliams:
Looks good - thanks!

PR #1786, opened 2022-05-02
Add new widget alarmeta

nxdefiant:
This widget displays the time to the next Alarm or Timer in hours and minutes.


apps/alarmeta/metadata.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "alarmeta",
```
alessandrococco, line 2: my 2 cents: this is a widget so the name [_should_ start with **wid**](https://github.com/espruino/BangleApps#example-applications)

apps/alarmeta/widget.js
```diff
@@ -0,0 +1,37 @@
...
+    if (next > 0 && next < 86400000) {
+      const hours = Math.floor((next % 86400000) / 3600000).toString();
+      const minutes = Math.floor(((next % 86400000) % 3600000) / 60000).toString();
+
+      g.reset(); // reset the graphics context to defaults (color/font/etc)
+      g.setFontAlign(0,0); // center fonts
+      g.clearRect(this.x, this.y, this.x+this.width-1, this.y+23);
+
+      // Use 'locale' module to get a shortened month name
+      // in the correct language
```
alessandrococco, line 16: You could use the new locale_utils, see https://github.com/espruino/BangleApps/pull/1784

nxdefiant: Leftover comment from the template in http://www.espruino.com/Bangle.js+Widgets

apps/alarmeta/metadata.json
nxdefiant: done

gfwilliams:
Thanks! Just merging, but just a note:

> `const alarms = require("Storage").readJSON("sched.json",1) || [];`

This is going to load all alarms into RAM and keep them there. Not a huge issue for Bangle.js 2, but on Bangle.js 1 maybe it matters more.

You *could* just work out the time to the next alarm at startup, and store that. Then to work around the issue with days of the week you queue another check just after midnight.

PR #1788, opened 2022-05-03
[layout & smpltmr] New interface for Simple Timer, and changes to layout to make it work.

sir-indy:
Update to Simple Timer, preview here: https://sir-indy.github.io/BangleApps/?id=smpltmr
Discussion here: http://forum.espruino.com/conversations/375342/newest/

This is a rewrite of the Simple Timer app with a new interface, including the ability to set time in seconds.

This required a few tweaks to the layout module, to:
- center images in their container
- allow image rotation
- allow setting font on buttons (this feature not used in the app anymore, but left in as I can see uses for it in the future).


modules/Layout.js
```diff
@@ -246,6 +231,8 @@ Layout.prototype.render = function (l) {
         g.setFont(l.font).setFontAlign(0,0,l.r).drawString(l.label, l.x+(l.w>>1), l.y+(l.h>>1));
       }
     }, "btn":function(l){
+      if (l.font && l.font.endsWith("%"))
```
gfwilliams, line 234: I don't think this code should be here. `update` gets called first which checks for `%` andoverwrites `font`, so this code should never get called.

modules/Layout.js
```diff
@@ -259,12 +246,22 @@ Layout.prototype.render = function (l) {
...
-    g.setColor(bg).fillPoly(poly).setColor(l.selected ? g.theme.fgH : g.theme.fg2).drawPoly(poly);
+      g.setColor(bg).fillPoly(poly).setColor(l.selected ? g.theme.fgH : g.theme.fg2).drawPoly(poly);
     if (l.col!==undefined) g.setColor(l.col);
-    if (l.src) g.setBgColor(bg).drawImage("function"==typeof l.src?l.src():l.src, l.x + 10 + (0|l.pad), l.y + 8 + (0|l.pad));
-    else g.setFont("6x8",2).setFontAlign(0,0,l.r).drawString(l.label,l.x+l.w/2,l.y+l.h/2);
+    if (l.src) g.setBgColor(bg).drawImage(
+      "function"==typeof l.src?l.src():l.src,
+      l.x + l.w/2,
+      l.y + l.h/2,
+      {scale: l.scale||undefined, rotate: Math.PI*0.5*(l.r||0)}
```
gfwilliams, line 255: Just so you know, setting a rotation does center the image, but it also knocks the rendering out of the 'fast path' so images will render mucg slower.

Layout is already slow to update I know, so probably this doesn't matter - but it we just need to be careful we don't make it so slow it's unusable.

sir-indy: Thank you for your time reviewing this Gordon.
Yes, I realised it would slow it down, but it is convenient to center the images. In the interest of speed though I will remove the option, and prepare rotated images outside of the layout.
Unless there is a fast way of rotating an image by increments of 90 degrees? Or for this use, flipping by 180?

gfwilliams: No, I think what you've done is the best option - I thought I'd better point it out though. Probably the best bet is to extend the 'fast path' inside Espruino if this becomes an issue.

modules/Layout.js
sir-indy: Thank you, I hadn't twigged that, removed as requested.

modules/Layout.js
sir-indy: Thanks Gordon, I'll leave it as it is then. At the moment it doesn't feel any slower than before, but they are quite small images (9px square and 24px square 1bpp).

sir-indy:
Changes made to layout as suggested, and a minor tweak to the theme colours (same actual colour, just changed the reference from fg2 to the standard fg, now I think I understand a little better how themes work).
I'm happy with this now, are there any tests that I've not broken the `Layout` module?

gfwilliams:
Thanks - yes, there's `tests/Layout/bin/runalltests.sh` which does show up some differences. There were some regressions on 'master', but it looks like for your branch the button font is changed.

I just checked and in `padding.js` it seems your changes actually give it the correct behaviour :)

gfwilliams:
Just spotted the issue with the other tests - it was just that we'd started relying on Bangle.appRect which needed faking properly. They all pass now

sir-indy:
Excellent! Thank you very much for your help Gordon! You've made something that is an absolute joy to play with, and having your help with it is invaluable.
I'm planning on using the configurable button font in a keyboard app, but by default it should still be 6x8, scale 2, though now written as '6x8:2'

PR #1790, opened 2022-05-04
[Alarms & Timers] forEach

alessandrococco:
Small code change,  [as suggested during the review](https://github.com/espruino/BangleApps/pull/1773#pullrequestreview-961431288).


gfwilliams:
Thanks! I think you can remove the `(i => { })(i)` bit too, but I'll do that this end

alessandrococco:
> I think you can remove the (i => { })(i) bit too

Good! I wasn't sure about it, my javascript skills are quite limited and I didn't fully understand that code block.

PR #1791, opened 2022-05-04
[NotAnalog Clock] Use Bangle.getHealthStatus("day").steps as fallback if no widget is installed.

peerdavid:
As Bangle.getHealthStatus("day").steps is available, it is now used as fallback if no widget is installed.


gfwilliams:
Thanks!

PR #1792, opened 2022-05-04
sched: add day of week check to getActiveAlarms()

nxdefiant:
See also discussion in #1780:

When having multiple alarms on a single day a successive alarm can display a previous alarm that should not fire because of its day of week configuration.

Example, today is Monday, alarm configurations are
Alarm 23:10, S_TWTFS, REPEAT YES
Alarm 23:11, SMTWTFS, REPEAT YES

Expected: Only one alarm at 23:11
Actual: Two alarms at 23:11 because sched things the first one was missed due to the check for day of week is missing.

This PR adds a check for day of week in getActiveAlarms()


nxdefiant:
Just not 100% sure if the dow check should happen in the function getActiveAlarms() or in the calling of the function in sched.js.

gfwilliams:
Great - thanks!

PR #1794, opened 2022-05-04
Calendar

nxdefiant:
- Do not register as watch, manually start clock on button (Compatibility with Quick Launch)
`Bangle.setUI({mode:"custom",back:()=>load()});' is not an option because there is no space for widgets.

- Read start of week from new system settings from #1773


gfwilliams:
Thanks!

PR #1795, opened 2022-05-05
[Bordle] Fix bug in valid word detection

nemesisridiculii:
Update to @berkenbu's bordle app

The valid word detection did not account for the fact that a valid word
could exist twice in the list: once at a non-multiple-of-five index
because it is the end of one word and the start of another, and again at
a multiple-of-five index for the word itself. An example is aglow agony
contains wagon. This prevented wagon from being considered a valid word
(even though it could be selected as the target word).

This fix loops through all detections of the word in the dictionary to
see if any are at a multiple-of-five index.

### DIFF
--- a/apps/bordle/ChangeLog
+++ b/apps/bordle/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New App
 0.02: app keeps track of statistics now
+0.03: Fix bug in valid word detection
--- a/apps/bordle/bordle.app.js
+++ b/apps/bordle/bordle.app.js
@@ -110,7 +110,12 @@ class Wordle {
     }
   }
   addGuess(w) {
-    if ((this.words.indexOf(w.toLowerCase())%5)!=0) {
+    let idx = -1;
+    do{
+      idx = this.words.indexOf(w.toLowerCase(), idx+1);
+    }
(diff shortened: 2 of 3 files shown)


nemesisridiculii:
This runs on the emulator, but not on a bangle2 watch. I'm closing this and will submit a new pull request once I've gotten it working for real. Sorry for the disruption.

berkenbu:
Thanks, nice catch! I originally used a newline separated word list but then thought I could save some space by optimizing the newlines away - I clearly didn't think it all the way through though...

nemesisridiculii:
OK. Tried it again today, and it worked without an error. I think I may have uploaded the source file directly instead of going through the toolchain in the ide. I was getting all kinds of weird errors regarding parsing failing, and/or the Layout module not being found.

I'm new to bangle.js coding and the online ide, so please let me know if I've done something wrong here. As for now, it looks like it's running, and I think it's correct so I'm going to re-open this PR and it can hopefully be merged.

gfwilliams:
Thanks! Looks good. You may have hit some issues with the module loading - there's some info at https://github.com/espruino/BangleApps/blob/master/modules/README.md

PR #1796, opened 2022-05-05
miclock2: Redraw only when seconds change

nxdefiant:
I had some trouble of undetected button presses when miclock2 is running. Redrawing every second instead twice a second helps to mitigate.


gfwilliams:
Thanks!

PR #1797, opened 2022-05-05
[GolfView] Course search

jdekarske:
This PR adds a more reasonable way to search for your course. Future improvements include the ability to see how well a given course is supported and the ability to store more than one course at a time.


gfwilliams:
Thanks!

PR #1798, opened 2022-05-06
Initial import of the tabanchi watch/app

trufae:
app


gfwilliams:
Nice! Just so you know, you can do screenshots right from the Web IDE (https://github.com/espruino/BangleApps#screenshots) which is generally a lot easier than having to take a photo :)

PR #1799, opened 2022-05-06
Update eslint

jdekarske:
ESLint was a major version behind. I've also included the Airbnb lint config which is the most popular. I'll use it for my app and I'd expect others to use it if it is helpful for them.


package.json
```diff
@@ -18,8 +23,5 @@
   },
   "dependencies": {
     "acorn": "^7.2.0"
-  },
-  "devDpendencies": {
```
jdekarske, line 22: Also a typo here

gfwilliams:
Thanks!

PR #1800, opened 2022-05-07
widgps: add intervall to update the widget

nxdefiant:
Periodically update the gps widget so the always on display of the BangleJS 2 does show current GPS fix


gfwilliams:
Thanks!

PR #1801, opened 2022-05-07
[NeoX] Better lock animation.

peerdavid:
This PR improves the animation when the bangle is locked / unlocked for the NeoX watch face. I tested both, NeoX and Neo IO X.


gfwilliams:
Thanks!

PR #1802, opened 2022-05-08
Slim battery widget with cells

idavydov:
Hi,
I know there are many batter widgets here already. But I think this is the only one which has cells. :-)

I hope this could be merged.

It borrows some code from other batter widgets, but I guess that's fine since it's all in one repo?


gfwilliams:
Looks good - thanks!

PR #1805, opened 2022-05-09
[terminalclock] Add altitude to terminalclock

asmaps:
Thanks to the Altimeter app, where I borrowed the altitude reading code from.


gfwilliams:
I think there may be an issue here - because this app also runs on Bangle.js 1, and that doesn't have barometer.

I think you'd need to put `setBarometerPower` inside `process.env.HWVERSION != 1` or this will break on the Bangle.js 1.

asmaps:
Thanks for the hint. I removed the setting from the menu on Bangle.js 1 and also as a failsafe disabled the baro enabling for hwversion 1

gfwilliams:
Great - thanks!

PR #1806, opened 2022-05-09
noteify: Add interface

nxdefiant:
Add web interface to add, edit and delete notes.

Version number is not incremented because the application itself is unchanged.

Notify @frigis1


gfwilliams:
Thanks!

frigis1:
@nxdefiant Oh wow, thanks!

PR #1813, opened 2022-05-10
Add 12h support, auto-cycle control to Rebble Clock

zachstr2:
I was annoyed by the lack of 12h support and the inability to turn off the auto-cycling of the right bar, so I added those features.

I also ran into a bug with the settings app where the 's' variable was being overwritten by something else which caused issues. It must be a global variable or is used internally somewhere. Because it was only used for one setting and immediately saved, the desired settings were always saved there successfully (sometimes along with other stuff from the global variable), but with a second setting the pervious settings could be lost. Changing it to 'localSettings' fixed it.

### DIFF
--- a/apps/rebble/ChangeLog
+++ b/apps/rebble/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Added dependancy on Pedometer Widget
 0.04: Fixed icon and png to 48x48 pixels
 0.05: added charging icon
+0.06: Add 12h support and autocycle control
--- a/apps/rebble/metadata.json
+++ b/apps/rebble/metadata.json
@@ -2,7 +2,7 @@
   "id": "rebble",
   "name": "Rebble Clock",
   "shortName": "Rebble",
-  "version": "0.05",
+  "version": "0.06",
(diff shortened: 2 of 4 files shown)


Rarder44:
could you also add the part relating to the location of the calendar?
(in order not to create 2 different pullrequests)

I tried the code and it should work (but double checking is always welcome xD)

>   let date = new Date();
>   let dy=require("date_utils").dow(date.getDay(),1).toUpperCase();
>   let dd=date.getDate();
>   let mm=require("date_utils").month(date.getMonth()+1,1).toUpperCase();
>   ....
>   drawDateAndCalendar(w3, h/2, dy, dd, mm);

you should put this code in the drawSideBar.


in the future i was thinking about making the interface a little more modular:
give the possibility (via web Browser editor, or options directly from the clock) to choose what to display in the various areas ...
but I think that for the moment these changes are enough!

gfwilliams:
That would be great! But actually you don't need a separate pull request. It looks like this one was done off your `master` branch, so all you need to do is make the changes in your GitHub fork and they'll appear here automatically.

PR #1817, opened 2022-05-11
[apps/widbt_notify] fix screen redraw

### LINKED ISSUE #1810: [widbt_notify] Incorrect redraw?
### Affected hardware version

Bangle 2

### Your firmware version

2v13

### The bug

Screen is not properly cleared after widbt_notify popup.

## Steps to reproduce
1. Run a dummy clock app.
```
function draw() {
  g.reset();
  g.clearRect(50,50,100,120);
  g.drawString("01:01", 50, 50);
}

g.clear();
draw();
var secondInterval = setInterval(draw, 1000);
```
2. Disconnect your watch from bluetooth.

Expected results: pop up appears, and the screen is showing 01:01 in a couple of seconds again.

Actual result: pop up appears, but the screen is not properly cleared.

I suspect the reason for that is how widbt_notify cleans up the screen:

```
    redrawCurrentApp: function(){
        if(typeof(draw)=='function'){
            draw();
        }else{
            load(); // fallback. This might reset some variables
        }
    },
``` 

Do I understand correctly, that widbt_notify is incorrect in assuming that if the app has a `draw()` function, it will correctly redraw the whole screen?

If this is true, the trivial fix would be just to call `load()` regardless of the `draw()` method.

```
            E.showMessage(/*LANG*/'Connection\nlost.', 'Bluetooth');
            setInterval(()=>{load();}, 3000);
```

CC @KungPhoo

idavydov:
Hi,

I noticed that @KungPhoo opened a PR KungPhoo/BangleApps#3, but I think in a wrong repo.

I cherry-picked his commits on top of the master branch.

This should fix #1810

### DIFF
--- a/apps/widbt_notify/ChangeLog
+++ b/apps/widbt_notify/ChangeLog
@@ -8,3 +8,4 @@
 0.09: Vibrate on connection loss
 0.10: Bug fix
 0.11: Avoid too many notifications. Change disconnected colour to red.
+0.12: Better notification and redraw.
--- a/apps/widbt_notify/metadata.json
+++ b/apps/widbt_notify/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "widbt_notify",
   "name": "Bluetooth Widget with Notification",
-  "version": "0.11",
+  "version": "0.12",
   "description": "Show the current Bluetooth connection status in the top right of the clock and vibrate when disconnected.",
   "icon": "widget.png",
   "type": "widget",
--- a/apps/widbt_notify/widget.js
+++ b/apps/widbt_notify/widget.js
@@ -19,7 +19,7 @@ WIDGETS.bluetooth_notify = {
     
     disconnect: function() {
         if(WIDGETS.bluetooth_notify.warningEnabled == 1){
-            E.showAlert(/*LANG*/'BLUETOOTH\nConnection\nlost.', 'BLUETOOH');
+            E.showAlert(/*LANG*/'Connection\nlost.', 'BLUETOOH');
             setTimeout(()=>{
                 E.showAlert();
                 load();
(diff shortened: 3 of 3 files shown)


idavydov:
Hi @KungPhoo , I noticed that the new showAlert behavior doesn't play well with the setTimeout.

I think I managed to solve this problem, but I would like you to first take a look at the code if you approve.

gfwilliams:
Looks ok to me - clocks/etc may still draw over it during those 3 seconds though. You could add a plain `clearInterval()` before the showAlert to stop that happening.

Also I see there's a `setTimeout('WIDGETS.bluetooth_notify.warningEnabled = 1;', 30000);` but that will never happen because you reload after 3 sec.

One other thing - because you force a reload, it's going to be frustating if you were using another app. Say you're using the calculator and typing a sum in, the alert goes off and 3 seconds later it reloads back to the clock. Probably not ideal. You could check if `Bangle.CLOCK!=0` and if so don't show an alert?

idavydov:
Thanks, @gfwilliams ! Makes sense. Not sure I'll update it quickly, since I'm using a different widget now.

But I would like to understand the problem regarding `warningEnabled`.
The only option here is something like saving the last popup event in storage?

E.g.,
```js
const DATAFILE = "myapp.json";
var appData = Object.assign({
  lastNotify: 0,
}, require('Storage').readJSON(DATAFILE,true)||{});
const time = getTime();
if(time-appData.lastNotify>30) {
  // show notification
  ...
  appData.lastNotify = time;
  require('Storage').writeJSON(DATAFILE,appData); 
}
```

gfwilliams:
> The only option here is something like saving the last popup event in storage?

Yes - right now you can't persist anything in RAM when you swap between apps. It's just that the relatively limited RAM available and variable quality of apps means allowing apps to store stuff in RAM all the time could very easily make Bangle.js unstable for everyone

gfwilliams:
Closing this for now as no updates (and i'm not sure forcing an app reload is going to be a good improvement for the reasons above).

Espruino 2v15 now includes the ability to draw 'overlays' so I think that would end up being the best solution (although maybe we should have a library to reduce code duplication)

PR #1821, opened 2022-05-12
ClockFace module

rigrig:
I figured this might be a good idea.

A module which handles watchface boilerplate:

* gets rid of watchface boilerplate
* adds `Bangle.pause`/`Bangle.resume`, so widgets/boot code can do this
   ```js
   if (Bangle.pause) { // check for support
       Bangle.pause(); // now Bangle.paused = true
       E.showAlert("Alert!").then(() => Bangle.resume());
   }
   ```
* we could extend it later with global options, like powersave mode or custom widget Stuff.

```js
// Just showing off: clock one-liner 
(new (require("Clock"))(t=>g.setFont('Vector:50').drawString(require("locale").time(t,1),50,100))).start();
```

### DIFF
--- a/apps/barclock/ChangeLog
+++ b/apps/barclock/ChangeLog
@@ -7,3 +7,4 @@
 0.07: Update to use Bangle.setUI instead of setWatch
 0.08: Use theme colors, Layout library
 0.09: Fix time/date disappearing after fullscreen notification
+0.10: Use ClockFace library
--- a/apps/barclock/clock-bar.js
+++ b/apps/barclock/clock-bar.js
@@ -11,13 +11,9 @@ let locale = require("locale");
   date.setMonth(1, 3); // februari: months are zero-indexed
   const localized = locale.date(date, true);
   locale.dayFirst = /3.*2/.test(localized);
-
-  locale.hasMeridian = false;
-  if (typeof locale.meridian==="function") {  // function does not exist if languages  app is not installed
-    locale.hasMeridian = (locale.meridian(date)!=="");
-  }
+  locale.hasMeridian = (locale.meridian(date)!=="");
 }
-Bangle.loadWidgets();
+
 function renderBar(l) {
   if (!this.fraction) {
     // zero-size fillRect stills draws one line of pixels, we don't want that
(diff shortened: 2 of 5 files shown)


gfwilliams:
Looks like a great idea! Maybe `Bangle.on('pause/resume/redraw',` would be better though, as it'd potentially allow more than one thing (widgets?) to respond. I guess maybe just have a `function draw() { Bangle.emit('redraw'); }` so it'll work as-is with the current widget code?

Also I think it might be an idea to rename it, because https://www.espruino.com/clock exists (even though it's old and unmaintained) and if someone used the wrong capitaliation thay'd get the wrong one. What about `clock_utils`?

I wonder whether the whole fullscreen notification thing needs some proper thought though, so we can be sure it works well with all apps - but it's not a reason not to have this module - we can just change the module later. What I'm thinking is we *could* just allocate a second buffer (for Bangle.js 2) and render everything there while the notification is up - so then it'd work with all apps without having to pause/resume/redraw. On Bangle.js 1 you could even do something similar if you were able to keep the notification to 80 pixels high (since that could be displayed fullscreen, while any draws went to the rest of the buffer).

rigrig:
> rename it, because https://www.espruino.com/clock exists ... What about `clock_utils`?

Oh, we definitely don't want to name it `clock` then. But I'd like to keep the name and class name the same to avoid confusion.  Maybe `WatchFace`/`ClockFace`? (Not overjoyed with those, let the bikeshedding begin ;-)


>  Maybe `Bangle.on('pause/resume/redraw'`
> ...
> the whole fullscreen notification thing needs some proper thought 

My idea was: 
- If an app sets up `Bangle.pause`, that means widgets can call it, and the app promises to sleep until you call `Bangle.resume()`. 
- Widgets should just keep working, because notifications should respect `Bangle.appRect`
- If widgets really need to, they can always check `Bangle.paused`
- I doubt we'd be able to make fullscreen notifications work universally without app support. Maybe through the firmware, but that would get complicated: apps might try *anything*: add/clear event listeners, turn the screen on/off (and expect it to stay that way), mess with `WIDGETS`...

But I agree it would be better to keep the `Bangle.pause`/`Bangle.resume` stuff out of this, we can have a separate issue about pausing/resuming.

rigrig:
Named `ClockFace` now, as [the tutorial](https://www.espruino.com/Bangle.js+Clock) is titled "Clock Faces". The documentation also shows how the tutorial clock would look like using this module.

I removed the `Bangle.pause`/`Bangle.resume` stuff, and also added a commit which converts `barclock`.

rigrig:
There is a forum thread about [separate locked/unlocked faces](http://forum.espruino.com/conversations/375791/)   
Maybe the library should also call `draw` when locking/unlocking?

gfwilliams:
Nah, I think that's something that each clock face should probably handle itself if it wants to do that.... I mean, we have almost 100 clock faces and this is the first one that's asked to do it so far... And doing it should only be a few lines of code if that.

gfwilliams:
This all looks good - I'll merge this now. One thing I would say though is `is12Hour` would be a good function to put in the library :)

PR #1828, opened 2022-05-14
[apps/widbthide] bluetooth widget which hides upon no connection

idavydov:
Hi,

It is rather hard to distinguish blue from gray on banglejs2. Especially, when the backlight is off.

So I added an option to widbt: `No conn. widget` (Show/Hide). It will hide the widget when there's no connection. I didn't spell the word "connection" fully, since this causes the menu to overflow.

I also had to add an `app-icon.js`. Without the icon the settings menu was not showing up.

Let me know if you would prefer to keep `widbt` simple. I can make it a separate app then.


idavydov:
Hi @gfwilliams ,
I now created a new widget instead. Does it look ok to you?

gfwilliams:
Looks great - thanks!

PR #1829, opened 2022-05-14
New app - Multi Timer

frigis1:
An app that lets users create multiple timers and chronographs that count down/up in real time. Also allows users to create alarms.


gfwilliams:
Looks great - thanks!

PR #1830, opened 2022-05-15
[kbmulti] Various changes

thyttan:
@sir-indy 

0.02: Introduce setting "Show help button?". Make setting firstLaunch invisible by removing corresponding code from settings.js. Add marker that shows when character selection timeout has run out. Display opened text on launch when editing existing text string.


gfwilliams:
Thanks!

PR #1831, opened 2022-05-15
Add homework app

dronesflier:
Add my own simple homework app with the ability to add and remove subjects with the app loader customizer.


dronesflier:
Sure, I'll update the app soon to compensate for that!

PR #1832, opened 2022-05-15
[dtlaunch]  Don't change page if doing swipe-to-exit

thyttan:
@jeffmer @hughbarney 

When doing a swipe to exit on v0.13 dtlaunch moves pages the instance before exiting. The changes in this PR (v0.14) makes it so that doesn't happen - but still respect left and right swipes when swipe to exit is turned off in settings.


gfwilliams:
Ahh, thanks! Sorry, I only looked in the diffs, not the full readme.

Merging now!

PR #1836, opened 2022-05-16
ClockFace: add is12hour

rigrig:
This is indeed nice to have in the library.

Also updates barclock to use it.


gfwilliams:
Thanks!

Issue #1835, opened 2022-05-16
[SleepLog] Infinite history eat all memory.


Stiralbios:
### Affected hardware version

Bangle 1, Bangle 2

### Your firmware version

2v13

### The bug

SleepLog record the history in sleeplog.log. When reading the file it decompress the data and load it as an Array. This can eat up all RAM for a long history (few weeks).

A possible solution would be a "Max history" setting to keep only the last few days.

storm64:
I am already working on a complete rewrite of sleeplog, but this takes much more time than I had expected.

But to have no one else searching for this bug I would like to implement a fix for this asap

Do you have a suggestion for the number of days to store?

Stiralbios:
I looked at your pull request and I don't think a 1500 length is low enough. I had the issue at a size of 1140 on a BangleJS2 and the BJS1 have less memory.
If you want to do it by a number of days I think a default at 3 would be good (that could maybe be override in the settings). It's low enough to not cause issue. I had the issue after ~3weeks but the BJS1 have less memory. Plus you can feel the watch is slower after 4 or 5 days store moving between menus/apps. So 3 days seems good to me

Stiralbios:
No, I use it on the BJS2. For the BJS1, my bad, I though it was compatible. The rest of my comment still hold thow

storm64:
I think this bug can be closed, now. :)

gfwilliams:
Thanks!

PR #1839, opened 2022-05-17
sleeplog: Fix LOW_MEMORY,MEMORY error #1835

storm64:
Check and if neccessary reduce logsize to 1500 entries in writeLog() to prevent low mem error.

### DIFF
--- a/apps/sleeplog/ChangeLog
+++ b/apps/sleeplog/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Fix crash on start #1423
 0.03: Added power saving mode, move all read/write log actions into lib/module
 0.04: Fix #1445, display loading info, add icons to display service states
+0.05: Fix LOW_MEMORY,MEMORY error on to big log size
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks! Just to add that you could use StorageFile (which handles appends) to store one record per line (still as JSON) and that'd mean you could store data much more easily - it'd be way faster and it'd hardly use any memory. The current method is rewriting a whole new file for each entry added too.

You could then 'stream' the data out, line by line, and again there would be no memory limits.

While 1500 would (I think?) work on Bangle.js 2 as long as there isn't much data it'd still kill Bangle.js 1.

storm64:
The "new" sleeplog is using StorageFiles and only writes old data into a seperate new file.

> While 1500 would (I think?) work on Bangle.js 2 as long as there isn't much data it'd still kill Bangle.js 1.

The sleeplog app is not supported on Bangle.js 1 for now because of the excessive memory needed.  
I hope the new app might be able to run on lower memory.

PR #1841, opened 2022-05-17
Clock & Calendar: Improve colours

emilecantin:
The red when disconnected is very hard to read (and not documented). I've switched it to have a slight blue tint when connected, and white when disconnected.


gfwilliams:
Thanks! But please can you add a line to the ChangeLog and bump the version number in `metadata.json` so that users know there's an update available?

emilecantin:
@emilecantin Took me a little while, but it's done!

gfwilliams:
:) thanks!

PR #1843, opened 2022-05-18
[Android & Sched] - small fixes and updates

frigis1:
For the Android app, it seems better to let Sched's default alarm settings handle things like vibration, auto-snooze, and repeat for GB alarms. 

As for Sched, if a nonexistent js file is specified in an alarm's js field, the watch will be stuck on a screen which says "Loading..." until auto-snooze kicks in, and then repeats 10 mins later (or however long your AS is set to). Sched will now check to see if the js file exists - if not, it will default to sched.js. It seems to handle alarms with no js field just fine as well.

### DIFF
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -7,3 +7,4 @@
 0.06: Option to keep messages after a disconnect (default false) (fix #1186)
 0.07: Include charging state in battery updates to phone
 0.08: Handling of alarms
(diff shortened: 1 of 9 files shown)


gfwilliams:
Thanks! Please can you bump the version numbers in metadata?

frigis1:
Metadata updated :)

gfwilliams:
Thanks!

PR #1844, opened 2022-05-18
[widbt_notify] Prevent repeated draw()

storm64:
Replace `setInterval` with `setTimeout` to clear the `Connection lost.`-message to prevent repeatedly calling of `draw()` from the current app/clock.


gfwilliams:
Thanks!

PR #1845, opened 2022-05-18
Upgrade tinydraw app with fixes and performance improvements

trufae:
* Fix segmented line drawing with all the pencils
* Dont use a timer to check if the screen is locked


gfwilliams:
Great - thanks!

PR #1847, opened 2022-05-18
Created Poketch clockface

elykittytee:
There's 54 commits because I couldn't figure out how to work in my respository. But it works on my app loader :) Only the **pokeclk** folder should be the one added


gfwilliams:
Looks great - thanks! Merging now!

Looks like the ChangeLog is in slightly the wrong format (no `:`) but I'll fix that here.

Looking at the images you have in there, it seems like at minimum you might benefit from enabling the 'compression' option in the image converter, but maybe also rather than 8 bits, you could use the 'optimal palette' or something like that to reduce memory usage.

elykittytee:
Thank you so much!
I'll work on adding the images with compression :)

PR #1850, opened 2022-05-18
[time_utils] Improve formatDuration with days and seconds

### LINKED ISSUE #304: Adding locale.duration
We have stuff for absolute time, distance, temperature, but I think it also makes sense to have a nicely formatted duration. For example in the stopwatch, or other apps...

* <2 minutes => `23 sec`
* <60 minutes => `1 min 23`
* `1 hr 23`

something like that?

alessandrococco:
This PR adds support for days and seconds to `time_utils`. It also improves the output of `formatDuration`.

Closes #304

### DIFF
--- a/modules/time_utils.js
+++ b/modules/time_utils.js
@@ -1,54 +1,78 @@
 // module "time_utils"
 //
 // Utility functions useful to work with time and durations.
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks!

PR #1851, opened 2022-05-19
[BW Clock] The color of widgets is now shown correctly.

peerdavid:
If the fullscreen mode was off, the color of widgets was wrong as the "dark" BW clock has a white header. This is fixed with this PR:

screenshot_3


gfwilliams:
Thanks!

PR #1852, opened 2022-05-19
[Rebble Clock] various fixes

Rarder44:
- fix autoCycle
- removed unused font
- font reconverted and compressed with heatshrink


Rarder44:
- imported suncalc and trimmed

gfwilliams:
Looks good, but please could you tweak the changelog and version? Looks like 2 lines are deleted and the version has gone back by one!

Rarder44:
by bad;
in the first two commits I increased the version (0.08 and 0.09) then I realized that it was useless to keep increasing it for each fix until the pull was finalized.
for this I then restored to 0.08 and merged the changelog lines.

gfwilliams:
Ok, thanks - probably best in future to just keep bumping it though. This'll be a bit confusing to anyone that installed it from the dev app loader :)

Rarder44:
ok, sorry!

PR #1822, opened 2022-05-13
[Alarms & Timer] New UI

alessandrococco:
**Discussion here: http://forum.espruino.com/conversations/375582/**, PR opened for a feedback :-)

---

In this PR the default **Alarms & Timers** has been heavily improved:

- Simpler main menu
- New labels for alarms and timer (with localized days of week)
- New "Repeat Alarm" menu with shortcuts (Once, Workdays, Weekends, Every Day, Custom)
- New icons
- Cleaner code

This PR also introduce the new `time_utils` module. Some functions have been moved from `sched` to it. This PR also update the dependent app.

The PR is 90% complete. Now I need to update the README and add some screenshot to the `metadata.json`.

You can try it in advance from by App Loader: https://alessandrococco.github.io/BangleApps/ (you need to install the Alarms & Timers app and the Scheduler library).

---

image image

image image

image image

image image image 

image

image

PR #1853, opened 2022-05-19
[Alarms & Timers / Scheduler] Fix bug with alarms not firing when repeat is set to "once"

alessandrococco:
In #1822 I merged the "repeat (Y/N)" and "days" menus. Before my PR a "repeat _once_ alarm" inherited a dow = 127 (_every day_) but in my updated app I allow a free customization of the days of the week so if the user select none of them then dow = 0 and the app assumes that the alarm will fire _once_.

I didn't noticed that an alarm with dow = 0 will never fire because the filter will discard it!

This PR fixes that (I also changed the filter conditions to be in the same order so they are easier to read)

---

Anyway, could the `rp `attribute be removed? The dow alone should be enough to understand if an alarm will run multiple times (dow > 0) or once (dow = 0). What do you think?

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -27,3 +27,4 @@
 0.25: Fix redrawing selected Alarm/Timer entry inside edit submenu
 0.26: Add support for Monday as first day of the week (#1780)
 0.27: New UI!
+0.28: Fix bug with alarms not firing when configured to fire only once
--- a/apps/alarm/app.js
+++ b/apps/alarm/app.js
@@ -50,7 +50,7 @@ function showMainMenu() {
   alarms.forEach((e, index) => {
     var label = e.timer
       ? require("time_utils").formatDuration(e.timer)
-      : require("time_utils").formatTime(e.t) + (e.dow > 0 ? (" " + decodeDOW(e)) : "");
+      : require("time_utils").formatTime(e.t) + (e.rp ? ` ${decodeDOW(e)}` : "");
     menu[label] = {
       value: e.on ? (e.timer ? iconTimerOn : iconAlarmOn) : (e.timer ? iconTimerOff : iconAlarmOff),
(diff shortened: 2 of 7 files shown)


storm64:
> Anyway, could the `rp `attribute be removed? The dow alone should be enough to understand if an alarm will run multiple times (dow > 0) or once (dow = 0). What do you think?

Sounds reasonable to me.

gfwilliams:
What if I want an alarm that fires once on Monday, but not every Monday? That feels like a useful thing to be able to do, but this logic would break that - a non-repeated alarm could no longer have days of the week set.

But it also seems to me this makes the logic more complicated when actually the fix is just to ensure that `dow` gets set to 127 rather than 0? That should be pretty much a one-liner?

gfwilliams:
> I can configure an alarm with repeat NO and days "MTW____": what should happen in this case? Today is Friday: the alarm will fire once on Monday and then it disables itself but then the TW days will be ignored because repeat is NO, not really clear imo.

Yes, that's true. It is a bit confusing. But I guess it's what you might expect.

How about for now we leave your menus as they are since they seem to make sense, but we leave the scheduler as-is so that if someone wants to do a custom app to enable no repeat with days of the week they can?

Then I think all you need to do is ensure that `if (!dow) dow=127` - which you could even maybe do in setAlarm?

alessandrococco:
> How about for now we leave your menus as they are since they seem to make sense, but we leave the scheduler as-is so that if someone wants to do a custom app to enable no repeat with days of the week they can?

Ok!

> Then I think all you need to do is ensure that `if (!dow) dow=127` - which you could even maybe do in setAlarm?

Ok but I need to make some tweaks in the alarm app because a dow of 127 will print the full MTWTFSS string. I'll update the PR.

gfwilliams:
That's great - thanks!

gfwilliams:
Great, thanks!

alessandrococco:
lol, it wasn't ready yet 😄 

I opened a new PR #1860

PR #1855, opened 2022-05-20
[Multitimer] fix for icon

frigis1:
Might've accidentally uploaded an older icon - this one should work on the Bangle. Don't think I should bother with a new version?


gfwilliams:
No, that's fine - thanks!

PR #1856, opened 2022-05-20
watch face with dino (t-rex from chrome) and weather

idavydov:
Hi,
I created a watch-face inspired by a watch-face from amazfit bit.

I slightly modified the 4x5 font (to make digits equal width). Days of the week were drawn by hand; I did it in the past for a custom amazfit watch-face; so shouldn't be any copyright issues with regards to the fonts.

I used weather icons from icons8 (set ios-glyphs).

Please let me know if you have any code style suggestions. I was thinking of splitting into multiple files, but it becomes difficult to use in emulator; perhaps there's a way.

screen1


gfwilliams:

Code style wise it looks good as well, especially the images defined in functions (so they don't take up RAM when not used).

Only thing I'd say is for the day of week you could use a normal font and the 'locale' module (http://www.espruino.com/Bangle.js+Locale#what-s-provided) and then it could display the shortened day in different languages. But it wouldn't look the same so I'm not sure it's really worth it.

NachtigallersLehrling:
I love the font and design of this app, so i edited the day-of-the-week font to display german abbreviations.
If anyone is interested, this font can be found in https://github.com/NachtigallersLehrling/BangleApps/tree/master/apps/dinoClock.

PR #1857, opened 2022-05-20
[BW Clock] Larger font size if colon is hidden to improve readability further.

peerdavid:
Its now possible to disable the colon in the time -- 2314 instead of 23:14. If the colon is disabled I also show the time with a larger font to improve readability further:

Colon enabled (fullscreen):
screenshot

Colon disabled (not fullscreen):
screenshot_3


gfwilliams:
Nice - thanks!

PR #1860, opened 2022-05-20
[Alarms & Timers] Fix repeat when user unchecks ALL the days

alessandrococco:
New PR because #1853 was merged too early 😄


gfwilliams:
Thanks! :)

PR #1861, opened 2022-05-21
added tags to the metadata

nerdmaennchen:
without those entries in the metadata the app won't show as watch face


gfwilliams:
Thanks! If you bump the version it can also be good to update the ChangeLog as well - but I'll do that now

gfwilliams:
Aha - there is no changelog :)

PR #1862, opened 2022-05-21
[Nifty Clock A&B] Bug fix and improvements

alessandrococco:
This PR fixes a couple of bugs in Nifty Clock A & B:

- Call setUI _before_ loading widgets (otherwise the "close widget" icon remains in the clock face)
- In Nifty Clock B there was a bug with black colour (the settings were read wrongly)

The PR also updates the clocks' Settings page.


gfwilliams:
Thanks!

PR #1863, opened 2022-05-21
[Sched & Multitimer] small fixes and updates (again)

frigis1:
Sched.boot.js didn't check for an alarm's custom js correctly so it always defaulted to sched.js - it SHOULD do so correctly this time, but that's what I thought it did during testing last time....

Multi Timer is updated for the time_utils module.

### DIFF
--- a/apps/multitimer/ChangeLog
+++ b/apps/multitimer/ChangeLog
@@ -1,1 +1,2 @@
 0.01: Initial version
+0.02: Update for time_utils module
--- a/apps/multitimer/alarm.js
+++ b/apps/multitimer/alarm.js
@@ -73,7 +73,7 @@ function showAlarm(alarm) {
   const settings = require("sched").getSettings();
 
   let msg = "";
-  msg += require("time_utils").formatTime(alarm.timer);
+  if (alarm.timer) msg += require("time_utils").formatTime(alarm.timer);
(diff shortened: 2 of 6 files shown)


gfwilliams:
Sorry, but this looks super weird. `active[0].js.slice(6, -2)` is very hacky

The docs show the JS field is just some JS to execute so it could be anything. This change will break anything that's not `load` in a very specific form:  https://github.com/espruino/BangleApps/blob/master/apps/sched/README.md#internals--library 

Why can't I just revert your previous change 2351340a73a354e377c504edec3c960d176137ae so the code is `Bangle.SCHED = setTimeout(active[0].js||'load("sched.js")',t);` - doesn't the old code work just fine while being much shorter?

frigis1:
Ooh, I apologize, I assumed the js field was only for loading a js file. In that case my changes were unnecessary, so yes, it should revert to the original code. Again, sorry.

gfwilliams:
No problem - thanks! I've just taken it out. If you remove the `apps/sched` changes from this PR I can merge in your multitimer changes?

frigis1:
If you don't mind, I'll re-submit a PR later, there are a couple more changes I'd like to make to multitimer before merging.

gfwilliams:
No problem :)

PR #1864, opened 2022-05-22
ClockFace: set Bangle.CLOCK=1 before loading widgets

rigrig:


alessandrococco:
lol, I just opened a similar PR (#1867), didn't notice this.

I thought CLOCK = 1 was set *in* the setUI method - Am I doing it wrong?

rigrig:
Yes, `CLOCK = 1` is set by `setUI`, but there's a bit of a chicken and egg problem:
* `Bangle.appRect` checks if `WIDGETS` is loaded, so we want to call `loadWidgets` _before_ `init`, in case the clock uses the Layout library. (or does other clever things with the widgets)
* We need to call `setUI("clock")` _after_ `init`, because if the clock does use the Layout library, that calls `Bangle.setUI()`, clearing the clock event handlers.

gfwilliams:
Ok, thanks. This definitely doesn't feel ideal and I'd prefer folks just used setUI of possible, but I guess this is in a library so it's not too bad. I'll add a comment to the lib explaining this though...

rigrig:
> I'd prefer folks just used setUI of possible

Well, maybe we could get rid of [this line (`Bangle.setUI(); // remove all existing input handlers`)](https://github.com/espruino/BangleApps/blob/5e4ad094e7eef048cb6fa94bb592ebff2f92982c/modules/Layout.js#L80) from the Layout library? So it only calls `setUI` if it needs to for Bangle.js 1 buttons or a `back` option.
Then we could just move `setUI` to the top in `ClockFace`.

But that would mean hunting down all code that now relies on Layout calling `setUI` :-(

alessandrococco:
Out of curiosity (I'm still learning the Layout module) : why the module needs to remove the handlers?

rigrig:
> why the module needs to remove the handlers?

If you define `btn` elements, the Layout library sets up event listeners to make them just work. The Bangle.js 1 doesn't have a full touch screen, so it uses `setUI({mode:"updown"...` to scroll through them with the physical buttons.
And for users of the library it's easier to just remember "Layout calls `setUI`", than "Layout might sometimes call `setUI`". Especially if it works differently between the Bangle.js 1 and 2.

So I guess simply removing that line won't really do, maybe make it conditional on there being any `btn` elements?

gfwilliams:
I've just had to refactor Layout a bit to fix the `run` app, so we could have an `options.noSetUI` quite easily. It feels like the one `CLOCK=1` hack in a library is better than the few tiny hacks we'd have to implement to do this nicely though ;)

PR #1865, opened 2022-05-22
widclose: remove widget if Bangle.CLOCK is set after loading the widget

rigrig:


rigrig:
Well, clocks using `ClockFace` (oops), [this PR](https://github.com/espruino/BangleApps/pull/1864) fixes that, but I figured I would try and mitigate it on the widget side as well.

But you're right: we should just fix clocks, and this only hides the problem.

Issue #1466, opened 2022-02-17
[Feature request] Hideable widget bar


gfwilliams:
Just a thought here, but I see a lot of clock apps now having an option to hide the widgets.

Maybe we could have a library (in modules) that adjusts the widget draw call such that it renders widgets to an offscreen buffer. Swipe-down could then make the widget bar slide out of the top of the screen, and after a few secs it could pop back up.

Having it as a library would mean we could avoid a bunch of code duplication too.

andiohn:
This is an excellent idea, I used a watchface without a widget zone was disappointed that they weren't still running. Great idea!

rigrig:
I wonder if this needs to be a library, or an installable boot code app that "just works" for all clocks?

gfwilliams:
It's an idea - it's just getting clocks to work with it. Many of them will just assume there's always a widget bar and leave 24px spare all the time.

Also, I guess we want it to only work for clocks, not *all* apps?

Rarder44:
`Also, I guess we want it to only work for clocks, not all apps?`
I think it is also useful to extend it to apps.

the downside is that the "Swipe-down" command cannot be used (unless overwritten) and I don't know if it could interfere with scrolling through the menus ....

in any case, I think it is good to have a "standard" that manages the visibility of the widgetbar (personally I don't use it and the apps I use / have written don't display the widgetbar in order to have more screen space)

rigrig:
> getting clocks to work with it. Many of them will just assume there's always a widget bar and leave 24px spare all the time.

Sure, but just having that empty space for a "cleaner" look would be a nice start, and then we can just convert clocks/apps to properly use `appRect` when it comes up.

>> I guess we want it to only work for clocks
>
> the downside is that the "Swipe-down" command cannot be used

Maybe we could split it into two separate options?
* Clock widgets: Show/Hide/Swipe
* App widgets: Show/Hide/Swipe -> If you choose "Swipe" and it interferes with an app, that's your own fault... (But probably we can make it so that you just won't get the widget bar at all in 90% of problem cases.)

> I don't use it and the apps I use / have written don't display the widgetbar

That should be possible by just not installing any widgets: if code is doing something it should go into a `boot.js` file.
(Not saying it works like that now, just how I'd like things to be)
    
For non-clock apps, I feel trading the widgetbar for screen space is fine, but for clocks it would be nice if they looked at `appRect`, and at least can work for users that do want widgets.

rigrig:
[...]

PR #1867, opened 2022-05-22
[ClockFace] Fix setUI/Bangle.loadWidgets order, new option "loadWidgets"

alessandrococco:
I moved the Bangle.loadWidgets call *after* the setUI so CLOCK gets defined (so widclose won't be shown in the clock). I also add a new option "loadWidgets" (default true) so a developer can easily enable/disable the widgets in their clock.

@rigrig I took the liberty of making a couple of changes, hopefully that's not a problem! I'm working on a custom clock and your ClockFace is really cool, thanks!

### DIFF
--- a/modules/ClockFace.js
+++ b/modules/ClockFace.js
@@ -8,6 +8,7 @@ function ClockFace(options) {
   Object.keys(options).forEach(k => {
     if (![
       "precision",
+      "loadWidgets",
       "init", "draw", "update",
       "pause", "resume",
       "up", "down", "upDown"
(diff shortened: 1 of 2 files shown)


rigrig:
I'm always happy to see my code improved, however I have some concerns about this PR:

As I mention in [this PR](https://github.com/espruino/BangleApps/pull/1864#issuecomment-1133971819), moving `loadWidgets` after `init` will mess up `Bangle.appRect`, meaning all clocks that use the Layout library.

But more importantly, I don't think we want _clocks_ to disable the widgets, I think it should be a global setting. That way clocks can use `Bangle.appRect` to work either way, and users can decide if they want widgets independently from which clock they install.

alessandrococco:
> As I mention in this PR, moving loadWidgets after init will mess up Bangle.appRect, meaning all clocks that use the Layout library.

This could explain the weird issues I'm encountering (it's my first time with both ClockFace and Layout). Good to know, thank you!

> I don't think we want clocks to disable the widgets, I think it should be a global setting.

A global setting (+ #1466) would be awesome!

Thank you for your comment

I close this PR.

PR #1868, opened 2022-05-23
Alternative waypoints file names

nujw:
Multiple GPS apps use the waypoints.json file for waypoints. Adds an option for alternative waypoints file names allowing other gps apps to be used without having to replace the speedalt2 waypoints.


gfwilliams:
Thanks! However I think you need to update the metadata to mention the different file types. I think for `data` files you can use a wildcard to deal with the different suffixes.

Is the issue that installing a new app overwrites the waypoints file? Because I guess the intention was that the waypoints could always be shared...

nujw:
Hi Gordon,
Thanks for the update.

I have two use cases :
- To be able to hold multiple sets of waypoints on the watch and choose which one to use for the current situation. eg. One set to use while sailing and other sets for different motorcycle trips.
- To have waypoint files with alternative names. A couple of times I have tried out other GPS apps to find that my waypoints have been removed when the other app is removed.

Would something like this achieve what you are suggesting ?
  "data": [
    {"name":"speedalt2.json"},
    {"name":"waypoints1.json"},
    {"name":"waypoints2.json"},
    {"name":"waypoints3.json"}
  ]

Thanks
Michael

gfwilliams:
Ok, thanks! Yes, that should work - although it seems wildcards are possible too.

> A couple of times I have tried out other GPS apps to find that my waypoints have been removed when the other app is removed.

Ahh, I wonder whether that is a separate issue - that the app loader should be smart enough to know if a data file is used by other apps before getting rid of it. 

One thing I'm wondering though: It's going to be a pain to have all apps require the same 'waypoint file selector', and suppose someone wants 4 waypoint files - it'll require changing all apps. It might be better to have a `waypoints` app and library that every app that needs waypoints depends on?

That way, you get to share code and potentially have as many waypoints files as you want, but also it fixes the above problem. As long as you have the waypoints app installed, you keep all your waypoints.

PR #1869, opened 2022-05-24
[Multi Timer] update

frigis1:
Update Multi Timer for time_utils module, plus a few tweaks.

### DIFF
--- a/apps/multitimer/ChangeLog
+++ b/apps/multitimer/ChangeLog
@@ -1,1 +1,2 @@
 0.01: Initial version
+0.02: Update for time_utils module
--- a/apps/multitimer/alarm.js
+++ b/apps/multitimer/alarm.js
@@ -73,7 +73,7 @@ function showAlarm(alarm) {
   const settings = require("sched").getSettings();
 
   let msg = "";
-  msg += require("sched").formatTime(alarm.timer);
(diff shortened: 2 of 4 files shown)


gfwilliams:
Great, thanks! Just wondering, why do you do:

```
a.js = "(require('Storage').read('multitimer.alarm.js') !== undefined) ? load('multitimer.alarm.js') : load('sched.js')";
// and not
a.js = "load('multitimer.alarm.js')";
```

Because to me it seems like you're **certain** `multitimer.alarm.js` exists, because it's part of your app. Or are you trying to deal with the case where someone's set a timer and then uninstalled your app in the mean time?

frigis1:
That's right, in case my app gets uninstalled and the user gets stuck with an alarm/timer that only shows a "Loading..." message. That was actually why I was mucking around with sched, not quite realizing I could do this in my own app.

gfwilliams:
Ahh, ok - thanks! Yes, that makes sense then... I guess maybe at some point we need to think about having the ability to run some 'cleanup' code when an app is uninstalled.

... or maybe actually since `appid` is set, `sched` should check if `require("Storage").read(a.appid+".info")` exists... It seems overkill for the boot code to have to do it though...

PR #1870, opened 2022-05-24
New app: Icon Launcher

TheGLander:
This launcher is inspired by smartphones which have a lot of icons per screen. As the README states, my code can almost use `E.showScroller`, but it doesn't report the X of the tap on the Bangle.js 2, which is required for the launcher. It also has settings which reuse the normal launcher config file, for easier adoption.


gfwilliams:
Looks great - thanks!

PR #1872, opened 2022-05-24
messages: Add message icons to the widget

TheGLander:
This PR makes the messages widget list (colored) icons instead of the simple mail icon.
It moves the icons from the app to the message library, as both the app and the widget use them now.
There's also a tangentially related bug fix for the Android integration app, as it's very obvious that there are two message types for one SMS message with the new icons.

### DIFF
--- a/apps/messages/lib.js
+++ b/apps/messages/lib.js
@@ -105,13 +105,6 @@ exports.clearAll = function(event) {
     WIDGETS.messages.update(messages);
 }
 
-function getNotificationImage() {
-  return atob("GBIBP//8H//4n//5z//z5//n8//H+P+f/H4//n5//Dw/+Jkf8YGP48PH5//nz//zn//5H//4P//8");
-}
-function getFBIcon() {
-  return atob("GBiBAAAAAAAAAAAYAAD/AAP/wAf/4A/48A/g8B/g+B/j+B/n+D/n/D8A/B8A+B+B+B/n+A/n8A/n8Afn4APnwADnAAAAAAAAAAAAAA==");
-}
-
 exports.getMessageImage = function(msg) {
   /*
   * icons should be 24x24px with 1bpp colors and 'Transparency to Color'


gfwilliams:
Thanks! And for the recent merge...

I like the idea of moving the message icons into the library, and the SMS fix seems good... However I'm a bit concerned about the widget image stuff - that it could be quite slow/cause memory problems (on Bangle.js 1) with it loading the whole message list each time it flashes.

Also we did have a `if (global.MESSAGES===undefined)` so the widget didn't buzz/flash when in the messages app, because at that point I guess you know you have new messages - but it looks like that check has gone now?

Maybe we could instead just pass a list if icon names into `WIDGETS.messages.show` from the messages library? Because at that point we already have the messages list loaded, and it would be easy to grab the `msg.src` out - and then the widget itself hopefully stays pretty simple?

What I'll do for now is merge the other bits of this though...

TheGLander:
I haven't thought about Bangle.js 1 performance, I think the icon name list idea is cool.

As for the message app check, I think I removed it on accident.

gfwilliams:
Thanks for your work on this - please can you let me know when you think it's ready?

I noticed it seems there are some changes in [apps/messages/lib.js](https://github.com/espruino/BangleApps/pull/1872/files#diff-73da835ccb0c190b9545584c57154f5d990da6450c16586486edd30a7afa8d8b) that look like they might revert stuff that was done recently (inlining the FB icon)

TheGLander:
@gfwilliams I think it's ready, sorry for the delay

apps/messages/lib.js
```diff
@@ -102,7 +102,14 @@ exports.clearAll = function(event) {
   if (inApp) return onMessagesModified();
   // if we have a widget, update it
   if (global.WIDGETS && WIDGETS.messages)
-    WIDGETS.messages.hide();
+    WIDGETS.messages.update(messages);
+}
+
+function getNotificationImage() {
```
gfwilliams, line 108: Thanks! However are `getNotificationImage` and `getFBIcon` needed now? I think they might be left over from before the most recent changes

TheGLander: Fixed in [f3d3091](https://github.com/espruino/BangleApps/pull/1872/commits/f3d309170675929ec667f0d44dba88e3eea83028)

gfwilliams:
Great - thanks!

PR #1876, opened 2022-05-24
R2D2 Clock v1

nh-99:
- Adds new clock face based on Interlaced Clock that has R2D2 on it

I apologize for the dirty commit history on this. I can take a stab at fixing that if you want before merge.


gfwilliams:
No worries about the commit history - this looks great, thanks!

PR #1878, opened 2022-05-24
[rebble]  remove "mylocation" dependency - custom html page

Rarder44:
since I don't like the idea of ​​installing an extra app just to be able to choose the necessary coordinates for sunrise and sunset, I decided to remove the "mylocation" dependency and create an html page that allows you to load the coordinates during installation.

I put this PR as a draft for a few reasons:
- I found a problem in the simultaneous use of the interface.js and customize.js libraries (described in the comments of the html page, I think I will open a dedicated issue)

- I want to understand how is the policy for the use of external libraries (in this page I used jquery), if we can use them but referring to an external hosting, if we download the library or if we don't use any libraries.

- it would be nice to be able to view, even after the "app" has been installed on the watch, the "3 bars" icon to open the app edit menu
(i have implemented the possibility to change the city without necessarily reinstalling the whole app)

### DIFF
--- a/apps/rebble/custom.html
+++ b/apps/rebble/custom.html
@@ -1,0 +1,143 @@
+<html>
+    <head>
+        <link rel="stylesheet" href="../../css/spectre.min.css">
+        <script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>
+    </head>
+    <body>
+        
+        Select the city (coordinates) for sunrise and sunset<br>
+        The service used to search the city is
+        <a href="https://nominatim.openstreetmap.org/ui/search.html" target="_blank">https://nominatim.openstreetmap.org/ui/search.html</a><br><br>
+        <p>Search: <input type="text" id="query" class="form-input" list="exampleList">
+
+            <datalist id="exampleList">
+            </datalist>
+      
+        </p> 
+        <p>
+            This data will be saved on your Bangle.js ( you can edit these fields )<br>
+            Leave blank if you don't use the service
+            <table>
+                <tr>
+                    <td>name:</td>
+                    <td><input id="cityName"></td>
+                </tr>
+                <tr>
+                    <td>lat:</td>
+                    <td><input id="lat"></td>
+                </tr>
+                <tr>
+                    <td>lon:</td>
+                    <td><input id="lon"></td>
+                </tr>
+            </table>
+        </p>
+  
+      <p><button id="saveOnly" class="btn btn-primary">Save on Bangle.js</button> <button id="upload" class="btn btn-primary">Upload</button></p>
+  
+      <script src="../../core/lib/interface.js"></script>
+      <script>
+        const LOCATION_FILE = "mylocation.json";
+
+        //TOFIX:
+        //Need this because adding interface.js break customize.js 
+        //"Identifier '__id' has already been declared"
+        //so i removed the customize.js and copied the piece of code to send a custom app
+        function MysendCustomizedApp(app) {
+          console.log("<CUSTOM> sending app");
+          window.postMessage({
+            type : "app",
+            data : app
+          });
+        }
+
+        //I don't know how is the policy for using external libs (like JQuery),
+        //at the moment i used it; in case i try to implement everything in standard js
+        $(document).ready(function(){
+
+            query.oninput=function(){
+                var val = document.getElementById("query").value;
+                var opts = document.getElementById('exampleList').childNodes;
+                for (var i = 0; i < opts.length; i++) {
+                    if (opts[i].value === val) {
+                        data =JSON.parse($(opts[i]).attr("data"));
+                        $("#cityName").val(data["name"]);
+                        $("#lat").val(data["lat"]);
+                        $("#lon").val(data["lon"]);
+                        NoSearch=true;
+                        return false;
+                    }
+                }
+            };
+
+
+            var NoSearch=false;
+            var timeoutSearch;
+            const timeSearchDelay=300; 
+            $("#query").on('input',function(){
+                if(NoSearch)
+                {
+                    NoSearch=false;
+                    return;
+                }
+                clearTimeout(timeoutSearch);         //I use a timer every time I press to avoid making too many requests
+                timeoutSearch = setTimeout(function(){
+                    $.get("https://nominatim.openstreetmap.org/search",{q:$("#query").val(),format:"json",addressdetails:1},function(data){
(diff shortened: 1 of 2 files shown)


gfwilliams:
I'm really not sure this is a good idea... So now every app that wants location is supposed to have a its own duplicate HTML page, and you have to set each one up to the same location manually each time you install a new app?

Also because it's custom you won't be able to apply updates to apps automatically any more because they need user intervention?

> I don't like the idea of ​​installing an extra app

The app itself is trivially small (less than 100 lines). What is it you don't like? Is it just the idea? Or is it just the entry in the App Loader that you have a problem with? Because we could just make it `mylocation.settings.js` so then the location menu is only accessible from `Settings->Apps` and is basically invisible to you.

> I found a problem in the simultaneous use of the interface.js and customize.js libraries

That's because interface.js is for interface.html files and customize.js is for custom.html files. You definitely shouldn't use one for the other type of app because it's entirely likely it'll break - if not now, in the future. I guess we should really add some kind of check in there to stop it.

If you need something maybe we can make sure it gets added to customize.js though. BUT....

> it would be nice to be able to view, even after the "app" has been installed on the watch, the "3 bars" icon to open the app edit menu

That's what the `interface.html` is for (changing/reading from an app after it's installed). Perhaps what we need is actually not a `custom.html` that can be called after, but an option in the metadata that says `install app and run interface.html right after`? It seems that'd likely solve the issue for you?

In fact, it seems to me that what might be preferable is having `mylocation` with the HTML page as an `interface.html` so you can actually set your location nicely from the app loader if you want to, maybe in addition to the on-bangle settings page rather than an app? Then everyone can benefit from it...

> I want to understand how is the policy for the use of external libraries (in this page I used jquery)

Well, generally I'd say:

* Try not to use stuff just for the sake of it. In this case I'd say your use of jquery could trivially be replaced by standard JS functions like `document.getElementById` (and in fact you seem already to do this in some places) 
* But if you do need something, just get it from a CDN
* ... unless it's something people are using a lot, in which case we can look at including it in https://github.com/espruino/EspruinoAppLoaderCore/tree/404e981834f2e8df9c505a8fab12ae12fe3bd562/lib

Rarder44:
> I'm really not sure this is a good idea... So now every app that wants location is supposed to have a its own duplicate HTML page, and you have to set each one up to the same location manually each time you install a new app?

ok I understand the problem, but it seems "wasted" to have to install an app (even if it's only 100 lines) to manage a single coordinate.

> In fact, it seems to me that what might be preferable is having mylocation with the HTML page as an interface.html so you can actually set your location nicely from the app loader if you want to, maybe in addition to the on-bangle settings page rather than an app? Then everyone can benefit from it...

I hadn't gotten to this point in your speech yet, and I was thinking the same thing xD
so I'd say it's a great solution!
I think the installation of the app should be "optional": since the coordinate is a little "dynamic" information (I set it once and never change it ...) it would be enough to be able to set it through interface.html and for those who want change the coordinate via GPS, can always install the app.


> That's because interface.js is for interface.html files and customize.js is for custom.html files. You definitely shouldn't use one for the other type of app because it's entirely likely it'll break - if not now, in the future. I guess we should really add some kind of check in there to stop it.

ok, my mistake!
I was not aware of the difference between custom.html and interface.html; I'm definitely going to look at the documentation again!

My "problem" is that I was thinking of an app that needs an ID related to an online service.
so I was thinking of setting it through the custom.html page but every time I update the app I would like to see that ID and choose whether to change it or not.
the custom.html page does not implement those quick features for reading a JSON file on the device.

I have seen that all of the features of customize.js are present in the interface.js (except the sendCustomizedApp) while the "message" listener is duplicated but in one version there is an extra control (I think there is some of inconsistency ...)
at this point isn't it better to keep one?


sorry if i can't be clear but i struggle to express myself in english.

gfwilliams:
Thanks! If you find there's something confusing/missing in the docs about interface/custom please let me know!

> the custom.html page does not implement those quick features for reading a JSON file on the device.

I've just added them. You also need to ensure you have `"customConnect": true,` in metadata to force a watch connection.

The issue is that in the App Loader, custom&interface are loaded in different ways. `custom` is sitting there waiting to get a `sendCustomizedApp` (in fact it doesn't even require a connection to the Bangle), but `interface` just loads the page (with an active connection to the Bangle). They do share some code but the pages work in different ways.

> every time I update the app I would like to see that ID and choose whether to change it or not.

Do you really mean that? There was a complaint recently about having customisable apps pop up during the updater process, so that had to be disabled. So if you were to use `custom` then the app would not auto-update, and you'd always have to update it manually.

But honestly this is a small thing, and if you want to make it a bit more painful for users of your app then it's not really up to me to stop you.

So if you tweak the html to use `customize.js` then I'll pull this in? 

Longer term I'd really prefer this just to be in `mylocation` though (and without jQuery) - I really don't see not wanting to install a very small extra app as a reason to make things more painful for users (and maintainers).

Rarder44:
> I've just added them. You also need to ensure you have "customConnect": true, in metadata to force a watch connection.
thank you!

> Do you really mean that? There was a complaint recently about having customisable apps pop up during the updater process, so that had to be disabled. So if you were to use custom then the app would not auto-update, and you'd always have to update it manually.
But honestly this is a small thing, and if you want to make it a bit more painful for users of your app then it's not really up to me to stop you.

i think i can rethink how the app works by implementing just interface.html

> (and without jQuery)

understandable

> Longer term I'd really prefer this just to be in mylocation though (and without jQuery) - I really don't see not wanting to install a very small extra app as a reason to make things more painful for users (and maintainers).

ok, understood, I don't know if you think an interface page is useful to be able to easily set the coordinates of mylocation without using the GPS (so you can also set the name of the custom city and remove the ???).
If so, I'll add it.


> Or is it just the entry in the App Loader that you have a problem with? Because we could just make it mylocation.settings.js so then the location menu is only accessible from Settings->Apps and is basically invisible to you.

maybe this could be a good compromise.
if you want to implement it.

gfwilliams:
> Because we could just make it mylocation.settings.js so then the location menu is only accessible from Settings->Apps and is basically invisible to you.

Ok, just done! I don't know if you want to just add your HTML code to that as an 'interface.html' file?

Rarder44:
> Ok, just done! I don't know if you want to just add your HTML code to that as an 'interface.html' file?
great!
yes; later undo the changes to rebble, clean up the html / js code and put it in mylocation.

I close the pr, I will open another dedicated later

PR #1881, opened 2022-05-25
[Alarms & Timers] Fix dow handling for new timers

alessandrococco:
I just noticed that a timer is saved with a wrong value for `dow` (254 instead of 127) if first day of week is Monday. This PR fixes the bug.


gfwilliams:
Thanks!

PR #1882, opened 2022-05-25
[r2d2clk] Add screenshot to metadata.json

nh-99:
The screenshot was missing from the metadata in the app store initially, so this just adds the screenshot in so it will show up.


gfwilliams:
Great - thanks!

PR #1884, opened 2022-05-26
WIP: android: added calendar sync events

glemco:
Quick and dirty writing a JSON on storage, in the future we could use a library (like messages and sched) providing everything needed for calendar events.
Leaving it like that could also be fine as most likely no other apps would be able to modify events (can they be synced back to GB at all?) and all they would need to do is to read from the JSON.


gfwilliams:
Thanks! Only thing I'd say is it'd be very easy to get out of sync - so events were on the Bangle but not on Gadgetbridge

Not sure what you think but maybe it makes sense to just send all the events in one big packet? We only want to send them for the next 48 hours or so anyway or we'll swamp the bangle

glemco:
I didn't dig into how GB is doing, but I inspired on how it is handled in the pebble (the only device supporting this), so I bet such problems are addressed already for it. It may not be the best way but I just didn't want to reinvent the wheel and used the implementation already present in GB.
I read somewhere that only events in the next 3 days are synchronised and this actually kind of works because if I move an event far away it gets deleted. I may need to test this more, also I'm not sure how events already in the calendar are handled, but for this to be more comfortable I guess I need a calendar app to show what's on the device.
If you have concerns don't merge the pull request until I'm done with the agenda and testing.

glemco:
After a few days running it and after checking GadgetBridge under the hood, that's basically how it works: GB doesn't know directly what is on the bangle's database, but it keeps a database table that supposedly mirrors its content. Whenever an event is added on the bangle, it's added on that table too and the same happens on deletion.
To sync with the actual android calendar, GB fetches periodically a list of the event from today to the next 7 days (maybe that could be changed, I read old issues talking about 3 days), then compares the internal database (the one mirroring the bangle) and adds/deletes events accordingly. Practically speaking every day the events of the previous days will be removed and the ones of 7 days from now added, as all events in the middle are supposedly already saved. Created/deleted or edited events in that time window are synchronized directly.

glemco:
Now I see some problems if the two tables (GB and bangle) get out of sync, but I could see this only in two conditions:
1. GadgetBridge sends the update and the bangle acknowledges it but doesn't apply it (happens if the GadgetBridge debug app is running on the bangle).
2. We connect to a new instance of GB when the database on the bangle is not empty (e.g. it synchronized earlier and for whatever reason we deleted the data on GB, synchronizing again might produce duplicates or not delete what ought to be).

In the end of the day we could play safe and delete all older events whenever we disconnect to GB (although GB should do it anyway in the majority of cases). Or ignore the 2 cases altogether as they happen in very specific cases and when users know what they are doing.
What do you think?
Would it be possible instead to catch the event as in 2. and clear the database. Or forward events caught by the GB debug app to the system to prevent also 1. from happening?

gfwilliams:
Sorry for the delay - I was on holiday. This looks good though.

In terms of keeping in sync, I guess what you've got with the transmission of IDs is good. I'd have said  you could maybe transmit *all* calendar IDs in one go when Gadgetbridge connects, but if we assume 10 events a day and 7 days, transmitting 70 events could be serious overkill.

glemco:
Thanks for the answer, hope you enjoyed your holidays!

I have been exploring the GadgetBridge code even more, some models (Huami and Mi bands), although they don't support calendar events, do synchronize all of them together as alarms, whenever the time is synchronize with GB.
I tried this approach but still don't really like it, the synchronization isn't really frequent (it basically happens only if I force it in the GB debug settings) and I don't know how big a BLE packet could get.

With the other implementation (pebble like) events are synchronized every time there's a change in the calendar, rich doesn't happen on Mi bands, for instance.
To cope with out of sync db, I implemented a new command that sends all event IDs to GB, the app will then check the internal database and update it accordingly, the next update from the calendar will hence be consistent with both the bangle and GB (I should actually force a synchronization too but still didn't manage to). This command can be sent periodically or forced in settings.
Currently I kept both implemented to do some tests (if calendar sync is enabled, GB uses the complex approach, otherwise the other), they could technically both stay, or perhaps the simpler one could work with banglejs 1, if you think it meaningful

glemco:
> _However_ do you think we could make `forceCalendarSync` something that is just called in response to a message from Gagdetbridge so Gadgetbridge is responsible for kicking things off? That way we're not having to set up any timeouts or hooks, so we're not using any extra RAM to add the calendar integration, which would be really handy.

That's a good idea, I can move the exact same timer to GB and use another message for this.

> Also I'm pretty sure `NRF.once` doesn't exist (just `.on`), so the code you have there won't work.

Damn, I remember it exists in node and didn't check Espruino's docs enough.. Good to know

glemco:
Moved the forceCalendarSync in GB and tested it a little, if it looks fine for you we can start to merge.
(after this merge I'll ask the GadgetBridge people for merging, until that gets approved too this functionality won't work, of course)

gfwilliams:
Looks good to me! Merging...

You can always do a PR to https://github.com/gfwilliams/Gadgetbridge and I'll merge that, which will then work its way into mainline in the end (but otherwise it's used for the Play Store app)

glemco:
Mmh, I actually opened one already in mainline a while ago, thinking it would have required time for reviewing..
But I'll keep that in mind for the next times!

PR #1886, opened 2022-05-26
Pong Clock

pidajo:
A Pong playing clock


gfwilliams:
Thanks! Looks like you've accidentally included `widday` as well though, and that has some issues - like having the `app.info` file and not having a `metadata,json` (I think you named it .info?)`

You also don't need apps/pongclock/pongclock.img

pidajo:
thanks

gfwilliams:
Great! Thanks for getting all that sorted - just merging now!

PR #1887, opened 2022-05-26
Baroalarm v0.04

myxor:
* Use Prompt with dismiss and pause
* Improve barometer value median calculation

### DIFF
--- a/apps/widbaroalarm/ChangeLog
+++ b/apps/widbaroalarm/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Do not warn multiple times for the same exceedance
 0.03: Fix crash
 0.04: Use Prompt with dismiss and pause
+      Improve barometer value median calculation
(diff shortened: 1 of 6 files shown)


trufae:
A similar app would be nice for paragliding. Maybe you can tweak that to support this use case. Let me give some context. When doing paragliding you rely on hot wind to gain altitude. When you don't have references or known hot spots it's useful to have a device that vibrates or beeps (in bangle usecase should be vibrating), when you gain altitude, so you know you must stay in this location to go up. When the altitude is going down nothing beeps or vibrates.

This could be an idea for another app. but just wondering if you had this in mind to integrate it here :)

gfwilliams:
Looks good - you happy to merge?

The whole 7x getPressure thing might be better accomplished by just turning the barometer on, responding to the 'pressure' event and then turning it off when you're done though?

myxor:
Thank you for your feedback. I will rework the barometer handling and some other things.

Issue #1873, opened 2022-05-24
[Feature requests]: Messages: Configuration of buzz strength & do not power on LCD on message-auto-close


myxor:
**Messages:**
1. I would like to have a configuration for the buzz strength.
2. It would be nice if the LCD would not turn on when an unread message is being auto-closed after the delay.

I can work on these myself but i am busy right now.
Leaving this here so i don't forget it.

alessandrococco:
> 1. I would like to have a configuration for the buzz strength.

I believe the current strength is already at its maximum. The [doc says](https://banglejs.com/reference#l_Bangle_buzz) 

> strength - Power of vibration from 0 to 1 (Default 1)

The messages app call the buzz with default values. What we can do is increase the _duration_ (default is 200ms but we can allow user to customize this value).

I can work on it later today.

gfwilliams:
> The messages app call the buzz with default values. 

How about we use the `buzz` library and then extend it, so you can have not just `.` and `-` but something else that's a lower power buzz. Then in all things (like alarms too) you could have a mix of powerful and less powerful buzzes.

myxor:
> The current strength is already at its maximum. 
 
I do not want it to be stronger but rather weaker :)

gfwilliams:
That would be good if it wasn't for existing apps that use it, but I guess it's not the end of the world to have them buzzing a bit weaker until they reinstall or update the settings

gfwilliams:
Great! But you do realise that's gone from 14 lines up to 54? And if you don't use one of `:.-=` it just fails to resolve the promise now?

I yours is more readable now, but it's going to use more memory and probably be slower. Couldn't we just do it by adding 2 lines to the if...else chain?

alessandrococco:
Done!

alessandrococco:
> It would be nice if the LCD would not turn on when an unread message is being auto-closed after the delay.

@myxor Does the fixed version work for you?

This is the fix I pushed last week: https://github.com/espruino/BangleApps/pull/1890/files#diff-97acef7b0efa75daea98d9fec30c299f193c5daaa88c7928c569915eb0a1e2f7I tried the change before I opened the PR and  I am 100% (99.999999%) sure it worked but I just noticed that now the screen turn on again. I reinstalled the latest version (from dev app loader) but the issue is still here - Does it only happen to me?

myxor:
@alessandrococco no, it does not work for me. The screen still turns on on auto closing messages.

[...]

PR #1889, opened 2022-05-26
[buzz] Add support for new patterns

alessandrococco:
As discussed in #1873 this PR adds two new vibration patterns: `:` and `=`.

`.` now is a "short and weak" vibration and `:` is a "short and strong" vibration. Similarly `-` is "long and weak" and `=` is "long and strong" where 
- "strong" means that the Bangle.buzz() is called with strength = 1 and 
- "weak" means the function is called with strength = 0.25.

Feel free to propose some common patterns for `buzz_menu`!

Current values are  `var patterns = ["", ".", ":", "..", "::", "-", "=", "--", "==", "=.=", "---"];`

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -10,3 +10,6 @@
 0.09: Move some functions to new time_utils module
 0.10: Default to sched.js if custom js not found
 0.11: Fix default dow
+0.12: Update default buzz patterns to new values
+      Improve timer message using formatDuration
+      Fix wrong fallback for buzz pattern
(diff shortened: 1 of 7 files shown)


nxdefiant:
Feature request: Can we have a 200ms buzz? I think ',' is still free.

alessandrococco:
> Feature request: Can we have a 200ms buzz? I think ',' is still free.

Sure!

I can add `,` and `;` for "medium weak" and "medium strong" :)

nxdefiant:
Cool, thanks.

nxdefiant:
One other thing: Shouldn't the Default Alarm & Timer of the sched module changed to the new strong value?

alessandrococco:
> One other thing: Shouldn't the Default Alarm & Timer pattern of the sched module changed to the new strong value?

You're right, done!

gfwilliams:
Looks great, thanks!

Just just in general for something like this, global (in the module) constants are not a good idea as unless they get minified and removed, they just sit there eating RAM.

PR #1890, opened 2022-05-26
[Messages] Don't turn on the screen after unread timeout expires

### LINKED ISSUE #1873: [Feature requests]: Messages: Configuration of buzz strength & do not power on LCD on message-auto-close

alessandrococco:
Closes #1873 

---

I didn't add an option for this because I think it is wrong that the screen turns on when the unread timeout expires but I'm open to discussion :-)


johan-m-o:
Yes! This has been bugging me without even realising!

gfwilliams:
Thanks!

PR #1892, opened 2022-05-27
Calibration - Scale the inputs to account for linear variation

halemmerich:
I have tested this with two bangles and was slightly surprised by the different results:
```
{
  "xoffset": 0,
  "yoffset": -19.5,
  "xscale": 0.90285714285,
  "yscale": 1.04290429042,
  "active": true,
  "reload": false
}
```
```
{
  "xoffset": -8,
  "yoffset": -7.5,
  "xscale": 0.94894894894,
  "yscale": 0.95468277945,
  "active": true,
  "reload": false
}
```
Despite this it seems work well so far. Will test a bit longer before removing the draft marker.

### DIFF
--- a/apps/calibration/ChangeLog
+++ b/apps/calibration/ChangeLog
@@ -1,0 +1,2 @@
+1.00: New App!
+1.01: Use fractional numbers and scale the points to keep working consistently on whole screen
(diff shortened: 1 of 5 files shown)


halemmerich:
So far no problems in daily use, seems to work well.

gfwilliams:
Great - thanks!

PR #1893, opened 2022-05-27
ClockFace: add settings support

rigrig:
The idea is you can specify a `settingsFile` in options, and it will be loaded directly into the clock object. (i.e. not inside a separate `settings` property, as I figured that only makes accessing settings more verbose.)

`loadWidgets` is special: it defaults to true, when `false` it prevents the call to `Bangle.loadWidgets()` (and we check if `WIDGETS` is loaded before calling `drawWidgets()`

`showDate` doesn't do anything special, but it does default to `true`

Also adds `ClockFace_menu`, although at the moment it only has two boolean menu options... Still cuts down on the boilerplate though.

### DIFF
--- a/modules/ClockFace.js
+++ b/modules/ClockFace.js
@@ -10,7 +10,8 @@ function ClockFace(options) {
       "precision",
       "init", "draw", "update",
       "pause", "resume",
-      "up", "down", "upDown"
+      "up", "down", "upDown",
+      "settingsFile",
     ].includes(k)) throw `Invalid ClockFace option: ${k}`;
   });
   if (!options.draw && !options.update) throw "ClockFace needs at least one of draw() or update() functions";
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Merging, but...

I notice when using the settings, you do:

```
  const menu = {
    "": {"title": /*LANG*/"Bar Clock"},
    /*LANG*/"Show date": require("ClockFace_menu").showDate(s.showDate, saver('showDate')),
    /*LANG*/"Load widgets": require("ClockFace_menu").loadWidgets(s.loadWidgets, saver('loadWidgets')),
   // ...
  };

  E.showMenu(menu);
```

Do you think it would make sense to just have:

```  
const menu = {
    "": {"title": /*LANG*/"Bar Clock"},
   // ...
  };
  require("ClockFace_menu").addMenu(menu, saver, { date:true, loadWidgets:true });
  E.showMenu(menu);
```

Then we keep all the names for menu items the same?

rigrig:
> Do you think it would make sense to just have `require("ClockFace_menu").addMenu(`

Thanks, that's much cleaner. I will rewrite it to do that. (Good thing I didn't get round to converting a ton of clocks yet ;-)

PR #1895, opened 2022-05-27
new clock: Cog clock

rigrig:
Actually made this a while ago, but now converted it to use `ClockFace`, so it needs [this PR](https://github.com/espruino/BangleApps/pull/1893) to land first.

Bangle.js 1 only (for now), because the custom font is too large to make it look good on the Bangle.js 2 :-(


gfwilliams:
Nice!

PR #1896, opened 2022-05-28
Added widshipbell

nxdefiant:
A widget that buzzes according to a nautical bell, one strike at 04:30, two strikes at 05:00, up to eight strikes at 08:00 and so on.

See https://www.westmarine.com/WestAdvisor/Ships-Bell-Time for bell pattern list


nxdefiant:
Hold until #1889 is merged.

gfwilliams:
Thanks!

PR #1900, opened 2022-05-29
[Alarms & Timers] Fix "Enable All"

alessandrococco:
I found a bug with a disabled timer + _enable all_ action: the _enable all_ does not update the `timer` field. This PR fixes the bug :-)


gfwilliams:
Thanks!

PR #1901, opened 2022-05-29
[Settings] Improve "Turn Off" user experience

alessandrococco:
The _Turn Off_ feels strange: it shutdowns the watch immediatly without asking the user (sometimes I turned off the watch by mistake while exploring the menus) and the last displayed screen (before the shutdown) is shown when the watch is turned on again. This feels really strange and "broken" imho.

This PR improves the "turn off" user experience:

- The user must confirm 
- A goodbye message will be shown ("_See you later!_" &rarr; other ideas?)
- The screen will be cleared

Maybe this is a trivial change but I think this is a step towards better overall usability :-)


gfwilliams:
Thanks! Yes, that looks like a great idea

PR #1902, opened 2022-05-31
Add low flash warning to LCARS

adamschmalhofer:
Similar to the low battery warning added a low flash storage warning to the LCARS clock.


gfwilliams:
Thanks!

PR #1905, opened 2022-06-01
miclock2: Fix typo in redraw check

nxdefiant:
For some reason a test with reversed equal comparison got committed and pushed.


gfwilliams:
Thanks!

PR #1909, opened 2022-06-02
calendar: Fix scope of let variables

nxdefiant:
To work with firmware 2v13.123.


gfwilliams:
Thanks! I'm surprised the linter didn't pick those up

PR #1911, opened 2022-06-03
Big Digit Clock - just the essentials, for those of us who need glasses

deirdreobyrne:
There are various big digit clocks on the app store, but none contained all the essentials. This app shows the time, the day-of-week, the day-of-month, and the battery state of charge, in a big easy-to-read format.


deirdreobyrne:
Messed up this PR

PR #1914, opened 2022-06-04
Allow reading custom themes from files

halemmerich:
This allows the settings app to apply themes from files named `*.theme`. Theme files currently have to be uploaded via IDE, but an UI theme designer as app in the apploader would be easy to do.

Merge after: https://github.com/espruino/BangleApps/pull/1915

Example for a theme file:
```
{
  "name": "Dark Accent Blue",
  "fg": "#000",
  "bg": "#fff",
  "fg2": "#fff",
  "bg2": "#004",
  "fgH": "#0ff",
  "bgH": "#004",
  "dark": false
}
```
grafik
grafik


gfwilliams:
Thanks - that seems like a nice idea!

PR #1915, opened 2022-06-04
Improve localization for de_DE

halemmerich:
* Improves some of the translations
* Changes settings app to allow default colors to be translated


gfwilliams:
Great - thanks!

PR #1917, opened 2022-06-05
Iconlaunch - Fix colors changed during scrolling by widgets

halemmerich:
Widgets can change the colors in between two drags and cause artifacts.


gfwilliams:
Thanks!

PR #1918, opened 2022-06-05
add cassioWatch clock face App

Santagain:


gfwilliams:
Looks awesome, thanks! 

Just so you know, you can get screenshots direct from the IDE using https://github.com/espruino/BangleApps#screenshots if you need to...

PR #1920, opened 2022-06-05
Big Digit Clock

deirdreobyrne:
A clock containing the essentials, designed for those of us with poor eyesight. Time-of-day, day-of-month and day-of-week, with
an easy-to-read battery icon thrown in.


gfwilliams:
Thanks!

PR #1921, opened 2022-06-05
Fixing the moon phase calculation

deirdreobyrne:
The moon phase calculation was quite poor - this one gives much better results.


gfwilliams:
Great - thanks!

PR #1924, opened 2022-06-06
Fullscreen watchface means no buzzing

pikipirs:
If watchface is fullscreen (ie it doesn't load widgets), messages fail to buzz. This patch fixes it.


gfwilliams:
Do you know why this line below in the same function isn't working though? It seems like it should, if your line above gets called?

```
if (!quiet && (!global.WIDGETS || !WIDGETS.messages)) return Bangle.buzz(); // no widgets - just buzz to let someone know
```

I'd really like to figure out why, rather than adding an identical line further up.

Just some background though - we do the buzz in a timeout so you don't get a bunch of buzzes queued up if more than one message arrives at once.

pikipirs:
That line isn't reached because the watchface is showing and the if-statement above returns:
`return load("messages.app.js");`

Good choice with buzzing in the timer though. It may be better if buzzing in the condition above is removed completely and instead my "patch" is applied inside the timeout.

So lines 72-74 would condense to:
```
if (!quiet && loadMessages && unlockWatch != false){
```

And the timeout function would be:
```
{
  exports.messageTimeout = undefined;
  if (!quiet) {  //no loadMessages condition here - buzz should happen regardless, right?
    if(global.WIDGETS && WIDGETS.messages)
      WIDGETS.messages.buzz();
    else
      Bangle.buzz();
  }
  // if we're in a clock or it's important, go straight to messages app
  if (loadMessages){
    return load("messages.app.js");
  }
  if (global.WIDGETS && WIDGETS.messages) {  // show messages if widgets are loaded
    WIDGETS.messages.show();
  }
}
```

I will test this solution for a couple of days.

gfwilliams:
That sounds great, thanks! I imagine it'll make things work a lot better when multiple messages are received on connection too

gfwilliams:
Hi - I'm really sorry about this - not sure how I missed your commits and forgot to merge this!

Looking at the above, I *think* the things you fixed here all got fixed now (albeit slightly different ways)...

* The watch now buzzes if the app isn't displaying widgets
* As you mentioned above, the messages app now buzzes rather than the widget to avoid the half-complete buzz (it was done by loading a special `messages.new.js` file rather than just `messages.app.js`). I think this works better as it means the message is displayed at the time you get the buzz
* I think repeat:off is fixed too

The buzz=>promise change is neat though so I'll just try and tweak this PR to add that (as there are merge errors for most other stuff now) and will merge it in

PR #1925, opened 2022-06-06
Bigdclock v0.02 bug fix, new features, and improvements

deirdreobyrne:
Fixing a bug with how setTimeout() was being used.
No leading zero on the date.
Implemented a light mode.
Implemented a 12 hour format.
Some code cleanup.


gfwilliams:
Great - thanks! Just a thought - looks like you manually set colors for black/white - but you *could* use `g.theme.fg` and so on, so you can use the watch's current theme - allowing for white or black backgrounds.

deirdreobyrne:
Thanks! I'm really enjoying the watch now I know how to program it. I actually do use `g.theme.fg` and so on - the only colours I set manually are the green, yellow and red for the battery charged, charging and low indications, and the green bar across the middle of the screen. I guess I could use `g.theme.fg2` for the green bar, but since green has to work with the theme anyway (for the battery) I'm not sure it's a correct use of the theme property. Alternatively I could write a `settings.js`, but I think that's going a bit far...

Next step - internationalisation, I think. :)

gfwilliams:
Great, thanks! I didn't spot that!

deirdreobyrne:
I'm trying Internationalisation, and on the emulator I get -
```
Uncaught Error: Module date_utils not found
 at line 32 col 36 in bigdclock.app.js
    var dows = require("date_utils").dows(0,1);
                                   ^
in function "draw" called from line 89 col 6 in bigdclock.app.js
draw();
```
Is there something wrong with the emulator? (I don't want to upload it to my watch in case it "hangs" the watch).

gfwilliams:
Ahh, sorry - that's because the IDE doesn't load the Bangle.js modules in by default. Try going to IDE settings -> communications -> module URL and change it to:

```
https://banglejs.com/apps/modules/|https://www.espruino.com/modules
```

PR #1930, opened 2022-06-07
Big Digit Clock - Internationalisation; small bug fix

deirdreobyrne:
(Apologies for the mess earlier on - I find git awkward to use, and it always seems to get in my way).


deirdreobyrne:
Is it my imagination or has the app disappeared from the app loader web site?

nxdefiant:
Which app loader? There is the [release version](https://banglejs.com/apps/) and the [development version](https://espruino.github.io/BangleApps/?id=bigdclock) I see your app in the development app loader. Looks like the app is new and was recently merged, so it probably never was in the release app loader at this time. The release app loader usually gets updates with some delay of a few days.

deirdreobyrne:
> Which app loader?

Huh. I thought I previously saw it on the release/official version. Apparently not :/

Thanks. :)

PR #1931, opened 2022-06-07
Clockface menu improvements

rigrig:
As discussed [here](https://github.com/espruino/BangleApps/pull/1893#issuecomment-1147301019), instead of having separate "menu item" functions, do `require("ClockFace_menu").addItems(menu, callback, items)` instead.

And then I figured that if clocks don't have any other settings, it would be nice to get rid of the `Storage` boilerplate as well, so you can just do `require("ClockFace_menu").addSettingsFile(menu, settingsFile, ["showDate", "loadWidgets"])`


gfwilliams:
That looks great, thanks! I reckon that'll make a huge difference to clocks.

PR #1934, opened 2022-06-07
bug fix (again!): confused setTimeout with setInterval

deirdreobyrne:
Sorry 'bout this...

### DIFF
--- a/apps/bigdclock/ChangeLog
+++ b/apps/bigdclock/ChangeLog
@@ -1,3 +1,4 @@
 0.01: Initial version
 0.02: setTimeout bug fix; no leading zero on date; lightmode; 12 hour format; cleanup
 0.03: Internationalisation; bug fix - battery icon responds promptly to charging state
+0.04: bug fix
--- a/apps/bigdclock/bigdclock.app.js
+++ b/apps/bigdclock/bigdclock.app.js
@@ -16,6 +16,7 @@ var drawTimeout;
 function queueDraw() {
     if (drawTimeout) clearTimeout(drawTimeout);
     drawTimeout = setTimeout(function () {
+        drawTimeout = undefined;
         draw();
     }, 60300 - (Date.now() % 60000)); // We aim for 300ms into the next minute to ensure we make it!
 }
--- a/apps/bigdclock/metadata.json
+++ b/apps/bigdclock/metadata.json
@@ -1,7 +1,7 @@
 { "id": "bigdclock",
   "name": "Big digit clock containing just the essentials",
   "shortName":"Big digit clk",
-  "version":"0.03",
+  "version":"0.04",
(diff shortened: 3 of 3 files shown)


gfwilliams:
Thanks! By the way, I noticed:

```
60300 - (Date.now() % 60000)); // We aim for 300ms into the next minute to ensure we make it!
```

Which obviously isn't what we do in all the other clocks. Did you notice a problem with the code we use for the other clocks, or you were just playing it safe? :)

deirdreobyrne:
> Did you notice a problem with the code we use for the other clocks, or you were just playing it safe? :)

There is, technically, a race condition in how it was done in the clock code I adopted. The re-draw could start at 12:03:59.999 and display "12:03". The next setTimeout() could then be called at 12:04:00.001, which schedules the next re-draw for 12:05:00.000

There are technically better ways of dealing with that race condition, but I don't think many people will notice a 0.3 second delay.

On the subject of time, I'm thinking about automatic DST changes. It's frustrating that a watch that could "take me to the moon" can't deal with the clock change. I should be able to do it in javascript, but I'm wondering - should it be relegated to the underlying OS? Would you prefer if I wrote a library in C?

gfwilliams:
> I'm thinking about automatic DST changes

Last time I looked there were quite a lot of different DST zones/arrangements (which I believe do occasionally change), that you might not want stored in the firmware (or even JS). It would be great if this functionality could be added to the `Languages` app - so as soon as you put in your country, it can figure out what dates it has to change over at...

I guess it may even be possible to abuse `Date` on the web browser to work out what the dates/amounts for daylight saving are, so we don't have to store/maintain a list?

gfwilliams:
If you have to maintain the list yourself, I think that's probably more hassle than just clicking 'set time' to update your Bangle twice a year.

What about what I suggested about figuring out from the Web Browser's Date implementation what the rules are? There seems to be some examples on Stack Overflow

deirdreobyrne:
I think you misunderstand. You only have to set your rules for your watch once. So I'd set my watch to DST starts `last Sunday in March at 2am` and DST ends `last Sunday in October at 3am`. I'd only have to do that once (presumably when I'm setting up my timezone anyway) -- and once again if the EU ever wakes up and gets rid of DST altogether!

Any other solution requires the watch to at least periodically connect to an external source of information. I'm aware the watch
periodically gets the time from an attached phone (though I don't know how often) - which is good, but it's just better (not to mention a whole lot cooler) to stay up late in March/October and see the watch correct itself on the hour. :)

deirdreobyrne:
... in case you are interested, I have written [the settings.js file](https://github.com/deirdreobyrne/BangleApps/blob/DST/apps/dst/dst.settings.js). I think it should be easy enough to follow, and it (so far) covers everyone but the Iranians. It should also be future-proof.

PR #1935, opened 2022-06-08
[Nifty-A Clock] ClockFace

alessandrococco:
This PR upgrades the **Nifty-A Clock** to the  **ClockFace** library (thank @rigrig!)

(I plan to update **Nifty-B Clock** and some other clocks as well)

PR #1938, opened 2022-06-08
[Nifty-A Clock] Read settings file via ClockFace

alessandrococco:
Part 2 of #1935 - I didn't realize it was that easy 😅

(I wouldn't upgrade the app version)


gfwilliams:
Thanks!

PR #1943, opened 2022-06-10
Daylight Savings Time

deirdreobyrne:
A widget which automatically updates the effective timezone of the watch immediately upon entering or leaving Daylight Savings.

However, there are a few problems with this code.

### Directory
In spite of the fact that it's a widget, it's not in a directory called `widdst`. I wasn't sure if this would end up being an app, a widget, or just some boot code. But this is the easiest problem to solve.

### setting.json
This widget needs to take control over the `timezone` field of `setting.json`, as that is where everything else on the watch gets the timezone. Hence when DST starts or end, the widget needs to update `setting.json`. This creates a big problem, however, in that `.boot0` will hang if `setting.json` doesn't pass an integrity check, and the only way of ensuring that integrity check will
pass is to call `bootupdate.js`.

So lines 56 and 57 in `dst.wid.js` in the code for `function setCurrentEffectiveTimezone(tz)` say -
```
eval(storage.read("bootupdate.js")); // re-calculate the hash of the config files
load(".bootcde"); // re-boot into the clock
```
I'm pretty dang sure that's not the right way to do it, and in any case it leads to a poor user experience where it looks like the watch is rebooting a few times.

### dstRulesUpdating callback

When the DST rules are updated by `dst.settings.js`, those changes need to be acted upon immediately by the widget, as the updated rules could mean a change in whether DST is active or not. Hence there needs to be a callback in `dst.wid.js` that `dst.settings.js` can call when the settings are changed.

However `dst.settings.js` doesn't know when the changes are complete. Hence all it can do is call the callback every time any of the fields are changed. So implementation of the callback (`function scheduleUpdate()` in `dst.wid.js`) waits for a minute before acting on the updates. Again there surely has to be a better way of doing it.


deirdreobyrne:
Searched through all existing apps on the Bangle - none of them depend on setting.js for current timezone data. So there is actually no need to "reboot" when the effective timezone changes.

As for the issue around the settings.js calling the wid.js - that's only important if a dst change is imminent. What I think I'll do is just include a caveat in the documentation for the widget!

So I'll try and change this PR to "draft" status.

deirdreobyrne:
This "works", though I'm still noticing strange behaviour just after a DST change when setting the time in the settings.

I've cloned the Espruino code, as DST stuff should really be in Date. I presume there are no objections if I work on the implementation of Date?

deirdreobyrne:
> It looks like you could be setting `E.setTimeZone` but then not updating it in the watch's settings file as well, so as soon as you change apps the change is lost?
> 
> For instance this is what the app loader does:
> 
> https://github.com/espruino/EspruinoAppLoaderCore/blob/c46b4edd2052d0df37fea41f8839af8175a78ec9/js/comms.js#L354-L355

I found that when I did something like that, then .boot0 would hang the next time it carries out an integrity check on setting.json. I maybe only ran that test on the emulator - I'll try it on my watch.

> > I've cloned the Espruino code, as DST stuff should really be in Date. I presume there are no objections if I work on the implementation of Date?
> 
> Well, it'd be good to check what you intend to do first?
>
> Espruino has to target some very restricted devices, so potentially if you're going to do something fancy then it may have to be done in such a way that it can only be included on Bangle.js but not all devices

Absolutely nothing fancy. Indeed I think there are some optimisations I can contribute to the current Date implementation. But regarding Daylight Savings - the code would be much the same as in the current widget.js.

So maybe the best thing to do is for me to attempt both and have you decide? You know this stuff, and its use cases, far better than I.

gfwilliams:
>  .boot0 would hang the next time it carries out an integrity check on setting.json

That's odd - it shouldn't, but it could be an emulator issue. When settings.json updates it does rewrite boot0 and reload your code, but that should all be ok.

> I think there are some optimisations I can contribute to the current Date implementation

That'd be great - if there are some simple things that can be done to get DST working for everyone without a bunch of complexity then that'd be great.

deirdreobyrne:
Converting back to draft until I have something in the implementation of Date in Espruino.

I'm pretty confident I should be able to get something simple which works. Probably next week before I can start on it.

PR #1946, opened 2022-06-11
[Scheduler] Delete a timer after expiration

alessandrococco:
I create many _one-time_ timers and the **Alarms & Timers** app is quickly filled with an endless list of timers... Sometimes I'd like to delete a timer as soon as it expires so I made a little change in the scheduler.

After the user touch the "stop" button a new prompt is shown:

image

- **Yes** will delete the timer from the alarms list
- **Keep** will ehm, _keep_ the timer 😅

What dou you think? (I didn't add a setting for it because I think it's usefull for all) but I'm open to discussion :-)


thyttan:
I feel like there could also/instead be a setting for if you never want timers to be saved. "Keep timers?/Always delete timers?" "Yes/No"

alessandrococco:
Thank you @thyttan 

I could add a new setting:

_Delete expired timers_
- always
- ask
- never (the default)

This applies to timers only.

thyttan:
@alessandrococco I think that's a good idea! 

I think I've seen it discussed before that having to many settings could be a bad idea from an UX and debugging standpoint. But talking for myself - I would like this setting.

gfwilliams:
Personally, I think a new menu that appears after each timer all the time is probably not a great user experience, but...

> I could add a new setting: Delete expired timers

This sounds like a good idea. I guess this could even be per-timer, defaulting to `always`? It might be a useful feature in the case where timers are used by other apps

alessandrococco:
> I guess this could even be per-timer, defaulting to `always`? It might be a useful feature in the case where timers are used by other apps

It would have been my next step. I'll update the PR in the next few days!

alessandrococco:
Don't merge yet, I need to do some more tests.

alessandrococco:
Done!

Quick recap:

- New `del` field added to the alarm json
  - `newDefaultAlarm` set it to `false`
  - `newDefaultTimer` set it to the value set in the settings (`defaultDeleteExpiredTimers`)
- In the **Alarms & Timers** app you can set the `del` flag of a timer

gfwilliams:
Thanks!

PR #1948, opened 2022-06-11
[cassioWatch] Version 0.10 - Show daily step count, temperature as well as heartrate.

peerdavid:
First of all, @Santagain thank you so much for the great cassio watch face. I have seen some todos that you added and implemented some of those: 

I added support for heart rate and the watch also shows the temperature now if weather data is available. Daily steps are also shown now.

Here is a screenshot:

download


Santagain:
Thanks! nice job.

PR #1954, opened 2022-06-13
Update Version

Santagain:
Update Version, for some Reason version with 0.10 dont appear to update on App Loader


alessandrococco:
0.10 is available on [development](https://espruino.github.io/BangleApps/?id=cassioWatch) apploader. The "stable" apploader will be updated in few days.

gfwilliams:
Yes, there's no need for this.

I usually wait a bit before updating the 'normal' app loader in case something got broken :)

PR #1960, opened 2022-06-15
Updates Github action and replaces TravisCI badge

fewieden:
Updated Github actions to v3 and replaced TravisCI badge in readme with badge for Github action.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,4 +1,4 @@
-name: Node CI
+name: build
 
 on: [push, pull_request]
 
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks!

However what is the rationale for package-lock and hard-coding the node.js version?

It feels like that's just going to be another thing to keep up to date.

fewieden:
@gfwilliams 

package-lock: The lock file allows to have the same versions of dependencies and sub-dependencies installed on every installation. This makes the code and issues reproducible. If you just install via `npm i` or `npm install` it is modifying the lock file and defeats the purpose. But installing dependencies with `npm ci` strictly installs the dependency version otherwise it throws an error. https://docs.npmjs.com/cli/v8/commands/npm-ci

hard-coded nodejs version: The version was already hard-coded before to version 16.x but it used the matrix notation, which just makes it harder to understand what's going on if it is not used for more than 1 node version, e.g.

```
  test:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        node-version: [12.x, 14.x, 16.x]
    steps:
      - uses: actions/checkout@v3
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v3
        with:
          node-version: ${{ matrix.node-version }}
      - run: npm ci
      - run: npm run test
```

In this example, the flow is executed for 3 different node versions. Since the current test job is only linting, one node version is enough. If you run automated tests like in the example above you want to make sure the tests are passing for all supported node versions of the project.

PR #1961, opened 2022-06-15
Add a simple watchface to learn Hiragana and Katakana

trufae:
Future plans for myself:
* Find better resolution for the katakana and hiragana images (those are thumbnails ripped from wikipedia)
* Improve performance (reduce full screen refresh, cleanup code)
* Reduce image size by using strings and a factory method instead of an object of objects

screenshot


trufae:
I've added the images/ directory which contains the the scripts i wrote to pull all the images and convert them. should i remove that? on a future release i would like to find images with a higher resolution than those wikipedia thumbnails

PR #1965, opened 2022-06-16
sleepphasealarm. misc changes

nxdefiant:
- fixes: do not dynamically change accelerometer poll interval
- Vibrate with configured pattern
 - Add setting to defer start of algorithm
 - Add setting to disable scheduler alarm


gfwilliams:
Thanks!

PR #1967, opened 2022-06-16
Document all widget areas

fewieden:
Documents all four possible widget areas

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ widget bar at the top of the screen they can add themselves to the global
 
 ```
 WIDGETS["mywidget"]={
-  area:"tl", // tl (top left), tr (top right)
+  area:"tl", // tl (top left), tr (top right), bl (bottom left), br (bottom right)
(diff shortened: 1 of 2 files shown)


fewieden:
@gfwilliams I found the possibility for those regions on this page http://www.espruino.com/Bangle.js+Widgets

If the bottom areas are no longer wanted I can revert the change here and remove them also from the:
- docs https://github.com/espruino/EspruinoDocs/blob/4b387fc4faa7ae326915db547d3b357034f69684/tutorials/Bangle.js%20Widgets.md#how-to-develop
- example https://github.com/espruino/BangleApps/blob/574736f3ff33e0f354eba1f7a4bcfd442fe23f66/apps/_example_widget/widget.js#L12

gfwilliams:
Ahh, thanks! I think it's good to document as the functionality is there - but if you're happy to, it'd be great to mention something like "not all apps support widgets at the bottom of the screen"

PR #1969, opened 2022-06-16
Pull weather from OpenWeatherMap via http

halemmerich:
This is a proof of concept for reading the weather data "directly" instead of implementing the code in Gadgetbridge.

OWM returns JSON and allows to filter elements to keep the answers relatively compact. More info here [https://openweathermap.org/current#data](https://openweathermap.org/current#data).

This needs work to be usable:

- [x] Implement settings
- [x] Decide on good way to allow entering API-Key (bangle keyboard or custom.html)
- [x] Handle pulling correctly (do not pull on every reload)
- [x] Handle location changes


gfwilliams:
Wow, this looks great - thanks!

Just thinking out loud here, but do you think maybe this could be worth including in the Bangle.js Gadgetbridge app instead? We could plug it in where the existing Weather app for Android would usually put data in?

halemmerich:
Yes, I had thought about that too. Actually there is no reason not to do both. Doing it on the Bangle is easy to modify and a whole bunch of weather APIs could in theory be used. One provider per API would keep them very small.

Gadgetbridge implementation would certainly save a bit of energy on the bangle side. I had mediocre experience using default Gadgetbridge and the weather providers from F-Droid with regard to stability, but directly including that would probably make it more stable.

Long term I would like to use a linux phone for my Bangle with amazfish or whatever app works best. So implementation on the bangle only needs http to work with the companion app. If I implement something in other apps, I would rather do http support than some specific code to get only weather working.

gfwilliams:
Good point - and it's also a good reference for others looking at using other web services :)

One thing I'd say is maybe it's worth building a simple `http` function/handler into the 'Android Integration' app? For instance internet-enabled Espruino devices provide an `http` library: http://www.espruino.com/Internet#client

While it's slight overkill we could implement a basic version of that - it would give us the option of 'chunking' retrieved webpages in the future, which might be a good idea for bigger pages.

I'm happy to do that if you want though?

halemmerich:
Having something to actually handle the http requests would be great. Especially some sort of queuing should be implemented, since the requests and the async response currently can not be easily matched.
Maybe it would be even better to (optionally) send an ID with the request and have Gadgedbridge mirror the ID back in the response if one was set? That would make filtering really easy and would not need serialization of the requests to match request to response.

dotgreg:
Adding my two cents here, it looks really interesting, I would definitely be happy of having access to easier ways handling http calls from the watch.

I also implemented the openweatherapi on my clock for a 4 days forecast and a scratchpad. So I needed some flexibility on the communication, more than what gadgetbridge seemed to offer right now. So I coded a simple webapp that can be used on the phone as replacement of a "companion app" 

here is the PR https://github.com/espruino/BangleApps/pull/1994

Rarder44:
> This would be perfect. I seem to recall someone said they were going to do this before and give me a PR, but they never did :( It'd be great to get it implemented though.
hi, I think I proposed it ...
I sent the code to GB which should have been implemented.
I have the snippet ready to implement the request on the bangleJS side.
(sorry if i couldn't upload everything before but i'm pretty busy at this time ...)

maybe I'll pass you the code so you can see where it is better to insert it (at the moment I inserted it in a modified version of the android app)

Rarder44:
added this in android boot.js
`global.httpRequest={};
  global.http = (url,id_custom)=>
  {
    var id;
    if(id_custom=== undefined)
       //TODO: find a way to crate a better unique id
      do{    id = new Date().getTime()+""+Math.random();  } while( httpRequest[id]!==undefined);
    else
      id = id_custom;

    //send the request
    gbSend({t: "http",  url:url,  id:id});

    //create the promise 
    var promise = new Promise(function(resolve,reject) {
      //save the resolve function in the dictionary and create a timeout ( 30 seconds? )
      var timeoutMillisec=30000;      //TODO: change this value???
      var timeoutVar= setTimeout(function(id){
          //if after "timeoutMillisec" it still hasn't answered -> reject
          delete httpRequest[id];
          reject("timeout");
      },timeoutMillisec,id,reject);

      httpRequest[id]={"r":resolve,"t":timeoutVar,"j":reject};  // j = reJect
    });
    return promise;
  }`

and this in the HANDLERS obj in andoroid boot.js
`"http":function(){
        //get the promise and call the promise resolve
        var objID=httpRequest[event.id];
        if( objID === undefined) return;    //already timedout or wrong id

        delete httpRequest[event.id];
        clearInterval(objID.t); //t = timeout variable

        if(event.err!==undefined) //if is error
          objID.j(event.err); //r = reJect  function  
        else
          objID.r(event); //r = resolve  function 
        
      }`

Rarder44:
the code work in this way:
` http(URL).then((data)=>{
      var response = data.resp;
  });`

gfwilliams:
Thanks! did you implement the Gadgetbridge changes? Mind dropping them here or doing a PR? While they should be tiny it always feels better to use pre-tested code :)

Rarder44:
yes I have already done a pr and already approved.
codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2683

I tested everything on an app that I was developing for the weather (I tried to recover json objects and "images")

gfwilliams:
> maybe it would be better to integrate the "http" functionality in the emulator.

That's a good plan. PRs welcome! :)

gfwilliams:
Looks good - ready to merge?

halemmerich:
Yes, works fine for me.

PR #1974, opened 2022-06-17
[new] A-GPS Data

pidajo:
Load assisted GPS data directly to the watch using the new http requests on Android GadgetBridge.

Currently proof of concept on Bangle2 only. Will eventually add a widget for automatic download.


gfwilliams:
Great - thanks!

PR #1975, opened 2022-06-17
Release kanawatch 0.03

trufae:


trufae:
I think all the clock apps should call setUI(), do you think it's worth adding such check in the ci?

gfwilliams:
Thanks! Please can you also update the version in `metadata.json`?

In most cases, yes, they should (although I think one or two try and do fancy things without it). A check of some kind could be added to sanitycheck.js though - maybe just a `string.includes("Bangle.setUI")` would do it

trufae:
fixed

gfwilliams:
Thanks!

PR #1976, opened 2022-06-18
Who uses yards?

Darin755:
I have switched from yards/miles to meters/miles for the us locale due to the fact that no one really uses yards. All of the other units are unaffected and were not changed.


gfwilliams:
Thanks!

PR #1977, opened 2022-06-18
tabanchi v0.03

trufae:


gfwilliams:
Thanks! Please can you also update the version in `metadata.json`?

trufae:
Fixed! maybe good to add a ci check for this stuff too :) sorry for that little mistake and thanks for the review

gfwilliams:
> maybe good to add a ci check for this stuff too

There is one if you check the CI log, but it just warns rather than erroring. Maybe it should error...

PR #1979, opened 2022-06-18
[TerminalClock] Use ClockFace module and add the ability to personalize the line order

Stiralbios:
There is also a small fix on the ClockFace module to really draw the widgets.
Ping to @rigrig if you want to check it's not a breaking change


modules/ClockFace.js
```diff
@@ -103,12 +103,13 @@ ClockFace.prototype.pause = function() {
   this.paused = true; // apps might want to check this
   if (this._pause) this._pause.apply(this);
 };
+
 ClockFace.prototype.resume = function() {
   if (this._timeout) return; // not paused
   delete this._last;
   this.paused = false;
   if (this._resume) this._resume.apply(this);
-  this.tick(true);
```
rigrig, line 111: Whoops, no idea how that `true` got there.

rigrig:
Is it not drawing the widgets for you now?   
The idea is to only draw them at startup/resume, with your changes it would redraw all widgets every time the clock updates.

Stiralbios:
It does initially draw it but at the first `tick()` it the widget bar become all black and stay this way

alessandrococco:
> It does initially draw it but at the first `tick()` it the widget bar become all black and stay this way

You use draw() and draw() clears all the screen. Try using update() instead.

rigrig:
I see what you mean, that does look like a ClockFace bug. But instead of redrawing widgets all the time, I think it's better to [just not erase them](https://github.com/espruino/BangleApps/pull/1985).

Stiralbios:
Ok, I tried it and it seems to work, I reverted my changes to the ClockFace module

PR #1980, opened 2022-06-18
Remove custom boolean formatters

alessandrococco:
This PR removes the 99% of the Yes/No, On/Off, etc.  custom boolean formatters.

See https://github.com/espruino/BangleApps/pull/1750#issuecomment-1157382452

### DIFF
--- a/apps/activityreminder/ChangeLog
+++ b/apps/activityreminder/ChangeLog
@@ -5,3 +5,4 @@
 0.05: Battery optimisation, add the pause option, bug fixes
 0.06: Add a temperature threshold to detect (and not alert) if the BJS isn't worn. Better support for the peoples using the app at night
 0.07: Fix bug on the cutting edge firmware
+0.08: Use default Bangle formatter for booleans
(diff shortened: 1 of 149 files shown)


Stiralbios:
Can you revert the change for  TerminalClock. I think it will create conflicts with #1979  ?
I saw your replaced `+ " min"` per + "m" too. The value displayed will be different then no ?

alessandrococco:
> Can you revert the change for TerminalClock ? I think it will create conflicts with #1979 ? 

Sure! Sorry, I didn't notice your PR 😅 

> I saw your replaced `+ " min"` per `+ "m"` too. The value displayed will be different then no ?

Yes, I noticed that many apps use "m" to abbreviate min/minutes in the settings menu.

I just reverted all the changes to TerminalClock ;-)

Stiralbios:
Ok thanks. No problem ;)

I didn't notice the "m" thing on other apps before. I made the change to use the default boolean on my pull request and use "m" instead of " min" in the setting.

gfwilliams:
Great - thanks for this!

PR #1981, opened 2022-06-19
apploader - Ignore empty folder instead of creating invalid json

halemmerich:
Git is not good at cleaning up empty folders on changing branches. The script created corrupt json on empty app folders.


gfwilliams:
Ahh - thanks!

Just FYI the app loader will now work without having to run this script (it just scans local files). Despite my initial thoughts if you're serving locally it's actually not too slow :)

PR #1987, opened 2022-06-21
[stardateclock] Fix let/const usage

fewieden:
Resolves issue described in http://forum.espruino.com/conversations/377003

Fixes let/const usage for stardate clock to work properly with firmware version >=2v14

Before BJS1

image

Before BJS2

image

After BJS1

image

After BJS2

image


gfwilliams:
Great - thanks!

PR #1988, opened 2022-06-22
[new app] MacWatch

blogmywiki:
A very simple clock app that shows a large, high-contrast numeric time and date in the old Mac Chicago font. The digits are easy to read in direct light and the code is simple enough for a beginner to modify and learn from.


peerdavid:
Is it possible that the metadata for macwatch2 is wrong? I just tried to install it from your store and I got a "App source not found" error. Also the entry in the launcher looks wrong:

download

Is it possible that macwatch2.app.js must be set rather than macwatch.app.js in metadata.json?

blogmywiki:
> Is it possible that macwatch2.app.js must be set rather than macwatch.app.js in metadata.json?
It certainly is - thank you for spotting the error in macwatch2's json file. Hopefully now fixed and I will update the icons later to differentiate the two apps. I am very new to this and learning as I go!

kilrah:
This looks very cool, kinda want to try a color version now!

peerdavid:
> > Is it possible that macwatch2.app.js must be set rather than macwatch.app.js in metadata.json?
> 
> It certainly is - thank you for spotting the error in macwatch2's json file. Hopefully now fixed and I will update the icons later to differentiate the two apps. I am very new to this and learning as I go!

It works :D Thank you so much!

I just updated a few things on my bangle:
- I set the theme constantly to "light" and ignore the bangle settings such that I can still use the dark theme with my bangle as I prefer this one and the mac watchface still looks nice.
- I also load the widgets and only hide them such that everything still works as expected -- otherwise the code for widgets is not executed.
- Instead of using an interval I use a timeout such that the time changes exactly at the second where the time really changes. Otherwise it could be that the time is one minute behind...

In case you want to see those / adapt a few of those changes -- here is a gist:
https://gist.github.com/peerdavid/415d363bb96486ccdea588d734c79a5d

blogmywiki:
Thank you so much @peerdavid - I have incorporated your changes, it works so much better now!

gfwilliams:
`macwatch2` looks great, thanks! 

Now all you need is a matching `launcher` ;)

PR #1989, opened 2022-06-22
CirclesClock v0.13

myxor:
- Load step goal from Bangle health app as fallback
- Memory optimizations


gfwilliams:
Great, thanks! Just to add, I see `function getImage(graphic, color)` is used here, but it'd probably be faster/neater to just include the image as an 'image string'. If you supply a 1bpp image without palette then it's actually drawn in the current forground/background color so there's no need to mess with supplying a custom palette for each image.

PR #1990, opened 2022-06-22
WidBaroAlarm v0.05

myxor:
- Show difference of last measurement to pressure average of the the last three hours in the widget
- Cleaner code and improved warning calculation


gfwilliams:
Thanks!

PR #1992, opened 2022-06-23
New app: tinyVario

xxDUxx:


gfwilliams:
Thanks - this looks great! It's something I always hoped the Bangle would be useful for :)

PR #1994, opened 2022-06-23
[NEW CLOCK] Advanced Casio Clock

dotgreg:
An over-engineered clock inspired by Casio watches. It has a 4 days weather, a timer using swipe and a scratchpad. Can be updated using a dedicated webapp.

You can check it here : https://dotgreg.github.io/BangleApps/?q=advanced%20casio

Hope all is good, Thanks :)


kilrah:
Looks like it's not uploading/creating a default `advCasioData.json` so it fails to load

dotgreg:
indeed, just updated and tested the code to include that file during installation

gfwilliams:
Looks great, merging now!

For the web app, is there any reason you can't just use the `interface.html` file in the app loader? That way everything is included together, but also you can still run it from within Gadgetbridge

dotgreg:
@gfwilliams  Great suggestion, I did not know about it. I'll add it on the next release then

nxdefiant:
@dotgreg It should be in the development app loader: https://espruino.github.io/BangleApps/

The release apploader is not updated that often.

dotgreg:
great thanks!

PR #1995, opened 2022-06-23
[HomeAssistant] Trigger HomeAssistant automations with your BangleJs

peerdavid:
This PR adds a new app that integrates the BangleJS into HomeAssistant.  All triggers can be selected when the app is uploaded from the Bangle App store e.g. I open the door and toggle lights in different rooms.

screenshot

How it works: It uses the really really cool (thank you so much @gfwilliams) new GadgetBridge BangleJs Android App and sends an intent. The HomeAssistent Android app forwards this intent to the HomeAssistant Server -- on the server simple automations can be created e.g. whenever the banglejs intent with key "DOOR" is triggered, the door is opened etc.

Detailed documentation for "one-time setup" (we have to tell the HomeAssistant Android App once the intent) and how to add triggers including a screenshot of the HomeAssistant configuration is added to the Readme.

Honestly, I'm not only surprised how easy it was to create this app, but also how fast the trigger works (in less than 1sec. I can toggle lights etc.) -- so thanks again for this great job @gfwilliams


myxor:
I was thinking about doing something similar. Really happy that this is working :) Thank you @peerdavid

myxor:
I have a suggestion:
It would be nice to have triggers to not just consists of the trigger name itself but if they would have a displayname too. 
Even better if the displayname could contain emoticons / symbols.

peerdavid:
> I have a suggestion: It would be nice to have triggers to not just consists of the trigger name itself but if they would have a displayname too. Even better if the displayname could contain emoticons / symbols.

Yea I also started thinking about a displayname already -- and nice idea with the icon, probably we could include some "build in" icons such as light etc.

peerdavid:
Ok @myxor  its implemented -- you can now specify a json file which is a list of objects containing the trigger, display name and icon name:
```json
[
      {"display": "Open", "trigger": "OPEN_DOOR", "icon":"door"},
      {"display": "Office", "trigger": "TOGGLE_LIGHT", "icon":"light"},
      {"display": "Living Room", "trigger": "OVEN", "icon":"fire"}
]
```

On your bangle its shown as follows:
screenshot
screenshot_2
screenshot_3

myxor:
@peerdavid it works good. It would be even better if the custom.html would load the existing JSON configuration from the storage and show it so that one has not to redo the whole JSON every time.

peerdavid:
>

peerdavid:
> @peerdavid it works good. It would be even better if the custom.html would load the existing JSON configuration from the storage and show it so that one has not to redo the whole JSON every time.

I just added this to custom.html -- whenever your bangle.js is connected the content will be updated as soon as the data is read from your bangle.

dmcavinue:
Nice, looks great, it's funny, I actually started mulling over the [same](https://github.com/dmcavinue/BangleApps/tree/master/apps/homeassistant)  thing in the IDE and came to a very similar design albeit not as polished with a slight difference to let me toggle entity states grouped by domain instead. | was curious if you had any thoughts on returning state? I'm considering leveraging the http function of the same GB bangleJS to call the hass api for entity state but not sure if it's viable or too expensive a transaction yet. My ultimate goal is to have specific 'cards' per entity type (switches/lights/media player) but it's definitely a WIP and more ambitious then my js skills.

peerdavid:
> Nice, looks great, it's funny, I actually started mulling over the [same](https://github.com/dmcavinue/BangleApps/tree/master/apps/homeassistant) thing in the IDE and came to a very similar design albeit not as polished with a slight difference to let me toggle entity states grouped by domain instead. | was curious if you had any thoughts on returning state? I'm considering leveraging the http function of the same GB bangleJS to call the hass api for entity state but not sure if it's viable or too expensive a transaction yet. My ultimate goal is to have specific 'cards' per entity type (switches/lights/media player) but it's definitely a WIP and more ambitious then my js skills.

I wonder why you decided to implement intents over HTTP calls when your goal was to also read states? I had some very good reasons not to use the API and my goal was always to implement a "trigger only"... so intents was for several reasons preferable in my case. And I totally agree with you @dmcavinue its really suprising how your test app evolved the last hours to the "same thing and very similar design" implementing the same intent function, retry logic, custom HTML etc.

dmcavinue:
Hah, Yeah, I'll admit to stealing your retry logic for the intent call but that's about it. This was entirely solely IDE based up until yesterday when I spotted your pr, you can actually see me also opening one and then immediately closing it when I spotted yours already in play. Honestly no intention to plagurize your work. I went with intents initially it seemed safer but hit the hurdle pretty quick on response.

peerdavid:
Nothing to worry - I only checked because you explicitly mentioned that they are very similar :)

gfwilliams:
This looks great, thanks! It's great how well the intents worked here (I just added them because someone was pushing for Tasker integration) but I wonder what else they could be used for now!

PR #1996, opened 2022-06-24
Rewrite Bowser clock to use ClockFace

seanhagen:
Not being able to have the clock face be in 12hr mode was bugging me, so I re-wrote the Bowser clock face to use the ClockFace module that takes care of finding the 12hr/24hr setting.


gfwilliams:
Great - thanks! Please could you update the metadata version and Changelog so users know there's a new version?

seanhagen:
Sorry, yes!

gfwilliams:
Thanks! But I'm afraid the ChangeLog has to be in the same format as other apps (see the _example_app folder) - in fact there was already a ChangeLog file in there that you just didn't update - but I'll make the change now

PR #1998, opened 2022-06-26
Screenlock Widget

CarlR9:
Hi,
This is a screenlock widget I've created for the Bangle 2. This is the first time I've used github, so hopefully I've got everything right.

All the best,
Carl Read.


CarlR9:
Thanks. I'll look into those.

PR #2000, opened 2022-06-27
Refactoring and improvements of HomeAssistant, BWClock and Info App.

peerdavid:
I improved and maintained some of my apps:
- HomeAssistant: I now expose a library which can be used by other apps/clocks.
- BW Clock: Stability improvement if clock is killed + images are loaded lazy to reduce the memory consumption (we don't need all the images in ram all the time). Also, it uses the HomeAssistant library (if installed) such that users can send trigger directly via the watch face. Here is an example (the triggers are those that are configured in the HomeAssistant BangleJs app):

screenshot_4

- Info App: Shows now also weather data if weather is installed. Some refactorings such as proper widget handling etc. is also included.


gfwilliams:
Great, thanks! With the `info` app - someone recently got pretty confused because the 'free storage' item reported kept going down and they were trying to slim down all their apps to improve matters.

Might be better to use `Storage.getStats` which gives you sizes for used,free, and trash - so then folks can know they actually have free+trash bytes available :)

PR #2005, opened 2022-06-28
Messages: Separate buzz pattern for incoming calls

myxor:
As suggested here: http://forum.espruino.com/conversations/376923/newest/


LeonMatthes:
@myxor I've noticed a bug in the current buzzing behavior, as it will ignore the 'Vibrate: Off' setting.
The same bug is in your version of the code, as it's very subtle and hard to track down.

See #2023 .

Would you mind adding that fix to your code, then we can kill two birds in one stone with your PR.

myxor:
@LeonMatthes  thank you for your feedback. I integrated your fix into my branch with commit https://github.com/espruino/BangleApps/pull/2005/commits/d82454f942f3790996a3aede3c62a01dbc7d2b71.

gfwilliams:
Looks good - thanks!

PR #2007, opened 2022-06-29
[BW Clock] Improved Menu

peerdavid:
I was not so happy after including HomeAssistant into BW clock with the "menu" of the clock (1D list with entries) -- so e.g. with many HomeAssistant triggers I had to click quite often until I reached e.g. the step counter etc. again. So I decided to implement a "2D-menu" (sry I have no idea how this is really called). My idea was that each integration (i.e. BangleData, Timer, Weather and HomeAssistant) can be selected via tab left/right and each submenu provides items via up/down. Whenever one integration is not available (e.g. not installed), it is simply not shown to keep everything as minimal as possible. An overview of the menu is shown here:

```
   Bpm               ...
    |                 |
  Steps            10 min.          ...                     ...
    |                 |              |                       |
  Battery           5 min          Temp.                 Trigger1
    |                 |              |                       |
  Bangle   -- Timer[Optional] -- Weather[Optional] -- HomeAssistant [Optional]
```

And here you can see where each entry is shown on the watch face:
screenshot_2

Clearly, it's also possible to select "nothing" and only display a large time. But overall, this new structure is relatively generic and should make it easy to extend it :)

### DIFF
--- a/apps/bwclk/README.md
+++ b/apps/bwclk/README.md
@@ -41,4 +41,4 @@ to e.g. send a trigger via HomeAssistant once you selected it.
 
 
 ## Creator
-- [David Peer](https://github.com/peerdavid)
+[David Peer](https://github.com/peerdavid)


gfwilliams:
Looks great - thanks!

PR #2010, opened 2022-07-01
[BW clock] Bugfix: Ignore clicks for widgets.

peerdavid:
Clicks < 24px are for widgets, if fullscreen mode is disabled. In this case, clicks are ignored such that only the widget click is executed.


gfwilliams:
Thanks!

PR #2012, opened 2022-07-01
Sliding Clock: Settings Screen

awkirk71:
By popular demand screen controls for the sliding clock can now be controlled from the new settings screen so that the clock is not constantly changing colour and language on the bangle 2.
By default the bangle 2 screen controls are switched off, but they can be re-enabled from the sliding clock settings by switching the 'live control' setting. 
The bangle 1 is still controllable from the buttons on by default (as the buttons are apt controllers), but can be switched off from the setting screen.

Have also added a white clock face so it can blend in with the white theme


gfwilliams:
Great - thanks!

Issue #2009, opened 2022-07-01
[bthrm] Internal HRM on by default all the time


gfwilliams:
### Affected hardware version

Bangle 2

### Your firmware version

2v14

### The bug

* Install `bthrm`
* Don't use heart rate monitor for anything
* Internal heart rate monitor is on. `Bangle._PWR` reports `HRM: [ "bthrm_fallback" ],`

I'm not quite sure what's meant to be going on here, but it seems like `switchInternalHrm` gets called at boot time, as well as a timer to check connection state - but I guess really nothing should really be happening until an app calls `setHRMPower`?

@halemmerich any thoughts? If I'm honest I'm a bit lost by all the stuff that'd going on now :)

halemmerich:
I wanted to have a look at the mechanics for replacing the HRM anyway. Will do some experimenting this weekend.

halemmerich:
I think this can be closed.

gfwilliams:
Ok, thanks!

PR #2013, opened 2022-07-02
BTHRM - Various improvements

halemmerich:
* Fix HRM fallback handling -> #2009
* Use default boolean formatter in custom menu and directly apply config if useful
* Allow recording unmodified internal HR
* Better connection retry handling

### DIFF
--- a/apps/bthrm/ChangeLog
+++ b/apps/bthrm/ChangeLog
@@ -25,3 +25,7 @@
 0.10: Use default Bangle formatter for booleans
 0.11: App now shows status info while connecting
       Fixes to allow cached BluetoothRemoteGATTCharacteristic to work with 2v14.14 onwards (>1 central)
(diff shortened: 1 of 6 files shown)


gfwilliams:
Thanks!

Do we really need `recorders.hrmint` though? We already have a recorder for `BTHRM` and `HRM`. Is there not a way to just ensure that the mode is set to `both` if `BTHRM` recording is enabled? Then it should all work ok?

halemmerich:
Not really. I added the hrmint-recorder primarily to be able to capture all existing event types. Besides debugging bthrm I do not really have an usecase for the unmodified internal sensor events (I would use hrmaccevents for that).
Setting the mode from the outside of the settings app would require extracting some code for that into a lib to avoid duplication. Since bthrm has gained a lot of code, maybe extracting that and some other functionality into libs/other apps would be worth a thought.

PR #2016, opened 2022-07-03
presentation_timer: initial release

glemco:
Simple timer based on stopwatch touch showing the current slide number for your presentation (after uploading a CSV file)


gfwilliams:
Thanks! The docs look really helpful too!

You could definitely add something pretty easily to allow the info to be changed from the app loader too if you wanted. There are a few examples but the todo list looks like a promising start: https://github.com/espruino/BangleApps/blob/master/apps/todolist/interface.html

PR #2017, opened 2022-07-03
Multitap Keyboard

halemmerich:
Implements additional mode for moving the cursor in the edited text. Swipe down for changing modes.


gfwilliams:
Thanks!

PR #2019, opened 2022-07-04
sensortools - Sensor data emulation tool

halemmerich:
Debugging/Development tool for everything that uses sensordata.
I was experimenting with compass and gps based code and had neither GPS signal nor valid magnetometer readings (those pesky electronic devices and walls everywhere).

Currently not even tested, more of a quick brainstorming/pseudo code notepad.

ToDo:
- [x] Get it working
- [ ] Configuration
- [ ] Support more Sensors
- [ ] Valid test data
- [ ] App to show real and emulated sensor data


gfwilliams:
Interesting - so the idea is to fake sensors, in case you're not somewhere you can test them out yourself? I wonder if it could be handy in the emulator too...

xxDUxx:
>I wonder if it could be handy in the emulator too...
Definitely!

gfwilliams:
What's the status of this now? This PR is 2 months old and it still says `Currently not even tested,` and only 1 out of 5 tasks is complete.

Can we close this and re-open when it's finished? I'd rather not have the repo filled with loads of 'draft' PRs

halemmerich:
I think we can close this for now. It is now in a state where I can use it, but it still requires lots of manual adjustments and interaction during use. Possibly the idea of configuring it an then have it running the whole time is not optimal. Maybe building it as a library with some simple calls for starting/changing would be more useful for actual development.

PR #2020, opened 2022-07-05
hidcam: Bangle.js 2 support

pebl-hank:


gfwilliams:
Looks great - thanks!

PR #2022, opened 2022-07-05
[Messagesmusic] Remove unnecessary line of code

thyttan:
... making the app a little more reliable. Specifically while testing @rigrig's rewrite of  the Messages app.

The functionality is intact for the official Messages app.


gfwilliams:
Thanks!

PR #2023, opened 2022-07-05
Fix "Vibrate: off" being ignored by messages app

LeonMatthes:
The messages app used this code:
``` js
(require('Storage').readJSON("messages.settings.json", true) || {}).vibrate || ":"
```
to determine the buzz pattern.

If I set the buzz pattern to "Off" in the settings app, it will return an empty string (`""`), which is falsey in JS, therefore the `|| ":"` will still return ":" in this case, causing a buzz, even though I explicitely disabled it in the settings.


gfwilliams:
Great - thanks!

PR #2024, opened 2022-07-05
Fixed existance of multiple change logs with different cases

buergi:
Cloning the repo on systems with case insensitive file system cause trouble, as multiple `ChangeLog` files differing only by case were existant. Restored the original change logs and converted all `Changelog` to `ChangeLog`


gfwilliams:
Great, thanks! I didn't spot that - maybe I should add a test to sanitycheck to make sure the case is correct

PR #2025, opened 2022-07-05
Teatimer display bug

buergi:
Most text messages drawn by using `E.showMessage` were not shown properly on Bangle.js 2. Only the title line was visible but the complete message space was drawn in white, no text was visible.
Added empty `E.showMessage` calls before the actual text rendering was added as a workaround. I'm not sure, however, why this is necessary, so probably there might be a better way to fix this.

### DIFF
--- a/apps/teatimer/ChangeLog
+++ b/apps/teatimer/ChangeLog
@@ -1,0 +1,2 @@
+0.01: New App
+0.02: Fixed rendering problem of several text messages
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! I just looked into this and I *think* it may be related to colors not being reset properly. It may need a slight tweak in the firmware too (i added a workaround) - I'll look into this now.

Were you using the dark theme?

buergi:
Thanks, yes I'm using the dark theme...

PR #2028, opened 2022-07-06
AGPSdata v0.02: Load AGPS data on app start and automatically in background

myxor:
- Load A-GPS data on app start and automatically in background (if enabled)
- Configuration of which GNSS types (GPS, BDS, GLONASS) will be used
- Exposes a library which can be used by other apps/clocks to load A-GPS data

@pidajo pinging you to let you know that i am working on this. And sorry that i removed your touch menu - i think it's much easier for the user if everything happens automatically without multiple UI interactions required.


pidajo:
Fine by me, I work on a widget anyway

PR #2029, opened 2022-07-07
Fixed FG/BG color for Teatimer

pebl-hank:


gfwilliams:
Thanks - `g.reset()` might be a quicker way of doing this (like the 0.02 update I did). I've put a fix in to the main Espruino firmware as well, so in cutting edge builds (or 2v15 when released) this should get fixed anyway

PR #2030, opened 2022-07-07
Add "minimal" versions of battery and Bluetooth widget

LeonMatthes:
I personally like my watch to be more context-sensitive and only show me certain information when I need to know about it.
e.g. I don't care whether my battery is 70, or 75 percent charged, it's simply "charged enough".

Therefore I've created two widgets:
- Minimal Battery: Only appears when the battery is running low, changes color from orange to red the lower the battery goes.
- Minimal Bluetooth: My watch is usually connected to my phone, this widget only appears if it loses the connection.

Future Work: A "Minimal" lock indicator, which would only appear when the watch is unlocked.


gfwilliams:
Looks great - thanks!

PR #2031, opened 2022-07-07
[New] AGPS Widget

pidajo:
Download A-GPS data to the Bangle2 automatically every 12 hours in the background


gfwilliams:
Thanks! Is it worth actually having the `agpsdata` app now? It feels like they are pretty much duplicates, just that the widget is a widget?

pidajo:
It's a question of once vs always. I still see a value in a separate app

gfwilliams:
Ok, it's just a bunch of code duplication.

You could just have one app that either allowed you to upload AGPS data once, or turn on the widget to upload it automatically

PR #2032, opened 2022-07-08
[Alarms & Timers] Allow hiding timers/alarms

LeonMatthes:
In my quest to hide as many widgets as I can; here is the ability to hide alarms or timers from the Alarms & Timers widget.

This feature was already available in the "sched" library & the widget, it just wasn't exposed in the menu.
So allow to change this from the menu now.
It mostly serves to hide the daily morning alarm that you might have set up.


gfwilliams:
Thanks - sounds like a great idea!

PR #2033, opened 2022-07-08
Drink counter Demo

pebl-hank:
Still work in progress - but usable for at least counting :)
This is only a demo for now that calculates a 0.33L Beer for a 80kg male for the BAC!


pebl-hank:
No demo anymore - 0.15 should work more or less :)
Tagged for BJS 2 only (yet)

gfwilliams:
Looks great - thanks!

PR #2037, opened 2022-07-10
Widget to set dst rules

deirdreobyrne:
Widget to set the daylight saving rules in E.setDST(...).

**NB the firmware _must_ support daylight saving time.** (I don't know how to check for this before upload).


myxor:
Maybe you could check `if (E.setDST)` to check if the method exists on the Espruino object?

gfwilliams:
Looks good! I'll update the description to be explicit about DST.

However I think the DST should probably be set in a `widdst.boot.js` file, because as-is, if a clock face doesn't load widgets I think the timezone won't end up getting set properly?

PR #2040, opened 2022-07-12
GPS Widget: Alternative marker icon (configurable via settings)

myxor:
Screenshots:

image

image

image

### DIFF
--- a/apps/widgps/ChangeLog
+++ b/apps/widgps/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Show GPS fix status
 0.05: Don't poll for GPS status, override setGPSPower handler (fix #1456)
 0.06: Periodically update so the always on display does show current GPS fix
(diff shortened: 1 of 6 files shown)


myxor:
Still todo: 
Visibility of icons on dark theme.

gfwilliams:
Please could you export the icons with https://www.espruino.com/Image+Converter as: 

* Transparency
* Inverted (if the preview doesn't end up white)
* 1 bit black/white
* Image String

I think you might be using palettes at the moment? If you use a simple 1bpp image without palettes Espruino will automatically use the FG/BG colors for it - so you don't need two separate images one for each type of theme - it'll just magically work

myxor:
@gfwilliams thank you for your help with the icon format.

I changed them to 1-bit, transparency and now they magically work on dark and light theme ;)

gfwilliams:
Great - thanks!

PR #2044, opened 2022-07-13
[thumbnailer.js] Output app name, reset lcd and intervals

BartS23:
flow app sets LCDMode to 120x120 wich will break image export (`TypeError: FUNCTION_TABLE[(HEAPU8[(($0 + 52) | 0)] | (HEAPU8[(($0 + 53) | 0)] << 8) | ((HEAPU8[(($0 + 54) | 0)] << 16) | (HEAPU8[(($0 + 55) | 0)] << 24)))] is not a function`)

Some applications use a very short timeout, which also break the export.


gfwilliams:
Thanks!

PR #2045, opened 2022-07-14
Adopted for BangleJS2

pebl-hank:


gfwilliams:
Great! Thanks for doing these conversions - it's great to be able to pull across some Bangle.js 1-only apps

PR #2046, opened 2022-07-14
Clockface: powersaving 

rigrig:
* Makes `tick` a "ram" function, so if clocks implement their `update()` in "ram" as well, (in theory) updating the clock doesn't need to wake flash. (Based on [this forum thread](http://forum.espruino.com/conversations/375995/))
* Adds a `powerSave` menu option. (It's a boolean for now, in the future maybe also allow integer options (`["Always","During Quiet Mode", "Never"]`), but that would mean clocks need to watch for quiet mode changes.)

### DIFF
--- a/modules/ClockFace.js
+++ b/modules/ClockFace.js
@@ -49,6 +49,7 @@ function ClockFace(options) {
 }
 
 ClockFace.prototype.tick = function() {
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks!

PR #2047, opened 2022-07-14
Barclock powersaving

rigrig:
(Needs the [Clockface: powersaving](https://github.com/espruino/BangleApps/pull/2046) PR)

* Optimize drawing: second updates (the bar) happen in `"ram"`, and only redraw the changed part of the screen. (Instead of calling `layout.render()` every second)
* (Bangle.js 2 only): adds a "Power saving" option: only draws the bar while unlocked, otherwise only update once a minute.


gfwilliams:
Looks great - thanks!

PR #2049, opened 2022-07-17
Matrix Clock: 24 Hour Format

awkirk71:
By popular request 24 hour time format is now available through the settings screen


gfwilliams:
Great - thanks!

PR #2053, opened 2022-07-19
Add a bunch more typings to Bangle.ts

qucchia:


gfwilliams:
Wow, thanks! There's a huge amount of work gone into this.

For things like the documentation or the type file for the editor (http://www.espruino.com/json/espruino.json) we auto-generate from the JSON blocks in the `jswrap_*.c` source files of https://github.com/espruino/Espruino 

If you're up for it I'd be really interested in trying to add this type information into that JSON, so that we can auto-generate the typescript decls and they always stay up to date.

qucchia:
That would be great! I suppose the JSON comments would have to be updated to include typing information?

gfwilliams:
Yes - there is certain type info there already but it's very Espruino-centric. Looking at how you've done it I guess the simplest method is to maybe have it create Typescript decls automatically, but then have a new field that might be something like: `"typescript":"() => string | undefined"` that would be used if it were specified...

qucchia:
I've tried it out and it's almost working (see https://github.com/qucchia/BangleApps/commit/805b5026da669f534ce76b1fb684f64c2ab913e5)! Only problem I've encountered is that the StorageFile and RstInfo types don't seem to be defined in the `espruino.json`.

gfwilliams:
Thanks! Just looked into this and some errors in the JSON were stopping everything from getting parsed! Try it now

PR #2054, opened 2022-07-20
Auto-generate types

qucchia:
Types are now automatically generated by a script in `typescript/generate.js` getting data from https://espruino.com/json/espruino.json. The result can be seen in `typescript/types/main.d.ts`.
However, the file won't compile yet as `StorageFile` and `RstInfo` are used but not defined in espruino.json (this is why the check failed). Hoping I can be directed as to what I should do about this.


qucchia:
That's great, thanks! The RstInfo is coming from here: https://github.com/espruino/Espruino/blob/master/targets/esp8266/jswrap_esp8266.c#L74

gfwilliams:
Ok, try now! That's been removed. It was a broken bit of typing info - `RstInfo` wasn't a proper object that was returned

qucchia:
It seems there are still errors: some globals such as `g` and `WIDGETS` are not defined and events are not implemented.

gfwilliams:
Ahh - I guess those could be added to Espruino as JSON, although it might be easier to add one typescript file that just helps to 'fix up' these issues. 95% of the API should be handled ok by the import.

qucchia:
That's odd, the `npm run build` script runs fine but `npm run build:types` doesn't? It looks like when running `npm run build:types` TypeScript is ignoring the `noLib` option. Perhaps we can remove the `npm run build:types` check as it's superfluous anyway?

gfwilliams:
That's a strange one... But yes, I'm very happy for that extra build step to be removed

gfwilliams:
Great! Are you happy to merge this now?

qucchia:
Yes!

gfwilliams:
Great!

PR #2056, opened 2022-07-22
Allow Noteify on Bangle.js 1

qucchia:
The Noteify app said it only supported Bangle.js 2, but it actually works fine on Bangle.js 1, as long as a different textinput library is used. (Although because kbmorse is the only such library for Bangle.js 1 and it's currently [broken](https://github.com/espruino/Espruino/issues/2240), it doesn't change much just yet).


gfwilliams:
great - thanks!

PR #2058, opened 2022-07-23
Fix kbmorse

qucchia:
As I said on #2056, kbmorse is currently broken because of what seems to be an [Espruino bug](https://github.com/espruino/Espruino/issues/2240). This small update fixes the library by changing the code order.


gfwilliams:
Thanks! As I mentioned in that issue, I believe this is already fixed - we just need a 2v15 release.

It doesn't hurt to change this to fix it on 2v14 though

PR #2059, opened 2022-07-23
messages: reduce widget memory usage

rigrig:
Don't keep entire messages in memory: we only need `src` for icon+color

I figured this doesn't warrant a version bump, as there aren't any functional changes


gfwilliams:
Thanks!

PR #2063, opened 2022-07-27
Move tsconfig.json

qucchia:
In the editor I use the TypeScript language server doesn't detect the `tsconfig.json` file when it's in the `typescript` folder, meaning I have to manually open the `main.d.ts` file. Moving it to the root folder fixes that.


gfwilliams:
Thanks!

PR #2064, opened 2022-07-27
widcw: fix memory leak

rigrig:
Repeated redrawing would create multiple redraw-timeouts

(Same as https://github.com/espruino/BangleApps/pull/1713 for `widcal`)


gfwilliams:
Nice - thanks!

PR #2065, opened 2022-07-27
messages: improve memory usage

rigrig:
Clear `layout` when going back from message to list

Map and music already do this, just the single message view missed it.

I also considered converting `layout` from a global to local variables, but figured that leaving it global makes debugging easier.

### DIFF
--- a/apps/messages/app.js
+++ b/apps/messages/app.js
@@ -286,6 +286,7 @@ function showMessage(msgid) {
     }
   }
   function goBack() {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - that's great! Yes, I bet that makes a big difference.

Yeah, global layout is just nice for debugging I guess. Maybe at some point the handler that gets called when a message changes will be smart enough to update just part of the layout too

PR #2066, opened 2022-07-27
weather: don't mark app as CLOCK.

rigrig:
Use setWatch() instead of setUI to make middle button open the launcher, checking HWVERSION to watch the correct button.

Otherwise widgets get confused and you don't get e.g. a close button.

### DIFF
--- a/apps/weather/ChangeLog
+++ b/apps/weather/ChangeLog
@@ -12,3 +12,4 @@
 0.13: Tweak Bangle.js 2 light theme colors
 0.14: Use weather condition code for icon selection
 0.15: Fix widget icon
+0.16: Don't mark app as clock
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -101,7 +101,11 @@ weather.on("update", update);
 
 update();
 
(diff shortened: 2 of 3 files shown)


gfwilliams:
Please could we just use `setUI("updown", b => { if (b===undefined) ... })`?

I'm trying to get rid of explicit usage of setWatch if at all possible, because it doesn't play well with other stuff (for instance if a widget wanted to use `E.showPrompt`) - not to mention if we added a new Bangle at some point with more buttons

rigrig:
> I'm trying to get rid of explicit usage of setWatch if at all possible

Ok, that makes sense.   

How about explicitly using the `clock` interface without registering as clock?
```
setUI("clock");
delete Bangle.CLOCK;
```

gfwilliams:
You could, yes. Although please can you just add a comment to it to hint at why it's being done? ;)

rigrig:
Oh definitely, otherwise I see myself making the same PR again in a couple of months ;-)

gfwilliams:
Thanks! :)

PR #2069, opened 2022-07-31
agenda: disable past events display from settings

glemco:
(Small feature hoping someone is actually able to use the agenda with GB)
Gadgetbridge doesn't synchronize events very frequently, unless there's a change in the calendar (new events), the update may take a day or more, depending on the connection status.
This means that before GB realizes there are old events and deletes them, we may see them in the calendar for a while.

With this update, by default the agenda won't show them. If enabled they can still be shown (grayed out). An event is considered old immediately after its finishing time, in a future update we could add a threshold for that (e.g. show events finished 1 hour ago).


gfwilliams:
Great - thanks!

PR #2070, opened 2022-08-02
New app: Altimeter widget

xxDUxx:


apps/tinyVario/app.js
```diff
@@ -266,6 +266,9 @@ function initAltMenu() {
         }},
         {type:"btn", font:"16%", pad:1, fillx:1, label:"SAVE", cb: ()=>{
           require('Storage').writeJSON("tinyVario.json", settings);
+          var o=Bangle.getOptions();
```
gfwilliams, line 269: You don't actually need this (and it's probably better not to use it) - just do `Bangle.setOptions({seaLevelPressure:qnh});`

xxDUxx:
Closing this for now, somehow the timing does not work...

apps/tinyVario/app.js
xxDUxx: Thanks!

PR #2082, opened 2022-08-09
Add `getMessages` and `status` to `messages` library, use in apps

### LINKED ISSUE #2081: 7x7dotsclock and hcclock shouldn't load messages.json directly
As noted in https://github.com/espruino/BangleApps/pull/1498#issuecomment-1205822196 - ideally non-messages apps should be able to use the library or widget to get info on messages

rigrig:
Instead of reading `messages.json` directly from apps.

If you update `17x7dotsclock` or `hcclock`, you also need to update `messages`, but I figured that's fine.

Fixes #2081


gfwilliams:
Why the closure? Because it's dependent on the messages-as-storage PR?

rigrig:
Whoops, no. It's based on master, I just deleted the branch by accident which closed this PR.

I set up Gitea to mirror my personal repository to GitHub, but then mistakenly pushed this branch (only) to GitHub, so the next sync deleted it.

gfwilliams:
Ahh :)

Just comparing this to https://github.com/espruino/BangleApps/pull/1498/files - that uses `MESSAGES = require("messages").load();` but this uses `var MESSAGES = require("messages").getMessages();` - do you think it's maybe worth using `load` here too so it's not such a pain to merge?

rigrig:
I was planning to change it in that branch once this lands, as `getMessages()` seems like a better name, especially if it also checks `MESSAGES`.

gfwilliams:
Thanks! Sorry for the delay merging.

IMO `var messages_installed = require("Storage").read("messages")!==undefined;` might be a bit faster/more readable than `var messages_installed = require("Storage").list(/^messages$/).length > 0;` but I'll make that tweak now.

PR #2083, opened 2022-08-09
Waypoint bugfix

mbks:
The current library for waypoints includes nested backticks for format string evaluation. However, when attempting to import this library (e.g. with `require("waypoints").load()` such as in the Way Pointer app) I get the error  

`Uncaught SyntaxError: Got EOF expected ':'
 at line 1 col 5
num?
    ^
 at line 1 col 5
num?
    ^
in function "load" called from line 44 col 38 in waypointer.app.js
var waypoints=require("waypoints").load();`. 

This is fixed by removing the inner backticks and replacing them with string concatenation, which still appears to evaluate the format string correctly.

### DIFF
--- a/apps/waypoints/lib.js
+++ b/apps/waypoints/lib.js
@@ -1,7 +1,7 @@
 exports.load = (num) => {
-  return require("Storage").readJSON(`waypoints${num?`.${num}`:""}.json`)||[{name:"NONE"}];
+  return require("Storage").readJSON(`waypoints${num?".${num}":""}.json`)||[{name:"NONE"}];
 };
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - that's very strange though since I'm pretty sure we've done similar things all over the place.

I'll merge this but will also see if I can figure out what's wrong with the interpreter itself that causes this error

gfwilliams:
Ok, just found an issue inside Espruino so this should be fixed in firmwares 2v15 and later.

Thanks for the PR and letting me know about the issue!

mbks:
Cool, thanks for fixing it!

PR #2084, opened 2022-08-10
Add option to hide GPS widget when GPS is off

khromov:
Small refactor that allows the user to switch off the icons when GPS is off.

Screenshot 2022-08-11 at 00 10 24

### DIFF
--- a/apps/widgps/settings.js
+++ b/apps/widgps/settings.js
@@ -23,6 +23,10 @@ var mainmenu = {
     value : !!settings.crossIcon ,
       onchange : v => { writeSettings("crossIcon", v); }
     },
+  "Hide icon when GPS off" : {
+      value : !!settings.hideWhenGpsOff ,
+        onchange : v => { writeSettings("hideWhenGpsOff", v); }
+      },
 };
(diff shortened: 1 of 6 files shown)


apps/widgps/settings.js
```diff
@@ -20,9 +20,13 @@ var mainmenu = {
   '' : {'title' : 'GPS widget'},
   '< Back' : back,
   "Cross icon" : {
-    value : !!settings.crossIcon ,
+    value : settings.crossIcon ,
```
gfwilliams, line 23: I guess in this case it's probably fine because default settings are loaded, but just in general please can we keep the `!!` in. It ensures that even if the settings get corrupted, the field is still turned into a boolean and so can be edited by the menu. If for instance somebody put `{crossIcon:"hello"}` in the settings file, users would be unable to change the setting if `!!` isn't there

gfwilliams:
Looks good - thanks!

apps/widgps/settings.js
khromov: Good to know, thanks for the heads up!

PR #2088, opened 2022-08-14
[BW Clock] Added option for a dynamic screen mode to show widgets only if unlocked.

peerdavid:
This PR adds an option for a dynamic screen mode to show widgets only if unlocked. If the bangle is locked, widgets are hidden. So BW Clock provides 3 different screen modes:
- Normal: Widgets are always shown (default)
- Dynamic: Widgets are only shown if unlocked
- Full: Widgets are always hidden.

Note: The dynamic screen is implemented through caching the draw function first into a _draw for each widget. Afterward, I overwrite draw with an empty function if fullscreen, and I use _draw otherwise to avoid the .loadWidgets call as its relatively slow.


gfwilliams:
Nice, thanks! There was some talk of adding a library that'd handle widgets, but unfortunately I haven't got around to that yet.

2v15 does at least provide some nice functionality that allows you to 'overlay' on top of graphics though

PR #2089, opened 2022-08-15
Update macwatch2 app to use time based on locale settings

eyecreate:
I liked the watch but wanted it to show 12/24 hour time based on locale settings.


gfwilliams:
Looks great - thanks! Please can you bump the version number in metadata.json and add a line to the ChangeLog for it? Then existing users will be able to see there's an update

gfwilliams:
Thanks!

PR #2090, opened 2022-08-17
Use /usr/bin/env as shebang everywhere

sersorrel:
It's not guaranteed that bash will be available at /bin/bash; it's even less likely that node will always be installed at /usr/bin/node{,js}. Always indirecting through /usr/bin/env means that there is only one such path that needs to be available on every system everywhere forever.

In particular, /bin/bash _is_ frequently available, but is not included in NixOS, which means I can't use the scripts unmodified for local development.

Note: `npm run update-local-apps` works for me after this, but I have not tested whether this breaks e.g. the github actions stuff.

### DIFF
--- a/bin/apploader.js
+++ b/bin/apploader.js
@@ -1,4 +1,4 @@
-#!/usr/bin/nodejs
+#!/usr/bin/env nodejs
 /* Simple Command-line app loader for Node.js
 ===============================================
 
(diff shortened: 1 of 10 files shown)


gfwilliams:
Thanks - just merging.

In general though I'm against this sort of thing, as realistically not too many people use Linux, and I reckon 99% of those have bash at `/bin/bash` - realistically this change is *just* for you, and as a NixOS user, you must hit this for pretty much every script you come across anyway

sersorrel:
Stuff that gets installed ends up being mostly ok, actually – there's a script which automatically patches shebangs to refer to a path in the Nix store. It's mostly just an issue for scripts like this that you can't really install, and a significant number of them use /usr/bin/env (or /bin/sh, which is the other such "compatibility" path which NixOS ships by default) already. Fair enough, though, I'll try and avoid submitting similar patches to you in future.

PR #2091, opened 2022-08-18
[BW Clock] Added agenda to menu

peerdavid:
The agenda can now be shown directly in BW clock if the calendar is synced with Gadgetbridge...


gfwilliams:
Great, thanks!

PR #2092, opened 2022-08-19
[Pomodoro] Display widgets and back-button

thyttan:
@ra100 @adamschmalhofer 

... and switch out g.getWidth/g.getHeight and use Bangle.appRect instead.
... and use g.clearRect instead of g.clear.

I don't know if the changes work on a bangle.js 1 as I don't have one to test with. If someone can test it that would be great! Load it from [here](https://thyttan.github.io/BangleApps/?q=pomodoro).

Converted to draft because displaying countdown doesn't behave. The font turns very small sometimes for some reason.

### DIFF
--- a/apps/pomodo/pomodoro.js
+++ b/apps/pomodo/pomodoro.js
@@ -25,12 +25,12 @@ class State {
 
   clear () {
     clearWatch();
-    g.clearRect(Bangle.appRect.x, Bangle.appRect.y, Bangle.appRect.x2, Bangle.appRect.y2);
+    g.clearRect(Bangle.appRect);
     g.setFontAlign(0, 0);
   }
 
   draw () {
-    g.clearRect(Bangle.appRect.x, Bangle.appRect.y, Bangle.appRect.x2, Bangle.appRect.y2);
+    g.clearRect(Bangle.appRect);
   }
 
   init () { }
@@ -112,8 +112,6 @@ class InitState extends State {
   }
 
   draw () {
-    g.clearRect(Bangle.appRect.x, Bangle.appRect.y, Bangle.appRect.x2, Bangle.appRect.y2);
-    g.setFontAlign(0, 0); // center font
     g.setFont("Vector", 50); // vector font, 80px
     drawCounter(this.timeCounter);
   }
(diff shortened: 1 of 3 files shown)


apps/pomodo/pomodoro.js
```diff
@@ -25,12 +25,12 @@ class State {
 
   clear () {
     clearWatch();
-    g.clear();
```
gfwilliams, line 28: Just for the future, I believe just `g.clearRect(Bangle.appRect)` will work here too

gfwilliams:
Looks good! I'd be pretty sure the font size is because a widget/something else is resetting the font size between calls to the countdown timer.

All you'd need to do is move:

```
    g.setFontAlign(0, 0); // center font
    g.setFont("Vector", 50); // vector font, 80px
```

into `drawCounter` - because that's what gets called in setInterval and where the `drawString` command that wants the font set is.

thyttan:
Thanks! I did as you suggested.

apps/pomodo/pomodoro.js
thyttan: Changed to that!

thyttan:
I think it's ok now.

thyttan:
... It seem's I have to do something similar when the countdown is finished. The app doesn't behave after pressing "again" or "break", e.g. the font colour is not white but light blue instead.

gfwilliams:
Ok - all you'd need to do then is (in the same way) call `g.setColor`/etc.

It's good practise to call `g.reset()` at the beginning of your draw function - that will restore everything to the 'default' (colors, font, etc). Then you can set things up after that, and you'll be sure that everything will be rendered the same way each time

gfwilliams:
you think this is ready now, or did you want to add that `g.reset` call?

thyttan:
Not ready for merging I think. Lost interest somewhat, the timer functionality on BWClock is a good substitute for me. Might come back to it though.

Should I close the PR?

gfwilliams:
Ok, no worries. Let's close this for now then - if someone wants to tweak it as mentioned above, feel free and re-issue a new PR :)

PR #2094, opened 2022-08-20
[BW Clock] Fix - Step count was no more shown in the menu.

peerdavid:
Unfortunately, I introduced a bug with the last update and the step-count was not shown correctly. Fixed in this PR


gfwilliams:
Thanks!

PR #2097, opened 2022-08-22
[LCARS] Add ability to disable alarms 

khromov:
Really like this watch face, but I would find myself accidentally starting alarms when trying to touch right to go to the activity view, and occasionally also have "ghost" alarms where I didn't feel like I pressed something but must have accidentally started an alarm. This adds a simple setting for disabling the functionality without having to uninstall the Scheduler library.


gfwilliams:
Looks great - thanks!

PR #2098, opened 2022-08-23
agenda: added awareness of allDay field

glemco:
Hide the time on `allDay` events, needs an up-to-date gadgetbridge to get the field right.


gfwilliams:
Great - thanks!

PR #2099, opened 2022-08-24
New App - Chimer

aaronrolls:
Chimer is a fork of Hour Chime that adds extra features such as:

- Buzz or beep on every 60, 30 or 15 minutes.
- Reapeat Chime up to 3 times
- Set hours to disable chime


gfwilliams:
Thanks!

PR #2100, opened 2022-08-24
[BWClock] Set timer for agenda entries

peerdavid:
With the previous version agenda entries were included which is extended in this version:
You can now click in the middle of the screen to immediately set a timer such that your BangleJs reminds you when a meeting takes place -- I hope that I now never forget a meeting again :-P The small symbol at the right shows if a timer is set. If the users clicks again, the timer is disabled. 

Timer disabled for agenda entry:
screenshot_3

Timer enabled for agenda entry:
screenshot_4

And clearly, everything works also if fullscreen mode is disabled:
screenshot_4


gfwilliams:
Thanks!

PR #2103, opened 2022-08-26
[BWClock] Make BWClock and "Digital clock widget" work together

thyttan:
@peerdavid

Moved setUI("clock") to before loadWidgets().

This builds on PR #2100, so should be merged after that one.


peerdavid:
Thank you so much for fixing this! For my understaning -- when I swap the order, widgets are aware that the current app is a clock?

thyttan:
My pleasure 😉

Yes, that's how understand it as well 🙂

Thank you for developing my favourite clock! 😃

gfwilliams:
Thanks! 

> widgets are aware that the current app is a clock?

Yes, that's right! I just added a (very simplistic) check to `bin/sanitycheck.js` and it now picks up a bunch of apps that do this. If someone has time it'd be really appreciated if you could have a stab at fixing some of them

PR #2106, opened 2022-08-30
Calclock

LeonMatthes:
CalClock (short for "Calendar Clock") is a watchface that displays the next calendar events right on the watchface.

@gfwilliams I'm currently daily-driving this watchface, I've noticed that my calendar events currently don't sync reliably. 
I'm using a Samsung Galaxy S9+ with Gadgetbridge stable release.
Most events from Google Calendar are synchronized correctly, but events from my CalDav calendar (synchronized using DAVX5) don't show up at all.
Maybe using this watchface could be a good option for you to debug any problems with the calendar synchronization.


gfwilliams:
Looks great, thanks!

Odd about the calendar sync though - although I'm not sure if you're following but it seems Gadgetbridge is going through quite a lot of development on the calendar side of things at the moment and I think some things did get broken, so I wouldn't be entirely surprised if this was fixed in the next release.

PR #2107, opened 2022-09-02
HRM: Add fft elimination algorithm to remove MA

nxdefiant:
HRM: Add fft elimination algorithm to remove MA

Remove Motion Artifacts by cutting out the frequencies from the HRM frequency spectrum that are noisy in acceleration spectrum.
Under motion this can report a heart rate that is closer to the real one but will fail if motion frequency and heart rate overlap.

The fft with size=256 is calculated over a window of 8 seconds with step size 2 seconds.
Sample rate of acceleration (12.5Hz) and PPG sensor (25Hz) is unchanged, so the algorithm runs with 12.5Hz.
Generally the maximum peak in fft spectrum in an interval of -5..+10BPM of last one is used.

The low fft resolution limits the accuracy to ~3BPM, but mean error is ~5-10BPM.

If firmware reports confidence >= 90% its value is used.

Also see http://forum.espruino.com/comments/16668193/

### DIFF
--- a/apps/hrm/ChangeLog
+++ b/apps/hrm/ChangeLog
@@ -8,3 +8,4 @@
 0.08: Don't force backlight on/watch unlocked on Bangle 2
 0.09: Grey out BPM until confidence is over 50%
 0.10: Autoscale raw graph to maximum value seen
+0.11: Add fft elimination algorithm to remove motion artifacts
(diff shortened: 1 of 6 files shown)


gfwilliams:
This looks great! Do you think you could just make this a separate app though? I'm not sure the built-in HRM app is the best place for this...

If it's just a proof of concept then a new app for this looks good, but if it's actually something that can be used then we could turn it into some boot code that'd attempt to overwrite the built-in algorithm.

nxdefiant:
Its more in the direction of proof of concept. So I will copy the hrm app into a new app.

gfwilliams:
Thanks!

PR #2109, opened 2022-09-04
Adds personal work-from-home health widget

splch:
Follow the 20-20-20 rule for better eye health when working at a computer. Additionally, stand up or sit down to keep good circulation at a desk!


gfwilliams:
Great idea - thanks!

PR #2111, opened 2022-09-05
[gpsnav] Adding support for b2

storm64:
Adding support for bangle.js 2


gfwilliams:
Thanks! Is there any possibility of using the same file for both bangle.js 1 and 2? It's just it makes it easier to maintain later on (eg if some feature gets added or a bug gets fixed)

storm64:
Using the same app.js is possible but would need some effort as the design should be rewritten using the layout module.
For now I'm getting used to the layout module by creating my own watch-face. Afterwards I will have a look into this again. ;)

gfwilliams:
great - thanks!

PR #2112, opened 2022-09-06
[ptlaunch] Add setUI custom clock recognition

storm64:
Update boot.js to support clocks in custom setUI mode, started with: `Bangle.setUI({ mode: "custom", clock: 1, ... });`


gfwilliams:
Thanks!

PR #2113, opened 2022-09-06
Clock apps tell clock widgets to hide.

thyttan:
EDIT: Fixed all clock apps.


thyttan:
I took it as practice on using vim 👍

PR #2114, opened 2022-09-06
[BWClock] Improved handling of callback functions in menu

peerdavid:
Before the menu directly showed data gathered from some function that returned a string (e.g. getSteps). Now the menu supports callback functions such that also data like the altitude can easily be shown: 

download


gfwilliams:
Nice, thanks!

Just a thought but I guess we could try and generalise this somehow - as this sort of thing seems pretty common to want to do in watch faces?

So for instance an app could supply `altimeter.clock.js` and then clocks could scan for any `.clock.js` files and load them into that array (we could have a library that handled it). It'd mean someone could easily add their own implementation to different clocks, and also other clock faces could easily add things that can be cycled through... although I guess theming is a problem.

Just a thought anyway.

peerdavid:
Oh yea that sounds like a great idea and I really like that this logic is not part of the BWClock as it would better fit to the corresponding app -- additionally, if a user installs an app and uses a clock that supports this it is automatically extended :D

Regarding theming I'm not sure if this is really a problem -- currently, this array only exports functions for "getText", "getImg", "next", "previous" and "click" without any theming... so this could be implemented for each `*.clock.js` while the theming is handled later on e.g. in each clock separately.

I'm working on my Bangle apps only as a hobby (and I really love it!) but I can definitely try to implement and provide a prototype for this if you want -- but I can not guarantee until when I'm able to do this :)

gfwilliams:
Please can you take a look at https://gist.github.com/gfwilliams/0477ad86b611182df8d2b731e8af1a90 and see if it works for you?

So the idea is that items are able to auto-update (eg HRM, altitude) but they only do so when they're actually shown. For stuff like altitude that might take a while, it gets shown first with `--` as the label, and then requests that it is redrawn as soon as data comes in. It seemed easier...

Hopefully this way, the `sched` library (for instance) can provide `sched.clkinfo.js` which will list all active timers

gfwilliams:
Ok, great! I'll commit this as a library now then.

So I'll add an optional `run` function to the docs, so then HA could implement that and your clock can call it if the screen is touched and it exists?

gfwilliams:
Just done! 20c4f06718fd5920b1662550a5ef204bf442eed2

peerdavid:
Wow thanks!

peerdavid:
Hi @gfwilliams, I simply want to let you know that I'm working on the changes and I think its really great as it greatly reduces the lines of code needed in BWClock and its so great that the BWclock is extended as apps are installed :)

ALso, I made some changes to clock_info and I have already some questions:
(I copied clock_info directly into my app.js source as its a bit easer to upload directly via the web ide -- this is my branch https://github.com/peerdavid/BangleApps/tree/feature/bw_use_clock_infos)

1.) I think per convention all items of the menu should call the listener immediately after show(), otherwise a draw is triggered when the first event occurs -- for the BW clock this would mean e.g. that steps are only shown if someone walks. I changed the clock_info accordingly (note that clock_info is currently part of the bw-clock app.js as its easier for me to upload via the webide -- I will later change this if you agree with those changes)

2.) I think it would be great to include ids to all items. E.g. if I include this in home-assistant, instead of exposing a list I now expose an object which contains an id (home_assistant) together with all items such that other apps that use clock_info could also exclude items using the id.

3.) The first element displays now an "overview" of what follows -- this could help users to navigate. As an example this is how you can navigate in BWClock via right/left and up/down...

```
  Bangle   -- HomeAssistant -- ...
    |             |
   Bpm         Trigger1
    |            
  Steps  
    |
  Battery
```

4.) Thats mostly an understanding problem - when are modules uploaded to my Bangle? Together with apps? Or with firmware updates? Currently, I added the clock_info source code in the app.js but that's only for me to test and develop...

5.) Apart from that, I think I have everything to completely refactor BWClock :) I will then continue to add clock infos also to other apps. Currently, I have the following in mind: HomeAssistant, Agenda, Weather such that this information will still be shown in BWClock, and its also provided to others that use the clock_info module :) 

Let me know what you think :)

With this, I want to say one more time - thank you so much for all the great work and this really really great open-source project!


Best, 
David

gfwilliams:
Thanks! I'm glad it's helping! I think the main benefit will be when more clocks start using it :)

> 1) all items of the menu should call the listener immediately after show()

I had actually expected that 'draw' would be called immediately after `show` anyway - like in  https://github.com/espruino/BangleApps/commit/1f3695efff96ddb3f894022f8f4d325b9a6e618e#diff-00d918fab679ddd859bdcd79e6c3e86474e1c74f8a25cc9c64d36784517f5146R111-R112 

But yes, you're right. Just calling it automatically after `show` would save on code complexity!

> 2.) I think it would be great to include ids to all items.

Yes, an ID makes sense! However I think you were proposing returning `{bat:{...},steps:{...}}`?

Only thing I'd say there is I think you'd got an implementation of timers, where each timer had a menu entry? If so, you'd have to give each one an ID to add it to the object.

It might be better keeping the array returned and having `[{id:"bat",....}, {id:"steps",....}]` and then you can have more than one item with the same ID. It's easy enough to filter then with `items = items.filter(itm => itm.id!="timer"`.

> 3.) The first element displays now an "overview" of what follows 

Sounds great!

> 4.) Thats mostly an understanding problem - when are modules uploaded to my Bangle? Together with apps? 

Yes - it's more or less what you've done manually, but it just happens automatically. https://github.com/espruino/BangleApps/blob/master/modules/README.md might help.

Right now I thought that made sense, but potentially we could make it a whole new type of app that gets uploaded separately with its own settings page. Changing that later wouldn't be a big deal as you'd still use it like a module.

> 5.) Apart from that, I think I have everything to completely refactor BWClock :) I will then continue to add clock infos also to other apps.

This sounds great! If you notice any common code that you end up writing the same in each app (like navigation?) then maybe we could add that to the library too?

peerdavid:
Thanks for your answer that helps a lot!

For (2) I meant an ID for the whole app (e.g. timer / homeassistant / bangle etc.) and not for each single item -- So I would propose to create an object for each app. E.g. for the bangle default entries in the lib I would use:
```javascript
var items = [{
    name: "Bangle",
    img: atob("GBiBAf8B//4B//4B//4B//4A//x4//n+f/P/P+fPn+fPn+fP3+/Px+/Px+fn3+fzn+f/n/P/P/n+f/x4//4A//4B//4B//4B//8B/w=="),
    items: [
    { name : "Battery",
      get : () => ({
        text : E.getBattery() + "%",
        img : atob(Bangle.isCharging() ? "GBiBAAABgAADwAAHwAAPgACfAAHOAAPkBgHwDwP4Hwf8Pg/+fB//OD//kD//wD//4D//8D//4B//QB/+AD/8AH/4APnwAHAAACAAAA==" : "GBiBAAAAAAAAAAAAAAAAAAAAAD//+P///IAAAr//Ar//Ar//A7//A7//A7//A7//Ar//AoAAAv///D//+AAAAAAAAAAAAAAAAAAAAA==") }),
      show : function() {
...
```
This way I can categorize different entries. E.g. battery, hrm etc. goes under "Bangle", timers go under "Timer", ...  I will refactor it this way and propose a solution afterwards - we can always refactor this easily :)


> This sounds great! If you notice any common code that you end up writing the same in each app (like navigation?) then maybe we could add that to the library too?

Yea sure I will definitely do this -- I planed to update my LCARS clock too and add screens dynamically - I hope that I then detect duplicated code with BW clock that could go into the lib later on... this way we also have already 2 clocks which make use of the clock infos :)

Best, David

PR #2115, opened 2022-09-06
Created a new watchface

elykittytee:
- the imported font distorted a bit, hence the "distortclk" name


gfwilliams:
Great, thanks!

If you get a moment, please could you remove the `secondInterval` code and use the `queueDraw` code from https://www.espruino.com/Bangle.js+Clock+Font#simple-clock instead ? I'll update the clock tutorial to mention this as well.

The reason is just that as far as I can tell you're only drawing hours and minutes? But with `secondInterval` you're drawing every second - so you're drawing 60x more than you need to, and that'll hurt your battery life :)

You might also find `require("date_utils").month(d.getMonth()+1)` is really handy instead of your `january` function too:

* developing with these modules: https://github.com/espruino/BangleApps/blob/master/modules/README.md
* date_utils module: https://github.com/espruino/BangleApps/blob/master/modules/date_utils.js#L44-L52

It'll work the same, but for someone in a different country it'll magically use the month name in their lamguage :)

elykittytee:
Ooooo that's handy! I'll make the suggested changes and submit another pull request when it's up and working. Thank you!

Issue #2118, opened 2022-09-08
[BWClock] Feature suggestion: Week number


thyttan:
@peerdavid 

I suggest introducing an option for showing week number along side date/day info. Or maybe have it as a part of the 2d menu on the default/'only time' menu entry.

peerdavid:
Yea I like that -- I think in the default menu (just before bangle) that would fit well :)

peerdavid:
I just opened the PR #2119 :
![download](https://user-images.githubusercontent.com/4281901/189142380-4357ea11-f9de-41a7-b5bd-8282bf21804c.png)

thyttan:
Very nice!

PR #2119, opened 2022-09-08
[BW Clock] Show calendar week

peerdavid:
Now the calendar week can be shown as suggested in #2118.:

download


gfwilliams:
Neat - thanks!

PR #2120, opened 2022-09-09
Apploader.js improvements

rigrig:
* Don't hide errors while loading `noble`:
  Loading the module can also fail when e.g. compiled against a different Node.js version, don't hide those errors.
* Add `-b1` param to upload to Bangle.js 1


gfwilliams:
Handy - thanks!

PR #2121, opened 2022-09-09
Use 10-minute acceleration and charging for isWorn()

splch:
removes temperature due to weather situations

this function achieved 97.3% [pr auc](https://stats.stackexchange.com/a/123370) on the dataset.

### DIFF
--- a/modules/wear_detect.js
+++ b/modules/wear_detect.js
@@ -11,7 +11,7 @@ exports.isWorn = function() {
   return new Promise(resolve => {
     if (Bangle.isCharging())
         return resolve(false);
-    if (Bangle.getHealthStatus().movement > 146)
+    if (Bangle.getHealthStatus().movement > 124)
         return resolve(true);
     return resolve(false);
   });
--- a/modules/wear_detect.js
+++ b/modules/wear_detect.js
@@ -11,9 +11,7 @@ exports.isWorn = function() {
   return new Promise(resolve => {
     if (Bangle.isCharging())
         return resolve(false);
-    if (E.getTemperature() > 24.625)
-        return resolve(true);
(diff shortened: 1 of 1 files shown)


modules/wear_detect.js
```diff
@@ -11,9 +11,7 @@ exports.isWorn = function() {
   return new Promise(resolve => {
     if (Bangle.isCharging())
         return resolve(false);
```
storm64, line 13: Nice idea to make a library.

I'd like to suggest to add a check for the global sleeplog object, as the [sleeplog app](https://banglejs.com/apps/?id=sleeplog) is tracking the wearing status too:
```
// sleeplog.status values:
// undefined = service stopped,  0 = unknown, 1 = not worn, 2 = awake, 3 = light sleep, 4 = deep sleep
if (global.sleeplog) {
  if (sleeplog.status) return resolve(sleeplog.status > 1 ? true : false);
}
```

gfwilliams: Sounds like a good plan. I'll merge this PR but we can add another one with this in

modules/wear_detect.js
```diff
@@ -11,9 +11,7 @@ exports.isWorn = function() {
...
-    if (E.getTemperature() > 24.625)
-        return resolve(true);
-    if (Bangle.getAccel().mag > 1.045)
+    if (Bangle.getHealthStatus().movement > 124)
```
gfwilliams, line 14: Actually worth noting that the docs mention `movement is the 32 bit sum of all acc.diff readings since power on` - so this reading keeps going up. If you call this right after the 10 minute timeslot changes it'll be 0, but 9 minutes after it'll be 9x bigger than the value 1x after - I think this line needs a bit more thought.

PR #2122, opened 2022-09-11
[Layout] Add additional options

storm64:
Add options for a x/y separated padding.
Add options for offsetting in x and y direction
Add option to do a pixel perfect clear by redrawing the value with its background color. (Needs to be tested with lazy rendering.)

- [x] readme edited
- [ ] minification to do


gfwilliams:
Hi - I'm not so sure about these... 

Do you have more info about what `pxClear` actually does? I mean, using negative padding values isn't something that's really supported anyway (since it would cause overlap) so I'm not sure we should be adding code to work around that.

Also, with padx/pady - any code that gets added to the layout code really slows it down (and lets face it, it's already too slow). Do you *really* need this for what you're working on? I guess if you absolutely do, we could just rename `pad` to `padx/y` in the layout code, and then do something like `if (l.pad) { l.padx=l.pady=l.pad; delete l.pad;}` in https://github.com/espruino/BangleApps/blob/236aa389c2f7d0640f0326598f16df35407c5287/modules/Layout.js#L70

But if it's not vital, I'm not sure it's worth slowing down everyone's user experience just so one app can be written a little more elegantly.

storm64:
I understand your concerns as I experienced that noticable loading time, too.

I do need the extra options for my [6tilesclk](https://github.com/storm64/BangleApps/tree/6tilesclk/apps/6tilesclk) to display as much data as possible.
For this purpose the spacing around the strings is to wide and I could easily reduce this with some negative padding and a little bit of offset.
On negative padding values the clearing obviously can't work out right, so I added the pixel perfect clearing as this just resets the altered pixels to the background color.

For now I added the changes as custom module to the app. Do you think this might be a better option?

gfwilliams:
Sorry for the delay getting back to you on this.

I think for the moment, yes, it's probably better to keep this as a custom module in the app. Realistically it won't make any difference to your app, because that's what happens internally anyway.

If lots of people keep having issues with this then we can definitely resurrect this though.

PR #2125, opened 2022-09-16
DNDtoggle - Initial release

pebl-hank:
The app itself seems to be working now as intended, but I am struggeling with the icon for the launcher.
Can you help me to find a conversion that works for Dark and Light mode? The icon is transparrent and basically 2 color. It always shows wrong - regardless of the settings I use in the image converter.


apps/a_dndtoggle/a_dndtoggle.app.js
```diff
@@ -0,0 +1,44 @@
...
+
+E.showMessage(modeNames[current] + " -> " + modeNames[bSettings.quiet]);
+setTimeout('exitApp();', 2000);
+
+
+function exitApp(){
+
+require("Storage").writeJSON("setting.json", bSettings);
+// reload clocks with new theme, otherwise just wait for user to switch apps
+if (Bangle.CLOCK) load(global.__FILE__);
```
rigrig, line 40: No need for this: we always call `load()` anyway in the next line.

apps/a_dndtoggle/a_dndtoggle.app.js
```diff
@@ -0,0 +1,44 @@
...
+		Bangle.buzz();
+		break;
+	case 2:
+		bSettings.quiet = 0;
+		Bangle.buzz();
+		break;
+	default:
+		bSettings.quiet = 0;
+		Bangle.buzz(); 
+}
```
rigrig, line 28: The last three cases are all the same, and we always buzz, so you could just replace it with
```js
if (current === 0) {
    bSettings.quiet = 2;
    setTimeout('Bangle.buzz();',500);
} else {
    bSettings.quiet = 0;
}
Bangle.buzz(); 
```

pebl-hank: This is for future use. I will make the behavior configurable.

apps/a_dndtoggle/a_dndtoggle.app.js
pebl-hank: Good point. Will change that.

gfwilliams:
Hi! Sorry for the delay. For the icon, you just want this?

image

So basically on a whilte background you want it black (as above) and on black, you want it white? But no other colors?

Or you say "2 color" - was part of the inner meant to be filled? Becayse I don't think it is in the source image.

I think in the image converter you want `1 bit black/white`, `transparent` and also `inverted` (you want the image to show up as white in the preview).

Try:

```
require("heatshrink").decompress(atob("mEwwJC/AAl/Agf/AAUAgIFDwEHAofgh/g/0Ag/wj+AnwVB/EegEfEIN4nkAh+AgE8vgVBAoV4Aoce/EAgfADQIFcjwpFHYIFCnxBFJopZBn5ZCMopxFPoqJFSowA/gA="))
```

pebl-hank:
> require("heatshrink").decompress(atob("mEwwJC/AAl/Agf/AAUAgIFDwEHAofgh/g/0Ag/wj+AnwVB/EegEfEIN4nkAh+AgE8vgVBAoV4Aoce/EAgfADQIFcjwpFHYIFCnxBFJopZBn5ZCMopxFPoqJFSowA/gA="))

This shows a totally white icon in light mode and a totally black icon in dark mode.
Actually I'd be totally fine if the inner of the image is also transparent. So technically the image is 1 color (black) with transparent background. The black should be shown as black in light mode and as white in dark mode.

gfwilliams:
This shows a totally white icon in light mode and a totally black icon in dark mode.
Are you sure? I just tried it in the emulator with `g.drawImage(require("heatshrink").decompress(atob("mEwwJC/AAl/Agf/AAUAgIFDwEHAofgh/g/0Ag/wj+AnwVB/EegEfEIN4nkAh+AgE8vgVBAoV4Aoce/EAgfADQIFcjwpFHYIFCnxBFJopZBn5ZCMopxFPoqJFSowA/gA=")))` and it seems to work ok for me? Black in the default light mode (basically using whatever the foreground color is)

pebl-hank:
I pushed the change and tried to re-load the app. Maybe something is cached somewhere.
It's hard to test for me. Can I somehow make a g.drawImage(require("heatshrink").decompress(atob( string with Windows out of a PNG? Otherwise I need to commit and push to wait until I can download the app from my app page.

gfwilliams:
You can just connect with https://www.espruino.com/ide/ and paste the code in on the left hand side?

Best to do a `reset()` first as since the draw call will use the current color, some other code might have set it up to something non-standard otherwise

pebl-hank:
It looks good when sending it through the ide. But the launcher icon doesn't look good though.

gfwilliams:
Ahh - I think I see what's happened - you have made a `dndtoggle.img` file.

Please can you just use the format that we do in the examples? https://github.com/espruino/BangleApps/tree/master/apps/_example_app

So `app-icon.js` and then `evaluate:true`: https://github.com/espruino/BangleApps/blob/master/apps/_example_app/metadata.json#L12

gfwilliams:
Great - thanks!

PR #2127, opened 2022-09-18
[slash] If midnight clear the display to clean up artifacts

lightnin:
Since slash is (wisely) very conservative about clearing / redrawing, artifacts from past dates that aren't overwritten tend to stack up. On my watch now where it says "Sun 9/18" there is some left over "T" from last Thursday sticking out to the left of the "S" in "Sun", and some leftover "6" sticking out to the right of the "8" in "18" from two days ago. This should clear everything once a day at midnight so these artifacts get cleared out.

### DIFF
--- a/apps/slash/ChangeLog
+++ b/apps/slash/ChangeLog
@@ -1,1 +1,2 @@
 0.01: First version for upload
+0.02: Fix for leftover date artifacts on display.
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Please can you update the version in the metadata and the changelog?

lightnin:
All set.

gfwilliams:
Thanks!

PR #2131, opened 2022-09-19
feature: set Brightness default to full

xxpasixx:
could not test it on Bangle.js 1 (don't have one) that everything works fine there.


gfwilliams:
Great plan - thanks!

PR #2133, opened 2022-09-19
[BWClock] Use the clkinfo module to dynamically add new features to the clock as apps are installed by the user

peerdavid:
Hi @gfwilliams ,

I just finished a first version of BWClock which now uses clkinfo. I also changed the clkinfo module, and included a clkinfo for the following apps:
- HomeAssistant
- Agenda
- Weather

Let me know if you think that something that I created is problematic -- I'm really not a JS expert. Also I'm not sure if I should add the module clkinfo somewhere to the metadata (currently I added nothing about it).

## Next steps (but in new PR in my opinion to keep PR small):
I want to add a clkinfo.js to the following apps:
- Simple Timer: For alarm / sched I think we should show the alarms and timers, but for SimpleTimer we could expose a timer menu i.e. enable / disable for n minutes

And I plan using clkinfo in the following apps:
- LCARS -- here I think I can show one screen for each app which would be nice. Not 100% sure when I have time to implement it but I will definitely do this ;)
- Info -- Could be interesting to add this to the Info app too...

Thanks again for all your work @gfwilliams 

David


modules/clock_info.js
```diff
@@ -37,57 +37,74 @@ exports.load = function() {
...
       alt = Math.round(data.altitude) + "m";
-      items[3].emit("redraw");
+      bangleItems[3].emit("redraw");
     });
   }
+  function deleteInterval(inter){
+    if(inter){
+      clearInterval(inter);
+    }
+    delete inter;
```
gfwilliams, line 56: This doesn't work I'm afraid. It will just delete the argument `inter` but not the actual field.

So `clearInterval(this.interval); delete this.interval;` is needed - or you could do `deleteInterval(this)` and then change this to `clearInterval(obj.interval);delete obj.interval;` - but it doesn't allow for >1 interval (all intervals would have to just be called `this.interval`)

gfwilliams:
Thanks! This looks amazing! And it's great to keep the PR small.

Before merging it'd be great if you could fix two things though:

* the change to `deleteInterval` mentioned above - as this won't actually remove the interval field
* `clock_info` returning a 'tree' of items...

Right now, you return:

```
var items = [{
    name: "Bangle",
    items : [ ... ]
```

So it's actually a tree rather than just a list. If that works better for you then great - but...

Please could you update the docs at the start of `clock_info.js` as well as the example code at the end?

I'm also thinking that maybe someone might want to add a plugin that adds more options to the standard `Bangle` list of items? If that happens right now we'll end up with two `Bangle` lists... So maybe instead of:

```
require("Storage").list(/clkinfo.js$/).forEach(fn => {
    items = items.concat(eval(require("Storage").read(fn))());
  });
```

in clock_info we need to do something fancier (so that we first check to see if `items` contains something with the same `name`...)

peerdavid:
Thank you for the detailed review and for the fixes. I will change the code accordingly :) And I really love the idea that clkinfos can be merged e.g. to extend the Bangle list -- I will add a more advanced merge code in the clkinfo too :)

modules/clock_info.js
peerdavid: I use this.interval right now. Thanks for this hint!

peerdavid:
Hi @gfwilliams,

I just updated the clk_info module: 
(1) I now use `this.interval` as it was before (sry I don't know why I changed your code). 
(2) I updated the doc in `clk_info.js` and try to explain menu objects and menu items.
(2) I changed the function to load clk info files as follows:
```js
  // In case there exists already a menu object b with the same name as the next
  // object a, we append the items. Otherwise we add the new object a to the list.
  require("Storage").list(/clkinfo.js$/).forEach(fn => {
    var a = eval(require("Storage").read(fn))();
    var b = menu.find(x => x.name === a.name)
    if(b) b.items = b.items.concat(a.items);
    else menu = menu.concat(a);
  });
```

So whenever a menu object with the same name already exists, the new object is not added, the items are appended. I tested it by renaming all to "test", then I got only a single menu object "test" containing all menu items :)

If there is anything else - let me know I'm happy to improve and work on this :)

gfwilliams:
That looks excellent - thanks for your work on this! Hopefully we'll see some others include this in clocks soon as well :)

peerdavid:
Cool thanks - was fun doing this :)
 --> I will definitely include this in some other clocks ;)

gfwilliams:
Great! Just thinking, I guess it might help to add some drawing/navigation code to `modules/clock_info.js` if it looks like we're doing the same thing in each clock :)

peerdavid:
Yes definitely -- if I implement this for LCARS and I find duplicated code, I will move this code into `modules/clock_info.js`. But I think its best to do this while implementing the next clock -- mainly because I'm not sure what code exactly can be shared :)

Issue #1971, opened 2022-06-16
[Bluetooth Widget with notification] FEATURE Request


andiohn:
### Affected hardware version

Bangle 2

### Your firmware version

2v14

### The bug

Simply, it would be nice to be able to have no icon when connected and red bluetooth when disconnected. I don't need to know when there isn't a problem, just when there is. :)

gfwilliams:
Do you care about the notifications? Why not just make a new Bluetooth widget based on the original simple widget, but with the logic you describe. It'd only take a minute :)

pebl-hank:
https://github.com/espruino/BangleApps/pull/2011 introduces Settings for the plugin. It should be no big deal to add this feature in future versions using the configuration.

pebl-hank:
Added this to the PR.

gfwilliams:
Closed with https://github.com/espruino/BangleApps/pull/2134

PR #2134, opened 2022-09-21
[widbt_notify] Simplify code, add some options

storm64:
Should close feature request #1971 (if it shouldn't be closed already ;) ).


gfwilliams:
Great - thanks! 

Just a thought but the notification could actually use http://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setLCDOverlay now (at least on Bangle.js 2) and then it can avoid having to redraw

storm64:
Good idea, I will have a look in to this.

PR #2135, opened 2022-09-21
[Swipe Menus] Initial release

thyttan:
Replace built in E.showScroller to act on swipe instead of drag. Navigate menus in discrete steps instead of a continuous motion.

Would it make more sense to have the name be something with 'Scroller' instead? Is the scroller used for stuff other then menus?


apps/swscroll/boot.js
```diff
@@ -0,0 +1,98 @@
...
+s.draw(); // draw the full scroller
+g.flip(); // force an update now to make this snappier
+Bangle.setUI({
+  mode : "custom",
+  back : options.back,
+  swipe : (hor,ver)=>{
+    pixels = 120;
+    var dy = ver*pixels;
+    if (s.scroll - dy > menuScrollMax)
+      dy = s.scroll - menuScrollMax-8;
```
thyttan, line 59: This is what causes the error ```Uncaught Error: Cannot read property 'title' of undefined...``` which is mention in the README.md

gfwilliams:
Nice - thanks! Now you know what causes the error, maybe it's worth putting a quick fix in before we merge?

As far I can see, for `Cannot read property 'title' of undefined` there's no `.title` access in your code so it's one of the bits of code that is calling `showScroller` (is it `E.showMenu`?). Maybe we should make it explicit that showScroller might call back with an out of range index and anything using it should handle that? But I guess that might need quite a few code changes!

thyttan:
>maybe it's worth putting a quick fix in before we merge?

Yeah, that makes sense 👍

gfwilliams:
Great - thanks!

PR #2138, opened 2022-09-22
Twenties - vibrates during work hours

splch:
The watch would vibrate all night, harming battery life. This PR makes the watch only vibrate from 8 AM - 5 PM.

### DIFF
--- a/apps/widtwenties/ChangeLog
+++ b/apps/widtwenties/ChangeLog
@@ -1,3 +1,3 @@
 0.01: New Widget!
 0.02: Fix calling null on draw
-0.03: Only vibrate during working hours
+0.03: Only vibrate during work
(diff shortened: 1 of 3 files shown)


splch:
Style-wise, if a widget doesn't draw, how should we handle the draw method?

Not including it throws an error during widget drawing times, so probably not that.

I'm debating between `draw: _ => { return null; }` and `draw: _ => { }`.

It might be worth making a PR to check for existence of a draw method in a widget before calling. An optional chain `?.` sort of thing.

gfwilliams:
Just setting `draw` to an empty function is fine. returning `null` is not needed.

**However** if you don't actually want to display anything in the widget area is there a reason this isn't actually just 'boot code'? Because then it'd work even on apps/clocks that don't display widgets

splch:
> **However** if you don't actually want to display anything in the widget area is there a reason this isn't actually just 'boot code'? Because then it'd work even on apps/clocks that don't display widgets

Boot code! That's exactly what I'm looking for :) Thanks for the direction

PR #2139, opened 2022-09-22
[About] Bangle2 add software back button via setUI

thyttan:


gfwilliams:
Thanks! Just checking - have you tested this? If the callback was called with an argument (I'm not sure if it is) then it might just end up going back to the Bangle.js logo rather than the clock

Also `back` also sets up the button to go back, so the `btn` field isn't really needed.

thyttan:
Thanks for fixing it!

PR #2140, opened 2022-09-22
[Activity Reminder] sofware back button via setUI

thyttan:
I'm unsure if this change breaks the app on Bangle.js 1?


gfwilliams:
> I'm unsure if this change breaks the app on Bangle.js 1?

No, this should be fine - thanks!

gfwilliams:
Just removed `btn:` as this doesn't work anyway (the callback is called with `1` as an argument so `load` would fail) and back via button is added automatically

PR #2143, opened 2022-09-23
[SmplTmr] Add smpltmr.clkinfo.js

peerdavid:
The Simple Timer now adds clock info -- so a timer can be used directly on the watch face :)
I also changed the clock info slightly: If run() returns true, the caller can feedback the user (e.g. with buzz). If its false then no feedback is necessary. I e.g. use a feedback for home-assistant when a trigger was send, but for the timer buzz feels a bit annoying IMO if you push the button a few times, so run returns false :)

And here a screenshot of the BWClock which now shows the timer:
image


gfwilliams:
Great - thanks!

PR #2144, opened 2022-09-23
Convert Twenties to boot code instead of widget

splch:
Twenties doesn't need to draw and should run regardless of watch-face. A widget isn't the correct format, so now it's boot code! :smile:


gfwilliams:
Excellent - thanks!

PR #2145, opened 2022-09-23
Bigdclock v0.06

deirdreobyrne:
- Better battery graphic - now has green, yellow and red sections
- battery status reflected in the bar across the middle of the screen
- better handling of locale

re that last one - not everyone has a locale set up, and when you set a locale up, it seems the date_utils.js disappears or becomes unavailable???

(I've incorporated https://github.com/espruino/BangleApps/commit/5b1b3553c5a028c595845434762147b28de317a6 as best I can).


gfwilliams:
> re that last one - not everyone has a locale set up, and when you set a locale up, it seems the date_utils.js disappears or becomes unavailable???

That shouldn't happen as date_utils gets included in the JS file as it is uploaded. It's just not possible for `locale` to remove it.

I imagine it just happened because you uploaded your app through the IDE without having the path set in the IDE settings? Info at https://github.com/espruino/BangleApps/blob/master/modules/README.md

> not everyone has a locale set up

Absolutely *everyone* has locale set up. It's included in the firmware, so if you don't have one installed you get the default english locale.

Please could you take out those `locale` changes? The original `var dows = require("date_utils").dows(0,1);` will work fine - this gets included all the time by the app loader, and uses locale so will be internationalized. There's really no reason for the `try` - I think it just stems from you uploading from the IDE without having the Bangle.js libraries in your search path

deirdreobyrne:
Yes you were right. Changed implemented. Also I now only check the battery state of charge every 15 minutes instead of every minute, giving me noticeably extra time between having to charge!

gfwilliams:
Great - thanks! Just merging...

PR #2146, opened 2022-09-24
Message events

rigrig:
See http://forum.espruino.com/conversations/379866/

- Emit "message" events for incoming messages.
- Allow event handlers to prevent default app from loading.
- Add setting to hide default widget, meant for people who want to install custom widgets.
- Move `buzz()` from the widget to the library.


gfwilliams:
Thanks! Nice idea moving the buzz logic out of the widget `draw` function

PR #2147, opened 2022-09-24
Add Power Save

kyleplo:


gfwilliams:
Thanks! Although I'd be interested in how much battery life this adds over a normal clock app that updated once a minute. It may not be too much.

It might be worth mentioning that it'll break wake-on-twist, step counting, and other features that rely on the accelerometer.

I'm interested by:

> Due to an Espruino bug, this does not take affect immediately when installed.

What do you mean here? If you upload with the app loader, you'll have to long-press the button to reload apps after it is installed, which I'd have hoped would have got this working...

kyleplo:
> Thanks! Although I'd be interested in how much battery life this adds over a normal clock app that updated once a minute. It may not be too much.
Maybe if I get a chance I'll do some experimenting. I do feel like I'm getting more than the advertised 4 week battery life with this, but I'd have to do further testing to confirm that.
> It might be worth mentioning that it'll break wake-on-twist, step counting, and other features that rely on the accelerometer.
That's a good point. I'll make that feature optional.
> I'm interested by:
> 
> > Due to an Espruino bug, this does not take affect immediately when installed.
> 
> What do you mean here? If you upload with the app loader, you'll have to long-press the button to reload apps after it is installed, which I'd have hoped would have got this working...
This might just be an issue with uploading through the web IDE, I haven't tried installing it from the app loader.

Issue #2124, opened 2022-09-15
[agpsdata] Debug info "error: No http method found"


thyttan:
@pidajo @myxor

### Affected hardware version

Bangle 2

### Your firmware version

2v15.14 (Also had it on 2v15)

### The bug

Description:
When A-GPS Data is installed and the Bangle is connected to Web IDE the message "error: No http method found" is often printed.

Expected behavior:
To there not to be an error.

Steps to reproduce:
1. Connect Bangle to Web IDE
2. Install [A-GPS Data](https://banglejs.com/apps/?q=a-gps%20data)
3. "error: No http method found" is printed in the Web IDE.

The message is often displayed when launching different apps as well.

myxor:
This is triggered in https://github.com/espruino/BangleApps/blob/674ef0aa133a22724c532b11e8c51445ed411c01/apps/agpsdata/lib.js#L72 because the WebIDE does not (yet*) support the HTTP call functionality.

Which confuses me is that you say you see this log message often. This should not be.



*: there is somewhere the idea to extend the WebIDE with this functionality.

gfwilliams:
Do you have the `Android Integration` app installed as well? You need that for the A-GPS Data app to work

thyttan:
@myxor 
If I am connected to the Web IDE on my laptop, I seem to get the error every time I load any app.

@gfwilliams 
Yes, Android Integration installed. When I'm connected to gadgetbridge everything seems to work as expected 👍.

Feel free to close this if we're done! :)

thyttan:
A little OT:
@myxor @pidajo 
Should "A-GPS Data" be added as a dependency for "AGPS Widget"? And maybe the descriptions could be clearer about if they are dependent on each other or not?

myxor:
> A little OT: @myxor @pidajo Should "A-GPS Data" be added as a dependency for "AGPS Widget"? And maybe the descriptions could be clearer about if they are dependent on each other or not?

They are independent from each other.

gfwilliams:
> because the WebIDE does not (yet*) support the HTTP call functionality.

The Android Integration app is loaded regardless of if you're in the IDE or not, so `Bangle.http` should exist and you shouldn't get the error.

Maybe when you're in the IDE you could check if `Bangle.http` exists?

>  And maybe the descriptions could be clearer about if they are dependent on each other or not?

[...]

PR #2149, opened 2022-09-27
AGPS-Data

myxor:
- Do not load AGPS data on boot
- Increase minimum interval to 12 hours

as suggested in https://github.com/espruino/BangleApps/issues/2124#issuecomment-1253497336


myxor:
Should we set the minimum update Interval to 6 hours as the agps data seems to be only valid for this timespan?

See http://forum.espruino.com/comments/16701648/

gfwilliams:
That's excellent, thanks! And sorry, in that post above I'd missed the `settings.refresh / 60` in your code, so I saw `settings.refresh * 1000 * 60` assumed those figures were in minutes :)

PR #2151, opened 2022-09-27
Improve accellog

asmaps:
* added maximum G display to the recording screen
* Log actual G values to CSV (was with factor of 8192, probably from some old API)
* Added total G / magnitude to CSV
* Don't exit immediately to menu after recording, but keep showing the stats after stopping, a second button press leaves to menu


gfwilliams:
Thanks! But please can you make the data format that is used an option in the menu?

x8192 was used because then we're storing the actual data that comes off the sensor (and we can store it without decimals so it doesn't use up extra space in the file). If you store in `g` with 2 decimal places you're losing quite a lot of accuracy.

While maybe many people don't care, the app is used already by things like https://github.com/gfwilliams/step-count / http://forum.espruino.com/conversations/359542/ and changing the format of data that it records would break that (and would cause problems for pretty much anyone else that is using it already).

asmaps:
I added a menu point in the main menu to opt into logging "raw" sensor values.

gfwilliams:
Great - thanks!

PR #2153, opened 2022-09-29
[LinuxClock] A new Clock

peerdavid:
Hi,

so I decided to create a new clock based on the new awesome clock_info. My idea was simple: Each "clock_info" is a "linux_command":

This is the home screen (date cmd), other screens can be selected via left/right:
screenshot

And here the home assistant screen -- items can be select via up/down (left >) and triggered via center click ... 
screenshot_2


gfwilliams:
Looks great - thanks!

PR #2155, opened 2022-09-29
[kbmulti] switch directions for Caps lock switch and Move cursor mode

thyttan:
I switched the direction for Caps lock switching and Move cursor mode. I have two rationalizations for this. One, I feel the top-to-bottom swipe motion is more natural/pleasant to perform and should therefore be used for the more frequently used task - for me, and I suspect most users, this is to switch Caps lock. Two, since we start with upper case it makes sense to make a downward motion to get to lower case letters.

A counter argument is doing this switch might interfere with established habits. Should this be a setting in that case? If so I suggest the default to be as what I change it to with this PR.

What do you think? @sir-indy @halemmerich


sir-indy:
You've made it very clear to the user what the new setting is, and while I had fun making this, I've not really used it in anger yet, so I wouldn't worry about established habits.
I'm happy for you to go ahead as you wish.

halemmerich:
What he said ;) Directions were essentially chosen by fair dice roll so no objections.

thyttan:
Thanks for your inputs! 🙂

This is ready for being merged then.

gfwilliams:
Great - and thanks for checking up with the original authors :)

PR #2156, opened 2022-09-30
Fast switch for imageclock and iconlaunch

halemmerich:
Launcher -> Clock works fast on 2v15, the other way around only works for 2v15.20 upwards.

Diff seems to be gigantic, but it's mostly whitespace changes by indention of the surrounding blocks for GC and var->let.


gfwilliams:
Great - thanks!

PR #2159, opened 2022-09-30
widcloselaunch - New widget

halemmerich:
Essentially the same as widclose, but closes to launcher instead of to clock


gfwilliams:
Nice idea - thanks!

PR #2160, opened 2022-09-30
imageclock - Fix the clock not refreshing completely after switch

halemmerich:
Clock did not refresh completely once after switching from launcher if the drawing in timeouts option in the customizer was deactivated.

### DIFF
--- a/apps/imageclock/ChangeLog
+++ b/apps/imageclock/ChangeLog
@@ -9,3 +9,4 @@
       Show/Hide widgets with swipe up or down
 0.08: Use default Bangle formatter for booleans
 0.09: Support new fast app switching
+0.10: Fix clock not correctly refreshing when drawing in timeouts option is not on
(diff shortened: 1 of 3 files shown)


gfwilliams:
There's a warning from the CI check here:

```
Clock imageclock file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)
```

Is it possible to swap the call order around?

Also `if (!global.WIDGETS)` isn't required because loadWidgets is now smart enough to check

halemmerich:
> Is it possible to swap the call order around?

Yes :)

halemmerich:
I had to re-add a loading screen, since some watchfaces take a few hundred ms to load. If the button is pressed in that loading time the launcher would be loaded again without any indication what happened.
Double button press now effectively starts loading the clock and stops it again showing the loading screen for the time in between presses. At least it is still a few times faster than the normal way, which takes around 3 seconds on my daily driver bangle.
This iteration is now ready to merge :)

gfwilliams:
Thanks! Maybe the double press needs some thought - although you'd hope that it should only take effect after setUI is called. Maybe the previous press gets queued as the device is busy - we could try and work around that by adding a setTimeout inside setUI

PR #2161, opened 2022-10-01
Improve Calendar Clock current time contrast

### LINKED ISSUE #2154: [Calendar Clock] Current time is hard to read
### Affected hardware version

Bangle 2

### Your firmware version

2v15

### The bug

The Calendar Clock app uses a cyan colour for the current time text. The cyan is sadly very hard to read on the white background, especially when the screen is dimmed. I absolutely love Calendar Clock, but this colour choice forces me to flick my wrist very often to be able to read the current time.

The future events, however, are quite easy to read.

The problem with design and colour is that it is quite personal, so I am not sure what to suggest for improving this. Naively I would say to "just use a darker colour".

Perhaps the most perfect solution would be to use a darker default colour and make it possible to choose the preferred colour in settings.

I am willing to take a look at this myself as this seems like a quite beginner-friendly issue to tackle if we can agree on the correct way forward :)

(P.S.: Are there any plans to make this repo join [Hacktoberfest](https://hacktoberfest.com/participation/#maintainers)? It would just mean adding a "hacktoberfest" topic to the repo and merging or labelling "hacktoberfest-accepted" PRs that fit your desired standards. I'm still willing to help regardless, but during October I will prioritize repositories that join Hacktoberfest)

TheLastProject:
Fixes #2154, see that issue for the reasoning of how I came to this colour.

It's raining outside today quite badly and it doesn't look like the weather will improve much, so I haven't tested this under more glare yet. But I can notice a big difference just leaving my watch lying on the table with the screen dimmed from not actively using it.

I'm marking this as draft for now and will un-draft it sometime during next week when I have tested it during a regular work day.

P.S.: I'm impressed by how easy the BangleJS tooling is to use. The Web IDE works great and setting up my own App Loader based on the README instructions were easy and helped me feel certain everything would work well before making this PR. Thanks for BangleJS, this project really shows how great Open Source smartwatches can be :)

### DIFF
--- a/apps/calclock/ChangeLog
+++ b/apps/calclock/ChangeLog
@@ -1,3 +1,4 @@
 0.01: Initial version
 0.02: More compact rendering & app icon
 0.03: Tell clock widgets to hide.
+0.04: Improve current time readability in light theme.
--- a/apps/calclock/calclock.js
+++ b/apps/calclock/calclock.js
@@ -68,7 +68,7 @@ function drawEvent(event, y) {
 var curEventHeight = 0;
 
 function drawCurrentEvents(y) {
-  g.setColor("#0ff");
+  g.setColor(g.theme.dark ? "#0ff" : "#006060");
   g.clearRect(5, y, g.getWidth() - 5, y + curEventHeight);
   curEventHeight = y;
 
--- a/apps/calclock/metadata.json
+++ b/apps/calclock/metadata.json
@@ -2,7 +2,7 @@
   "id": "calclock",
   "name": "Calendar Clock",
   "shortName": "CalClock",
-  "version": "0.03",
+  "version": "0.04",
   "description": "Show the current and upcoming events synchronized from Gadgetbridge",
   "icon": "calclock.png",
   "type": "clock",


gfwilliams:
Thanks! This all looks great - glad it all went so well!

All I would say though is that Bangle.js 2 only has a 3 bit screen, so 8 colors. Generally if a color is too far off what can naturally be displayed it is dithered, so your choice of `"#006060"` will cause dithering, eg:

image

Maybe it doesn't matter, but on smaller text that may make things harder to read

TheLastProject:
I've been using `#006060` all day and I can say that it is quite readable, but the moment you pointed this dithering out I took a closer look at the text and I suddenly couldn't unsee the text quality being lower than that of the pure black text. The dithering definitely managed to trick me for quite a while though.

Given the original colour is cyan (`#00ffff`), which is both 100% Green and 100% Blue, blue and green would be the closest colours. Green fails the contrast check, so I switched it to pure blue (`#0000ff`) and... that feels quite natural and looks even better than `#006060`.

gfwilliams:
Great - thanks! I'll get this merged then!

PR #2164, opened 2022-10-04
[gpsnav] Corrections wpindex=0, draw + nm calc for b2

storm64:


nxdefiant:
Does waypointer need the same fix?

storm64:
Just looked into waypointer, yes in deed, I will add this fix too.

gfwilliams:
Thanks! Please could you bump the waypointer version too so that folks know they can get an update?

storm64:
> bump the waypointer version
Oops, forgot about it.
Done.

gfwilliams:
Thanks!

PR #2166, opened 2022-10-05
Imageclock - Improvements

halemmerich:
- Generated drawing code creates less promises
- Additional option in customizer to force drawing directly
- Fix possible early clearing of timeouts created by delay function
- Allow calls to setMatchedInterval without delay parameter to work correctly


gfwilliams:
Neat - thanks!

PR #2171, opened 2022-10-06
iconlaunch - Directly eval apps instead of loading them

halemmerich:


gfwilliams:
I'm really not sure this is a good idea (at least as a default)... For instance if you load any app that doesn't load widgets, widgets will stay loaded and it'll break.

Do you think you could make it an option?

halemmerich:
Thats a valid point. I had looked at this from the standpoint of using the upcoming hidable widget bar. But I see how it would cause problems in the current state of things. Option it is.

halemmerich:
Direct eval is now an non-default option.

gfwilliams:
Great - thanks!

PR #2175, opened 2022-10-13
gpstrek - Improve background/widget handling

halemmerich:
Reduces number of file writes and makes the options a bit more explicit by moving them into their own submenu.


gfwilliams:
Great - thanks!

PR #2177, opened 2022-10-15
[AI Clock] New Clock

peerdavid:
Just for fun, I asked StableDiffusion (https://github.com/CompVis/stable-diffusion) to generate a banglejs watchface. I got the following, generated, image for the input prompt `A rectangle banglejs watchface`

orig

 So I implemented this watchface and called it the AI clock :-P

impl


gfwilliams:
Looks quite good! Maybe even at some point we can even have a custom.html page that creates a clock face from a text prompt :D

andiohn:
I wonder if the AI clock could regenerate with an option in the app loader? Like "re-AI this" in the hamburger menu.

PR #2178, opened 2022-10-15
[Sliding Text Clock] Added Styling and new watch faces

awkirk71:
- Added styling to faces so rows can move be configurable in size, direction, speed, etc
- Added new watch faces using the new styling  
- Added German 24 hour clock face
- Fixed the error with one o' clock with the 12 hour German watch face


gfwilliams:
Great, thanks!

PR #2182, opened 2022-10-20
Allow setting a launcher

halemmerich:
Changes to make use of [#2274](https://github.com/espruino/Espruino/pull/2274)
Launcher is configurable like a clock.


gfwilliams:
Thanks!

PR #2183, opened 2022-10-20
Iconlaunch - Cache apps for faster startup

halemmerich:


gfwilliams:
Great - thanks! How much difference does that make?

halemmerich:
Scanning 40 apps has gone from 1.5 seconds to under 0.5 seconds when the cache is used.

gfwilliams:
Great! And I bet removing:

```
apps.forEach((app) => {
    if (app.icon) app.icon = s.read(app.icon);
  });
```

and instead loading the icons on demand (like in the launcher change) would probably knock a bit more off that too!

halemmerich:
Yes, did not try that yet. Although it could make the scrolling a bit more choppy on first load of each new row as it has to load three icon per row.

thyttan:
> Can it be done in two steps maybe? 

1. Load only the first nine icons
2. Display them
3. Then immediately when they are displayed load the rest of the apps as before

gfwilliams:
Yes, that'd work... You could even load them one by one with a setInterval

PR #2184, opened 2022-10-21
Store clock as setting if none is found on boot

halemmerich:


gfwilliams:
Great - thanks!

gfwilliams:
Actually just spotted a minor issue here:

```
clockApp = require("Storage").read(clockApp.src);
s.clock = clockApp.src;
```

But will push a fix in now

PR #2185, opened 2022-10-21
Iconlaunch - Cache icons on demand

halemmerich:
With ~40 apps this takes about 140ms of the start of iconlauncher.


gfwilliams:
Great - thanks!

PR #2186, opened 2022-10-21
new widget: widmsggrid: Messages Grid Widget

rigrig:
Displays multiple notification icons in a grid.

I only now noticed that `settings.showRead` ("Show widget even for read messages") isn't present in the default messages app. I commented it out in the README for now, and I'll make a PR to add it Soon. (Unless you already think it's a bad idea)


gfwilliams:
Looks great! So you have to disable the built-in messages widget? Or does this manage to do it for you?

rigrig:
> you have to disable the built-in messages widget?

Yes, I couldn't think of a good way to temporarily disable it, and didn't want 
[install `widmsggrid`]->[uninstall `widmsggrid`]->[*default widget is somehow missing*]

Maybe the Readme could be a bit clearer...

gfwilliams:
> Maybe the Readme could be a bit clearer...

Actually yes, it'd be great if it could mention in there how you disable the widget

PR #2187, opened 2022-10-21
[Icon Launcher] Swipe to exit enable via settings

thyttan:
@halemmerich 

Add functionality swipe to exit to clock and setting to enable it (disabled by default).

This conflicts with another open PR - I'll adjust this PR once the other one is merged.


gfwilliams:
Thanks! That PR should be merged now? The only conflict seems to be the ChangeLog? If that's the case I can tweak that here

thyttan:
@gfwilliams sure! If that's all that conflicts please do!

gfwilliams:
Great! Just merging - I'm keeping the version the same as it was only bumped about 30 minutes ago :)

PR #2190, opened 2022-10-23
[AI Clock] The probability of horizontal lines is correlated with the battery level

peerdavid:
The prob. of horizontal lines is now correlated with the battery level. So if the watch-face shows only a few lines its time to charge the bangle again ;)


gfwilliams:
:)

PR #2191, opened 2022-10-23
GPS Trekking - Improve UI and widget

halemmerich:
* Changes to custom UI to prevent menu opening method being called multiple times caused by drag handling of leftright-UI.
* Correctly add and remove listeners for sensors on starting and stopping background mode
* Fix listener for accel beeing active all the time

### DIFF
--- a/apps/gpstrek/app.js
+++ b/apps/gpstrek/app.js
@@ -331,7 +331,9 @@ function onSwipe(dir){
 function setButtons(){
   let options = {
     mode: "custom",
-    swipe: onSwipe
+    swipe: onSwipe,
+    btn: nextScreen,
+    touch: nextScreen
   };
   Bangle.setUI(options);
 }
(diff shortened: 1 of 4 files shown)


apps/gpstrek/app.js
```diff
@@ -318,16 +318,24 @@ function triangle (x, y, width, height){
...
-      nextScreen();
-    }
-  });
+  let options = {
+    mode: "custom",
+    swipe: onSwipe,
+    btn: nextScreen,
+    touch: nextScreen
+  };
+  Bangle.setUI(options);
```
gfwilliams, line 338: I just wanted to check, does this break Bangle.js 1? Or does it still work on both?

gfwilliams: Forget that - it never supported the 1 anyway ;)

gfwilliams:
Thanks! Just merged.

PR #2193, opened 2022-10-24
New medical info app

jt-nti:
Initial implementation to enable custom text in the medical alert widget

Possible future updates could include better editing UI instead of raw json, more information eg. medications, ability to set info from other apps, etc.

Signed-off-by: James Taylor <jt-git@nti.me.uk>


gfwilliams:
Thanks! Am I right in thinking that right now all this does on the watch is display `MEDICAL ALERT` though?

Shouldn't there at minimum be a `medicalinfo` app on the Bangle which shows the medical info that you've put in?

Doing that would mean you could just have  `medicalinfo` as an app (not app type `medicalinfo`) and so wouldn't have to add it as a new app type in the sanity checker.

Right now your readme says:

> You edit the contents using the Web IDE to load and then save the file from and to watch storage.

But I guess you don't need that now as you can edit it from the app loader?

It's not a big deal, but personally I'd avoid `exports.get*` - it's kind of a waste of space when you could just return the JSON object, and it means that any app that wants all the data is could to have to load the file at least 5 times or more.

jt-nti:
Thanks for the review!

> Am I right in thinking that right now all this does on the watch is display MEDICAL ALERT though?

At the moment, if you have the medical alert widget, the watch will display the contents of `medicalAlert`, or "MEDICAL ALERT" if that's not set. The plan is for the widget to display more details if you touch the alert text but I've not figured that bit out yet.

> Shouldn't there at minimum be a medicalinfo app on the Bangle which shows the medical info that you've put in?
>
> Doing that would mean you could just have medicalinfo as an app (not app type medicalinfo) and so wouldn't have to add it as a new app type in the sanity checker.

I was definitely thinking of another app to display the info but was kind of following the waypoint example of letting different apps display some or all of it in different ways, e.g. a blood group widget. Maybe that's a bit overly complicate though- happy to just have medicalinfo as an app if you think that would be better?

> I guess you don't need [edit the contents using the Web IDE in the readme] now as you can edit it from the app loader?

Ooops, yes, I just pinched that from the Waypoints readme- I'll update it.

> It's not a big deal, but personally I'd avoid exports.get* - it's kind of a waste of space when you could just return the JSON object, and it means that any app that wants all the data is could to have to load the file at least 5 times or more.

Good point, I'll change that.

gfwilliams:
Thanks!

> happy to just have medicalinfo as an app if you think that would be better?

I think for now that'd be best - we can always change things around later if we need to, but I imagine that for most people (and emergency responders trying to get info) if it all possible it'd be best to have just a single app.

>  The plan is for the widget to display more details if you touch the alert text but I've not figured that bit out yet.

I think you basically just want this at the bottom: https://github.com/espruino/BangleApps/blob/master/apps/widclose/widget.js#L12-L17

so along the lines of:

```
  }, touch: function(_, c) { // if touched
    const w = WIDGETS.widmedabl; // if in range, go to the app
    if (w && c.x>=w.x && c.x<w.x+w.width && c.y>=w.y && c.y<=w.y+24) load("medicalinfo.app.js");
  }
};
Bangle.on("touch", WIDGETS.widmedabl.touch);
```

... but if you need any help or advice just let me know!

jt-nti:
That makes sense. Thanks for the pointer for opening the app. I'll try and update the PR in the next week or so.

jt-nti:
@gfwilliams I think this is almost ready for another review, except I can't seem to get the medical info app to open when touching the top right widget! Let me know if you have any comments or can see what I'm doing wrong with the touch handler. Thanks.

jt-nti:
Actually, it does seem to launch after all... maybe it's my fat fingers, or perhaps the detection is slightly out?

gfwilliams:
This looks great - thanks for putting all those changes in!

Under Settings in new firmwares you can now calibrate the touchscreen, so that could be worth a try if you're finding the widgets hard to press? The coordinates of taps are clipped to the edge, so if you tap further towards the edge than you need to, you should be fine

PR #2194, opened 2022-10-25
Circlesclock v0.14

myxor:
Support to display weather in a big form next to the clock and date

image


gfwilliams:
Neat, thanks!

If you get a change to add a `remove` option to `Bangle.setUI` (see Anton clock) then you could also make the transition from clock->launcher a lot quicker :)

PR #2197, opened 2022-10-26
Circlesclock v0.15

myxor:
Use Bangle.setUI({remove:...}) to allow loading the launcher without a full reset on 2v16

### DIFF
--- a/apps/circlesclock/ChangeLog
+++ b/apps/circlesclock/ChangeLog
@@ -27,3 +27,4 @@
 0.13: Load step goal from Bangle health app as fallback
       Memory optimizations
 0.14: Support to show big weather info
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! Although actually both `let` and `const` are block-scoped, so there's no reason to change (it's just when there's `var` where it's needed).

The CI checks are complaining about mixed spaces/tabs (https://github.com/espruino/BangleApps/pull/2197/files) - please could you try and get rid of those?

myxor:
I replaced the tabs with spaces.

Should i revert the `const` to `let` changes?

gfwilliams:
> Should i revert the const to let changes?

If it's not hard, that would be preferable I think - as then lint (or espruino) can more easily tell if there are any bugs

myxor:
Done.

PR #2199, opened 2022-10-26
widmsggrid: improve README (installation: disable default widget)

rigrig:
No code change/version bump

Discussed in https://github.com/espruino/BangleApps/pull/2186


gfwilliams:
Thanks!

PR #2200, opened 2022-10-26
Grocery Augmented app

fapeg:
I finally got around to adding a text field to my app to be able to enter the API key on mobile 
Already introduced the app on the forums: https://forum.espruino.com/conversations/373509/


fapeg:
sorry my first time, I think I committed some wrong files

PR #2201, opened 2022-10-27
Grocery Augmented app with Handwriting OCR

fapeg:
I finally got around to adding a text field to my app to be able to enter the API key on mobile
Already introduced the app on the forums: https://forum.espruino.com/conversations/373509/


gfwilliams:
Nice! The gif is very cool too - although I think maybe the URL could be relative (not `https://fapeg.github.io/BangleApps/apps/groceryaug/groceryaug_preview.gif` - but I'll fix that after merging).

Would it have made sense to add this to the original Grocery app rather than having a whole new app, or were there big differences in the Bangle.js app itself as well?

fapeg:
Thanks! I wanted to keep it simple and I don't have quantities so I created a separate app. But for the future I'd like to have one "ultimate" app as well haha

fapeg:
@gfwilliams just curious, when will the app appear on the official Apploader?

gfwilliams:
I kick the update off manually so it depends if there's other stuff in there I think could be problematic and it pays to wait on. I might do it this afternoon, but then if something breaks it'll be all weekend before I fix it

Changes will still be in the development app loader now

PR #2202, opened 2022-10-27
CirclesClock v0.16

myxor:
- Fix const error (i was seeing lots of "Uncaught TypeError: Assignment to a constant" errors)
- Use widget_utils if available


gfwilliams:
Ok. Ideally `Assignment to a constant` errors are good though - you have `const` for everything you expect to be a constant, and then `let` otherwise. And if you try to assign to something that wasn't supposed to change then it errors.

Ideally we'd have just figured out *what* was being assigned, and if it was ok to be assigned you use `let`

PR #2203, opened 2022-10-27
Quicklaunch - Improve performance

halemmerich:
* Do not scan all apps on every boot
* Do not override setUI


gfwilliams:
Excellent - thanks!

PR #2205, opened 2022-10-28
messages: pass all messages through "message" event

rigrig:
Instead of directly calling from the lib into the app/widget, always use "message" events.      
Also adds some optimization so we only read messages from storage once per event.

This also gets rid of the `update` method, see https://forum.espruino.com/comments/16740860/

Marking this as draft because it's friday night, and I don't have time to properly test it, but other than that it is finished ;-)

### DIFF
--- a/apps/messages/widget.js
+++ b/apps/messages/widget.js
@@ -7,15 +7,15 @@ function filterMessages(msgs) {
     .filter((msg, i, arr) => arr.findIndex(nmsg => msg.src == nmsg.src) == i);
 }
 
-WIDGETS["messages"]={area:"tl", width:0, draw:function(recall) {
+WIDGETS["messages"]={area:"tl", width:0, msgs:[], draw:function(recall) {
   // If we had a setTimeout queued from the last time we were called, remove it
   if (WIDGETS["messages"].i) {
     clearTimeout(WIDGETS["messages"].i);
     delete WIDGETS["messages"].i;
   }
   Bangle.removeListener('touch', this.touch);
-  if (!this.width) return;
-    let settings = Object.assign({flash:true, maxMessages:3},require('Storage').readJSON("messages.settings.json", true) || {});
+  if (!this.width || this.checkInApp()) return;
+  let settings = Object.assign({flash:true, maxMessages:3},require('Storage').readJSON("messages.settings.json", true) || {});
   if (recall !== true || settings.flash) {
     var msgsShown = E.clip(this.msgs.length, 0, settings.maxMessages);
     g.reset().clearRect(this.x, this.y, this.x+this.width, this.y+23);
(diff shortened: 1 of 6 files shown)


apps/messages/widget.js
```diff
@@ -51,5 +53,6 @@ WIDGETS["messages"]={area:"tl", width:0, draw:function(recall) {
 /* We might have returned here if we were in the Messages app for a
 message but then the watch was never viewed. */
 if (global.MESSAGES===undefined)
-  WIDGETS["messages"].update(require("messages").getMessages());
+  WIDGETS["messages"].onMsg('',{});
+  Bangle.on("message", WIDGETS["messages"].onMsg);
```
gfwilliams, line 57: If MESSAGES is defined and we're in the app, don't we still want to register our `Bangle.on("message"` handler?

gfwilliams:
> Marking this as draft because it's friday night, and I don't have time to properly test it, but other than that it is finished ;-)

How are you feeling about this now? Did you have a chance to test it properly?

Apart from that event handler registering I just posted, this looks good to me

rigrig:
> Did you have a chance to test it properly?

Whoops, I totally forgot, sorry. I'll have a look at it tonight.

apps/messages/widget.js
rigrig: I think I'm not really sure what's going on here actually...
Do we even want to load the widget while in the messages app?

gfwilliams: That's a good question, maybe not... I just pushed a change to the messages lib that makes it 'fast load' the app if it can (which is really neat because it makes it much more responsive). But, that means that if the messages widget was already loaded, it'll stay loaded - so would have to cope with being around for the messages app anyway.

rigrig:
Ran into a bug where `msgs` wasn't set, no idea when I broke that, but fixed now.
Also rebased, and added a `__FILE__` check to hide the widget while the app is open

rigrig:
Converted back to draft: this morning I didn't get any music notifications, so I'll have to look into that first :-(

gfwilliams:
Thanks! This is looking good though

rigrig:
Fixed it: there was a race when two events arrived in a row, causing the first one to clean up `MESSAGES` while the second was still using it. Changed it so that a) that doesn't happen anymore, and b) it won't break even if it does happen.

rigrig:
One thing I just thought about: it now saves all messages before emitting the event.   
Should it maybe not do that? Something like if any listener sets `event.handled`, only store the message if it also sets `event.save`?   
That could be a bit tricky though, as other listeners (like the widget) will assume the message now exists...

rigrig:
> I thought the idea was that stuff like the music app could intercept the message before it got to be stored, so it didn't end up in the list of messages.

It was, but apparently I lost track of that somewhere.   
(And it's a bit tricky, because we do need to store the first one before launching the app)

> What about updating the message list, and (...) pass it as an argument to the event

I guess that makes sense, as we're loading `MESSAGES` anyway.    
But ideally we wouldn't need to load them all to handle incoming events: usually a listener is going to launch the app or ignore the event anyway. (Widgets could load them once at startup, and then keep track of event `src`s internally, instead of re-filtering them all every time.)


> make it a one-argument event that passes an object, so it's more readable/flexible?

I could do that if you want, although I'm not really a fan of the idea. I think it makes sense for arguments with optional fields, but if we're always passing `(type,event,all)` arguments, I don't see the benefit of wrapping those in an object?


How about for now I'll adjust this PR so 
  - it passes the full message list in events
  - if an event is `handled`, it won't be saved unless `store` is also set
  - `music` events are not stored by default, unless we are going to launch the app
 
And a follow-up PR to try and separate the "default" code from the library.

gfwilliams:
>  if we're always passing (type,event,all) arguments, I don't see the benefit of wrapping those in an object?

It's just for readability really - but yes, let's keep it with args for now.

> How about for now I'll adjust this PR so

That sounds great. but... 

> music events are not stored by default, unless we are going to launch the app

You mean the messages app?  Because at the moment the messages app needs to store them...

rigrig:
> at the moment the messages app needs to store them...

Urgh, juggling too many message apps in my head.   
I guess for now no special-casing for `music` events. (But aiming for the app to eventually mark them as `handled` and keep them only in memory until `kill`.)

rigrig:
Closing in favour of https://github.com/espruino/BangleApps/pull/2297    
(Seemed cleaner to start a new branch)

PR #2208, opened 2022-10-30
Boot time tweaks

halemmerich:
* bthrm: Move code into lib and require during boot only if enabled, saves about 130 ms during boot if disabled
* sensortools: As bthrm, about 250ms less during boot if disables
* owmweather: Updates 5 seconds after boot if needed, saves ~60ms during boot
* quicklaunch: Do not at all hash info files, just check if the app to be started exists and either start or remove from config, saves 180ms with ~40 apps installed


gfwilliams:
Looks great! Thanks for your work on this!

PR #2209, opened 2022-10-30
Iconlaunch - Back to slow loading apps

halemmerich:
Since there are problems with all full screen apps it's probably better to only directly load the clock. When fastloading full screen apps widgets will draw over those without a way of recovering without restarting the app.
Discussion on ongoing development for correctly fast loading apps can be found here: [https://forum.espruino.com/conversations/379724/](https://forum.espruino.com/conversations/379724/)


gfwilliams:
Thanks! Just wondering - do you think this is reliable enough now that we even need an option? I believe I enabled this by default for the default launcher

PR #2213, opened 2022-11-01
sleepphasealarm

nxdefiant:
Sleepphasealarm fix: Do not wake when falling asleep.

I do not think that the gpsautotime changes require a version increment.


gfwilliams:
Ok, thanks!

PR #2214, opened 2022-11-01
iconlaunch - Remove no longer needed fastload option

halemmerich:
Always try to fastload the clock as .bootcde now checks if it has widgets.


gfwilliams:
Great - thanks!

PR #2215, opened 2022-11-02
waypointer

nxdefiant:
Add settings file with the option to disable the slow direction updates

I'm not a fan of the 5 step update, so I made it configurable, leaving the original behavior as default. 

@hughbarney Hope this is ok with you


gfwilliams:
Thanks!

PR #2216, opened 2022-11-02
Slope Clock ++

myxor:
I really like the new Slope Clock but i was missing some data - so i cloned it.

So we can have a minimal (maybe soon default?) Slope Clock and one with some extended features.

I hope this is okay.


gfwilliams:
Thanks! This seems to be quite common now though - I'm wondering whether there isn't something we can do to reduce all the code duplication (like all the weather icons/etc and even duplication of clock code).

I've just started an issue at https://github.com/espruino/BangleApps/issues/2226 and I'd be really interested to hear your thoughts. `clock_info` is really cool and it'd be nice to make it super-easy for clocks to implement it

Issue #1757, opened 2022-04-27
Alarm does not buzz


nxdefiant:
With versions firmware=2.13, alarm=0.23, sched=0.07 a test alarm at 7.53 did not buzz at that time. Just nothing happened.

sched.json:
```
[{"t":25200000,"on":true,"rp":true,"as":false,"dow":62,"last":26,"vibrate":".."},{"t":28380000,"on":true,"rp":false,"as":false,"dow":127,"last":0,"vibrate":".."}]
```

Also, the debugging code comment in the bottom of https://github.com/espruino/BangleApps/blob/master/apps/sched/boot.js does not work because Bangle.SCHED is undefined.

nxdefiant:
If I execute the code in sched/boot.js manually, Bangle.SCHED is getting set?

gfwilliams:
Just checked and the alarm is for 7:53 not 7:23?

If I update your code to a time where I can try it here (8:20):

```
require("Storage").writeJSON("sched.json",[{"t":25200000,"on":true,"rp":true,"as":false,"dow":62,"last":26,"vibrate":".."},{"t":30000000,"on":true,"rp":false,"as":false,"dow":127,"last":0,"vibrate":".."}])
```

It all seems to work great for me - and Bangle.SCHED is set too.

When did you set the alarm up? Yesterday? And then you left the watch over night and the alarm just didn't fire this morning? It's possible that for some reason the 'recheck at midnight' didn't work...

But the odd thing is, if you look at https://github.com/espruino/BangleApps/blob/master/apps/sched/boot.js

There's really no way at all that Bangle.SCHED can end up as `undefined` unless it's because `sched.boot.js` is never run.

Please could you download `.boot0` and post it up here? It might shed some light on why it's not running

gfwilliams:
Ok, looks like boot0 got corrupted. That's your problem then :) If you go to `Settings->Utils->rewrite settings` it should fix it.
 
So what firmware are you on? I know this was an issue on earlier firmwares but I was pretty sure than in 2v12 and later this got fixed.

gfwilliams:
Also... did this not give you some kind of error in the Web IDE when swapping between apps? Usually it'd say something like `Got [ERASED] expected ...`

nxdefiant:
"rewrite settings" worked, thanks. I updated the firmware from 2.12 to 2.13 two days ago.

[...]

PR #2217, opened 2022-11-02
Added bgtimer, gallery, infoclk, pomoplus, random, rpnsci, stlap, stlapview apps

bruceblore:
Descriptions of the apps included with this pull request:

bgtimer: A timer that runs in the background and can be set with a keypad
gallery: An app to present a list of images in a menu and allow the user to pick from them
infoclk: A clock face that displays more information and some app shortcuts when unlocked
pomoplus: A configurable pomodoro timer that runs in the background
random: Random number generator
rpnsci: Scientific calculator using reverse polish notation
stlap: A stopwatch with lap timer that runs in the background
stlapview: An optional app that displays data collected by stlap

If this is too much for one pull request, I can split it into multiple pull requests. Additionally, not all of my apps are at version 0.01, because I've gone through multiple versions before deciding to submit them. If you would prefer to have them all be at version 0.01 when they get merged, I can make that change.


gfwilliams:
Thanks! These look awesome!

Sorry for the delay replying but I wanted to give these a proper look. It seems some of the pre-merge tests do throw up minor errors though: https://github.com/espruino/BangleApps/actions/runs/3380426169/jobs/5638069919

```
Warning: App bgtimer has no ChangeLog
Warning: App gallery has no ChangeLog
Warning: App infoclk has no ChangeLog
Error: JS icon infoclk.img does not match the pattern 'require("heatshrink").decompress(atob("..."))'
Warning: App pomoplus has no ChangeLog
Error: App pomoplus icon doesn't exist
Warning: App random has no ChangeLog
Warning: App rpnsci has no ChangeLog
Warning: App stlap has no ChangeLog
Warning: App stlapview has no ChangeLog
```

Please could you add ChangeLogs? Just make them say `0.02: Submitted to app loader` - but it means that at least if someone modifies them next time they will put a note in there.

Also it seems that  `infoclk.img` has a `//TODO replace app icon with my own` and pomoplus may not have an icon at all? 

One thing I would say is there's a scheduler app/library (https://github.com/espruino/BangleApps/blob/master/apps/sched/README.md) that handles background timers nicely - and the idea is you don't end up having to have boot code for each separate timer you have (which would slow down the boot). I guess at some point `bgtimer` might benefit from that, but it's very nicely written and lightweight as-is

bruceblore:
Yes, I can add changelogs.

Not sure why it complains about pomoplus not having an icon. I'll look into that. I forgot about this when I submitted, but I know that I stole pomoplus's icon from pomodo, and I vaguely remember stealing infoclk's icon from somewhere but I don't remember where. Stealing other app's icons does not seem like a very nice thing to do, so I'll give them their own icons.

And thanks for the suggestion to use sched for bgtimer. I think that can be done in a future version, as it would probably require pretty significant changes to the app, but it would be nice to have the timer integrate properly with everything else. I remember alarm/timer fragmentation being an issue complained about here. If I do change bgtimer to use sched, would it be acceptable to change the app's displayed name to something else that implies that the special thing about this app is the fact that it uses a keypad, not the fact that it runs in the background?

Also on the topic of using sched, some alarms set using sched, especially the first alarm of the day, haven't been going off. I think it might be related to #1757 because the discussion claims it was caused by a corrupted .boot0 and fixed by rewriting settings, I use qmsched to stop buzzing overnight, and I've noticed that settings get rewritten the first time I load an app after qmsched has made a change. I'll have to do some testing to see if that's the cause for sure.

gfwilliams:
Great, thanks! Even if you just add a `+` icon to an existing icon that'd be good as at least then users scrolling through the app loader can tell them apart at a glance.

> If I do change bgtimer to use sched, would it be acceptable to change the app's displayed name to something else that implies that the special thing about this app is the fact that it uses a keypad, not the fact that it runs in the background?

Yes, absolutely! Maybe before we merge this we should change the ID to `keytimer` anyway? If that's a pain for you to do I can do it here?

Interesting about the alarm - does it always happen? If you can get some steps to reproduce reliably please file an issue and I'll see if I can look into what's happening. It is possible that `qmsched` is causing issues I guess - is it just that alarms during quiet mode don't go off at all??

bruceblore:
> Great, thanks! Even if you just add a + icon to an existing icon that'd be good as at least then users scrolling through the app loader can tell them apart at a glance.

That's a good idea. I'll do that.

> Yes, absolutely! Maybe before we merge this we should change the ID to keytimer anyway? If that's a pain for you to do I can do it here?

That's a good idea. Far better to do this now before we potentially have other users. It should not be a pain for me to do.

> Interesting about the alarm - does it always happen? If you can get some steps to reproduce reliably please file an issue and I'll see if I can look into what's happening. It is possible that qmsched is causing issues I guess - is it just that alarms during quiet mode don't go off at all??

Right now, all I can say with *complete* certainty is that the alarm I set to wake me up never goes off. (Though luckily I didn't miss anything important because I have multiple redundant devices to wake me up.) Sometimes a second alarm in the same day goes off, and if it does I get both alarms at the same time. For now, the connection with qmsched is just a guess. Testing will probably be done later because it is the middle of the night for me.

gfwilliams:
> Sometimes a second alarm in the same day goes off, and if it does I get both alarms at the same time.

Ahh, well that in itself is interesting. Maybe `qmsched` does something that means that the 'boot code' for `sched` never gets run...

bruceblore:
Changelogs have been added, the app icon issues have been fixed, and bgtimer has been renamed to keytimer.

(The missing icon warning for pomoplus seems to have been because the icon's file name was not the same as the file name in the metadata, so I have no idea why exactly one of my two personal app loaders was able to load it correctly. Infoclk now has its own icon that isn't just stolen from Anton Clock, and I discovered that I stole the icons for stlap and stlapview too so that has been fixed.)

Haven't yet done serious testing for the alarm issue, but should that be discussed somewhere else? It seems a little bit off topic for here.

PR #2219, opened 2022-11-03
[Calculator] Exit with short press to BTN1 on Bangle.js 2

thyttan:
@fredericrous


fredericrous:
@thyttan I don't have BangleJs2 and cannot test.
This part of the code is not mine :/ I don't use double quotes in JS xD
@gfwilliams what do you think of this commit?

untbu (approved):
Looks great.
Also using the swipe event is more straightforward.
@fredericrous I've tested it.

gfwilliams:
Looks good to me - the new way is definitely the best way to handle swipes, and the code seems written so that Bangle.js 1 functionality is unaffected. Thanks!

PR #2221, opened 2022-11-03
Enton clock based on Anton clock

siatwe:
Hello there, 
I really like the Anton clock face!

But I changed the main font and added/removed some information for my need. I decided to publish it, maybe someone else is also interested :)


gfwilliams:
This looks great, thanks!

Not sure if you knew but you can also install widgets for the step count/etc if you need and they'll appear at the top of the screen on almost any clock. But obviously if you don't want them there then building it into the clock works well

gfwilliams:
Hi! Just to let you know, I've seen a few cases recently where a clock's been modified to show extra data.

I tweaked an existing library to make it super easy to display extra information on the clock face (including weather/timers/etc if they're installed): https://github.com/espruino/BangleApps/issues/2226

I'd be interested to see what you think about this - might be useful? Maybe you need it to be able to display more than one thing at a time?

Issue #2196, opened 2022-10-26
Convert apps to use `require("widget_utils").hide()/show()` rather than duplicating code


gfwilliams:
Currently every app that wants to hide widgets does so with its own code. I just added a library and it'd be nice to change apps to use this.

Apps that look like they need changes right now are:

- [x] apps/lcars/lcars.app.js
- [x] apps/advcasio/app.js
- [x] apps/neonx/neonx.app.js
- [x] apps/pebbled/pebbled.app.js
- [x] apps/notanalog/notanalog.app.js
- [x] apps/aiclock/aiclock.app.js
- [x] apps/macwatch2/app.js
- [x] apps/limelight/limelight.app.js
- [x] apps/cassioWatch/app.js
- [x] apps/daisy/app.js
- [x] apps/pebble/pebble.app.js
- [x] apps/thering/app.js
- [x] apps/rebble/rebble.app.js
- [x] apps/imageclock/app.js
- [x] apps/widviz/widget.js
- [x] apps/bwclk/app.js
- [x] apps/circlesclock/app.js (still includes fallback code)
- [x] apps/90sclk/app.js
- [x] apps/widviztime/widget.js ?

Darin755:
Did you mean convert?

myxor:
Circlesclock is done with https://github.com/espruino/BangleApps/pull/2202

gfwilliams:
Just to add - for anyone else looking at this, widget_utils now contains `slideOn` which allows the widgets to be slid onto the screen with a downwards swipe

thyttan:
#2342 for aiclock

Rarder44:
#2227 for rebble

thyttan:
Maybe this can be closed since #2492 was merged?

gfwilliams:
Yes,great - thanks!

PR #2227, opened 2022-11-04
[Rebble] implemented widget_utils

Rarder44:
https://github.com/espruino/BangleApps/issues/2196


Rarder44:
I had done this pr for rebble and then I did a commit for iconlauncher;
I wanted to do a separate pr but I find them all here 2. 
is that a problem?

gfwilliams:
That's fine - thanks! Usually two PRs would be preferable but when you make a PR in GitHub it's based on your current branch - so if you're using `master` and then you make more changes, those appear too

PR #2228, opened 2022-11-04
powermanager - New option to automatically calibrate the battery

halemmerich:
Sets the battery calibration value after 3 hours of charging without reloads.


gfwilliams:
Nice idea - thanks!

PR #2231, opened 2022-11-05
added PrimeTimeLato app

hughbarney:


gfwilliams:
Neat - thanks! Out of interest, how long does the prime calculation take each time?

PR #2233, opened 2022-11-05
[Torch] exit by pressing upper left corner

thyttan:
If the screen is unlocked pressing upper left corner exits to watch face.


gfwilliams:
Thanks!

PR #2234, opened 2022-11-06
new app to follow gpx traces : gipy

wagnerf42:
hi,

i think my app is now good enough for a public release.
it has been tested over a few hundreds of kilometers now and is pretty usable.
you can see it in action here : https://wagnerf42.github.io/BangleApps/?q=gipy

you should note the following things:

- I am a complete javascript noob : it is my first javascript program. I don't know javascript and since it is a subset of it on the watch I must surely use it in a pretty awful way.
- I have some binary code in the repo because the file converter used in gipy_uploader is written in rust and compiled to javascript in wasm. this code is only for the custom webpage and does not run on the watch. the original source is here : https://github.com/wagnerf42/gpconv
- I think these wasm files break the current tests. could you give me a hand here ?

it's been a fun project and i still have lots of ideas for it. since it's the most useful app for me on my watch as of now I wanted to share.

feel free to ask me for modifications.


nxdefiant:
Cool, I need to test this. 

> (it uses your movement to figure out your orientation
and walking is too slow).

How about dropping to compass for orientation when speed drops below e.g. 5km/h?

wagnerf42:
yes, that's on my todo list

gfwilliams:
Thanks! This looks great! I guess personally I'd have used JS to convert GPX to your GPX format, but if this works ok then great - and I really like the idea of the binary format to cram as much data in as possible.

Just wondering, is there a reason that the `gipy_uploader` isn't just part of the `gipy` app? You can make the `custom.html` file an `interface.html` file so then you just install `gipy` and you click the icon next to it to upload/change its data?

wagnerf42:
i've been thinking we can maybe remove the blob:
 - have a git submodule
 - add rust + wasmpack to the docker
 - compile it at the same time as the web pages

this would be cleaner but i'm not sure how easy it is to do.

for the interface.html that's interesting. i would have preferred to have it as a single app but i found no app like that so didn't know how to do it.
i'll give it a try.

gfwilliams:
I'd actually prefer the blob (with a link to the repo where you built it) than to add extra stuff to the GitHub actions here. It's quite important that I can quickly verify PRs don't break stuff, and if I've got to wait for rust+wasmpack to install and run every time I want to check something it'll be annoying.

> i would have preferred to have it as a single app but i found no app like that

waypoints could be a good example? https://github.com/espruino/BangleApps/tree/master/apps/waypoints

The only reason it's not a 'full' app in its own right is so the list of waypoints could be shared between apps.

... Or https://github.com/espruino/BangleApps/tree/master/apps/openstmap does it too - but that uses 'custom' like you're doing - so you have to choose what you want when you upload the app. At some point that'll change though as for openstmap I'd like to allow multiple areas to be included.

wagnerf42:
could you give me a hand for the file upload ? i have some troubles i guess because i want to upload some binary.
i guess i need to escape more things. i tried to escape the simple quotes.

nxdefiant:
Maybe look at the restore-code of the app loader.

also do you know, that you can run the apploader locally with npm? No need to test each change on github.

wagnerf42:
oh thank god. no i don't know. how do I proceed ?

nxdefiant:
`npm ci` and `npm start` should do it after you checked out the submodules.

wagnerf42:
hey, thanks for the tip. i had to generate a certificate and key but it worked. it is so much better than uploading to github.

also,  if you don't mind i'd like to polish the interface a bit more. I think i should be able to get openstreetmap data even from there.

wagnerf42:
so, good news, openstreetmap data is now fetched from the interface.html.
it's much nicer like that, thanks for your help.

i'm going to tune a little bit this page to give more choices to the users as to what they want on their map.

gfwilliams:
That's great, thanks! Just let us know when you think you're ready to merge

wagnerf42:
yes i will. note that the upload seems actually broken. i had a quick test this morning and only the start of the files is actually uploaded.

i wonder, do i need to use chunks ?
like in core/lib/espruinotools.js ?

gfwilliams:
> i had a quick test this morning and only the start of the files is actually uploaded.

Argh, that's a pain. How big are the files?

I see you're using `Util.writeStorage`, but actually it looks like the implementation of that is pretty stupid (I guess it was previously just used for writing small files) and as you say it's not chunking the data (and it should be).

Do you think you'd be able to breakpoint that bit of your code in devtools (or add a `console.log`) and paste a copy of the exact data you're calling `Util.writeStorage` with in here? Then I can fix the function here and make 100% sure it'll write all your data correctly.

wagnerf42:
hi, so, here is the array i receive from the converter. this is an array of u8

[log.txt](https://github.com/espruino/BangleApps/files/9974676/log.txt)

it is then turned into a string :

179     let gpc_string = String.fromCharCode.apply(String, gpc_content);

i think outside of that the interface.html seems good. i'll test it a bit more tomorrow though.

gfwilliams:
Hi! I just committed some code which should fix `writeStorage`.  Please can you give it a try? If you're happy it works I guess we could merge now?

wagnerf42:
hi, it works great. thanks !

i tested on a very large file and while it took a minute to convert it worked nicely.
i think it is ok for release.

thank you both very much again for your help.

gfwilliams:
Excellent - thanks!

PR #2235, opened 2022-11-06
[Desktop Launcher] Fast load to clockface on B2

thyttan:
Fast loading requested [here](https://forum.espruino.com/conversations/381203/#comment16751193).

I changed most instances of var to let.

There is some small ram leak going back and forth between this dtlaunch and the slope clock, but pretty minor by my judgement. Edit: This has been solved with @halemmerich's help.

Edit: Added in the timeout to clock functionality from Icon Launcher.

### DIFF
--- a/apps/dtlaunch/app-b2.js
+++ b/apps/dtlaunch/app-b2.js
@@ -13,7 +13,7 @@ let settings = Object.assign({
 }, require('Storage').readJSON("dtlaunch.json", true) || {});
 
 if (settings.oneClickExit) {
-  let buttonWatch = setWatch(_=> returnToClock, BTN1);
+  var buttonWatch = setWatch(_=> returnToClock, BTN1);
 } 
 
 let s = require("Storage");
--- a/apps/dtlaunch/app-b2.js
+++ b/apps/dtlaunch/app-b2.js
@@ -1,189 +1,189 @@
 { // must be inside our own scope here so that when we are unloaded everything disappears
+
+/* Desktop launcher
+*
+*/
+
+let settings = Object.assign({
+  showClocks: true,
+  showLaunchers: true,
+  direct: false,
+  oneClickExit: false,
+  swipeExit: false,
+  timeOut: "Off"
+}, require('Storage').readJSON("dtlaunch.json", true) || {});
+
+if (settings.oneClickExit) {
+  var buttonWatch = setWatch(_=> returnToClock, BTN1);
+} 
+
+// taken from Icon Launcher with minor alterations
+if (settings.timeOut!="Off"){
+  let time=parseInt(settings.timeOut);  //the "s" will be trimmed by the parseInt
+  var timeoutToClock = setTimeout(returnToClock,time*1000);  
+}
+let s = require("Storage");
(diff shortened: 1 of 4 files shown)


apps/dtlaunch/app-b2.js
```diff
@@ -67,54 +68,60 @@ function draw_icon(p,n,selected) {
       }
     }
     g.drawString(line.trim(),x+36,y+54+lineY*8);
-}
+};
 
-function drawPage(p){
+let drawPage = function(p){
```
thyttan, line 73: I tried this in order to have something to delete when fastloading, but there's still ram leaking so don't know if it helped any.

Was inspired by this: https://forum.espruino.com/conversations/379724/?offset=25#comment16739560

apps/dtlaunch/app-b2.js
```diff
@@ -1,28 +1,30 @@
...
   showLaunchers: true,
   direct: false,
   oneClickExit:false,
-  swipeExit: false
+  swipeExit: false,
 }, require('Storage').readJSON("dtlaunch.json", true) || {});
 
 if( settings.oneClickExit)
-  setWatch(_=> load(), BTN1);
+  setWatch(_=> returnToClock, BTN1);
```
halemmerich, line 16: You need to store this in a variable and call `clearWatch` with it in `returnToClock` to remove it.

apps/dtlaunch/app-b2.js
```diff
@@ -125,12 +132,47 @@ Bangle.on("touch",(_,p)=>{
...
+  delete lineY;
+  delete line;
+  delete c;
+  delete O;
+  delete x1;
+  delete x2;
+  delete i;
+  delete drawIcon;
+  delete drawPage;
+  delete isTouched;
```
halemmerich, line 166: Most of these are not needed, everything defined in the block scope with let/var can be garbage collected automatically. It can be useful to delete some big vars beforehand to have more memory for the app that should be loaded.

thyttan: And the garbage collector should be as fast having delete statements like this, or probably faster?

apps/dtlaunch/app-b2.js
thyttan: Thanks! With that it seems there's no ram leak anymore :)

apps/dtlaunch/app-b2.js
thyttan: Removed a bunch of delete statements. Unsure if those left should be removed as well.

gfwilliams: Yes, the GC will be faster - when the block goes out of scope, everything will be freed in one go

gfwilliams:
Excellent - thanks! You could also look at adding the 'app cache' we've used in the other launchers too - that can really speed up the loading if you have a bunch of apps

thyttan:
> Excellent - thanks! You could also look at adding the 'app cache' we've used in the other launchers too - that can really speed up the loading if you have a bunch of apps

I might look at it! If anyone beats me to it, that will just be great ;)

PR #2236, opened 2022-11-06
Iconlaunch - Better timeout handling

halemmerich:
Resets the timeout while actively using the launcher.


gfwilliams:
I think there are some issues here - you have two `.orig` files added in GitHub?

Rarder44:
@halemmerich  you are right! I missed a few things!

gfwilliams:
Thanks!

PR #2237, opened 2022-11-06
Bthrm - Stability and power use improvements

halemmerich:
* Prevent mixing of internal and external HRM events
* Do not scan every 100ms if no device is found
* Run some steps of the connection process in timeouts


gfwilliams:
Great - thanks!

PR #2238, opened 2022-11-06
Imageclock

halemmerich:
* Use widget_utils module
* Some caching for performance
* Add UI when drawing is finished to prevent drawing over the launcher
* Fix custom color setting inside generated code


gfwilliams:
Great - thanks!

PR #2240, opened 2022-11-06
Widedit

halemmerich:
* Wrap Bangle.loadWidgets instead of overriding it
* Allow all areas to be set (bottom and top widget bar)


gfwilliams:
Looks great - thanks!

PR #2243, opened 2022-11-07
[Torch] Move interactions inside setUI

thyttan:


gfwilliams:
Thanks! On B2 the torch doesn't seem to load widgets - so I guess this won't *show* the back button. I guess the physical press (and tap in top-left?) will work?

gfwilliams:
> So they are loaded for me when clicking upper left corner, just not visible.

Ahh. Are you using @halemmerich's 'fast load' or something like that? If so I'd be a bit wary that maybe your setup doesn't always  match what other users have, and you might end up making changes that just don't work for others.

thyttan:
Yes, I'll do what I did last time and test it out on a factory reset watch. Just gonna make a backup first.

thyttan:
Yes, just tested it on factory reset and then the widget field gets drawn with a back button. So there'll be a black bar on top if the user uses dark theme.

gfwilliams:
Ahh, interesting, thanks! That's actually kind of an issue in Bangle.js I guess (if setUI is called, apps that weren't expecting widgets to be loaded may still have `WIDGETS` set).

But it's great that the back icon gets drawn anyway. We could actually just add `g.setTheme({bg:"#fff",fg:"#000"})` to the torch app and it'd fix the black widget bar issue.

PR #2245, opened 2022-11-07
[Desktop Launcher] move interactive code into setUI...

thyttan:
replace 'click to exit' option with back function in setUI (always on).


gfwilliams:
That's great, thanks! That really tidies things up!

PR #2248, opened 2022-11-08
[Dragboard] Performance and stability

thyttan:


apps/dragboard/lib.js
```diff
@@ -1,12 +1,10 @@
-//Keep banglejs screen on for 100 sec at 0.55 power level for development purposes
-//Bangle.setLCDTimeout(30);
-//Bangle.setLCDPower(1);
-
 exports.input = function(options) {
+  "ram";
```
gfwilliams, line 2: Please can you take this one out? Adding `"ram"` to everything doesn't always help - this function is only called *once* when the keyboard is first to be displayed, so really it's just using a whole bunch of RAM (while keyboard is displayed *and* afterwards until a new app is loaded) for no reason

The best thing is to do is figure out which functions get called a lot or have to be faster and tag those with `"ram"` - I can see that had been done a bit already but those mentions have been removed now?

I can imagine that `drag: function(event) {` could really benefit with being tagged with `"ram"` - that might provide you with the performance you're after?

thyttan: Thank you, will look into that.

thyttan: I removed that one and added "ram" to three sub-functions instead.

gfwilliams:
Great, thanks!

PR #2249, opened 2022-11-09
[kbswipe] Keep in ram for higher performance

thyttan:
This is just a suggestion.

I think it makes sense to have keyboards where draging is used for input to be all in ram to maximize performance. It helps a lot with being able to more reliably enter characters.

Here's without "ram" keyword:
```
>require("textinput").input().then(result => {}).then(load)
=Promise: {  }
>process.memory().usage
=1326
```

Here's with "ram" keyword:
```
>require("textinput").input().then(result => {}).then(load)
=Promise: {  }
>process.memory().usage
=1666
```

So there's not a very big difference in ram usage.

And by the way, it feels like the speedups in the newer cutting edge firmwares really do wonders for kbswipe!


gfwilliams:
Are you *sure* this makes a difference? Because as far as I can tell, `getStrokes` is a function that only gets called when the keyboard is first shown and when the 'hint' page is shown. Once it's called, the stroke data it generates is in RAM anyway

I'm all for loading stuff in RAM that is called often and needs to be fast, but I'm not sure that in this case it will actually make any difference.

thyttan:
Ok, I'll investigate some more before coming back. It's possible it's a placebo effect for me.

thyttan:
I removed the "ram" on top and added it to the drag function instead, could that make sense? I have not tested that any yet.

gfwilliams:
Yes, absolutely - that'll definitely help!

PR #2250, opened 2022-11-09
[Torch] Widget field background follows torch colour

thyttan:
Inspiration: https://github.com/espruino/BangleApps/pull/2243#issuecomment-1308513138


gfwilliams:
Thanks!

Issue #2226, opened 2022-11-04
Clock Info -> extra data on the clock face


gfwilliams:
I'm seeing quite a few clocks now that provide extra data on the watch face - like steps or weather. Often there may even be an existing clock that is then cloned to make it show that extra data - and then the clock's code is duplicated, as well as the code to show the data.

It'd be nice to:

* make the process of adding extra data to clock faces easier and more flexible
* reduce the code duplication

The idea of widgets was to display this kind of data, but right now they're relegated to a small 24px strip at the top and bottom of the screen.

**We also have clock_info now** which is a really neat system that allows a clock to display information that can be scrolled through (and added to with plugins): https://github.com/espruino/BangleApps/blob/master/modules/clock_info.js

I'm open to ideas here - just thinking aloud really, but:

* We have `tl,tr,bl,br` widget areas - we could have a new widget area `clk` or something, and then a library that allowed those widgets to be displayed wherever made sense on the clock face. Widgets would then need to look at `this.w` and `this.h` to figure out how much space they had available.
  * Downside is clock widgets would get loaded (and then ignored) when there wasn't a clock showing
* Maybe we have another library like `clock_info` that handles non-interactive clock widgets?
* **Or maybe these clocks should just use clock_info?** It seems in many cases being able to swipe through different info would actually be a big improvement 
  * but then what if >1 type of info was to be displayed? Maybe we could allow a tap to select which one swipes on the screen would affect
  * currently clock_info leaves user-interaction up to the app, but this could be part of the library too

gfwilliams:
... also if using clock_info it should probably remember what setting it was scrolled to last time the clock was shown

gfwilliams:
Ok, just added something... So as an example:

```
var clockInfoMenu = require("clock_info").addInteractive(require("clock_info").load(), (itm, info) => {
  var x = 20, y = 20, w=80, h=48;
  g.reset().clearRect(x,y,x+w-1,y+h-1);
  g.drawRect(x,y,x+w-1,y+h-1); // debug - just to show where we are
  g.drawImage(info.img, x+w/2-12,y+4);
  g.setFont("6x8:2").setFontAlign(0,0).drawString(info.text, x+w/2,y+36);
});
// then when clock 'unloads':
clockInfoMenu.remove();
delete clockInfoMenu;
```

So this should be really easy to add to clock faces now...

gfwilliams:
[...]

PR #2251, opened 2022-11-09
[WIP] circlesclock: load circles from clkinfo

glemco:
Simple view of the closest running timer in the clock.
I cannot figure out how to properly generate the icon though, I created a 16x16 image and converted in b/w but it only works if I don't set a colour, otherwise the picture is weirdly mirrored (I bet it has something to do with the size, any hint is appreciated).
TODO:
* differentiate among alarms and timers
* _something else_?

### DIFF
--- a/apps/circlesclock/ChangeLog
+++ b/apps/circlesclock/ChangeLog
@@ -30,3 +30,4 @@
 0.15: Use Bangle.setUI({remove:...}) to allow loading the launcher without a full reset on 2v16
 0.16: Fix const error
       Use widget_utils if available
+0.17: Load circles from clkinfo
--- a/apps/circlesclock/app.js
+++ b/apps/circlesclock/app.js
@@ -1,3 +1,4 @@
+let clock_info = require("clock_info");
 let locale = require("locale");
 let storage = require("Storage");
 Graphics.prototype.setFontRobotoRegular50NumericOnly = function(scale) {
@@ -18,6 +19,7 @@ let settings = Object.assign(
   storage.readJSON(SETTINGS_FILE, true) || {}
 );
 
+ //TODO deprecate this (and perhaps use in the clkinfo module)
 // Load step goal from health app and pedometer widget as fallback
 if (settings.stepGoal == undefined) {
   let d = storage.readJSON("health.json", true) || {};
@@ -29,7 +31,7 @@ if (settings.stepGoal == undefined) {
   }
 }
 
-let timerHrm;
+let timerHrm; //TODO deprecate this
 let drawTimeout;
 
 /*
@@ -44,9 +46,8 @@ let showWidgets = settings.showWidgets || false;
 let circleCount = settings.circleCount || 3;
 let showBigWeather = settings.showBigWeather || false;
 
-let hrtValue;
+let hrtValue; //TODO deprecate this
 let now = Math.round(new Date().getTime() / 1000);
-
 
 // layout values:
 let colorFg = g.theme.dark ? '#fff' : '#000';
@@ -91,8 +92,20 @@ let circleFontSmall = circleCount == 3 ? "Vector:14" : "Vector:10";
 let circleFont = circleCount == 3 ? "Vector:15" : "Vector:11";
 let circleFontBig = circleCount == 3 ? "Vector:16" : "Vector:12";
 let iconOffset = circleCount == 3 ? 6 : 8;
-let defaultCircleTypes = ["steps", "hr", "battery", "weather"];
+let defaultCircleTypes = ["Bangle/Steps", "Bangle/HRM", "Bangle/Battery", "weather"];
 
+let circleInfoNum = [
+  0, // circle1
+  0, // circle2
+  0, // circle3
+  0, // circle4
+];
+let circleItemNum = [
+  0, // circle1
+  1, // circle2
+  2, // circle3
+  3, // circle4
+];
 
 function hideWidgets() {
   /*
@@ -111,7 +124,7 @@ function hideWidgets() {
 function draw() {
   g.clear(true);
   let widgetUtils;
-  
+
   try {
     widgetUtils = require("widget_utils");
   } catch (e) {
@@ -177,6 +190,15 @@ function drawCircle(index) {
   let w = getCircleXPosition(type);
 
   switch (type) {
+    case "weather":
+      drawWeather(w);
+      break;
+    case "sunprogress":
+    case "sunProgress":
+      drawSunProgress(w);
+      break;
+    //TODO those are going to be deprecated, keep for backwards compatibility for now
+    //ideally all data should come from some clkinfo
     case "steps":
       drawSteps(w);
       break;
@@ -189,13 +211,6 @@ function drawCircle(index) {
     case "battery":
       drawBattery(w);
       break;
-    case "weather":
-      drawWeather(w);
-      break;
-    case "sunprogress":
-    case "sunProgress":
-      drawSunProgress(w);
-      break;
     case "temperature":
       drawTemperature(w);
       break;
@@ -205,9 +220,12 @@ function drawCircle(index) {
     case "altitude":
       drawAltitude(w);
       break;
+    //end deprecated
     case "empty":
       // we draw nothing here
       return;
+    default:
+      drawClkInfo(index, w);
   }
 }
 
@@ -303,6 +321,102 @@ function getImage(graphic, color) {
     };
   }
 }
+
+function drawWeather(w) {
+  if (!w) w = getCircleXPosition("weather");
+  let weather = getWeather();
+  let tempString = weather ? locale.temp(weather.temp - 273.15) : undefined;
+  let code = weather ? weather.code : -1;
+
+  drawCircleBackground(w);
+
+  let color = getCircleColor("weather");
+  let percent;
+  let data = settings.weatherCircleData;
+  switch (data) {
+    case "humidity":
+      let humidity = weather ? weather.hum : undefined;
+      if (humidity >= 0) {
+        percent = humidity / 100;
+        drawGauge(w, h3, percent, color);
+      }
+      break;
+    case "wind":
+      if (weather) {
+        let wind = locale.speed(weather.wind).match(/^(\D*\d*)(.*)$/);
+        if (wind[1] >= 0) {
+          if (wind[2] == "kmh") {
+            wind[1] = windAsBeaufort(wind[1]);
+          }
+          // wind goes from 0 to 12 (see https://en.wikipedia.org/wiki/Beaufort_scale)
+          percent = wind[1] / 12;
+          drawGauge(w, h3, percent, color);
+        }
+      }
+      break;
+    case "empty":
+      break;
+  }
+
+  drawInnerCircleAndTriangle(w);
+
+  writeCircleText(w, tempString ? tempString : "?");
+
+  if (code > 0) {
+    let icon = getWeatherIconByCode(code);
+    if (icon) g.drawImage(getImage(icon, getCircleIconColor("weather", color, percent)), w - iconOffset, h3 + radiusOuter - iconOffset);
+  } else {
+    g.drawString("?", w, h3 + radiusOuter);
+  }
+}
+
+function drawSunProgress(w) {
+  if (!w) w = getCircleXPosition("sunprogress");
+  let percent = getSunProgress();
+
+  // sunset icons:
+  let sunSetDown = atob("EBCBAAAAAAABgAAAAAATyAZoBCB//gAAAAAGYAPAAYAAAAAA");
+  let sunSetUp = atob("EBCBAAAAAAABgAAAAAATyAZoBCB//gAAAAABgAPABmAAAAAA");
+
+  drawCircleBackground(w);
+
+  let color = getCircleColor("sunprogress");
+
+  drawGauge(w, h3, percent, color);
+
+  drawInnerCircleAndTriangle(w);
+
+  let icon = sunSetDown;
+  let text = "?";
+  let times = getSunData();
+  if (times != undefined) {
+    let sunRise = Math.round(times.sunrise.getTime() / 1000);
+    let sunSet = Math.round(times.sunset.getTime() / 1000);
+    if (!isDay()) {
+      // night
+      if (now > sunRise) {
+        // after sunRise
+        let upcomingSunRise = sunRise + 60 * 60 * 24;
+        text = formatSeconds(upcomingSunRise - now);
+      } else {
+        text = formatSeconds(sunRise - now);
+      }
+      icon = sunSetUp;
+    } else {
+      // day, approx sunrise tomorrow:
+      text = formatSeconds(sunSet - now);
+      icon = sunSetDown;
+    }
+  }
+
+  writeCircleText(w, text);
+
+  g.drawImage(getImage(icon, getCircleIconColor("sunprogress", color, percent)), w - iconOffset, h3 + radiusOuter - iconOffset);
+}
+
+/*
+ * Deprecated but nice as references for clkinfo
+ */
 
 function drawSteps(w) {
   if (!w) w = getCircleXPosition("steps");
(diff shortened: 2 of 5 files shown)


thyttan:
Just thinking out loud: Could [clock_info](https://github.com/espruino/BangleApps/blob/master/modules/clock_info.js) be good for this? 

Also there's this discussion #2226.

Sorry, don't have a solution to your problem with the icon.

glemco:
Well, maybe. The thing is that I wrote this piece of code a while ago when clkinfo wasn't there yet. 
The library looks promising but I am not familiar with it.

thyttan:
Yeah, it's the same for me as well ;) 

Good idea about adding timer! :)

glemco:
Anyway, I am still struggling with the icon: here's an example (you can try in the IDE) with the timer image I created (16x16, 1bbp, image string, inverted) and with another icon from circlesclock. Without colour they both look fine, but if I change colour the timer looks bad and doesn't even take the right colour (the code reproduces how images are rendered in the clock).
```javascript
const timerIcon = atob("EBAB/////+AH8A/zz/GP+B/8P/5//D/4H/AP8A/gB/////8=");
const pressureIcon = atob("EBCBAAAAAYADwAJAAkADwAPAA8ADwAfgB+AH4AfgA8ABgAAA");
function drawPics(icon, x) {
  g.drawImage(icon, 0+x,0);
  const colorTimerIcon = {
        width: 16,
        height: 16,
        bpp: 1,
        transparent: 0,
        buffer: E.toArrayBuffer(icon),
        palette: new Uint16Array(['#fff', g.toColor('#f0f')])
      };
  g.drawImage(colorTimerIcon, 16+x,0);
}
drawPics(timerIcon, 0);
drawPics(pressureIcon, 32);
```

gfwilliams:
Hi - I think there are two issues here...

* The image shouldn't be inverted when you convert it - in the example you have, if your theme background is white, you want the image itself to be black (right now the image background is black and the image foreground is transparent)
* To change the color of the 1bpp image you don't need to redefine the image (that's doesn't work for image strings - the string you need has to be created using the 'image object' export mode). All you need to do is use setColor:

```
const timerIcon = atob("EBAB/////+AH8A/zz/GP+B/8P/5//D/4H/AP8A/gB/////8=");
const pressureIcon = atob("EBCBAAAAAYADwAJAAkADwAPAA8ADwAfgB+AH4AfgA8ABgAAA");
function drawPics(icon, x) {
  g.reset().drawImage(icon, 0+x,0);
  g.setColor('#f0f').drawImage(colorTimerIcon, 16+x,0);
}
drawPics(timerIcon, 0);
drawPics(pressureIcon, 32);
```

So this would work great if the images were converted 'uninverted'. But if you post up the source images I could do it for you.

## BUT

I'm really not a fan of shoving loads of code into clock faces for extras. That's why we created `clock_info` and have that issue (referenced above). It makes the clocks unmaintainable.

For example, you should probably use the `sched` library to access `sched.json` - that's what it's for. Right now you're checking the `t` field, but not anything else (whether the alarm has already fired, what date it's set for, whether it's enabled, etc) so in all but the most minimal cases this code won't display the correct value.

... and then someone else will copy it into a new clock face and so on, and before we know it we have this not-quite-working code all over the place and it's a nightmare to fix.

Also, as far as I can tell if you install https://banglejs.com/apps/?id=smpltmr then this functionality is already part of clock_info (and it deals correctly with the different states of alarms)... If you're dead set on not using clock_info, you should probably still copy the https://github.com/espruino/BangleApps/blob/master/apps/smpltmr/clkinfo.js code to get active timers

So I guess is there a way we can modify clock_info to make it do what you need? For example as you're drawing these circles maybe we need to return the value as a number, and min/max values for each circle?

glemco:
Well, as I'm not the developer of circlesclock, I tried to keep things as close how they are implemented already. I'd rather not change all the icons as I don't have the original pictures. `pressureIcon` is already there and works _somehow_ fine both with and without colour, as I'm adding `timerIcon` I'd like it to behave just like the others (unless the solution is so bad and there's a valid reason to change them all).
Here's the timer icon I'm using, in case you want to try something with it.
small_timer

Regarding the data source, at the time I wrote that code (months ago), I thought it could be great to get all the circles in this clock in a standardised way (somehow like complications work in WearOS), I couldn't find anything implemented and had to do it manually.
I'm supposing `clkinfo` works like that so I'd definitely be up to use it, just I'm wondering why cannot the `clkinfo` for timers be embedded in the default sched library (or in the `Alarms & Timers` app) rather than in yet another app to install.

The idea I had (based on how this watchface works), is that the common interface should return a number (for the gauge) and a short text (to display inside). In this way any kind of progress-bar-like complication can be implemented easily on any watchface, I'm not sure `clkinfo` can do that already (is there a proper documentation for it?).

gfwilliams:
This works how I'd expect now?

```
const timerIcon = atob("EBABAAAAAB/4D/AMMA5wB+ADwAGAA8AH4A/wD/Af+AAAAAA=");
const pressureIcon = atob("EBCBAAAAAYADwAJAAkADwAPAA8ADwAfgB+AH4AfgA8ABgAAA");
function drawPics(icon, x) {
  g.reset().drawImage(icon, 0+x,0);
  g.setColor('#f0f').drawImage(icon, 16+x,0);
}
drawPics(timerIcon, 0);
drawPics(pressureIcon, 32);
```

image

>  I'm wondering why cannot the clkinfo for timers be embedded in the default sched library (or in the Alarms & Timers app) 

Yes, I think it probably *should* be. I can't remember 100% but I think it's probably because it's @peerdavid who really started using clock-info, and since the simpletimer app is his I think he just added the code in there to avoid messing with a 'default' app - but maybe we could look at moving it...

> the common interface should return a number (for the gauge) and a short text (to display inside). In this way any kind of progress-bar-like complication can be implemented easily on any watchface

The existing docs are at the top of https://github.com/espruino/BangleApps/blob/master/modules/clock_info.js

Right now we return an icon and text, but we could easily extend it without breaking anything to include a min/max and numerical value too... In fact I just pushed a change to add this.

peerdavid:
Hi, 
 >  can't remember 100% but I think it's probably because it's @peerdavid who really started using clock-info, and since the simpletimer app is his I think he just added the code in there 

exactly I added this to the simpletimer app as I was not sure if we want this in the default sched library. But I totally agree with you that it would fit better in the sched library in case we want this :)

glemco:
Well, I can have a check then, at least to implement it for the timer in this watchface. Thanks for fixing the icon, however if the clkinfo returns an icon too, it could be even easier (although this clock requires a 16x16, the 24x24 may be just too big).
Another nice thing of circlesclock is that some icons change based on the values (e.g. the weather condition), if that can be done too, then this watch can be redesigned to rely just on clkinfo for all data (not sure what the original developer @myxor thinks about it, I may give it a shot if I find some time).
@peerdavid if it's just a matter of moving the code from one app to the other and nothing breaks, then I guess it can help!

peerdavid:
Yea I think moving the file from one to another app is enough :)

glemco:
> Yea I think moving the file from simpletimer to sched is enough :)

It's a little trickier as the simple timer has just one single timer (and gets it through its id), I'm playing around to see what can I do with it.
Ideally I guess it should show all active timers and alarms, but in cases like the circlesclock, we are just interested in the closest timer/alarm. The simplest trick would be to set this new field `hasRange` (or any other flag) only in the one we are interested in from the `clkinfo` itself.

gfwilliams:
clock_info can return more than one item, so I guess it could sort them by the time until they'll fire, and then list them all, nearest first. So then you get one shown, but if you swipe you can cycle through others

myxor:
@glemco feel free to make your changes to circlesclock if you want to. So far your changes look fine for me.

gfwilliams:
How's this looking now? Is it still a work in progress?

glemco:
Thanks @myxor , hope not to screw it up 😅
@gfwilliams I have a better version on my master branch (don't want to mess this up with tons of forced pushes), it kind of works but still has some bugs, as soon as it looks fine I'll let you know.

glemco:
I am still struggling with the icons, right now I'm getting them from the clkinfo and it works fine for the bangle stuff, however the ones from the weather clkinfo don't really work.. they always get the background color and I cannot manage to change it
I'm not sure if the images themselves are not fine for this usage or I'm doing something wrong.. (in bwclock they all work though..).
Here's the usual minimal example with those, if I don't change the background they aren't even visible (always white).
```javascript
const icon1 = atob("GBiBAf/D//+B//8Y//88//88//88//88//88//8k//8k//8k//8k//8k//8k//4kf/5mf/zDP/yBP/yBP/zDP/5mf/48f/8A///D/w==");
const icon2 = atob("GBiBAf/7///z///x///g///g///Af//Af/3Af/nA//jg//B/v/B/H+A/H8A+D8AeB8AcB4AYA8AYA8AYA+A4A/B4A//4A//8B///Dw==");
const icon3 = atob("GBiBAf4f//wP//nn//Pn//Pzg//nAf/meIAOfAAefP///P//+fAAAfAAB////////wAAP4AAH///z///z//nz//nz//zj//wH//8Pw==");
function drawPics(icon, y) {
  g.reset().drawImage(icon, 0,y*24);
  g.setColor('#f0f').drawImage(icon, 24,y*24);
}
g.setColor("#888").fillRect(0,0,512,512);
drawPics(icon1, 0);
drawPics(icon2, 1);
drawPics(icon3, 2);
```

thyttan:
>they always get the background color

Is that true whatever theme you set/whatever the background color is?

glemco:
Well, they are invisible both on my bangle (dark theme) and on the emulator. I didn't try anything else but the clock is supposed to change their colour according to settings (e.g. each circle has its own colour) and I cannot manage that

gfwilliams:
Just a note that you can use `g.imageMetrics(img)` to get some info about an image.

But it looks like the issue is that:

* Those images are inverted
* The transparent color is the foreground!

So I don't see how this could ever have worked properly :( I've just pushed a fix for that though!

If it helps anyone else, you can fix it with this code.

```
img= E.toUint8Array(img)
img[3]=0;
for (var i=4;i<img.length;i++) img[i]^=255;
img=E.toString(img);
g.reset().drawImage(img);
print(btoa(img));
```

thyttan:
BW Clock inverts the theme at some point. Maybe that's the origin?

gfwilliams:
> BW Clock inverts the theme at some point. Maybe that's an explanation?

Ahh, that might do it! Although I'd have thought that with clock_info either weather would work and all the others are broken, or the others would be work and weather would be broken.

But maybe the icons were copied out of BW clock and not tested?

glemco:
Thank you both very much! Finally I'm starting to understand a little more about the image format.
So I guess I can carry on with the clock, ignoring the images representation.

glemco:
I put in this pull request only the modifications on circlesclock, in the meanwhile I changed something in some `clkinfo`s to display data as they were in this clock before the update. Some more pull requests will follow.
I made some assumptions:
* the field `text` is short enough and works fine on the circles (not really bulletproof, would it make sense to have a `short` field for this case?)
* if items don't have a name, I assume we are in the case in which they are going to change over time (e.g. timers or calendar events), then I read just the first.
* Since this watch doesn't allow user interaction after set-up from settings, `clkinfo`s with a name are assumed to always be there (e.g. `Weather/temperature`) while in other cases the first item should be the most meaningful to show (e.g. closest timer, if there).

gfwilliams:
This looks great, thanks! Sounds like a good plan about detecting items changing - but if it's a worry then we could add a `dynamic:true` field I guess?

> would it make sense to have a short field

I think if this becomes a problem, maybe - however the idea of clock_info is to display the stuff in quite a small area (I'm using 50px wide in  my tests at the moment) so hopefully all the infos will be designed not to be that wide.

gfwilliams:
>  the T- are taking space and I'd avoid them on the circlesclock.

I'd be very happy if you wanted to do a PR to add an optional `small` field? It wouldn't break anything and as you say it could be useful

glemco:
> I'd be very happy if you wanted to do a PR to add an optional `small` field? It wouldn't break anything and as you say it could be useful
Great! I'll add it to the pipeline.

gfwilliams:
Just a note that something came up in the forum about this, and it seems that these changes broke the HRM and probably anything that would auto-update: https://forum.espruino.com/conversations/382172/#comment16789739

I've made some *big* changes to try and strip out all the old code now and to just use the addInteractive function. It's not ideal, but I think at least it's a better staring point for it now.

glemco:
Well, it clearly did (and probably the same with other sensor-based data), but wouldn't it be cleaner if the clkinfo itself was managing the activation of the sensors?
Maybe I'm playing wrong with the `show()`/`hide()` methods, but I think it would be ideal if the two were enough to have the data ready. Does it make sense?

gfwilliams:
> but wouldn't it be cleaner if the clkinfo itself was managing the activation of the sensors?

Well, it does if you use `addInteractive`, which we're now using.

The issue is that some sensors (like HRM) don't have the data ready instantly - they have to go away and come back some time later.

If you want to interface with the clock infos at a lower level than addInteractive, you have to tell them when they're visible, and when they're not, and respond to their requests to redraw. There's been a lot of work put in to ensure they handle everything else based on that.

The real issue I see with what's done now is the colors are still per-circle, but now you can cycle through what gets shown. So I guess really the settings for color need to be per type of circle (eg, HRM, etc).

thyttan:
One way around the specific problem with Quick Launch could be to add actions on double swipes in Quick Launch. That way you can reserve single swipe actions to the clock, although the clock would act on the first of the two swipes in that case. I have though about adding this to Quick Launch but haven't got around to it as I currently use Slope Clock which doesn't do anything on swipes.

glemco:
Yeah that could be an option too, but honestly I'd still prefer using swipes for Quick Launch.
What about using the same taps as in bwclock (e.g. instead of a swipe up, tapping on the upper part of the screen) also here in circlesclock? Maybe with a visual feedback like little arrows once a circle is selected and starting the selection with a long-press of a circle instead of single to avoid accidental taps?

gfwilliams:
> plus it adds a quite noticeable delay every time it loads

That delay has always existed, but maybe it's more noticeable now the clock draws first - the 'ring draw' algorithm just draws a gazillion circles, and takes 0.25 sec PER ring. If there are 3x full circles that ends up being 6x0.25 sec just to draw the rings!

So really someone ought to change that - when this came up I'd actually posted a ring draw algorithm to the forum but if I change it and the rings a bit different someone will likely complain.

> Swiping doesn't seem to go well with QuickLaunch

Normally, swiping has no effect on clock_info, it's only when you tap and focus one of the circles. Perhaps clock_info could set `handled=true` on each event and QuickLaunch could check that to ensure that it doesn't fire if the event has been 'used'.

Or I believe when @peerdavid gets time he's planning to fix some outstanding issues with the clock_info in BW clock and then maybe add that code to clock_info itself, so when that happens maybe the code could be used here instead.

I think clock_info is getting big enough that maybe it could be an item in the app loader in its own right, with a settings page. In that case, what appears in each element (and the mode of interaction) could be configured.

glemco:
> That delay has always existed, but maybe it's more noticeable now the clock draws first - the 'ring draw' algorithm just draws a gazillion circles, and takes 0.25 sec PER ring. If there are 3x full circles that ends up being 6x0.25 sec just to draw the rings!
> 
> So really someone ought to change that - when this came up I'd actually posted a ring draw algorithm to the forum but if I change it and the rings a bit different someone will likely complain.

Damn, never noticed!
Well, you can't make everyone happy, unfortunately. Would that new algorithms make the circles much uglier? IMO since the bangle already doesn't have a particularly sharp resolution, faster circles are better. (Would it make sense a poll? or a setting in the app for accurate vs fast circles?)
 
> Normally, swiping has no effect on clock_info, it's only when you tap and focus one of the circles. Perhaps clock_info could set `handled=true` on each event and QuickLaunch could check that to ensure that it doesn't fire if the event has been 'used'.

That looks neat. I could have a look on that perhaps.

> I think clock_info is getting big enough that maybe it could be an item in the app loader in its own right, with a settings page. In that case, what appears in each element (and the mode of interaction) could be configured.

Agree, also there's no way to update the clkinfo module (or any module) without updating the app itself using it, right?
I think it's much more maintainable as standalone library as well.

PR #2253, opened 2022-11-09
[Torch] Comply w FW change back/btn

thyttan:


apps/torch/app.js
```diff
@@ -11,14 +11,16 @@ Bangle.setLCDBrightness(1);
...
+bgBackup = g.theme.bg;
 g.setTheme({bg:settings.bg,fg:"#000"});
 g.setColor(settings.bg);
 g.fillRect(0,0,g.getWidth(),g.getHeight());
 Bangle.setUI({
   mode : 'custom',
-  back : load, // B2: Clicking the hardware button or pressing upper left corner turns off (where red back button would be)
-  btn : (n)=>{ // B1: Any button turns off
-    if (process.env.HWVERSION==1 && (n==1 || n==2 || n==3)) {
+  back : ()=>{g.setTheme({bg:bgBackup});load();}, // B2: SW back button
```
gfwilliams, line 20: Is this needed? The theme should be reset every time a watch app loads?

I think this is a case of you using a non-standard firmware again. I think really rather than bodging all the apps, we should just add a fix to set the default theme in @halemmerich's loader?

apps/torch/app.js
```diff
@@ -11,14 +11,16 @@ Bangle.setLCDBrightness(1);
...
+  btn : (n)=>{ // B1&2: Any HW button turns off
+    if (n==1 || n==2 || n==3) {
```
gfwilliams, line 22: Again, do we need this? I think n can ONLY be 1/2/3 so this check seems redundant?

thyttan: Since I don't have a B1 I'm not sure. But I assumed presses of the screen would fire the `btn`-function with `n==4 `or `n==5`. To me it made more sense to only have the HW buttons exit the torch and not the screen presses, and that's also how it was coded before.

thyttan: Reading this on the HW reference indicates you are right though:
 > btn : function(n) {}, // optional - handler for 'button' events (n==1 on Bangle.js 2, n==1/2/3 depending on button for Bangle.js 1)

I guess I got confused by the `n`-variable in:
>touch : function(n,e) {}, // optional - handler for 'touch' events
Where I think I've read `n` can be `n==4` or `n==5`

apps/torch/app.js
thyttan: Yeah sorry, I tested it on both my usual setup and a factory reset and saw a little better experience on mine and no change in experience for standard, so I rationalized it that way. But it's gone now, I'll refrain from using halemmerich's stuff for a while as to not have this happen in PRs any more...

apps/torch/app.js
```diff
@@ -16,10 +16,6 @@ g.setColor(settings.bg);
...
 Bangle.setUI({
   mode : 'custom',
-  back : load, // B2: Clicking the hardware button or pressing upper left corner turns off (where red back button would be)
-  btn : (n)=>{ // B1: Any button turns off
-    if (process.env.HWVERSION==1 && (n==1 || n==2 || n==3)) {
-      load();
-    }
-  }
+  back : load, // B2: SW back button to exit
+  btn : ()=>{load();}, // B1&2: HW button to exit. // A simple 'load' as on the line above did not work for btn???
```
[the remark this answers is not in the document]
gfwilliams: It's because 'btn' passes an argument (1/2/3) so you're basically saying `load(1)` - and there's no file called `1` for it to load

gfwilliams:
Thanks! The fillRect after setUI is overwriting the whole screen including the back button so I guess that's why you get it clear.

Personally I think having 24x24px as the red back button is nice as it shows where to tap, so I'm pretty happy with it as-is. Thanks for your work on this!

apps/torch/app.js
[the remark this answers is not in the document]
thyttan: Thanks!

PR #2255, opened 2022-11-11
gpstrek - Improvements

halemmerich:
* Better handle state in widget
* Slightly faster drawing by doing some caching
* Reconstruct battery voltage by using calibrated settings.batFullVoltage
* Averaging for smoothing compass headings
* Save state if route or waypoint has been chosen


gfwilliams:
Thanks!

PR #2256, opened 2022-11-11
boot - Add polyfills for Bangle.load and Bangle.showClock

halemmerich:
These are minimal polyfills, just delegating the new calls to load(). Is this enough or should they mimic more of the new calls?


halemmerich:
Simple polyfills are not enough for #2264, is it better to work around this in the app or should the polyfills be functionally identical to the firmware methods? This should only be a problem for code depending on `Bangle.load/showClock/showLauncher` actually fastloading.

gfwilliams:
Thanks! I think that's ok for now - if all looks good I'll probably release 2v16 this week so hopefully it won't be a big issue.

PR #2261, opened 2022-11-12
[Desktop Launcher] Use Bangle.load() and Bangle.showClock()

thyttan:


gfwilliams:
Great - thanks!

gfwilliams:
Yes, sorry, I think that won't work reliably at the moment - a PR for that would be good.

I'll update the docs for `Bangle.load()` as well to explicitly mention this.

PR #2263, opened 2022-11-12
[Torch] use Bangle.showClock

thyttan:
Weirdly exiting with HW button behaves a little differently from using the red back button.
 
- Red back button: the clock and widgets are displayed at the same time when everything is ready.
- HW button: the widgets are displayed immediately, followed shortly thereafter by the clock.

Open question: should we just leave widgets shown i.e. we could do without `widget_utils` here?

### DIFF
--- a/apps/torch/app.js
+++ b/apps/torch/app.js
@@ -1,21 +1,31 @@
-const SETTINGS_FILE = "torch.json";
-let settings;
+{
+  const SETTINGS_FILE = "torch.json";
+  let settings;
 
-function loadSettings() {
-  settings = require("Storage").readJSON(SETTINGS_FILE,1)|| {'bg': '#FFFFFF', 'color': 'White'};
+  let loadSettings = function() {
+    settings = require("Storage").readJSON(SETTINGS_FILE,1)|| {'bg': '#FFFFFF', 'color': 'White'};
+  };
+
+  loadSettings();
+
+  let brightnessBackup = require("Storage").readJSON('setting.json').brightness;
+  let optionsBackup = Bangle.getOptions();
+  Bangle.setLCDBrightness(1);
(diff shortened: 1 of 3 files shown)


apps/torch/app.js
```diff
@@ -1,21 +1,34 @@
...
-g.reset();
-g.setTheme({bg:settings.bg,fg:"#000"});
-g.setColor(settings.bg);
-g.fillRect(0,0,g.getWidth(),g.getHeight());
-Bangle.setUI({
-  mode : 'custom',
-  back : load, // B2: SW back button to exit
-  btn : ()=>{load();}, // B1&2: HW button to exit. // A simple 'load' as on the line above did not work for btn???
-});
+  let brightnessBackup = require("Storage").readJSON('setting.json').brightness;
```
thyttan, line 12: A general note: Should brightness be part of the options returned by Bangle.getOptions() instead?

gfwilliams:
Odd about the change in the way it works.

Personally I'd just leave widgets shown - I feel like in the torch it really doesn't have that great an impact

thyttan:
I'll remove widget_utils here then! Can this be an exception from the rule regarding not having `remove : ` for apps? Or do you want me to close the PR instead, @gfwilliams?

gfwilliams:
Thanks! Yes, I think it's fine to leave the `remove:` stuff in here and merge it. I think it's so small that we could hopefully end up reasonably confident things will be ok.

... longer term we can definitely put more of the fast-switching in (and maybe we'll be able to come up with some kind of test). I just think that especially before we've even got `Bangle.load` in an official release, it's too early to start pushing it out to a bunch of apps.

gfwilliams:
Thanks!

PR #2264, opened 2022-11-12
swp2clk - Implement compatibility with Bangle.load/Bangle.showClock

halemmerich:
This implements an additional path for firmware newer than 2v15.135 while for now keeping compatibility with older firmware.

### DIFF
--- a/apps/swp2clk/boot.js
+++ b/apps/swp2clk/boot.js
@@ -107,18 +107,23 @@
 
   // start main function
 
-  if (!Bangle.load)
+  if (!Bangle.load){
+    log("no Bangle.load available");
     main();
-  else {
+  } else {
+    log("Bangle.load available");
     Bangle.load = (o => (fileName) => {
+      log("load " + fileName);
       var settings = readSettings();
+      log("remove swipe handler");
       Bangle.removeListener("swipe", swipeHandler);
       if (fileName && fileName != ".bootcde") {
-
+        log("loading a not clock app " + fileName);
         if ((settings.mode === 1 && settings.whiteList.includes(fileName)) || // "White List"
         (settings.mode === 2 && !settings.blackList.includes(fileName)) || // "Black List"
         settings.mode === 3) { // "Always"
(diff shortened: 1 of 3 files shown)


thyttan:
Something weird happens for me. I have Slope Clock, Quick Launch and Torch (last two with my newest changes in PRs) set up so:

Starting after an initialization of .bootcde:
1. swipe l-t-r starts torch with fast switching
2. swipe l-t-r shows the clock with fast switching
3. **swipe l-t-r shows the "Loading..." message after which the clock is shown again (<--- the strange part!)**
4. go to 1, loop ad infinitum.

halemmerich:
Hm, something strange is happening while registering a swipe handler inside a swipe handler. It seems both handlers get executed, the first clears Bangle.uiRemove, and the second then has to normal load the app because no uiRemove method is set...

halemmerich:
I think these commits fix the problem by adding the new swipe handler in the next idle. That should prevent the newly added handler running directly and creating the problem.

thyttan:
Tested and works for me! Thanks 👌

halemmerich:
This will probably fail on older firmware with polyfills, need to check that later this day.

apps/swp2clk/boot.js
```diff
@@ -19,17 +19,19 @@
   var DEBUG = false;
   var FILE = "swp2clk.data.json";
 
+  var swipeHandler = (dir) => {
+    log("swipe");
+    log(dir);
+    if (dir === 1) {
+      if (!Bangle.showClock) load();
```
gfwilliams, line 26: Do we need this now? showClock should always be defined after the latest bootloader tweaks

halemmerich: If we assume that all apps are up to date, this is not needed.

halemmerich:
This seems to work for me on 2v15 as well as 2v15.138 in combination with quicklaunch.

gfwilliams:
Wow, this looks great - thanks! This is miles cleaner!

PR #2265, opened 2022-11-12
[Remote for Spotify] Use Bangle.showClock

thyttan:
We've said before to keep fast switching between clock and launcher as to keep memory issues at bay. Does that change now when there're Bangle.load and Bangle.showClock? If so I think this can be merged.

### DIFF
--- a/apps/spotrem/ChangeLog
+++ b/apps/spotrem/ChangeLog
@@ -3,3 +3,4 @@
 0.03: change handling of intent extras.
 0.04: New layout.
 0.05: Add widgets field. Tweak layout.
+0.06: Add fast switching functionality.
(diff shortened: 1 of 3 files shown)


gfwilliams:
> We've said before to keep fast switching between clock and launcher as to keep memory issues at bay. Does that change now when there're Bangle.load and Bangle.showClock?

Not really - I feel like the same still applies at least for the moment. For this to work, every app is going to have to be really thoroughly tested (and *stay* well tested). I'm concerned that I'm going to end up with a whole raft of PRs to change apps, and then months of users reporting issues because of those changes - either memory leaks that have been found, or actual instability which will be almost impossible to track down - because now it's not just what apps you have installed, but what apps you used before as well and *when* you chose to switch them.

... and as I'd said in the forum thread i'm not sure this is really in most users' best interests. If I could spend my time making Espruino/Bangle.js faster in general then everyone would benefit.

thyttan:
I'll close this then! 👍

PR #2267, opened 2022-11-12
[Podcast Addict Remote] Use Bangle.showClock

thyttan:


gfwilliams:
Is this being a draft because of what you said on https://github.com/espruino/BangleApps/pull/2265 ? I'll answer on there

PR #2271, opened 2022-11-14
[Icon launcher] Use Bangle.load and Bangle.showClock

thyttan:
Not tested yet.

### DIFF
--- a/apps/iconlaunch/ChangeLog
+++ b/apps/iconlaunch/ChangeLog
@@ -14,3 +14,4 @@
       added timeOut to return to the clock
 0.11: Cleanup timeout when changing to clock
       Reset timeout on swipe and drag
+0.12: Use Bangle.load and Bangle.showClock
--- a/apps/iconlaunch/app.js
+++ b/apps/iconlaunch/app.js
@@ -94,7 +94,7 @@
     const appId = id * appsN + iconN;
     if( settings.direct && launchCache.apps[appId])
     {
-      load(launchCache.apps[appId].src);
+      Bangle.load(launchCache.apps[appId].src);
       return;
     }
     if (appId == selectedItem && launchCache.apps[appId]) {
@@ -102,7 +102,7 @@
       if (!app.src || s.read(app.src) === undefined) {
         E.showMessage( /*LANG*/ "App Source\nNot found");
       } else {
-        load(app.src);
+        Bangle.load(app.src);
       }
     }
     selectedItem = appId;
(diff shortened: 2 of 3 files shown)


gfwilliams:
Looks good - let me know if it works ok.

Only think I'd say is that `remove` generally gets called from `Bangle.setUI`, so calling `Bangle.setUI();` from it might cause issues. If it does let me know though as that'd be a good thing to fix

thyttan:
>Only think I'd say is that remove generally gets called from Bangle.setUI, so calling Bangle.setUI(); from it might cause issues. If it does let me know though as that'd be a good thing to fix


Adding some prints like this:
```
  let mode = {
    mode: "custom",
    drag: onDrag,
    touch: (_, e) => {
      if (e.y < R.y - 4) return;
      updateTimeout();
      let i = YtoIdx(e.y);
      selectItem(i, e);
    },
    swipe: (h,_) => { if(settings.swipeExit && h==1) { Bangle.showClock(); } },
    remove: function() {
      print('1');
    Bangle.setUI();
      print('2');
      if (timeout) clearTimeout(timeout);
    }
  };
```

After trying to open an app, I get this output to console field:
```
>
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
Before Bangle.setUI() inside mode
```

And then the Bangle reboots of it's own accord.

gfwilliams:
Thanks! Ok, I'll have to fix that then.

.. but if you can remove the setUI and this is tested with those other changes I'm happy to merge

thyttan:
Will do.

One first thing I got now, If i have "one click exit" enabled on Icon Launcher and press HW button on clock to load it, it will immediately load the clock with a "Loading..." message. I'll come back when I figured that out.

thyttan:
Seems to work fine now!

`if (settings.oneClickExit) mode.btn = Bangle.showClock();`

was the problem

gfwilliams:
Great - thanks!

thyttan:
@halemmerich Do you think the implementation of 'One click exit' should be kept as is, or replaced by a 'back' function in setUI?

Rarder44:
@thyttan do you mean by restoring the red button in the widget?
I had implemented those features just to make the app fullscreen and to be able to go back to the clock without having to hold down the button for 3 seconds or enter a random app.

thyttan:
Ah, yes I see! Good point.

Issue #2241, opened 2022-11-06
[Slope Clock] Ram leak when loading launcher during animation


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v15.93 (Reopened 2v15.136)

### The bug

Video showing the bug on this link: https://sprend.com/download?C=6ce80d7aee83402299727f2620d02e0f

Description of video:
1. On the clock RAM Widget will show 26% or 27% typically.
2. If the slope animation has finished and I load dtlaunch, the RAM Widget will say 13% on the launcher.
3. If I go back to Slope Clock and the switch back to launcher before the animation has finished, the RAM Widget will say 19% on the launcher.

The ram leak is consistent and fairly inconsequential, but it's there.

gfwilliams:
I'm actually really struggling to reproduce this with the default launcher, slope clock, and running `process.memory().usage` in the IDE console. Please could you try connecting with the IDE and running `process.memory().usage`? I'm wondering whether the widget just happens to update at the wrong time, and whether actually a few ms after it updates the extra memory usage has gone

thyttan:
On Slope clock when animation is finished:
```
>process.memory().usage
=2365
>process.memory().usage
=2365
```

On dtlaunch launched after the animation on Slope clock was finished:
```
>process.memory().usage
=1469
>process.memory().usage
=1469
```

On dtlaunch launched during the Slope clock animation:
```
>process.memory().usage
=2190
>process.memory().usage
=2190
```

On Slope clock during the animation:
```
>process.memory().usage
=3105
>process.memory().usage
=3105
```
I ran the command twice for every case to see if it was constant.

Before you do any more testing, I'll do a factory reset and install just the things you tested with.

thyttan:
> I'm actually really struggling to reproduce this with the default launcher, slope clock, and running process.memory().usage in the IDE console.

After a factory reset and installing Slope clock, Ram widget and Quick launch I don't get the ram leak either.

gfwilliams:
Thanks for checking that out after a factory reset! Yes, that's going to be a tricky one to track down... If you get this again (especially with stuff from the app loader) definitely let me know though!

thyttan:
Closing this since the problem doesn't seem to arise when using the official app loader.

thyttan:
I get this when only using stuff from official development app loader (+ my changes in open PR's) now as well, with the introduction of Bangle.load and the likes.

halemmerich:
[...]

PR #2272, opened 2022-11-14
Slopeclock/pp - Reset graphics to remove font from memory

halemmerich:


thyttan:
Just a note to say that #2241 seems solved even before this PR, at least from looking at the RAM Widget (not tested ++)

halemmerich:
Yes, but that depends on whatever is loaded after the clock doing a `g.reset()` somewhere.

gfwilliams:
Thanks! It's a good point - hopefully apps won't assume that graphics will be reset but it'd be a good idea to do it in case. We just don't want to get into a position where we're implementing all of `reset/load()` just in JS each time we do `Bangle.load()` :)

... what about calling it in `setUI` instead? That might be quite a good idea in general as it pretty much always results in the screen being redrawn.

I'll merge this, but will actually tweak it slightly after - because IIRC the font is only used to redraw once in the `draw()` function, and after that it's not needed (even during the animation). We might as well reset state right after drawing and then the Bangle isn't sitting there for seconds with the font allocated

halemmerich:
Since `Bangle.load()` calls `setUI()` anyway in case a fast load is done that would not make any difference. I think it is OK to have the apps being responsible to clear out changes to the global graphics objects.

PR #2273, opened 2022-11-14
[Analog Clock Widget] clear before redraw

thyttan:


apps/widanclk/widget.js
```diff
@@ -12,7 +12,8 @@ WIDGETS["wdanclk"]={area:"tl",width:Bangle.CLOCK?0:24,draw:function() {
   let x=this.x+12, y=this.y+12,
       ah = (d.getHours()+d.getMinutes()/60)*Math.PI/6,
       am = d.getMinutes()*Math.PI/30;
-  g.drawCircle(x, y, 11).
+  g.clearRect(this.x, this.y, this.x+this.width, this.y+24).
```
thyttan, line 15: Would it be better to use `Bangle.appRect.y` rather than `this.y+24`, or maybe something else?

thyttan: Also, is clearRect the fastest way to clean up the previous draw? Or would doing `drawLine(x,y, x+Math.sin(ah-1)*7, y-Math.cos(ah-1)*7).`, with changed foreground colour be faster?

thyttan: Also, is clearRect the fastest way to paint out the previous draw? Or would it be faster to find the hands and paint only them out, with `drawLine(x,y, x+Math.sin(ah-1/60)*7, y-Math.cos(ah-1/60)*7).` and `drawLine(x,y, x+Math.sin(am-1)*9, y-Math.cos(am-1)*9);`

gfwilliams: I think what you've done looks great - except maybe it needs to be `y+23` and `.width-1` because the coordinates for clear/fillRect are inclusive, and we don't want to go outside of our widget's area?

> Also, is clearRect the fastest way to clear the previous draw?

Yes, I think so given the overhead of the JS execution and maths - it's nice and safe too.

gfwilliams:
Great, thanks for this spot - that was a stupid mistake of mine ;)

apps/widanclk/widget.js
thyttan: Thanks! Ready for merge now.

gfwilliams:
Thanks!

PR #2274, opened 2022-11-15
- implemented Bangle.emit into android/boot ( remove direct call to m…

Rarder44:
I implemented a demo of what I was trying to explain here: [](https://forum.espruino.com/conversations/379866/#comment16758750)
 
it is a proof of concept (theoretically working).

I say theoretically working because on the basis of these modifications I implemented my "message" app and (for now) it seems to work without problems.

  I think there are a few things that need to be changed:
- implementation of other events, or make them more specific (music for example)
- cleanup of the code and choice of the essential parameters to be passed by event
- creating a separate library for buzz/stopbuzz/getMessageImage/getMessageImageCol/...

It could be a solution? Are there any downsides I'm missing out on?

### DIFF
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -28,12 +28,18 @@
           event.id = lastMsg.id;
         }
         lastMsg = event;
-        require("messages").pushMessage(event);
+        Bangle.emit("message", "add", event);
       },
       // {t:"notify~",id:int, title:string} // modified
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - let's discuss this on https://forum.espruino.com/conversations/379866/#comment16762840 as there's already a long thread there

Rarder44:
I'm closing this PR because any solution that will be adopted will almost certainly not be this one.

PR #2275, opened 2022-11-15
Initial version of Sleep Status Widget

myxor:


gfwilliams:
Thanks! Do you think this should have a `  "dependencies" : { "sleeplog":"app" } ` in metadata so it requires that sleeplog is installed?

myxor:
Yes you are right. I added the dependency.

gfwilliams:
Thanks!

PR #2284, opened 2022-11-16
bthrm - Fix restarting issue and more debug logging

halemmerich:
* Fixes an issue where powering down the HRM and then powering on would prevent BTHRM from reconnecting
* Better logging for errors
* Additional setting in debug menu for active scanning


gfwilliams:
Thanks!

PR #2288, opened 2022-11-17
[rebble] - convert var/function to let

Rarder44:


gfwilliams:
Thanks! Do you need `function log_debug(o) {` ? I think that will still be defined globally.

Also, it doesn't matter, but in future you don't have to worry about changing the settings page - because it's a function itself there's no problem with memory

Rarder44:
>Thanks! Do you need function log_debug(o) { ? I think that will still be defined globally.

I had missed it! thank you!

>Also, it doesn't matter, but in future you don't have to worry about changing the settings page - because it's a function itself there's no problem with memory

i like to keep the same style for all files :)

if that the log_debug function is useless but it allows me to work much faster when I need to debug something.

(maybe it exists and I don't know)
would be nice a way to use:
#define DEBUG
#ifdef DEBUG
//code
#endif
as in c

so that during "minification"/"upload" these pieces are removed.

gfwilliams:
> #define DEBUG

Yes... well, you'd hope that what you're doing (maybe `const log_debug = function(){}`) would get removed by a decent minifier. I'm not sure if Esprima *is* that good though

PR #2289, opened 2022-11-18
[rebble] cleanup code and fix fastload issue

Rarder44:
sorry, I noticed that an event was not removed during the "remove".


gfwilliams:
Great - thanks!

PR #2290, opened 2022-11-18
[Rebble] fix draw before widget hide

Rarder44:
excuse me! I didn't realize that before I introduced a problem in the graphics!
I hope I haven't forgotten anything!


gfwilliams:
Thanks! :)

PR #2291, opened 2022-11-18
taglaunch: New launcher

nxdefiant:
Fork of the default launcher v0.19 but puts all applications in a submenu by their tag. This can result in a faster application selection than the linear access of the default launcher.


apps/taglaunch/app.js
```diff
@@ -0,0 +1,133 @@
...
+           };
```
nxdefiant, line 9: Probably a bad idea regarding memory,
I'll put them in separate files when more 
tags will be added.

gfwilliams: Yes, although for these few tags I guess you'll be ok at least on Bangle.js 2 :)

gfwilliams:
Thanks! This seems like a nice idea!

apps/taglaunch/app.js
nxdefiant: I had to remove support for Bangle JS1 because
out of memory :(

thyttan:
@nxdefiant @jeffmer 

> I'm thinking about maybe doing a version of Tag Launcher based on Desktop Launcher. It could be introduced in one of three ways: as an update to Tag Launcher, as an option in Desktop Launcher or as a new app. Personally I lean towards an update to Tag Launcher.

https://forum.espruino.com/conversations/381378/?offset=25#comment16933788

PR #2292, opened 2022-11-18
[Messages light] - new message GUI app

Rarder44:
a new app that replaces the message app GUI.


gfwilliams:
Thanks! Let's wait until https://github.com/espruino/BangleApps/pull/2297 goes in and then we should be able to tweak and merge this and the module renaming won't be needed

Rarder44:
yes yes, no problem!
since perhaps the work on the message could be long, I thought of starting to publish this (also for my order) and then, when the new methodology is completed, modify this app.
(I also uploaded it, so in splitting the message, there is a reference to what could be needed in a "custom" GUI message)

gfwilliams:
Yes, ok - makes sense. I'll merge this now and I think we can remove the extra bits pretty easily later

In future, it'd be handy if you could make your version numbering have 2 decimal places like other apps though? - even if you want to start with `1.xx` instead of `0.xx` is just seems less confusing for users.

Rarder44:
oh ok, no problem!

PR #2293, opened 2022-11-20
Update README.md to hint about how to require Bangle modules.

lauzonhomeschool:
It's frustrating for newcomers to try to tweak an existing app in the Web IDE and get "Module ... not found" and not know why, when this section just seem to state that any "require" should work.  The extra note clarifies this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,7 +458,10 @@ It should also add `myappid.json` to `data`, to make sure it is cleaned up when 
 ## Modules
 
 You can include any of [Espruino's modules](https://www.espruino.com/Modules) as
-normal with `require("modulename")`. If you want to develop your own module for your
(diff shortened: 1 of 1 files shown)


gfwilliams:
Great! Thanks for updating the documentation - it's a huge help!

lauzonhomeschool:
Thanks.  And what would be the URL to include in the Web IDE other standard libraries that are in apps folders instead of modules, e.g. "sched"?  I'll add it to a readme somewhere.

gfwilliams:
> And what would be the URL to include in the Web IDE other standard libraries that are in apps folder

There isn't one - you need to install the app that provides that library (as potentially more than one app can implement the same library, just in different ways).

Then you set in your app's metadata that it needs to have an app installed that provides the functionality - for example  `"dependencies": {"scheduler":"type"},`

PR #2294, opened 2022-11-20
Update README.md, state the IDE error that we are fixing

lauzonhomeschool:
Newcomers see the error e.g. "Module time_utils not found", Bangle freezes, a search on this repo doesn't list this error in the existing readmes, yet I suspect it is a very common error.  This PR will have it listed and hopefully help the next newcomer :)


gfwilliams:
Thanks!

PR #2296, opened 2022-11-20
[ios] Add letters to unicodeRemap

Kedlub:
Added letters with caron to unicodeRemap, to properly display messages in Czech language


gfwilliams:
Thanks!

gfwilliams:
Thanks! I'll do it this time, but in future please can you update the version and ChangeLog so users know they can upgrade and get new stuff?

PR #2297, opened 2022-11-20
messages: split library/gui/widget into separate apps

rigrig:
Some issues I ran into:
* I was having connection problems to http://puck-js.com/, so my app loader broke -> no testing yet :-(
* I've set up dependencies as "modules": `messagegui`/`messagewidget`, but the sanitycheck complains that the filenames don't match module names:
```
::error file=apps/messagegui/metadata.json::App messagegui has provides_modules messagegui but it doesn't provide that filename
::error file=apps/widmessages/metadata.json::App widmessages has provides_modules messagewidget but it doesn't provide that filename
::error file=apps/widmsggrid/metadata.json::App widmsggrid has provides_modules messagewidget but it doesn't provide that filename
```
* Settings are a  tricky, do we split them across the library/GUI/widget apps, or keep them together (inside the library)? But then what about custom apps/widgets with their own settings?


gfwilliams:
Thanks for this! Sorry about the app loader issue - the server had problems but it's fixed now.

> App messagegui has provides_modules messagegui but it doesn't provide that filename

The idea is that if an app says `provides_modules` then when it's installed you should be able to do `require("messageGUI")...` so it needs to provide a file called `messageGUI` in https://github.com/espruino/BangleApps/pull/2297/files#diff-e4329cd45d63616839f74226cff93ac933bc2cf5b359f417de099f7eb3787eb5

I reckon since there can only be one message gui, instead of `Bangle.emit("messageGUI"` we should just do `require("messagegui").open()` - and then you can remove `require("messages").openGUI()`?

That way it's faster and it's not using any memory at all until the gui is needed

> widmessages/widmsggrid

I think these probably just need to provide a `messagewidget` file in the same way as above. Maybe it doesn't even do anything at the moment - or I guess it could provide a `hide` method so that the `messagesgui` can ensure they're not shown when viewing messages?

> Settings are a tricky, do we split them across the library/GUI/widget apps, or keep them together (inside the library)? But then what about custom apps/widgets with their own settings?

Yes, this is a pain. I'd hit this with the messsageicons and left them in the `messages` app for now.

I guess longer term maybe they need splitting up?

rigrig:
> just do `require("messagegui").open()` - and then you can remove `require("messages").openGUI()`?
> provide a `messagewidget` file

Yeah, that makes sense. Definitely happy about getting rid of the `"messageGUI"` event.

> provide a `hide` method so that the messagesgui can ensure they're not shown when viewing messages?

I like that better then what I've did now (setting a global `Bangle.MESSAGES=1`), and I expect it'll also work nicer with fast app (un)loading.


> longer term maybe [settings] need splitting up?

How about splitting up the menus, but still saving everything in `messages.settings.json`? And just accept that switching between GUIs/widgets might leave some orphaned settings?

gfwilliams:
Sounds great!

> How about splitting up the menus, but still saving everything in messages.settings.json? 

That sounds good to me - it'll be faster too as it seems pretty much every time a notification comes in we load the settings, so at least it's less files to load.

If that's the case, let's just leave everything as-is for now to get the scope of this PR down? Then we can change that all later.

Hopefully as far as this PR is concerned, once merged the Bangle will behave basically identically to before, just with more apps in the app loader.

gfwilliams:
Are you happy with this now? I do see one reference to `Bangle.MESSAGES` which I think might be kicking around from before?

But otherwise it looks pretty good?

rigrig:
I think I'm happy, just been too busy to test it/look it over properly (or I'd at least have found that `MESSAGES`).
But I've got plenty of time tomorrow.

rigrig:
Ran into a bunch of bugs, but I think I fixed them all now.   
(First one: circular dependencies caused an infinite loop in the loader, I figured "don't use circular dependencies" is good enough for now ;-)

gfwilliams:
Ok - just merged! Had to tweak a few things to ensure it actually auto-updated if you had normal messages on before, but it does appear to work for me as well :)

PR #2299, opened 2022-11-20
apploader - Load puck.js locally

halemmerich:
Is this a valid way to make the locally hosted apploader usable if puck-js.com is not reachable? It did work for me, but I do not know if there are other implications of this.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "EspruinoAppLoaderCore"]
 	path = core
 	url = https://github.com/espruino/EspruinoAppLoaderCore.git
+[submodule "webtools"]
+	path = webtools
+	url = https://github.com/espruino/EspruinoWebTools.git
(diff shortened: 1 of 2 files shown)


myxor:
This is a good idea from my point of view as puck-js.com is down a lot since the weekend and therefore even local AppLoaders stopped working.

gfwilliams:
Argh, sorry about the downtime - looks like there have been so many failed login attempts the log files filled up all the server's storage! Yes, that would work ok.

Right now we host some of the WebTools stuff inside AppLoaderCore (like the imageconverter). I wonder if we should:

a) copy the puck.js lib inside apploadercore instead
b) Do what's suggested here and actually remove those other libs from AppLoaderCore and include the webtools repo instead?

halemmerich:
I don't like duplication, so my vote goes to removing the libs and using the webtools repo.

gfwilliams:
Ok, let's give this a try then, and I'll update the apploadercore

PR #2301, opened 2022-11-20
widget_utils - Prevent initial draw clearing on global g

halemmerich:
This fixes `Bangle.drawWidgets` clearing the top widget bar on the global graphics object instead of the newly allocated buffer for the widgets.


gfwilliams:
Thanks! I shouldn't have missed that ;)

PR #2302, opened 2022-11-21
MyLocation v0.08: Allow setting location from webinterface in the AppLoader

myxor:
Allow setting location from webinterface in the AppLoader

image


gfwilliams:
looks great - thanks!

PR #2304, opened 2022-11-21
[Android] Add handling of phone gps data

LukasEdl:
In the [Gadgedbridge PR](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/2976), i add the possibility to use the gps data of the phone.

With this PR, the android app gets extended, so that it can handle the communication with the gadegtbridge part.
To enable to do that i changed the following.
* I added a handler for a "gps" and a "is_gps_active" GB message.
   - The "gps" message retrieves the gps position of the phone and publishes it as a "gps" event onto the "Bangle" object.
   - The "is_gps_active" messages is a request of the phone to retrieve the current power status of the gps. This is used on the phone to stop sending GPS data if the gps is turned off on the bangle device to perserve the battery of the phone.
* I added a logic to overwrite the "setGPSPower" method and the "isGPSOn" method. This is needed so that, if the phone gps data is being used, the gps of the banglejs device is turned off.
* I added a setting to enable the overwrite. This must be set or other wise the messages from the phone will be ignored.

* The received gps data will be provided under the "gps" event, like if you would use the normal gps chip of the banglejs. However if the phone data is being used, the event "GPS-raw" will not be fired. (I don't know how to convert the gps data to raw data, if anyone can and want to do it, feel free to add this)
* The phone gps also dose not provide the amount of satelites and the course. Those two values will be set to the NaN value. Also the fix value will be set to 1 ( If someone knows how to set the correct "fix" value, feel free to add it)
* GPS data, which was provided through the phone will have an extra flag called "externalSource".

### DIFF
--- a/apps/android/settings.js
+++ b/apps/android/settings.js
@@ -1,4 +1,13 @@
 (function(back) {
+
+  function onGpsOverwriteChange(newValue) {
+    if (newValue) {
+      Bangle.setGPSPower(false, 'android');
+    }
+    settings.overwriteGps = newValue;
+    updateSettings();
+  }
+
   function gb(j) {
     Bluetooth.println(JSON.stringify(j));
   }
@@ -22,6 +31,10 @@
         settings.keep = v;
         updateSettings();
       }
+    },
+    /*LANG*/"Overwrite GPS" : {
+      value : !!settings.overwriteGps,
+      onchange: onGpsOverwriteChange
     },
     /*LANG*/"Messages" : ()=>load("messages.app.js"),
   };
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -130,6 +130,13 @@
           request.j(event.err); //r = reJect function
         else
           request.r(event); //r = resolve function
+      },
+      "gps": function() {
+        delete event.t;
+        event.satellites = NaN;
+        event.course = NaN;
+        event.fix = 1;
+        Bangle.emit('gps', event);
       }
     };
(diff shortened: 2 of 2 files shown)


myxor:
Looks good but i suggest to add some description for the "Overwrite GPS" setting to the README file.

apps/android/settings.js
```diff
@@ -1,4 +1,13 @@
 (function(back) {
+
+  function onGpsOverwriteChange(newValue) {
+    if (newValue) {
+      Bangle.setGPSPower(false, 'android');
```
gfwilliams, line 5: I'm not sure this is needed... The way `setGPSPower` works this would only ever turn off the GPS if it had been turned on with `Bangle.setGPSPower(true, 'android');` - if any other app requested it, it'd stay turned on

gfwilliams:
This looks great, thanks! And I really appreciate you adding the setting for this - as until we get this all merged and in the app store we don't want users finding their GPS stops working ;)

Just a thought, but since when the setting is changed we end up restarting and reloading the `boot.js` file, maybe we should check `currentSettings.overwriteGps` and then *only* overwrite the existing GPS power functions if that is set?

Do you see a problem with doing that? If not I'm happy to merge this and make the change here

gfwilliams:
@LukasEdl  just checking on what I said above - is it ok if I do that?

LukasEdl:
Sorry for the late response, i was away the last few days. If you want to do that, please go ahead.

PR #2305, opened 2022-11-21
Imageclock - Better widgets and performance

halemmerich:
* Uses the new `slideOn` from widget_utils to get a lot cooler and smoother handling of widget hiding
* Combine code files into one to make minification possible, saves about 50% file size and 10-15% render time


gfwilliams:
Great - thanks!

PR #2306, opened 2022-11-21
quicklaunch - Respect appRect on touch events

halemmerich:


halemmerich:
Swiping away widgets forces full screen appRect because apps need behave as if there were no widgets. This state is not detectable only using appRect. The same applies to apps implementing other methods of widget hiding. In those cases the easiest way is to ignore widget bars always.

halemmerich:
Then apps using swipeOn need to ignore the appRect and assume full screen. That seems feasible, apps should know if they are using swipeOn. What happens with other apps/widgets using appRect?

gfwilliams:
> Then apps using swipeOn need to ignore the appRect and assume full screen

Well, we couldn't easily strip out touch events that happen on the widget bar, but what I mean is we can change appRect to y=0 when no widgets are shown, and after the swipe y=24.

Then your current code here (that is checking appRect) would be perfect, but you wouldn't need the `restrictTouch` bit?

If that sounds good I'm happy to implement it now

gfwilliams:
Just updated - so I reckon we don't need `restrictTouch` now, just the appRect check?

halemmerich:
Lets without that, it's easily done later if actually needed :)

gfwilliams:
Thanks!

Issue #2189, opened 2022-10-22
[messages] Missing icons


rigrig:
I've kept a list of apps that sent a notification but don't have an icon yet.
I plan to (eventually) submit a PR to add them, but adding them one-by-one felt like way too much churn.

- [ ] KDE Connect
- [ ] Aurora Store
- [ ] Rabobank
- [ ] booq Personeelsplanner
- [ ] NS
- [ ] Network manager
- [ ] f-droid build status
- [ ] KeePassDX
- [ ] phonetrack
- [ ] Mull
- [ ] 1Password
- [ ] Tasks
- [ ] PostNL
- [ ] VLC

@gfwilliams How do you feel about adding lots of apps? Should we just "add them all", or maybe be a bit selective about which apps to include? (I doubt there are many people using "booq Personeelsplanner" or "Rabobank" outside the Netherlands...)

Rarder44:
since the list of apps could become long, why not create an online repository with images and the app downloads the icons it needs via a network connection and keeps them in cache?
this would avoid having to commit exclusively to images and the app would be much lighter

gfwilliams:
Personally I'd just include all the icons for now - we shouldn't be calling the function that often (and it sits in flash) so I think we'll be fine. If it gets really bad we can always look at including it in binary form.

The problem with the network connection is folks using Apple phones don't have any access to it.

DZPM:
How about having broad categories of icons? In example, all the banks could share the same icon, could even be the same bank emoji: 🏦 
There could be categories for password managers, games, etc.

I'm thinking in maintainability. Maintaining this categories would be way easier, as users would only insert "LocalBank" to the list of banks, without having to generate/convert a custom icon each time, or having to look at repeated icons.

gfwilliams:
> How about having broad categories of icons?

That sounds like a great idea to me... Given most icons are black and white 24x24px there should really be a lot of overlap.

Maybe one option to make this more maintainable is to have a folder of PNG images named by the app name, and then a simple script that loads them (handling duplicate images) and creates the code from them. It makes it easier to check icons at a glance, but also if at some point we wanted to go for a binary format changing over would be easy too.

m-p-3:
Sorry to drop in, but is there some documentation and/or guidelines on the icon format? I might add a couple if possible.

Thanks!

gfwilliams:
[...]

PR #2308, opened 2022-11-23
Add 24x24px icon source files under "messageicons/icons"

m-p-3:
As discussed in #2189, here are the original files I plan to use (not all of them, but I make them available in here just in case we decide to use more, as they're ready to be used) to better distinguish visually the origin of a notification.

I'll start updating `lib.js` shortly with those, preferring generic icons whenever possible to minimize the amount of icons required.


gfwilliams:
Looks great - thanks!

PR #2311, opened 2022-11-25
Add support for local time display to slopeclockpp

eyecreate:


gfwilliams:
Thanks! Looks like this fixes https://github.com/espruino/BangleApps/issues/2317 - I'll apply it to the original slopeclock too

PR #2312, opened 2022-11-25
[New App] Quick Center

Kedlub:
A small launcher-like app made for quick access to your favorite apps, inspired by control centers of other watches. It's designed to be used with other quick-launching apps, as an additional menu.

While developing I found out that Layout.js doesn't change the button width and height while changing the image scale, and I had to copy the button rendering code and use it as a custom element to circumvent it. So is there any better way to force button size?


gfwilliams:
Looks great - thanks! So the issue is that when you have a button with a scaled image, the size isn't right? It looks like scaled images are ok, but maybe that code didn't get copied over properly.

I think it's best to just get a fix in Layout for that?

PR #2314, opened 2022-11-26
Sched clkinfo and other improvements

glemco:
Various changes in some clkinfo mainly to support the `hasRange` functionality.

### DIFF
--- a/modules/clock_info.js
+++ b/modules/clock_info.js
@@ -2,9 +2,9 @@
 that can be scrolled through on the clock face.
 
 `load()` returns an array of menu objects, where each object contains a list of menu items:
-
 * `name` : text to display and identify menu object (e.g. weather)
 * `img` : a 24x24px image
+* `dynamic` : if `true`, items are not constant but are sorted (e.g. calendar events sorted by date)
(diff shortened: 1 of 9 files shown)


modules/clock_info.js
```diff
@@ -48,6 +49,15 @@ example.clkinfo.js :
 
 */
 
+let storage = require("Storage");
+let stepGoal = undefined;
+// Load step goal from health app and pedometer widget
+let d = storage.readJSON("health.json", true) || {};
```
gfwilliams, line 55: Thanks! Maybe we should make `wpedom` use `health.json` for this as well? Seems like a shame if apps have to pull this from two places

gfwilliams:
Great - thanks!

modules/clock_info.js
glemco: Definitely. Maybe also document it somewhere (besides the Health's readme) so that everyone would that instead of implementing their own settings?

gfwilliams: Yes, sounds like a good plan! Just filed an issue for this: https://github.com/espruino/BangleApps/issues/2318

PR #2315, opened 2022-11-27
hrmmar: Add fft elimination algorithm to remove MA

nxdefiant:
Remove Motion Artifacts by cutting out the frequencies from the HRM frequency spectrum that are noisy in acceleration spectrum. Under motion this can report a heart rate that is closer to the real one but will fail if motion frequency and heart rate overlap.

The fft with size=256 is calculated over a window of 8 seconds with step size 2 seconds. Sample rate of acceleration (12.5Hz) and PPG sensor (25Hz) is unchanged, so the algorithm runs with 12.5Hz. Generally the maximum peak in fft spectrum in an interval of -5..+10BPM of last one is used.

The low fft resolution limits the accuracy to ~3BPM, but mean error is ~5-10BPM.

If firmware reports confidence >= 90% its value is used.


gfwilliams:
Thanks! That's a really neat idea!

PR #2316, opened 2022-11-27
launch - Use Bangle.showClock

halemmerich:
Reduces code size a bit and uses firmware methods for cleanup/loading.


gfwilliams:
Neat - thanks!

PR #2320, opened 2022-11-28
Fastload - Allows fast loading on more loads

halemmerich:
This allows fast loading if the current app has a `remove` method and the target app contains `Bangle.loadWidgets`.


apps/fastload/metadata.json
```diff
@@ -0,0 +1,16 @@
...
+  "shortName" : "Fastload Utils",
+  "version": "0.03",
+  "icon": "icon.png",
+  "description": "Enable experimental fastloading for more apps",
+  "type":"bootloader",
+  "tags": "system",
+  "supports": ["BANGLEJS2"],
+  "readme": "README.md",
+  "storage": [
+    {"name":"fastload.5.boot.js","url":"boot.js"},
```
thyttan, line 12: should it be fastload.5.boot.js and not fastload.boot.js?

thyttan: Ok, thanks for explaining!

gfwilliams:
Thanks! It's nice to have this so neatly packaged up for users to install

PR #2323, opened 2022-11-28
[sleeplogalarm] New widget to trigger alarms early on light sleep or awake

storm64:
This widget searches for active alarms and raises an own alarm event up to the defined time earlier, if in light sleep or awake phase.


gfwilliams:
Looks great, thanks! Please could you add a dependency on `"scheduler":"type"` to the metadata, as I think this app needs that too?

gfwilliams:
Thanks!

PR #2324, opened 2022-11-28
Fix messagesmusic

### LINKED ISSUE #2321: [Messages Music] Can't install/update
@rigrig

### Affected hardware version

Bangle 2

### Your firmware version

2v16.11

### The bug

Can't install Messages Music v0.03, error message:

>Device connection failed, Error: Dependency of 'messagelib' listed, but nothing satisfies it!

I think there is a confusion of 'messageslib' and 'messages' maybe? The former of the two doesn't seem to exist, and the name is still messages for the library. I think?

Reproduce:
1. Connect to the development app loader
2. Try installing or updating 
3. Get the error message quoted above.

'messageslib' is also mentioned in the [README.md for Messages](https://github.com/espruino/BangleApps/blob/master/apps/messages/README.md):

>`require("messages")` (provided by `messagelib`) calls `Bangle.emit("message", "text", {/** the message */})`

rigrig:
Fix #2321
- Fix the dependency for `messagesmusic` (whoops)
- Wrap the app code in a timeout, as otherwise it didn't work for me (no idea why).

### DIFF
--- a/apps/messagesmusic/ChangeLog
+++ b/apps/messagesmusic/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Remove one line of code that didn't do anything other than in some instances hinder the function of the app.
 0.03: Use the new messages library
 0.04: Fix dependency on messages library
+      Fix loading message UI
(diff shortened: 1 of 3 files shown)


thyttan:
>Wrap the app code in a timeout, as otherwise it didn't work for me (no idea why).

When trying 0.03 by loading it via the web ide I got the same result.

PR #2327, opened 2022-11-29
weather: added weather condition to clkinfo

glemco:
The picture is dynamically generated with the weather lib, maybe not better than having pre-generated pictures but looks quite good on my tests.

### DIFF
--- a/apps/weather/clkinfo.js
+++ b/apps/weather/clkinfo.js
@@ -3,6 +3,7 @@
         temp: "?",
         hum: "?",
         wind: "?",
+        txt: "?",
     };
 
     var weatherJson = require("Storage").readJSON('weather.json');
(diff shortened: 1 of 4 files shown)


myxor:
i planned to do the same in the upcoming days. Great to see that you did it. Thanks! :)

apps/weather/clkinfo.js
```diff
@@ -14,33 +15,50 @@
...
+                show: function() { weatherItems.items[0].emit("redraw"); },
+                hide: function () {}
+            },
             {
                 name: "temperature",
                 hasRange : true,
                 get: () => ({ text: weather.temp, img: atob("GBiBAAA8AAB+AADnAADDAADDAADDAADDAADDAADbAADbAADbAADbAADbAADbAAHbgAGZgAM8wAN+wAN+wAM8wAGZgAHDgAD/AAA8AA=="),
                   v: parseInt(weather.temp), min: -30, max: 55}),
-                show: function() { weatherItems.items[0].emit("redraw"); },
+                show: function() { weatherItems.items[1].emit("redraw"); },
```
gfwilliams, line 45: Actually I think you can just do `this.emit("redraw");` for these. Not that it matters now, but it might make this more maintainable

gfwilliams:
That looks really nice - thanks!

glemco:
@myxor feel free to add your changes ;)
At the moment it's still missing the sun/moon pictures as it isn't aware of it (also the sun isn't so nice on a single colour), after that I guess it could be used to implement the current weather circle in circlesclock (which now reads from the json).

@gfwilliams right, I am just copying the way they were, I guess in all clkinfo it's done like that rather than with `this` (which is by the way one of the nightmares of javascript).

PR #2331, opened 2022-11-30
layout - Allow providing a remove handler in options

halemmerich:
This allows to push a remove handler through to `setUI` and allows cleanup after using a layout.


gfwilliams:
Thanks! That's a good idea...

PR #2332, opened 2022-11-30
qcenter - Fastload the clock on swipe

halemmerich:
qcenter is essentially a launcher and should be able to go back to the clock like one. Needs #2331 to actually trigger fast loading because else the remove method is not set for `setUI`. Should do close to nothing without #2331.


gfwilliams:
Looks great - thanks!

PR #2333, opened 2022-11-30
gallery: Add interface

nxdefiant:
Just for fun Quick & Dirty interface for gallery. Basically just copied imageconverter.html and added some Util.writeStorage()/Util.eraseStorage()


nxdefiant:
@bruceblore FYI

gfwilliams:
Thanks - that's great! Definitely what this needed :)

PR #2334, opened 2022-11-30
[Torch] add loadWidgets to work with Fastload Utils app

thyttan:


nxdefiant:
I would like to make a feature request: Show widgets only optional (configurable with settings?). Torch should be just white on my Bangle, without any widgets.

halemmerich:
How about hiding them by default with `require("widget_utils").swipeOn()` ?

thyttan:
@halemmerich I get some problems navigating back and forth Slope Clock <-> Torch, using Quick Launch.

1. Reset watch with long press hardware button to get a "Loading..." message.
2. Slope Clock is loaded.
3. Swipe to trigger loading Torch via Quick Launch.
4. Torch is loaded with a completely white screen. SlideOn works as expected.
5. Go back to Slope clock with red back button.
6. Slope Clock is loaded, 24pix reserved for widget but they don't respect the theme and they are hidden - not shown as expected. They are shown by a swipe down.

Sometimes the Red back button will show permanently on Torch after going back and forth between it and Slope Clock.

If I exit Torch via hardware button click the widget field will be hidden and the clock take up all screen space. SlidOn seem to persist and widgets are shown on a swipe down in this case.

One solution to this is probably to add slideOn to Slope Clock as well - might try that. But I thought this should be interesting for you if there is something to consider in Fastload Utils in some way.

halemmerich:
You probably need to restore the normal widget state during remove using `show` from `widget_utils`. If this is not done, slopeclock (or any clock for that matter) gets modified widgets it does not expect. Adding `swipeOn` to slopeclock would only cover up the problem.

thyttan:
Great, thanks - will try.

thyttan:
I think it makes sense to check if Fastload Utils is installed, because then we don't need the widgets at all. I changed to using read instead of readJSON.

Your tip regarding using widget_utils show in the remove-function helps, but one issue remains:

- If exiting Torch when the swipeOn widgets field isn't visible Slope Clock loads with widgets behaving as expected. 
- But if exiting when the swipeOn widget field is visible it will persist and show a white background when it should now be black (I use dark theme), and the Analog Clock Widget still show when it should not. Doing g.dump() from Web IDE gives this:
image
But on the watch the white widget field is there as described insted.

thyttan:
I landed on just loading widgets and hiding them. I think this is ready for merging.

gfwilliams:
>  But if using a clock without widgets, e.g. AI Clock, some widgets, e.g. Light Switch Widget, will interfere with the clock (and this would be alleviated by keeping the .hideinstead...).

You shouldn't ever be using fast load with a clock without widgets? The loading code should actually check this and if there are no widgets it will load the clock normally.

Is it an issue you were experiencing or just that you thought might happen?

gfwilliams:
This looks good though - merging...

gfwilliams:
> I guess AI Clock could be modified with a widget_utils .hide

Yes, it should ideally do that instead of `for (let wd of WIDGETS) {wd.draw=()=>{};wd.area="";}`, but the code that it there should work.

halemmerich:
> e.g. Light Switch Widget, will interfere with the clock (and this would be alleviated by keeping the `.hide`instead...).

I would suspect that the widget calls it's drawing code directly from some handler/event/timeout instead of calling the `draw` method. This would cause this problem when `.hide`ing the widgets, since that essentially only replaces the draw method of the widget. Changing those direct calls to drawing code to the widgets `draw` would help to catch those cases.

gfwilliams:
> Changing those direct calls to drawing code to the widgets draw would help to catch those cases.

Yes, that'd do it. Just looked at `lightswitch` and it's doing all kinds of stuff - I've just made some changes (see above) and that should fix it

PR #2336, opened 2022-12-01
ClockFace: support Fast Loading, by adding a remove() function to clocks

rigrig:
It also updates `setting.json` with the correct `clockHasWidgets`, because clocks relying on `ClockFace` don't call `loadWidgets` themselves.

### DIFF
--- a/modules/ClockFace.js
+++ b/modules/ClockFace.js
@@ -9,7 +9,7 @@ function ClockFace(options) {
     if (![
       "precision",
       "init", "draw", "update",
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! `clockHasWidgets` looks like a bit of a hack though :(

I'm not sure if it's better, but the easiest solution would just be if clocks were responsible for calling `Bangle.loadWidgets()` themselves. Calling it twice wouldn't actually matter as now it'll bail out immediately.

PR #2340, opened 2022-12-01
Weather clkinfo: weather condition with temperature

myxor:
Weather clkinfo: weather condition with temperature


gfwilliams:
Thanks!

PR #2341, opened 2022-12-01
Widget Sleepstatus: fix potential crash

myxor:
Crash happened when trying to edit this widget with the Widget Editor

### DIFF
--- a/apps/widsleepstatus/ChangeLog
+++ b/apps/widsleepstatus/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Load settings only once
       Better icons
       Read sleep status on every draw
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - although honestly `w` should be defined all the time, so if this fails in the widget editor, it's the editor that really needs fixing :)

gfwilliams:
... so I'm not sure we should really merge this - I'm sure there are a bunch of other widgets out there that will fail too. Best to fix where `draw` is called and ensure it gets called correctly

myxor:
I will have a look at the Widget Editor code.

PR #2342, opened 2022-12-01
[AI Clock] Change to using widget_utils .hide()

thyttan:
https://github.com/espruino/BangleApps/pull/2334#issuecomment-1333636754


gfwilliams:
Thanks!

PR #2345, opened 2022-12-02
[Multiple apps] autotranslate some strings

thyttan:
Same drill as last time, I haven't tested this.. Ok to merge by me.

### DIFF
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -16,10 +16,10 @@ var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
         {type: "txt", font: "12%", valign: -1, id: "tempUnit", label: "°C"},
       ]},
       {filly: 1},
-      {type: "txt", font: "6x8", pad: 2, halign: 1, label: "Humidity"},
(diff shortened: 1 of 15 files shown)


apps/weather/app.js
```diff
@@ -57,7 +57,7 @@ function draw() {
   layout.windUnit.label = wind[2] + " " + (current.wrose||'').toUpperCase();
   layout.cond.label = current.txt.charAt(0).toUpperCase()+(current.txt||'').slice(1);
   layout.loc.label = current.loc;
-  layout.updateTime.label = `${formatDuration(Date.now() - current.time)} ago`;
+  layout.updateTime.label = `${formatDuration(Date.now() - current.time)} ago`; // How to autotranslate this and similar?
```
gfwilliams, line 60: This should ideally use the `locale` module? If duration isn't handled then maybe we should add a new method to it for that... But right now with the tags you have minute/day/etc will get converted on

thyttan:
I dropped the commit regarding welcome app.

gfwilliams:
Thanks!

PR #2348, opened 2022-12-04
SlopeClock++ README update and debug feature for PrimeTimeLato

hughbarney:


apps/slopeclockpp/README.md
```diff
@@ -10,6 +10,8 @@ to change (top right or  bottom left). It should change color showing
 it is selected.
 
 * Swipe up or down to cycle through the info screens that can be displayed
+  when you have finnised tap again towards the centre of the screen to unselect.
```
myxor, line 13: typo. should be "finished".

gfwilliams: thanks - just tweaked

gfwilliams:
Thanks!

gfwilliams:
Just to say, if you update the README for things, don't worry about bumping the version/changelog - as there won't be any point in the user updating the app itself

hughbarney:
Thanks Gordon, makes sense.  Will do that in future.

PR #2349, opened 2022-12-04
[AIClock] Added clkinfos

peerdavid:
This PR adds clkinfo to the AIClock. As I prefer touch events rather then gestures (which I use for pattern launch) I implemented it classical and not via `addInteractive`.

I also updated some "base" icons which are currently used by BWClock and AIClock as the coloring was for some reasons not correct -- now it seems to work fine for both clocks.


peerdavid:
> >the coloring was for some reasons not correct -- now it seems to work fine for both clocks.
> 
> I suspect it had to do with [this (comment)](https://github.com/espruino/BangleApps/pull/2251#issuecomment-1321749135)?

Yea true I simply copied things from BW -- now it should be fine :)

gfwilliams:
Looks great, thanks for tweaking those icons too!

If you're finding that you're duplicating that clkinfo navigation code for your clocks then we could always add that to the library as well? I guess at some point we could look at making the interaction mode configurable

PR #2352, opened 2022-12-04
gpstrek - Workaround minifier issue

halemmerich:
I could not find an actual solution to this problem so this at least makes gpstrek usable with minification enabled.


gfwilliams:
Thanks!

PR #2356, opened 2022-12-05
sleepphasealarm

nxdefiant:
Support javascript command to execute as defined in scheduler 'js' configuration

### DIFF
--- a/apps/sleepphasealarm/ChangeLog
+++ b/apps/sleepphasealarm/ChangeLog
@@ -12,3 +12,4 @@
       Add setting to disable scheduler alarm
 0.10: Fix: Do not wake when falling asleep
 0.11: Minor tweaks
+0.12: Support javascript command to execute as defined in scheduler 'js' configuration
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! You might want a `if (nextAlarmConfig.js) { ... } else ...` though? Because I think if JS is supplied you're not meant to do anything else (like buzzing).

It's a shame there isn't a way to call into `sched` and use the code that's already in there

gfwilliams:
> I would like to avoid the else path. If you call 'load()' in the eval() the sleepphasealarm.buzz() is not executed anyway.

But that means that if someone installs your code, suddenly alarms with JS work differently if they don't call `load()`?

That just seems like a really bad idea. And when this breaks for someone - which it will - it'll be me that ends up having to try and debug it :(

nxdefiant:
ok, done

gfwilliams:
Thanks!

PR #2358, opened 2022-12-05
gallery: Do not invert image colors

nxdefiant:
or else depending on the theme settings black on white images end up as white on black.


gfwilliams:
Thanks - good spot!

PR #2363, opened 2022-12-06
[Remote for Spotify] remove only specific listeners

thyttan:
When fastloading Quick Launch listeners would be removed as well - this fixes that.


gfwilliams:
Thanks!

PR #2366, opened 2022-12-06
noteify: Remove duplicate alarm widget

nxdefiant:
Notify had a copy of the alarm widget. I see no reason for so I removed it.


lang/de_DE.json
```diff
@@ -20,7 +20,7 @@
         "On": "Ein",
         "Off": "Aus",
         "Ok": "OK",
-        "New Timer": "Neuer Kurzzeitwecker",
```
nxdefiant, line 23: While "Kurzzeitwecker" seems to be an
accurate translation I never heard this word
before. Everyone here understands "Timer".

gfwilliams:
Thanks! yes, that seems odd - maybe it was a copy/paste issue?

PR #2369, opened 2022-12-07
BangleJS2: Swipe down to rotate screen. So you can show the time to a friend real quick.

pebl-hank:
As in https://forum.espruino.com/conversations/382176


nxdefiant:
Note: This looks it might bite with quicklaunch: Both do handle swipe down.

pebl-hank:
I actually tried that out and found no issue.
However I could react only for the bottom portion of the screen modifing
`if (Math.abs(dy)>Math.abs(dx)+10) {`
when there are actually issues.

gfwilliams:
Neat - thanks!

PR #2373, opened 2022-12-07
messagegui: don't write messages to flash quite as often

rigrig:
* Fast load messages without writing them to flash  
    it checks `if (Bangle.load === load) // write to flash anyway` to prevent breakage on old firmwares   
    Still writes to flash if we decide not to load the app
* In-app: we already saved `MESSAGES` in `E.on("kill"`, so just remove all earlier calls to `saveMessages()`

### DIFF
--- a/apps/messagegui/ChangeLog
+++ b/apps/messagegui/ChangeLog
@@ -79,3 +79,5 @@
       Move widget to widmessage
 0.56: Fix handling of music messages
 0.57: Fix "unread Timeout" = off (previously defaulted to 60s)
+0.58: Fast load messages without writing to flash
+      Don't write messages to flash until the app closes
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -48,6 +48,11 @@ to the clock. */
 var unreadTimeout;
 /// List of all our messages
 var MESSAGES = require("messages").getMessages();
+if (Bangle.MESSAGES) {
+  // fast loading messages
+  Bangle.MESSAGES.forEach(m => require("messages").apply(m, MESSAGES));
+  delete Bangle.MESSAGES;
+}
 
 var onMessagesModified = function(type,msg) {
   if (msg.handled) return;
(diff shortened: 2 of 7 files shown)


gfwilliams:
Thanks! I hadn't really twigged that we don't even load the messages JSON from Storage any more unless it's a 't:"modify"' message - so this works *really* nicely.

Is this all tested and good to go?

digitalcircuit:
After upgrading from the stable Messages app to this version, most (all?) messages forwarded from the Bangle.js Gadgetbridge app to my Bangle.js 2 result in circlesclock being replaced by Message UI app opening up, only to say "No messages" and close.  I'm able to get messages working again by downgrading Messages (removing the split-out widget and UI apps) and downgrading the Android integration app.

If I should file a new issue on the forum, if there's any sort of debugging that would help, etc, let me know!

I have noticed occasionally seeing `FIFO_FULL` but that may be unrelated.

Apps: `antonclk`, `widbat`, `widbt`, `agenda`, `mylocation`, `agpsdata`, `gpsautotime`, `locale`, `widalarmeta`, `quicklaunch`, `multitimer`, `boot`, `launch`, `health`, `setting`, `about`, `widlock`, `messageicons`, `sched`, `alarm`, `circlesclock`, `weather`, `kbswipe`, `clkinfosunrise`, `owmweather`, `messages`, `android`
(This is after downgrading - there were separate message UI and message widget apps before.)

gfwilliams:
Thanks for letting us know! 

I'm just looking again at this code and it seems that while it'll work with a 'fast load' capable clock, if the clock can't fast load then it would fail. I believe I have now fixed this if you update messages to 0.59

> (I have noticed occasionally seeing FIFO_FULL, but that may be unrelated as it still happens even with the downgraded messaging apps.)

I think that's probably unrelated - it happens when Gadgetbridge sends data to the watch too fast for it to be processed, and in particular circlesclock is very bad as it can take over 1.5 seconds to redraw itself sometimes - see https://github.com/espruino/BangleApps/pull/2251#issuecomment-1342839671

gfwilliams:
Maybe a forum thread would be better - maybe you can find the OWM weather developer and mention them (or do it on GitHub if you can't find the maintainer there).

There *is* a PR for flow control in Gadgetbridge, but it hasn't been merged yet, so this will go away at that point. So it some ways you could consider the FIFO_FULL error fixed.

... but OWM weather might be making matters worse by requesting a whole bunch of data that it doesn't actually need. It's possible that by changing the HTTP request slightly it could get just the data it needs and not end up transferring a lot of stuff to the Bangle which causes the error

PR #2375, opened 2022-12-08
widmsggrid: provides a "message" widget

rigrig:
Looks like 3e5c0538a9776014a7f99d8c3521d1372dde8a90 missed this, I was wondering why the default widget kept getting reinstalled ;-)


gfwilliams:
Great - thanks!

PR #2376, opened 2022-12-08
Unlibrify message widgets

rigrig:
Now that `messages` can depend on widgets, we don't need fake `messagewidget` libraries to pull them in.

The widgets do always need to provide `WIDGETS["messages"]`, so the library can show/hide them, but we get rid of duplicate filenames, so installing/uninstalling in random order won't break things anymore.


gfwilliams:
Thanks!

PR #2377, opened 2022-12-08
Added Simplest++ and Lato Clocks, both support fast load and clock_info's

hughbarney:


gfwilliams:
Looks great - thanks!

PR #2379, opened 2022-12-08
sched - improve support for date timezone, toLocalISOString

lauzonhomeschool:
Test scenario that this patch fixes:
```
E.setTimeZone(-8);
date = new Date("2022-12-05T23:00");
date.toLocalISOString(); // -> "2022-12-05T23:00:00.000-0800"
getActiveAlarms([{on: true, t: 0, dow: 0b1111111, date: "2022-12-05"}], date); //should be marked as active, but isn't
getActiveAlarms([{on: true, t: 0, dow: 0b1111111, date: "2022-12-06"}], date); //should not be marked as active, but it is because it's already "tomorrow" in UTC
```

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -16,3 +16,4 @@
 0.13: Ask to delete a timer after stopping it
 0.14: Added clkinfo for alarms and timers
 0.15: Automatic translation of some string in clkinfo
+0.16: Improve support for date timezone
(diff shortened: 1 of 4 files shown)


gfwilliams:
Great - thanks! Please could you also bump the version and add a note to the ChangeLog so it'll get auto-updated for users?

lauzonhomeschool:
Good point, done.

gfwilliams:
Thanks! Merging now!

PR #2383, opened 2022-12-11
Simplest++ dropped the word clock from the app name

hughbarney:


hughbarney:
withdrawn, need to change title.

PR #2384, opened 2022-12-11
ClockFace: fix fast loading

rigrig:
I don't know why this is needed, but it was definitely broken for at least `barclock`: after fast loading a messages app, it would still draw over it once.

I tried adding a bunch of logging: it *does* call `clearTimeout` with the correct ID, but somehow the timeout fires anyway.   
No idea if it's a bug in `barclock`/`ClockFace` I missed, or some subtle problem with `clearTimeout` 🙁, but at least this fixes it.

### DIFF
--- a/modules/ClockFace.js
+++ b/modules/ClockFace.js
@@ -58,6 +58,7 @@ function ClockFace(options) {
 
 ClockFace.prototype.tick = function() {
   "ram"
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! That's an odd one though - I'm not sure why it'd be an issue

PR #2388, opened 2022-12-11
Added 'Firmware ClockInfo' and some teaks on screenshots and icons

hughbarney:


apps/simplestpp/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "simplestpp",
-  "name": "Simplest++ Clock",
```
gfwilliams, line 3: I think it might be better to use `shortName` here (which appears on the Bangle) - I'll do that now though.

gfwilliams:
Great - thanks!

PR #2389, opened 2022-12-11
astrocalc

nxdefiant:
- Compatibility with Bangle.js 2
- Get location from My Location

### DIFF
--- a/modules/suncalc.js
+++ b/modules/suncalc.js
@@ -279,7 +279,7 @@ function hoursLater(date, h) {
 // calculations for moon rise/set times are based on http://www.stargazing.net/kepler/moonrise.html article
 
 SunCalc.getMoonTimes = function (date, lat, lng, inUTC) {
-    var t = new Date(date);
+    var t = typeof(date) === "object" ? date : new Date(date);
     if (inUTC) t.setUTCHours(0, 0, 0, 0);
     else t.setHours(0, 0, 0, 0);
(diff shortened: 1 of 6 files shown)


modules/suncalc.js
```diff
@@ -279,7 +279,7 @@ function hoursLater(date, h) {
 // calculations for moon rise/set times are based on http://www.stargazing.net/kepler/moonrise.html article
 
 SunCalc.getMoonTimes = function (date, lat, lng, inUTC) {
-    var t = new Date(date);
+    var t = typeof(date) === "object" ? date : new Date(date);
```
nxdefiant, line 282: Espruino does not seem to support the
new Date(object) constructor

nxdefiant:
Notify @paulcockrell

paulcockrell:
@nxdefiant looks good but what is setting the GPS coordinates in the mylocation.json?

nxdefiant:
@paulcockrell You go to settings->Apps->My Location and set them there, see https://banglejs.com/apps/?id=mylocation

modules/suncalc.js
gfwilliams: Thanks for spotting that! I thought I'd tried to diff most of the clones of this file to see if there were differences, but I must have missed this one

gfwilliams:
Thanks for your work on this!

PR #2396, opened 2022-12-14
0.31: Tweaking the swipe option; Added mylocation as a dependency.

pebl-hank:


gfwilliams:
Thanks - just a note that I have removed the loadWidgets calls since they're not needed (you just need *one* at startup) and were causing warnings in the sanity check

pebl-hank:
Great, thanks!

PR #2397, opened 2022-12-14
circlesclock: fixed crash if item has no image and cutting long text

glemco:
Quick and dirty as my watch was crashing every time I was getting to the calendar clkinfo. In this case I guess the `addInteractive` could also populate the img field with the one from the menu item (still i think we ought to check it).
The text slicing is also preventing to draw outside the circle so that the text doesn't stay there while rotating, again this could be handled better splitting words, but I'd rather have clkinfos using the short field in this case (I'm planning to do it myself for the agenda, when I have some time).

### DIFF
--- a/apps/circlesclock/ChangeLog
+++ b/apps/circlesclock/ChangeLog
@@ -38,3 +38,4 @@
       Add fast load capability
 0.21: Remade all icons without a palette for dark theme
       Now re-adds widgets if they were hidden when fast-loading
+0.22: Fixed crash if item has no image and cutting long overflowing text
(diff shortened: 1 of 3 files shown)


glemco:
Sorry, my bad, wasn't aware of this one. I mean the agenda clkinfo. Basically the icon is set for the main agenda menu but left empty for events (which made sense for bwclock I guess). In this case it doesn't really make sense to set the same image to all events, does it?

gfwilliams:
Ahh, interesting. Well, in the docs we *don't* say it's optional - so really agenda should do something IMO as I bet it could break other stuff too.

I've just pushed a change to Agenda that puts the standard icon there.

> In this case it doesn't really make sense to set the same image to all events, does it?

In Circlesclock I think that makes a lot of sense? If you see a random item on the screen, it would be really helpful to have an icon to show you what it is. Same for SlopeClock++/etc.

It doesn't actually take up any more RAM either as we just reference the same icon we used before

gfwilliams:
But I guess a null check doesn't hurt anyway :)

glemco:
Sure, if the specs say it, better have it!

PR #2399, opened 2022-12-14
new widget (HW stats)

dapgo:
A basic HW/performance monitor widget that shows on real time some technical info, such as free mem, free storage, trash mem, files, FW version.


gfwilliams:
Thanks!

PR #2401, opened 2022-12-14
widmessages: Fix messages not showing if UI auto-open is disabled

### LINKED ISSUE #2390: [widmessages] App switch needed to show new messages if UI doesn't auto-open
### Affected hardware version

Bangle 2

### Your firmware version

2v16

### Describe the bug
When `Disable auto-open` (or disable in quiet mode with quiet mode active) is enabled for the Message UI, new messages won't be shown in the message widget until switching apps.

**However, the watch still vibrates** so messages are still being received and stored.

#### Additional bug

When in an app that is *not* fast-load enabled (e.g. the `Weather` app or `Settings` app), **new messages won't be received at all** regardless of whether `Disable auto-open` enabled.  This is likely a separate bug, which I'll file if desired (it is rather late for me to be awake).

### Expected behavior
When new messages are received, the message widget shows even if auto-opening is disabled without needing to switch apps.

### Steps to reproduce
1.  Using development Bangle.js App Loader…
     * Install the latest `messages` updates (v0.58), including the split `messagegui` (v0.60) and `widmessages` (v0.02)
2.  Check `Disable auto-open` for Messages so messages will not automatically display
     * I also disabled `Flash Icon`, but I don't think that matters
4.  Switch to the fast-load enabled clock (`antonclk`, `circlesclock`, etc)
     * Might apply to any other fast-loaded app?
5.  Send a test message via Bangle.js Gadgetbridge `Debug` menu or any other mechanism
6.  Feel for new message vibration
7.  Check for new message indicator
8.  Switch/launch apps
9.  Check for new message indicator

rigrig:
Fixes #2390

### DIFF
--- a/apps/widmessages/ChangeLog
+++ b/apps/widmessages/ChangeLog
@@ -1,3 +1,4 @@
 0.01: Moved messages widget into standalone widget app
 0.02: Fix 'srcs' being defined in global scope
       Remove library stub
+0.03: Fix messages not showing if UI auto-open is disabled
--- a/apps/widmessages/metadata.json
+++ b/apps/widmessages/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "widmessages",
   "name": "Message Widget",
-  "version": "0.02",
+  "version": "0.03",
   "description": "Widget showing new messages",
   "icon": "app.png",
   "type": "widget",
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks!

PR #2404, opened 2022-12-15
temp monitor app and correction on widgets

dapgo:
verified also on emulators


apps/widclkbttm/widclkbttm.wid.js
```diff
@@ -1,10 +1,12 @@
 WIDGETS["wdclkbttm"]={area:"br",width:Bangle.CLOCK?0:60,draw:function() {
   if (!Bangle.CLOCK == !this.width) { // if we're the wrong size for if we have a clock or not...
-    this.width = Bangle.CLOCK?0:60;
+    this.width = Bangle.CLOCK?0:60;      
     return setTimeout(Bangle.drawWidgets,1); // widget changed size - redraw
   }
   if (!this.width) return; // if size not right, return
-  g.reset().setFont("6x8", 2).setFontAlign(-1, 0).setColor("#0ff"); // cyan
+  if (process.env.BOARD=='BANGLEJS'||process.env.BOARD=='EMSCRIPTEN')  v_font_size=16
```
gfwilliams, line 7: It looks like v_font_size is defined as a global variable here? Please could you define it locally?

Also I'm a bit confused about this to be honest - both Bangles have 24px widget bars, and we're still only leaving 60px for the clock on both - so why do the sizes need to be different?

Also, I'm not sure if a 13pc vector font is really so readable compared to the 16px bitmap font used before?

apps/widhwbttm/widhwbttm.wid.js
```diff
@@ -1,7 +1,8 @@
 (function() {
+  let intervalRef = null;
     var v_switch; // show stats
-    if (process.env.BOARD=='BANGLEJS'||process.env.BOARD=='EMSCRIPTEN') v_font_size=16
-     else  v_font_size=13;
+    if (process.env.BOARD=='BANGLEJS'||process.env.BOARD=='EMSCRIPTEN') v_font_size=15
```
gfwilliams, line 4: It looks like v_font_size is defined as a global variable here? Please could you define it locally with `var`?

apps/widhwbttm/widhwbttm.wid.js
```diff
@@ -35,11 +36,12 @@
...
-  g.reset().setFontVector(v_font_size).setFontAlign(-1, 0).setColor("#0ff");
-  //clean a longer previous string
-  g.drawString("                        ", this.x, this.y+11, true);
+  g.reset().setFontVector(v_font_size).setFontAlign(-1, 0); 
+  //clean a longer previous string, care with br widgets
+  g.drawString("               ", this.x, this.y+11, true);
   g.drawString(v_str_hw, this.x, this.y+11, true);  
   } //end draw
 
 WIDGETS["wdhwbttm"]={area:"bl",width:Bangle.CLOCK?0:60,draw:draw};
```
gfwilliams, line 45: This looks wrong here actually - why is the width dependent on Bangle.CLOCK? And isn't the whole thing a lot wider than 60px anyway?

gfwilliams:
Thanks! The temperature monitor app looks great - but I'm a bit concerned about the clock mods (and I spotted an issue with the hardware info widget).

Also, you don't need the `.info` files in folders at all. They are auto-generated.

For the temperature monitor, you could even have an 'interface' file in the app loader to read the data: https://www.espruino.com/Bangle.js+Storage#reading-with-the-app-loader

apps/widhwbttm/widhwbttm.wid.js
dapgo: Sure, thanx

apps/widhwbttm/widhwbttm.wid.js
dapgo: Iong strings were the cause, but I'll trim and convert into shorter measure units. Mistakes from an amateur with hurry :|

dapgo: BTW, i just copied this widget parameters (width:Bangle.CLOCK?0:60,) from the v.004  of widclkbttm. I was assuming that it hides the widget when a clock is open

gfwilliams: Ahh, yes - but you also need to set the width to the actual width of the widget, or you'll end up overlapping with other widgets that want to draw in the same area ('bl'). It looks a lot more than 60 here - maybe 160 at least on Bangle.js 1?

... and if you're changing the font between Bangles you probably need to change that width too

gfwilliams:
Thanks for making those changes - just merging!

PR #2406, opened 2022-12-15
clkinfo: repeating selection if menu is empty

glemco:
Not sure if this has been discussed somewhere.
I can happen that some clock infos don't have elements (e.g. those with `dynamic` like alarms and calendar events, if the user simply doesn't have any). Swiping around in the `addInteractive` opens also those pointing to the first element (`undefined`) and crashes.
With this PR we are simply going on in the same swipe direction until we find a menu with at least an item (which is surely there as the module itself provides some items).
It looks to me more tidy than handling the case displaying an empty menu element.


gfwilliams:
Great! Thanks for tracking this one down!

PR #2408, opened 2022-12-17
Add Xiaomi Mijia temperature sensor display

dashavoo:
Adds an app to display the readings from Xiaomi Mijia temperature/humidity sensors.

This first release only supports the pvvx custom format for advertisements. Other formats (and perhaps other types of device) are planned for the future, as are temperature alerts.

The scan length is a bit lengthy - but I've found I catch more of my devices with a longer scan - I plan to make this configurable in future.


gfwilliams:
This looks great, thanks! And it's really good to use filters on findDevices too so the Bangle doesn't get overwhelmed by devices.

I know those Xiaomi sensors don't tend to advertise too quickly

PR #2410, opened 2022-12-18
[Hank's World Clock] Setting for single extra timezone small

rnist:
First, offtopic: Awesome work Gordon with the bangle! My garmin fenix 6 is sitting on my desk unused since I got the banglejs2 ! Awesome! 
Have a nice vacation!

Great work with updating this @pebl-hank ! I think in the future this setting could go in the initial pre-upload setup page, but this is my first time getting my fingers wet with the bangle and it was clear how to add it to the watch settings.

 
I don't want to step any boundaries with modifying someone else's watch face unless that's fine for the original author, and I can create a rworldclock, but I'd rather not. @pebl-hank ? :D 

This adds an entry in the settings menu that allows the user to show an extra single timezone small. Just like there would be more than 1 timezone. I need an extra timezone, and only one, but I do not need it that big. 

Happy holidays everyone!

### DIFF
--- a/apps/hworldclock/ChangeLog
+++ b/apps/hworldclock/ChangeLog
@@ -16,3 +16,4 @@
 0.30: BJS2: swipe seems to be working now
 0.31: Tweaking the swipe option; Added mylocation as a dependency.
       Remove calls to Bangle.loadWidgets as they are not needed and create warnings
+0.32: Added setting to show single timezone small, like where multiple ones are configured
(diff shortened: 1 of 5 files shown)


pebl-hank:
Actually i am more than happy if someone is contributing. I do not like forks that much because it will confuse the user when there are many versions of an app or watchface.
I can't test your pr right now but if this is just another optional setting i think this would be not a problem: )

gfwilliams:
Thanks! This looks like a really neat solution! Merging now...

It makes me think that maybe extra timezones in ClockInfo would be pretty neat as well.

PR #2412, opened 2022-12-18
[WIP][TerminalClock] Adding fast loading

Stiralbios:
Not to merge right now cause it currently have a big memory leak.


gfwilliams:
Does it still have a memory leak? If so and there's not work on this then maybe we should close it

Stiralbios:
Yep still a memory leak. I am quite busy lately but I think I will take another look in two weeks. If I don't find where it's come from I will ask some help from the community

PR #2417, opened 2022-12-20
add message input and dated Events

lauzonhomeschool:
1. Users can now create Events (triggered based on time and date, vs alarms, which are triggered based on time and day of week).  In this first version, there is no "Repeat" option for Events (e.g. "20th of each month"), but I hope to add that in a future version.
2. Users can now input a message to be shown for alarms, timers and dated events.  The message input is optional: I don't add a dependency to textinput, but juste remove the Message entry if it isn't detected.

Eventually, if we want to allow people to download just the functionalities they want from this "default" alarm app (e.g. if they don't need dated Events and want to spare resources), I could use the "custom" and "content" options of metadata.json to read and customize app.js based on user's choice.
e.g. tag the source code with comments such as the following, and remove everything in between if the user does not want a feature:
```
/*Option1->*/
...
/*<-Option1*/
```

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -37,3 +37,4 @@
 0.34: Add "Confirm" option to alarm/timer edit menus
 0.35: Add automatic translation of more strings
 0.36: alarm widget moved out of app
+0.37: add message input and dated Events
--- a/apps/alarm/README.md
+++ b/apps/alarm/README.md
@@ -1,15 +1,18 @@
 # Alarms & Timers
 
-This app allows you to add/modify any alarms and timers.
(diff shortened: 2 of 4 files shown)


gfwilliams:
Great, thanks! 

> e.g. tag the source code with comments such as the following, and remove everything in between if the user does not want a feature:

I think for now it's best to leave that - I feel like the proportion of users that would ever do that is pretty small, and the actual space/speed saving may not be that significant.

... but there is a more general problem of users forking apps and making a tiny change, which makes it really difficult to maintain things - so it was suggested that something like 'patches' could be added - but again I do wonder how many people would actually contribute changes that way, as it's just more complexity

PR #2418, opened 2022-12-20
Setting circles colours per clkinfo in circlesclock and improved agenda

glemco:
The agenda clkinfo was still missing the `dynamic` field (which is needed for the circles clock colour to work properly), I also added a range and `short` fields.

Now circles in circlesclock are coloured based on the information inside (as set by the user), the next step could probably be to have _suggested_ colours by the clkinfo itself. This could be useful for the agenda (use the colour of the calendar the displayed event belongs to) or the weather condition (e.g. yellow for sunny, cyan for rainy, etc.). Maybe gradients could be implemented directly like this, rather than inside circlesclock.
Would it be useful? At the moment I think only circlesclock actually gives colours to infos.

### DIFF
--- a/apps/agenda/ChangeLog
+++ b/apps/agenda/ChangeLog
@@ -9,3 +9,4 @@
       Fix clkinfo icon
 0.09: Ensure Agenda supplies an image for clkinfo items
 0.10: Update clock_info to avoid a redraw
+0.11: Added dynamic, short and range fields to clkinfo
--- a/apps/agenda/agenda.clkinfo.js
+++ b/apps/agenda/agenda.clkinfo.js
@@ -1,7 +1,14 @@
 (function() {
+  function getPassedSec(date) {
+    var now = new Date();
+    var passed = (now-date)/1000;
+    if(passed<0) return 0;
+    return passed;
+  }
   var agendaItems = {
       name: "Agenda",
       img: atob("GBiBAAAAAAAAAADGMA///w///wf//wAAAA///w///w///w///x///h///h///j///D///X//+f//8wAABwAADw///w///wf//gAAAA=="),
+      dynamic: true,
       items: []
     };
   var locale = require("locale");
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks!

Yes, clock_info could suggest colour - seems like a good idea. I guess the only issue is if someone wants a different one, but I guess in that case it could just be overwritten?

Did you ever see http://www.espruino.com/Reference#l_Graphics_blendColor for choosing a colour? I seem to recall circlesclock changed colours directly, and that function may make things a lot easier!

glemco:
> Yes, clock_info could suggest colour - seems like a good idea. I guess the only issue is if someone wants a different one, but I guess in that case it could just be overwritten?

Sure, I thought that circlesclock should use the default color (passed by the info) only if the user didn't set any other before, however this would mean that if a user selects a colour now, after the update they would need to unset it manually to use the colour provided by the info, a little tricky since user may not know that..

> Did you ever see http://www.espruino.com/Reference#l_Graphics_blendColor for choosing a colour? I seem to recall circlesclock changed colours directly, and that function may make things a lot easier!

Yeah it looks pretty much what the clock is trying to do manually, it would definitely help! By the way, is it a new function? It doesn't seem to work on my IDE via emulator (I cannot really test things comfortably on the watch as just my phone supports BLE).

glemco:
> Do you know about the [Web IDE relay feature](https://forum.espruino.com/conversations/382580/#16810030)?

Wow. Quite brilliant!

gfwilliams:
> By the way, is it a new function? It doesn't seem to work on my IDE via emulator

Thanks - good spot! I think it's been in there a while, but it looks like it wasn't included in the Bangle.js2 emulator for some reason. It's in there for Bangle.js 1 - I've just fixed that but we'll have to wait for the next firmware version before it hits the emulator

PR #2420, opened 2022-12-20
New Clock: Space Resist

TimWue:
Add space clock watch face for banglejs2, which is inspired by a casio watch.


gfwilliams:
Looks great - thank you!

Issue #2361, opened 2022-12-05
[BW Clock]  hangs and locks when sunrise_clockinfo is installed.


hughbarney:
### Affected hardware version

Bangle 2

### Your firmware version

2.16.7

### The bug

CONFIGURATION:
Bangle 2.
Firmware 2.16.7
BW clock 0.23
Slopeclock++ 0.5
Sunrise Info Clock 0.1
Latest everything else.

Steps to reproduce

install slopclock++ and observe that sunrise and sunset options are not displated in the clock info menu's
install BW clock and scroll through the Bangle menu without issue
install sunrise_clockinfo and observe sunrise and sunset options now in the clock info menus.
start up BW clock and select the Bangle menu. At some point it will hang.
You will see the following on the IDE console:

Uncaught Error: Unhandled promise rejection: undefined
in function "altUpdateHandler" called from line 1 col 2110 in bwclk.app.js
...;alt="--";altUpdateHandler();
                              ^
in function "show" called from line 29 col 190 in bwclk.app.js
...ings.menuPosY-1];item.show();
                              ^
in function "drawMenuAndTime" called from line 38 col 509 in bwclk.app.js
....length+1);drawMenuAndTime();}if(is_upper){if(e.y<widget_size)...
                                ^
in function called from system
>
You will not see a sunrise or sunset option in the Bangle Menu.
Thereafter a black bottom half of the screen is displayed whenever you load BW clock.

peerdavid:
I wonder if this clock_info works in slopeclock++ when installed? seemsike a bug in sunrise to me...

Still BW clock should not hang I will lokk into this :)

gfwilliams:
I just checked and this works fine in slopeclock++

**I can also reproduce this exact error in BW Clock without the sunrise widget** - just right after starting the clock, keep tapping really quickly, and then after a few taps it stops updating.

In the stack trace posted initially it sounds like the error is from altUpdateHandler - and I wonder whether the issue is actually that when flicking through items quickly, altUpdateHandler calls the 'redraw' event a second or so late but the redraw handler hasn't been removed even though the altitude has since been hidden? 

Interestingly I also get a different error in BW Clock sometimes:

```
Uncaught Error: clearInterval(undefined) not allowed. Use clearInterval() instead.
 at line 1 col 2449 in bwclk.app.js
clearInterval(this.interval),delete this.interval
                                    ^
in function "hide" called from line 22 col 81 in bwclk.app.js
[...]

PR #2421, opened 2022-12-21
[BW Clock] Add fastloading

thyttan:
RAM leaks gone thanks to @BartS23, thanks!

Bug #2361 will not be fixed with this PR.

### DIFF
--- a/apps/bwclk/ChangeLog
+++ b/apps/bwclk/ChangeLog
@@ -22,3 +22,4 @@
 0.22: Use the new clkinfo module for the menu.
 0.23: Feedback of apps after run is now optional and decided by the corresponding clkinfo.
 0.24: Update clock_info to avoid a redraw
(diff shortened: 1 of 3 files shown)


thyttan:
Hmm, I've tried to get this working but can't seem to do do it. If anyone has an idea what I might be missing please say :)

BartS23:
@thyttan it is the Kill handler.

If you replace Line 445-451 
```
E.on("kill", function(){
  try{
    storage.write(SETTINGS_FILE, settings);
  } catch(ex){
    // If this fails, we still kill the app...
  }
});
```

with 
```
let save = () =>  storage.write(SETTINGS_FILE, settings);
E.on("kill", save);
```

and then add in the remove method:
```
      // save settings
      save();
      E.removeListener("kill", save);
```

thyttan:
Thank you @BartS23! That did it :)

gfwilliams:
Looks great - thanks!

PR #2422, opened 2022-12-21
new clock app and minor updates

dapgo:
improving device compatibility


apps/widclkbttm/widclkbttm.wid.js
```diff
@@ -1,6 +1,11 @@
 WIDGETS["wdclkbttm"]={area:"br",width:Bangle.CLOCK?0:60,draw:function() {
+  console.log("clock? "+Bangle.CLOCK);
```
gfwilliams, line 2: Was this intended to be committed?

dapgo: no idea, the condition was no written by me but just copied  from other widget clock (apps\widclk) code. It seems useless, and specially when there a return in case of CLOCK.
The testing consolelog should not be there.. :|

gfwilliams:
Thanks - but there is still an extra `console.log`, and I don't see what `if (Bangle.CLOCK){` fixes? There's no version bump or changelog...

As far as I can see of CLOCK=1 width is set to 0 and it returns anyway, so there is no need for this?

I'll merge this but just revert the other changes

dapgo:
Sorry I did a very quick action without paying attention to any other line of code... I just remove the other console log and some useless lines

PR #2423, opened 2022-12-21
messagesdebug: new app, log all `message` events to storage file

rigrig:


gfwilliams:
Neat - thanks! Just wondering - would it make sense just to add writing to a file to https://github.com/espruino/BangleApps/tree/master/apps/gbdebug or were you looking at using this for iPhone?

rigrig:
I thought about logging from `gbdebug`, but I figured this way we get the exact messages that are emitted. (and some vague idea about maybe in the future we'll add apps that emit their own messages)

PR #2424, opened 2022-12-25
[clock_info] Catch error when Bangle.getPressure() returns...

thyttan:
... undefined. This makes it so BW Clock doesn't hang when running in the emulator and is trying to display altitude.


gfwilliams:
Thanks - Is this specifically for the emulator? It doesn't seem like getPressure would throw an exception normally

thyttan:
Yes, I only encountered it in the emulator.

PR #2425, opened 2022-12-25
[entonclk] Add timer function

siatwe:
For my watchface I urgently needed a timer function. In the past, I have used the built-in timer or 'a_clock_timer' from @alainsaas for this. I looked at the code for it and just unceremoniously added it to my watchface. I left the increase over the right touch. The decrease over the left touch, however, I have reduced to 1 minute. So it is also possible to give more precise information like 3, 7 or 12 minutes. Thanks for the code from @alainsaas. So it was very easy for me.

### DIFF
--- a/apps/entonclk/ChangeLog
+++ b/apps/entonclk/ChangeLog
@@ -1,1 +1,2 @@
 0.1: New App!
+0.2: Now with timer function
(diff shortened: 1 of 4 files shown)


alainsaas:
Thanks for the mention, glad to hear my code helped :)

gfwilliams:
Thanks! I guess a timer could be a good candidate for a clock_info module (https://github.com/espruino/BangleApps/issues/2226) so that it could be used by any participating clock faces as well

PR #2428, opened 2022-12-27
[dragboard] add settings screen for colors and caps lock

lauzonhomeschool:
Allow users to make characters more visible to them (by choosing different colors, or forcing caps lock / all uppercase).


thyttan:
If I read the code correctly there's no way to reset the color to follow the system theme? Could you add an option for resetting defaults?

lauzonhomeschool:
I added a "Default" color.  Thanks for Dragboard, it's the fastest way to input text.

gfwilliams:
Thanks! Maybe it makes sense to just store the color string like `#0f0` in settings rather than the numeric value of the hex bit? It might be more readable I guess.

PR #2429, opened 2022-12-28
Option to disable quiet mode after alarm/timer dismiss

Kedlub:
This adds an option to alarm and sched that automatically disables quiet mode after stopping an alarm or a timer
Meant to be used to automatically disable quiet mode after waking up

### DIFF
--- a/apps/alarm/metadata.json
+++ b/apps/alarm/metadata.json
@@ -2,7 +2,7 @@
   "id": "alarm",
   "name": "Alarms & Timers",
   "shortName": "Alarms",
-  "version": "0.37",
+  "version": "0.38",
   "description": "Set alarms and timers on your Bangle",
   "icon": "app.png",
   "tags": "tool,alarm",


Kedlub:
Rebased on the latest commit

gfwilliams:
Hi,

Sorry for the delay sorting this, but I'm not entirely sure about this one. Does anyone else have any thoughts about it?

It feels like adding yet an extra option to the list of things in the alarm list is going to end up being confusing for most people (does it really need to be there for timers?), and `qmsched` is capable of scheduling when to turn quiet mode on and off anyway?

So what's the main use for this? Is it because you didn't want to have to set it up via the `qmsched` app, or is the idea that you would have quiet mode turn off at different times on different days?

Kedlub:
Yeah, I thought that adding it directly to the `alarm` isn't the greatest idea, I originally wanted to put it into an external app, but right now there is no event that would run code after an alarm. Maybe there could be some kind of a submenu with checkboxes where there would be different kinds of functions that could run after an alarm.

> does it really need to be there for timers?

No, it doesn't need to be there for timers, that was just an experiment

> So what's the main use for this?

Yes, it's for turning off quiet mode at different times on different days because `qmsched` can't do that right now, and also so you don't have to change the schedule every time you change an alarm.
Maybe an alternative way would be to sync alarm times into `qmsched` directly? Because even if `qmsched` allowed making a schedule that only runs on a specific day, it would be extra work having to set the alarm and then manually set the schedule in `qmsched`.

gfwilliams:
Sorry about the delay - hopefully this is all fixed by https://github.com/espruino/BangleApps/pull/2510 now

PR #2430, opened 2022-12-28
[Clock Info] Revert "Update clock_info to avoid a redraw/general tidyups"

thyttan:
This reverts commit 8425f0b50aea39a9083b95ccc9f4211d9c5d556a, fixing bugs that appeared in BW Clock.

As per https://github.com/espruino/BangleApps/issues/2361#issuecomment-1366693191.

Maybe this should be fixed in some other way, then just disregard this PR.

### DIFF
--- a/apps/agenda/ChangeLog
+++ b/apps/agenda/ChangeLog
@@ -8,4 +8,3 @@
 0.08: Fix error in clkinfo (didn't require Storage & locale)
       Fix clkinfo icon
 0.09: Ensure Agenda supplies an image for clkinfo items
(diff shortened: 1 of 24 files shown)


nxdefiant:
Possible conflict with #2429

peerdavid:
I pushed a fix that works without reverting -- I now use .get instead of .show as .show is not always implemented. Fixed AI Clock and Linux clock too.

peerdavid:
So IMO (and if confirmed) this PR can be closed?

PR #2433, opened 2022-12-29
Add info about git submodules

siatwe:
I'm not sure about this. But I had a hard time to get the project running locally on my computer. I realized that I have to initialize and update the git submodules first. If this is already documented somewhere and I missed it, then you can delete the PR I think. Or is the approach to get the project running locally different?


gfwilliams:
Thanks! Yes, that's a good idea - submodules can be a pain, and what you're doing there looks good.

I guess in a way this could really be somewhere more prominent as well - although I guess in https://www.espruino.com/Bangle.js+App+Loader we only deal with using Github's website, not the CLI.

PR #2434, opened 2022-12-29
Fix mapping of Gadegtbridge GPS data and adjust event name

LukasEdl:
Hi everyone i found some bugs in the handling of the gadgetbridge gps data.

**The current state**

If we retrieve the gps data from the Gadegtbridge app, we receive a JSON object with the different values. The object has a key "long" which contains the longitude gps value. The received object will then be published under the "gps" event on the global Bangle object.

**Changes in this PR**

To prevent breaking of code if you start using the gps data provided by Gadgetbridge, i adjusted the event name from "gps" to "GPS" and added a mapper to map the longitude value from "long" to "lon" so that it is similar to a gps object, which you wouuld receive if you use the normal gps chip of the banglejs device.

### DIFF
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -18,3 +18,4 @@
 0.18: Use new message library
       If connected to Gadgetbridge, allow GPS forwarding from phone (Gadgetbridge code still not merged)
 0.19: Add automatic translation for a couple of strings.
+0.20: Fix wrong event used for forwarded GPS data from Gadgetbridge and add mapper to map longitude value correctly.
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -134,6 +134,9 @@
         event.satellites = NaN;
         event.course = NaN;
         event.fix = 1;
+        event.lon = event.long;
+        delete event.long;
+
         Bangle.emit('gps', event);
       },
       "is_gps_active": function() {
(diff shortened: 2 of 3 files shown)


thyttan:
You should probably update version number in metadata.json and ChangeLog :)

LukasEdl:
I updated the version number and added the changes to the change log.

halemmerich:
Why map the long/lon value on the bangle instead of just changing the event send by GB? The code was only included/usable in nightlies so far, I think we could just correct this at the source and be done with it.

gfwilliams:
Thanks! I guess ideally Gadgetbridge would be changed to send `lon` not `long` but we have to modify this anyway to send the correct event type, so for now I think this is fine.

My concern is I wonder how well this whole thing was tested now ;)

halemmerich:
Maybe it would make sense to hold off on a new GB release until this has been fixed. The current code would remove the correct value the moment GB would be changed to send `lon` as key by overwriting it by the no longer existing value of `long`.

I have tested a bit with manually doing the changes of this PR. It seems to work OK, turning GPS on at the Bangle instantly shows the location indicator on my phone and I am getting GPS events on the watch. The main problem I currently see is the phone showing a notification for sending GPS to two devices and it not switching off when the bangle stops using GPS. Maybe we need a timeout after a few seconds in boot code that checks if `_PWR` contains GPS and sends the power off message
to GB otherwise?

PR #2440, opened 2023-01-02
[BW Clock Lite] New version using bitmap font making it quicker

thyttan:
@peerdavid BW Clock loads quicker if using bitmap fonts, unsuprisingly :p I'm thinking I could maybe add [app customization](https://www.espruino.com/Bangle.js+Custom) so the user can choose between "Fancier Looks" or "Faster Loading". What do you think?

The changes as they are now is what I have tested performance with. It is not intended for merging as is.


peerdavid:
Oh wow thats great. Is there a huge difference how it looks when "fast loading"? If not I would guess fast loading should be the default right?

thyttan:
Here's a graphics dump with bitmap font:
download

thyttan:
Here's with vector font:
download (1)

peerdavid:
Yea it looks quite different, I honestly liked the original font -- in this case I prefer the "slow" solution which we should still provide

thyttan:
Yes, I think that would work well using app customization via the app loader as mentioned in the opening comment :)

I don't know how that's done yet, but I can try to learn if you don't particularly feel you want to implement it? ;)

gfwilliams:
> BW Clock loads quicker if using bitmap fonts, unsuprisingly :p 

Do you have any numbers?

Just thinking aloud here - if all the clock_info code apart from rendering ends up as part of the clock_info library, it feels like there's not actually *that* much code in the clock itself - and you'd likely have to change all the positioning for different fonts anyway.

It might be worth moving all the complicated stuff out into utility libraries so it's just 90% rendering code and then just forking the app?

Another option is to look at loading fonts from Storage. It could be done quite easily, and while rendering might be a smidge slower, startup time would be a lot better.

thyttan:
>Do you have any numbers?

Not more than counting in my head, shy of maybe 0.7 seconds vs shy of 0.5 seconds. I could record a video for comparison. I don't know how to do timings programmatically.

>It might be worth moving all the complicated stuff out into utility libraries so it's just 90% rendering code and then just forking the app?

I don't understand precisely what this means, so I don't have an opinion on it, maybe @peerdavid has one?

>Another option is to look at loading fonts from Storage. It could be done quite easily, and while rendering might be a smidge slower, startup time would be a lot better.

That sounds like it could be sensible to me.

gfwilliams:
> Not more than counting in my head,

I'd be very vary of making quite big sweeping changes without any numbers to back them up. Just add `t=getTime()` at the start of the app, and `print(getTime()-t)` at the end - and you'll get the time in seconds it took to load the app.

> I don't understand precisely what this means,

I just mean that if there's very little code in the clock app itself (so it's just ~100 lines of code that is rendering the clock face) then potentially it just makes sense to create a copy that doesn't use fonts, since almost everything in the clock would change anyway in order to tweak the font positions.

thyttan:
Timings based on PR #2471:
Regular load(): 1.58590698242 s
Fast loading:    0.91567993164 s

Timings when using bitmap font in this PR:
Regular load():  1.22711181640 s
Fast loading:     0.55581665039 s

gfwilliams:
Do you think this is ready to merge?

It's still marked as draft but as it's a separate app I'd think it's fine to pull in?

thyttan:
I've been meaning to tweak font sizes and positions, but will probably not get around to it soon.

I've used it for a while myself.

So sure, go ahead 👍

gfwilliams:
Ok, thanks!

PR #2441, opened 2023-01-03
Health: Show step goal in daily step chart

myxor:
- Show step goal in daily step chart
- Documentation fixes

Screenshot having 8000 as daily step goal:

image


gfwilliams:
Thanks! That's a great idea!

PR #2442, opened 2023-01-03
Health: Step goal reached notification

myxor:
**Open for discussion :)**


apps/health/boot.js
```diff
@@ -92,3 +96,22 @@ Bangle.on("health", health => {
...
 });
+
+function handleStepGoalNotification() {
+  var settings = require("Storage").readJSON("health.json",1)||{};
+  const steps = Bangle.getHealthStatus("day").steps;
+  if (settings.stepGoalNotification && settings.stepGoal > 0 && steps >= settings.stepGoal) {
+    const now = new Date(Date.now()).toISOString().split('T')[0]; // yyyy-mm-dd
+    if (!settings.stepGoalNotificationDate || settings.stepGoalNotificationDate < now) { // notification not yet shown today?
+      Bangle.buzz(200, 0.5);
+      /*LANG*/E.showPrompt("You reached your daily goal of " + settings.stepGoal + " steps!", {
```
thyttan, line 107: The /\*LANG\*/ tag should be just before the two strings inside E.showPrompt I believe.

apps/health/boot.js
```diff
@@ -92,3 +96,22 @@ Bangle.on("health", health => {
...
+        /*LANG*/title: "Step goal reached!",
```
thyttan, line 108: Similarly I think the /\*LANG\*/ tag should be just before the string and not before title, or will this work?

thyttan:
For Bangle 2 I think using a [LCD overlay](http://www.espruino.com/Reference#l_Bangle_setLCDOverlay) ([forum post](https://forum.espruino.com/conversations/378282/)) could maybe make sense. That way whatever else was going on on the watch will not be terminated.

But if implemented there will have to be a fallback to showPrompt for Bangle 1.

Nice one, I would probably enable this on my watch once it's done :)

myxor:
> For Bangle 2 I think using a [LCD overlay](http://www.espruino.com/Reference#l_Bangle_setLCDOverlay) ([forum post](https://forum.espruino.com/conversations/378282/)) could maybe make sense. That way whatever else was going on on the watch will not be terminated.
> 
> But if implemented there will have to be a fallback to showPrompt for Bangle 1.
> 
> Nice one, I would probably enable this on my watch once it's done :)

I tried to use LCDOverlay for Bangle.js 2 and it looks like this now:

image

What do you say?

thyttan:
Nice, I'd use that 😉

myxor:
If Gordon approves this as well i will add ChangeLog and metadata changes so it can be merged after the other open health related MRs.

gfwilliams:
I would argue that maybe we should add a `notify` library for Bangle.js 2 overlay notifications and make it the default.

This is the kind of thing it was intended for, and it should then work 'out of the box' on Bangle.js 1 as well? Right now I think the only implementation is the 'fullscreen' one for Bangle.js 2, so it would work it's just not ideal

myxor:
> I would argue that maybe we should add a `notify` library for Bangle.js 2 overlay notifications and make it the default. 

And with library you mean a module like clkinfo and the others?

> This is the kind of thing it was intended for, and it should then work 'out of the box' on Bangle.js 1 as well? Right now I think the only implementation is the 'fullscreen' one for Bangle.js 2, so it would work it's just not ideal

Which fullscreen implementation are you talking about?

gfwilliams:
Using this: https://github.com/espruino/BangleApps/tree/master/apps/notify

There's a readme there with some info in. I could tweak it so that it'll use overlays on Bangle.js 2, and then it feels like it could be a good solution?

myxor:
Sounds like a good solution to me.

@gfwilliams Can you update the notify app to support the Bangle.js 2?

gfwilliams:
Just done - be8b402dc79c631db2f430819c141ed991d85d8a

myxor:
Thanks @gfwilliams, i just pushed my changes. It now uses the `notify` library for B1 and B2 and i added the lib as dependency. 
For me this is ready to be merged now.

gfwilliams:
Great - thanks!

PR #2443, opened 2023-01-03
Health: View monthly data per day via Web interface in table and bar charts

myxor:
I see this as a start into showing the health data in a nice way via the Web interface later on.

**Screenshots:**

image

image

image


myxor:
I am not quite sure why the pipeline complains about the Chart.js file

gfwilliams:
This looks excellent - thank you!

PR #2448, opened 2023-01-04
Agenda: Setting to use "Today" and "Yesterday" instead of dates

myxor:
Setting to use "Today" and "Yesterday" instead of dates


gfwilliams:
Thanks!

PR #2449, opened 2023-01-04
[BWClock, AIClock and LinuxClock] Fix instability by using .get() instead of .show()

peerdavid:
It seems that .show is not always implemented. As quick-fix I use .get() instead which works fine currently. Later on I plan using `addInteractive` -- at least for BWClock as well as AI clock this should not be a problem :)

This should also fix your report @gfwilliams:
```I can also reproduce this exact error in BW Clock without the sunrise widget - just right after starting the clock, keep tapping really quickly, and then after a few taps it stops updating.```

### DIFF
--- a/apps/aiclock/ChangeLog
+++ b/apps/aiclock/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Indicate battery level through line occurrence.
 0.04: Use widget_utils module.
 0.05: Support for clkinfo.
+0.06: ClockInfo Fix: Use .get instead of .show as .show is not implemented for weather etc.
--- a/apps/aiclock/aiclock.app.js
+++ b/apps/aiclock/aiclock.app.js
@@ -29,7 +29,6 @@ var H = g.getHeight();
 var cx = W/2;
 var cy = H/2;
 var drawTimeout;
-var lock_input = false;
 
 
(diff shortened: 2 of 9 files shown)


peerdavid:
I hope that those also fixes #2361

gfwilliams:
Ok, thanks - although I think while this may help some things, it's not going to work with a lot of stuff.

`.show` and `.hide` are needed so clock_infos know when to start/stop the relevant sensors. For instance you would only want the barometer running when you're displaying the barometer clock_info.

If you were previously calling .show but never .hide then there would have been a big memory leak anyway so this is probably an improvement on that, but things like the altitude/BPM are probably unlikely to work properly now?

I think maybe we should add an option to `addInteractive` to choose the *method* of interactivity - and then you can implement the navigation method you have for BW Clock as an option for that?

gfwilliams:
>  but .show was not always implemented as I expected which lead to this wrong behaviour...

Do you remember which clock_info it was? I thought they were all supposed to have that, and it could break addInteractive too.

> Later we can let users decide but then I would add settings to clk_info or so?

Sounds great - thanks! I think at some point soon I'll split clock_info out into its own app and make all the clocks depend on it, and then we can do things like having a settings screen

PR #2450, opened 2023-01-04
Informational Clock delete emulator functionality because it does not work

Ronin0000:
when opened in the emulator the app gives several errors so I deleted the ability to open the app in the emulator. I also added a dependency to the weather app because without it the app does not work.


gfwilliams:
Great - thanks!

PR #2451, opened 2023-01-05
gipy: new release

wagnerf42:
hi, a new release for gipy with a lots of features:

	* When lost indicates nearest point on path.
	* Rescale display if lost and too far.
	* New setting to hide points and increase display speed.
	* Speed optimisations.
	* Estimated time of Arrival/Going back.
	* Display current and next segment in red so that you know where to go.
	* Avoid angles flickering at low speed at the cost of less refresh.
	* Splash screen while waiting for gps signal.


gfwilliams:
Thanks!

PR #2453, opened 2023-01-05
[Desktop Launcher] widget field follow correct theme immediately

thyttan:
When fast loading from BW Clock to Desktop Launcher the widget field would keep the inverted theme background then quickly resetting to follow non-inverted theme. 

With this PR this is no longer the case - I have inner peace.

### DIFF
--- a/apps/dtlaunch/ChangeLog
+++ b/apps/dtlaunch/ChangeLog
@@ -23,4 +23,6 @@ button to exit is no longer an option.
 facilitate 'fast switching' of apps where available.
 0.20: Bangle 2: Revert use of Bangle.load() to classic load() calls since
 widgets would still be loaded when they weren't supposed to.
+0.21: Bangle 2: Call Bangle.drawWidgets() early on so that the widget field
+immediately follows the correct theme.
 
--- a/apps/dtlaunch/app-b2.js
+++ b/apps/dtlaunch/app-b2.js
@@ -84,6 +84,7 @@
     g.flip();
   };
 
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks! 

So just so I understand though, 'BW Clock' inverts the theme, and this then messes up desktop launcher?

Does this not mess up other launchers too? Or is it just that desktop launcher draws the widgets a bit later?

If other launchers get messed up too, maybe we should modify BW Clock and put a redraw in there?

thyttan:
Mess up is maybe an overstatement, but first the appRect will be updated with a black background while the widget field is still white, and the next instant the widget field will fill with black instead. I want all background to change color at the same time.

>Does this not mess up other launchers too? Or is it just that desktop launcher draws the widgets a bit later?

I have not checked this.

>If other launchers get messed up too, maybe we should modify BW Clock and put a redraw in there?

I tried this with my setup, adding Bangle.drawWidgets() at the end of BW Clocks `remove` function. But then what would happen is 
1. The BW Clock is active with its GUI.
2. Click button initialize switch to launcher
3. The widget field background is painted black while on top of BW Clock graphics
4. Desktop Launcher is opened.

Step 3 above is for me not good either, it's basically the same problem in the opposite direction. I ideally want there to be only one visible update of the screen to get that instantaneous feel.

No hurry merging this if we want to think of another way to do it.

gfwilliams:
Thanks - no, that's fine - I just didn't want to have to put fixes in for everything if it was the clock that was the issue :)

PR #2454, opened 2023-01-05
grocery: sort, interface.html

nxdefiant:
Changes:
- Sort checked items to bottom as I have seen that behavior on most or all shopping/todo apps.
- Enable Widgets, the back button takes the space anyway.
- Allow to edit items without reuploading entire app with interface.html which is a modified grocery.html.


gfwilliams:
Thanks!

PR #2455, opened 2023-01-05
[BWClock, AIClock] Use clock_info.addInteractive instead of a custom implementation

peerdavid:
With this PR I changed the behavior as I use `.addInteractive` -- and I (personally) really like it. It also simplifies the code a lot and the code basically contains only rendering things.

Here a screenshot how it looks like when the clk_info is selected:
download

download

Still, 2 things are missing but I think we can add those after merging this PR:
1. Its no more possible to trigger an action (.run) if clicked -- but I think this should be implemented generically in the `addInteractive` fun of clock_info such that this is available for all clocks
2. Before it was possible to hide clock_infos completely and show a larger time instead -- I'm not 100% sure how I should implement this now but I guess it could also be an option of `addInteractive` e.g. hide if pos 0 or so?

Again, thanks @gfwilliams -- its so much fun coding on my bangle -- and your clkinfo module simplifies things a lot and I can still implement my custom design with it :D


gfwilliams:
This looks great, thanks!

> Its no more possible to trigger an action (.run) if clicked

That's a shame - sorry, I thought that was implemented: https://github.com/espruino/BangleApps/blob/master/modules/clock_info.js#L282

You need focus, and *then* to tap, but if that doesn't work I think it'll need some debugging.

> Before it was possible to hide clock_infos completely and show a larger time instead

That's a point. You can call `.remove` on it to get rid of it completely, and then recreate it later if that helps?

How would you normally select when it was visible or not? Or I guess you could add your own clock_info for the app: http://www.espruino.com/Bangle.js+Clock+Info#adding-clock-info-just-to-your-clock

and then detect when that was displayed and not render anything and instead make the clock bigger? I guess then getting focus is more of a pain.

peerdavid:
> Its no more possible to trigger an action (.run) if clicked [...]
Oh no sry my fault -- I expected a vibration as I had this before (my custom implementation) -- it works totally fine :)

> [...]  add your own clock_info for the app
Great idea I will implement it like this -- one question: clk_info stores the x and y position into the clock_info.json file:
```
{"default:0":{"a":2,"b":0}}
```
but if I use a custom clock_info, "a" is different for different clocks right? Additionally max size of a can be different between different clocks I guess -- is this a problem?

Thanks again!

gfwilliams:
> but if I use a custom clock_info, "a" is different for different clocks right?

That's a good point, yes... Right now all clocks share the same settings, but I have some provision in there for that not being the case, so maybe clock_info could be modified to take an `appName:"bwclock"` option?

peerdavid:
> > but if I use a custom clock_info, "a" is different for different clocks right?
> 
> That's a good point, yes... Right now all clocks share the same settings, but I have some provision in there for that not being the case, so maybe clock_info could be modified to take an `appName:"bwclock"` option?

Oh yea thats a great idea!

PR #2456, opened 2023-01-05
widminbat: Show while charging

notEvil:
I've changed a few parts so it would show while charging.

@LeonMatthes what do you think?

### DIFF
--- a/apps/widminbat/ChangeLog
+++ b/apps/widminbat/ChangeLog
@@ -1,1 +1,2 @@
+0.02: Display while charging
 0.01: Initial Version: Display at under 30% battery
--- a/apps/widminbat/metadata.json
+++ b/apps/widminbat/metadata.json
@@ -1,7 +1,7 @@
 { "id": "widminbat",
   "name": "Minimal Battery",
   "shortName":"MinBat",
-  "version":"0.01",
+  "version":"0.02",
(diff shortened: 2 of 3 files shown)


gfwilliams:
This looks like a good idea, but probably you will also want to respond on the 'charging' event to check whether the widget should be shown? Also changelog/version bump would be needed

notEvil:
True, even though I wouldn't mind the consequence: widget remains visible up to 10m after disconnect from charger. Also, it is easy to force update widgets (long press BTN3) if one desires the widget gone. I can add the event if requested, or describe the behavior in a README.

Regarding changelog and version, thanks for reminding

gfwilliams:
Thanks! I think adding the handler might be better - then it works the same way as other power widgets

notEvil:
Done

I still don't get why we can't just pass `WIDGETS.minbat.update` to `Bangle.on`, JS is a mystery to me xD

gfwilliams:
:) Yes, JS is great like that. The handling of `this` is pretty confusing!

gfwilliams:
This looks good, but the test fails because the ChangeLog is in the wrong order. I'll merge and fix here though

LeonMatthes:
@notEvil I'm a bit late to the party xD
Thanks for implementing this was bugging me for a while as well :D

PR #2464, opened 2023-01-06
[Podcast Addict Remote] Make sure widgets are hidden when fast loading...

thyttan:
... when using Fastload Utils.


gfwilliams:
I'm really not sure about this - under what circumstances does this not work? What app are you loading from, and does it have widgets shown? Does it happen just for these apps, all apps, or any apps that use widget_utils.hide()?

I feel like we need to fix this somewhere else rather than adding it to any affected app. For instance fastload could override Bangle.loadWidgets and just do it there...

PR #2466, opened 2023-01-07
Messagelist fix

rigrig:
Finally figured out why music would sometimes(well, most of the time...) revert to `<unknown>`: calling `setUI` to refresh/switch screens within the app also called `Bangle.uiRemove` every time.

Also a minor cleanup: remove the `MESSAGES!==undefined` check, as we don't handle failure anyway. (and the code that loads MESSAGES does have an error handler)
(This causes the diff to look huge, while it's actually just 4 lines+indentation changes)

### DIFF
--- a/apps/messagelist/ChangeLog
+++ b/apps/messagelist/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New app!
+0.02: Fix music updates while app is already open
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks!

PR #2467, opened 2023-01-07
sched: update clock_info to refresh periodically on active alarms/timers

glemco:
After the last modification, the text and progress bars were not updated from the clkinfo. With this PR the items are updated with a frequency that depends on how far will the timer fire.
I hope I didn't leave any dangling timeout around :)

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -19,3 +19,4 @@
 0.16: Improve support for date timezone
 0.17: Fix midnight in local timezone (alarms wouldn't always fire as expected in timezone != 0)
 0.18: Update clock_info to avoid a redraw
+0.19: Update clock_info to refresh periodically on active alarms/timers
--- a/apps/sched/clkinfo.js
+++ b/apps/sched/clkinfo.js
@@ -49,6 +49,52 @@
     return val;
   }
 
+  /*
+   * Returns the array [interval, switchTimeout]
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks!

glemco:
I've got a couple of questions on this topic:
* I am planning another modification of this kind for the weather app (trigger an update on the `update` event from the weather library) because it looks to me that updates aren't triggered unless manually or with a reset, at least looking at the code.
However I see on my watch that sometimes the clkinfo is actually updated (I don't know exactly when, perhaps on notifications) without intervention, am I trying to solve a non-problem or I'm just lucky and there should be an explicit redraw on update?
* With this PR, the watch is waking up (in the worst case) every minute, considering the code runs on a clock which is likely refreshing every minute too, would it be better for power consumption to somwhow align/merge the timeouts (in this case the alarm clkinfo and the clock's), or is it again a pointless observation with minute granularity?

gfwilliams:
... actually I just realised what this is for - mainly it's because you want the 'value' to update for circlesclock, rather than actually having to refresh the alarm items that are displayed? In which case this seems fine, especially with the update every minute I don't see it being a big deal at all

gfwilliams:
>  I have no way to know if it would have refreshed, say, 30 seconds later

Exactly - some clocks might do it automatically, some might not - so even if you tested with one clock it might still break on another!

PR #2469, opened 2023-01-07
ClockFace: use widget_utils instead of not loading widgets

rigrig:
ClockFace now always calls `loadWidgets`, clocks can add a `hideWidgets` setting to hide widgets (Bangle.js 2: or only show on swipe)
This also means we can get rid of the `clockHasWidgets` hack 🙂

* Default: show widgets
* If `hideWidgets` setting is undefined, but `loadWidgets` was still false, default to hiding widgets instead
* Also updated (I think) all clocks that were using ClockFace to hide widgets


gfwilliams:
Thanks!

PR #2470, opened 2023-01-07
[health] fix settings.js + boot.js

thyttan:
Settings would not be saved correctly since the settings variable wasn't always passed on to setSettings. This fixes that.

EDIT: moved a LANG-tag, I think this will fix a bug where the step goal notification would say 8000 regardless of what was actually set.

### DIFF
--- a/apps/health/ChangeLog
+++ b/apps/health/ChangeLog
@@ -18,3 +18,4 @@
 0.17: Add automatic translation of bar chart labels
 0.18: Show step goal in daily step chart
 0.19: Can show notification when daily step goal is reached
+0.20: Fix the settings page, it would not update settings correctly.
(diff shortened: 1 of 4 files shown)


thyttan:
Yes, it was actually that setSettings was called without argument when toggling "Step Goal Notification", and rather than adding an argument there I opted to redo setSettings to work without arguments as it's done in the tutorial on espruino.com. 

So heart rate settings did surely work before. :)

thyttan:
By the way, do boot.min.js have to be updated manually or is it generated automatically?

gfwilliams:
If there's a '.min.js' file then that'd have to be updated manually...

thyttan:
Ok... will see to it.

Issue #2459, opened 2023-01-06
[BW Clock] Updated time draws over Clock Info selected boarder


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v16.45

### The bug

When the clock info is selected there's a boarder around it. This overlaps with the time so it ends up painted over with black when the time updates. See image.

![download (2)](https://user-images.githubusercontent.com/97237430/211001030-3bde37d5-7942-4ef5-8408-fc64cb4e2122.png)

gfwilliams:
Fixed with https://github.com/espruino/BangleApps/pull/2471

PR #2461, opened 2023-01-06
[BW Clock] don't draw over clock info boarder when updating time

thyttan:
Fixes #2459

PR #2471, opened 2023-01-08
[BW Clock] Added option to only show the time as well as the week number.

peerdavid:
With this PR its again possible to (1) only show the time and (2) also show the week number. This PR also fixes fixes issue #2459 -- its a bit different than the fix of PR #2461 as the size that is shown for the clkinfo is smaller, if only the time is shown:

download


apps/bwclk/app.js
```diff
@@ -270,6 +325,14 @@ let lockListenerBw = function(isLocked) {
...
 Bangle.on('lock', lockListenerBw);
 
+let charging = function(charging){
+  // Jump to battery
+  clockInfoMenu.menuA = 0;
+  clockInfoMenu.menuB = 2;
+  clockInfoMenu.redraw();
+  drawTime();
+}
+Bangle.on('charging', charging);
```
thyttan, line 335: `Bangle.removeListener('charging', charging);` needs to be added to the `remove` function in setUI to not get ram leak.

apps/bwclk/app.js
```diff
@@ -100,21 +100,63 @@ let imgLock = function() {
  * Clock Info
  */
 let clockInfoItems = clock_info.load();
+
+// Add some custom clock-infos
+function weekOfYear() {
```
thyttan, line 105: Needs change to `let weekOfYear = function() {` to not give ram leak.

thyttan (changes requested):
Some changes needed to avoid ram leaks because of fast loading

apps/bwclk/app.js
peerdavid: Oh wow I was not aware of that! Its fixed now, thanks a lot!

apps/bwclk/app.js
peerdavid: Oh wow I was not aware of that! Its fixed now, thanks a lot!

apps/bwclk/app.js
peerdavid: Feel free to unresolve in case something is still wrong :)

apps/bwclk/app.js
peerdavid: Feel free to unresolve in case something is still wrong :)

apps/bwclk/app.js
thyttan: Looks good i think!

@gfwilliams has provided a tutorial regarding fast loading [here](https://www.espruino.com/Bangle.js+Fast+Load) :)

gfwilliams:
Looks great, thanks!

> I think the only thing that would still be nice IMO is to buzz whenever a clkinfo run is executed (or some kind of feedback)

I'd be very happy to add that to clock_info if you wanted to do a PR - I guess we just buzz if the .run function is defined?

Not many clockinfo have run so right now it won't matter too much, and at some point maybe it'll be an app with a settings page where it could be configured

PR #2472, opened 2023-01-08
remove null check, since mapping is now down on app side

LukasEdl:
With the [Gadgetbridge PR](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3019), the mapping of "long -> lon" will be done on the app side and therfore this guard is not needed anymore.

If you want to keep it as backward compatibility, just close this PR.

### DIFF
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -134,10 +134,8 @@
         event.satellites = NaN;
         event.course = NaN;
         event.fix = 1;
-        if (event.long!==undefined) {
-          event.lon = event.long;
-          delete event.long;
-        }
+        event.lon = event.long;
+        delete event.long;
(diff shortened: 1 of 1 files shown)


halemmerich:
This will overwrite the event.lon with undefined if the GB PR is merged. I think as soon that is merged any lines regarding event.lon/long could just be removed.

LukasEdl:
yeah i forgot about that, i removed the mapping

gfwilliams:
Thanks!

gfwilliams:
Sorry - actually I didn't realise this wan't in the PR that's merged - so reverting for now: https://forum.espruino.com/conversations/380297/?offset=25#comment16820037

PR #2473, opened 2023-01-08
android - Handle reconnects with fallback to internal GPS

halemmerich:
This handles connection events and switches all GPS clients over from GPS to GBGPS and viceversa.
@gfwilliams I have written some manually executed code to test the transition, is there a way to run something like that automatically?

### DIFF
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -20,3 +20,4 @@
 0.19: Add automatic translation for a couple of strings.
 0.20: Fix wrong event used for forwarded GPS data from Gadgetbridge and add mapper to map longitude value correctly.
 0.21: Fix broken 'Messages' button in menu
+0.22: Handle connection events for GPS forwarding from phone
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -221,19 +221,24 @@
     NRF.on('disconnect', ()=>{handleConnection(1);});
 
     // Work around Serial1 for GPS not working when connected to something
+    let serialTimeout;
     let wrap = function(f){
       return (s)=>{
+        if (serialTimeout) clearTimeout(serialTimeout);
         origSetGPSPower(1,"android_gpsserial");
         f(s);
+        serialTimeout = setTimeout(()=>{
+          serialTimeout = undefined;
+          origSetGPSPower(0,"android_gpsserial");
+        },5000);
       };
-    }
-    Serial.println = wrap(Serial1.println);
-    Serial.write = wrap(Serial1.write);
+    };
+    Serial1.println = wrap(Serial1.println);
+    Serial1.write = wrap(Serial1.write);
 
     // Replace set GPS power logic to suppress activation of gps (and instead request it from the phone)
     Bangle.setGPSPower = (isOn, appID) => {
       // disable our own request for GPS power first and always
-      if (!isOn) origSetGPSPower(0,"android_gpsserial");
       // if not connected use internal GPS power function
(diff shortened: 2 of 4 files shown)


halemmerich:
@gfwilliams I have seen https://github.com/espruino/BangleApps/blob/master/bin/runapptests.js, maybe that could be extended with a command to execute a script file for more complex tests?

gfwilliams:
That's the idea, yes.

Well, the plan is it'll search for a `test.json` file in each app's folder, and it'll run the commands in that. So that would be your script file.

I'm not sure having a separate JS file vs the JSON with individual commands in is such a big deal, but it could be added later I guess. The benefit of the JSON is you can just have commands like 'eval' that will evaluate something and test the response.

halemmerich:
The benefit of the script file would be being runnable in the IDE for testing. As well as having syntax highlighting etc. in editors. The test.js file currently has 148 lines, that is not great to handle in json while it contains practically only js.

What do you think of the latest commit, working around Serial1 not being setup when connected because of GPS used from GB?

gfwilliams:
Thanks for this - I'll need to look into it a bit more though. It feels like this adds quite a lot of complexity, and that's not going to be great for boot times or maintainability.

Why do we need separate `setGBGPS`/etc functions here? Are you envisaging some apps using Gadgetbridge GPS separately? I'd be very wary of that - you know in BTHRM there was all that extra complexity and I'm not sure any apps (apart from your logging app) ever used it?

I'd have though the switchover on connect/disconnect could be handled a bit easier than maintaining a separate GBGPS list, like maybe:

```
var x = Bangle._PWR.GPS;
delete Bangle._PWR.GPS;
origSetGPSPower(0);
Bangle._PWR.GPS = x;
```

For the Serial1 case - you're specifically thinking about updating AGPS data? Maybe the AGPS app should be aware of it somehow? Apart from Serial1.write handler currently calling Serial1.println, I'd be a bit concerned that once turned on, the internal GPS never gets turned off... I guess ideally you want to call the 'disconnect' handler so you disable Gadgetbridge GPS and switch back to internal at that point (but only call it once).

For testing, yes. We could add a script file - but remember it is *just* for the emulator. My concern there is just that the script files may end up being completely unmaintainable if everyone just writes them in their own style doing whatever they want - but I guess some tests is better than no tests. I'd still like to encourage use of the JSON where possible

halemmerich:
I primarly did it that way, because I was not sure if `Bangle._PWR.GPS` was the only place tracking apps using the GPS. In case there was something else keeping that information there would be chaos when just ignoring that. Seems that is not a problem from your answer. That would make this a lot easier for sure. GBGPS is only kept for this reason.
Boot times could be optimized by moving the code into a lib and requiring it only if the setting is enabled.
As for tests, I'm all for standardizing a json format to describe the test cases. When this stabilizes a bit I will think about converting the test.js file to that. Actually removing GBGPS would already remove about half of that file :)

halemmerich:
Removing GBGPS tracking actually makes my test script more or less useless, since the expected behaviour is now exactly the same for using GB GPS or not. Can I somehow check if the internal GPS is active in another way? Maybe some property or behaviour on Serial1? I have seen errors caused by missing RX/TX pins when internal GPS is not on while writing to the serial port, maybe that can be exploited for the test script?

halemmerich:
@gfwilliams is `getPinMode((process.env.HWVERSION==2)?D30:D26) == "input"` a valid way for checking the internal GPS state in the test script?

gfwilliams:
Thanks - I'm going to pull this in anyway as it seems the existing implementation is completely broken and this fixes that :)

PR #2479, opened 2023-01-09
[health] update boot.min.js

thyttan:
Continuation from https://github.com/espruino/BangleApps/pull/2470#issuecomment-1375560533.

Updates and fixes boot.min.js.

However, using this doesn't take into account LANG-tags in the non-minified boot.js. So maybe I should change the metadata to use boot.js when installing to watch instead? @gfwilliams @myxor  @alessandrococco 

Preminified code seems to have been added first with https://github.com/espruino/BangleApps/commit/0b4d2b5b5d811bf11d908576b15a8fbf63adeda6


gfwilliams:
Yes, I wonder if we should just get rid of the .min.js file now that we pretokenise apps (I think by default?).

It might be worth checking the commit history and also testing well - it may be the min.js was done for a reason (eg avoiding the built-in minifier/pretokeniser because of some issue)

PR #2481, opened 2023-01-09
remove not need lines

dapgo:
mini change


gfwilliams:
It's ok - I don't think this is needed now, I already made the changes :)

PR #2482, opened 2023-01-09
[clock_info] Buzz on run()

peerdavid:
This change updates clock_info to buzz whenever run() exists. Note that I have not updated the version of all apps that use clock_info. So I guess this update is installed whenever a user updates or re-installs the clock right? But also in this case I guess its fine to not update all clocks?


gfwilliams:
Thanks!

> So I guess this update is installed whenever a user updates or re-installs the clock right? But also in this case I guess its fine to not update all clocks?

Yes, exactly - anyone that installs after this will have it.

It's a reason to turn it into an 'app' in the app loader at some point soon I guess. When I get a moment I'll do that

PR #2484, opened 2023-01-09
Berlin clock - Enable fast loading

halemmerich:
This is a draft since I did not yet thoroughly test for memory leaks. The changes seem simple enough though ;)

### DIFF
--- a/apps/berlinc/berlin-clock.js
+++ b/apps/berlinc/berlin-clock.js
@@ -18,9 +18,9 @@ let time_digit = [];
 
 // schedule a draw for the next minute
 let queueDraw = () => {
-  if (drawTimeout) clearTimeout(drawTimeout);
-  drawTimeout = setTimeout(function() {
-    drawTimeout = undefined;
+  if (global.drawTimeout) clearTimeout(global.drawTimeout);
+  global.drawTimeout = setTimeout(function() {
+    global.drawTimeout = undefined;
     draw();
   }, 60000 - (Date.now() % 60000));
 }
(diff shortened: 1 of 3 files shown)


halemmerich:
Tested a bit and seems to be ok.

apps/berlinc/berlin-clock.js
```diff
@@ -1,3 +1,6 @@
+// timeout used to update every minute
+let drawTimeout;
```
gfwilliams, line 2: Out of interest, why did you put this here and then change to use `global.drawTimeout`? Every other clock I've seen just put it inside the curly braces and it's fine?

halemmerich: Hm, the thought was that the remove method and timeouts are called from global scope and would then not be able to find a non-global var. Did not explicitly test that, should it work?

gfwilliams: It should - but IMO 'if it's not broken, don't fix it'. It'd be good if you could just keep the pattern the same as other clocks.

In JavaScript a function has access to all variables in the scope it's defined in, so because `remove` is defined inside the curly braces too, it's fine.

halemmerich: You're right, seems to work just fine inside the scope.

halemmerich:
@gfwilliams Is something still missing for this PR?

gfwilliams:
No, looks good - thanks! I just didn't spot you'd committed the change :) 

Thanks for getting that in!

PR #2486, opened 2023-01-09
stopwatch: Save state to storage

nxdefiant:
This keeps the Stopwatch Touch running in the background.

I sometimes accidentally pressed the hw-button which killed the timer.

Notify @hughbarney


gfwilliams:
Great idea, thanks!

When you're adding a config file, please could you list it in a `data` section in the metadata though? I'll do that now after merging this time

nxdefiant:
Thanks, sorry I forgot the metadata entry.

PR #2487, opened 2023-01-10
Home Assistant: Allow swiping to switch triggers, standardize spellings

myxor:
- Allow swiping to switch triggers
- Standardize spellings


gfwilliams:
Neat - thanks!

PR #2488, opened 2023-01-10
Home Assistant 0.09

myxor:
- Improve web interface: use interface instead of custom, fix loading and saving of JSON
- Draw left/right arrow in UI if further triggers are available


peerdavid:
Pretty cool thanks for the improvements -- on thing that I found, when you upload HomeAssistant and not update triggers, the screen is only black. Do you think it would help to show a message "please set triggers in the web-ui" or so?

myxor:
Yes sure that totally makes sense

PR #2490, opened 2023-01-11
widbgjs - Get your xDrip Blood Glucose values on the bangle

phrogg:
I attempted to create a widget to display blood glucose values on the bangle. 

I did some testing and so far, for me, it works. Furthermore, I will soon make the companion app needed for this to work available on f-droid.


gfwilliams:
This looks great - thanks!

Issue #2478, opened 2023-01-09
[widget utils/Fastload Utils] Initial .hide() will not hide widgets when fast loading, needs a .show() first


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v16.49

### The bug

First discussed in https://github.com/espruino/BangleApps/pull/2464#issuecomment-1375354642.

I experience this with Podcast Addict Remote and Remote for Spotify, when fast loading from BW Clock (using Quick Launch) and from Desktop Launcher (both showing widgets) using Fastload Utils installed on the watch. The bug is not present if not using Fastload Utils.

I have yet to see if this bug is present when switching to other apps than the two mentioned using .hide(). I have also not tried launching them from other clocks or launchers.

Please propose other apps utilizing .hide() for testing this bug.

Reproduce:
1. Install Desktop Launcher and Fastload Utils
2. open Podcast Addict Remote
3. Notice widgets are not hidden.

### Installed apps

kbmulti (0.05)
widbthide (0.01)
widram (0.03)
calculator (0.07)
activityreminder (0.10)
pomoplus (0.05)
widanclk (0.02)
kbtouch (0.02)
run (0.14)
heartzone (0.01)
wohrm (0.09)
sevenmin (0.01)
banglexercise (0.03)
sleeplog (0.12)
swp2clk (0.04)
torch (0.11)
hrm (0.11)
lightswitch (0.06)
boot (0.55)
recorder (0.20)
spotrem (0.07)
podadrem (0.07)
sleepphasealarm (0.12)
annoy (0.01)
messages (0.58)
widmsggrid (0.04)
messageicons (0.03)
messagesmusic (0.05)
slopeclock (0.04)
tetris (0.01)
smpltmr (0.07)
weather (0.23)
circlesclock (0.23)
alarm (0.37)
android (0.20)
setting (0.56)
dragboard (0.07)
aiclock (0.07)
widminbat (0.02)
notify (0.12)
quicklaunch (0.08)
taglaunch (0.02)
altimeter (0.02)
messagelist (0.02)
health (0.20)
bwclk (0.28)
dtlaunch (0.22)
sched (0.19)
widalarm (0.01)
swscroll (0.02)
fastload (0.03)

thyttan:
```
/*
1. Upload this script to a file "testWU.js" via Web IDE.
2. Navigate the watch to a clock/launcher/app with remove-function specified and widgets shown (I use BW Clock).
[...]

PR #2491, opened 2023-01-11
[BW Clock] use widget_utils

thyttan:
#2478 #2196

Works well for me with Screen: Normal, Dynamic and Full.

### DIFF
--- a/apps/bwclk/app.js
+++ b/apps/bwclk/app.js
@@ -6,7 +6,7 @@
 const locale = require('locale');
 const storage = require('Storage');
 const clock_info = require("clock_info");
-
+const widget_utils = require("widget_utils");
 
 /************************************************
  * Globals
@@ -277,9 +277,9 @@ let drawLock = function() {
 
 let drawWidgets = function() {
   if(isFullscreen()){
-    for (let wd of WIDGETS) {wd.draw=()=>{};wd.area="";}
+    widget_utils.hide();
   } else {
-    Bangle.drawWidgets();
+    Bangle.drawWidgets(); // Should this be replaced with widget_utils.show(); ?
   }
 };
 
@@ -318,7 +318,7 @@ let lockListenerBw = function(isLocked) {
   if(!isLocked && settings.screen.toLowerCase() == "dynamic"){
     // If we have to show the widgets again, we load it from our
     // cache and not through Bangle.loadWidgets as its much faster!
-    for (let wd of WIDGETS) {wd.draw=wd._draw;wd.area=wd._area;}
+    widget_utils.show();
   }
 
   draw();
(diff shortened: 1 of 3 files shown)


apps/bwclk/app.js
```diff
@@ -277,9 +277,9 @@ let drawLock = function() {
 
 let drawWidgets = function() {
   if(isFullscreen()){
-    for (let wd of WIDGETS) {wd.draw=()=>{};wd.area="";}
+    widget_utils.hide();
   } else {
-    Bangle.drawWidgets();
+    Bangle.drawWidgets(); // Should this be replaced with widget_utils.show(); ?
```
thyttan, line 282: Should this be replaced with widget_utils.show(); ?

apps/bwclk/app.js
```diff
@@ -370,7 +370,8 @@ Bangle.loadWidgets();
 
 // Cache draw function for dynamic screen to hide / show widgets
 // Bangle.loadWidgets() could also be called later on but its much slower!
-for (let wd of WIDGETS) {wd._draw=wd.draw; wd._area=wd.area;}
+//for (let wd of WIDGETS) {wd._draw=wd.draw; wd._area=wd.area;} // Not sure how to replace this or if it should just be dropped?
+widget_utils.show(); // I replace with this to begin with, should I just delete it?
```
thyttan, line 374: // Not sure how to replace this or if it should just be dropped?

widget_utils.show(); // I replace with this to begin with, should I just delete it?

apps/bwclk/app.js
thyttan: Tried this on another clock app, didn't work very well so I'll not change to .show() here.

gfwilliams:
Hi - please can you add `widget_utils.show()` in the remove handler?

That's the real issue here. Basically even if you weren't using widget_utils, just showing the widgets again in the remove handler would have fixed it.

As-is, even with these changes I think that if the clock fast-loads to a launcher that doesn't use widget_utils, the widgets will still stay hidden?

thyttan:
Sorry! Just replaced already existing implementation and didn't think to do that even though I should have remembered that's how it should be done.

gfwilliams:
No worries -  thanks!

thyttan:
There, .show() in remove handler, also removed 371-374 as per my review comments above.

>edit: It'd be nice to have some way of testing this, but I'm not too sure how. I think we'd have to spin up an emulator, and it all gets a bit tricky.

Would it be possible to search for if the string ".show()" is between the setUI's curly braces? That could help many cases I think?

gfwilliams:
Great, thanks!

> Would it be possible to search for if the string ".show()" is between the setUI's curly braces? 

It would be with a bit of work, but then I guess we'd have to check for if widget_utils was used - plenty of faces *don't* hide widgets.

But that is only one issue - there's loaded libraries, memory usage, etc - like what's in http://www.espruino.com/Bangle.js+Fast+Load - I'm not sure it's reasonable to expect everyone to test all that all the time.

I think as we get more fast-loaded apps someone making a change that breaks one is going to get more and more common, and it could end up being quite a big issue. I did start something at https://github.com/espruino/BangleApps/blob/master/bin/runapptests.js and maybe when I get time I'll see if I can add a fast-loading check in there

PR #2492, opened 2023-01-11
[lcars/notanalog/limelight/daisy/thering/widviz/90sclk/widviztime/neonx] Add widget utils

thyttan:
I've converted the remaining clocks and widgets in #2196 to use widget utils.

I have not tested any of them and don't really intend to. If someone feels up for it please go ahead.

Especially widviz and widviztime I don't know if I've messed up or if they are ok. I can drop them from the PR if that's better.

For the others I feel it could be ok to merge them and if bugs come up we can fix them then?

### DIFF
--- a/apps/widviz/widget.js
+++ b/apps/widviz/widget.js
@@ -1,26 +1,19 @@
 (() => {
+  let widget_utils = require('widget_utils');
 
   var saved = null;
 
   function hide(){
     if (!Bangle.isLCDOn() || saved) return;
     saved = [];
-    for (var wd of WIDGETS) {
-      saved.push({d:wd.draw,a:wd.area});
-      wd.draw=()=>{};
-      wd.area="";
-    }
+    widget_utils.hide();
     g.setColor(0,0,0);
     g.fillRect(0,0,g.getWidth(),23);
   }
 
   function reveal(){
     if (!Bangle.isLCDOn() || !saved) return;
-    for (var wd of WIDGETS) {
-      var o = saved.shift();
-      wd.draw = o.d;
-      wd.area = o.a;
-    }
+    widget_utils.show();
     Bangle.drawWidgets();
     saved=null;
   }
--- a/apps/widviz/widget.js
+++ b/apps/widviz/widget.js
@@ -1,11 +1,11 @@
 (() => {
   let widget_utils = require('widget_utils');
 
-  var saved = null;
+  var saved = false;
 
   function hide(){
     if (!Bangle.isLCDOn() || saved) return;
-    saved = [];
+    saved = true;
     widget_utils.hide();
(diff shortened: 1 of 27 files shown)


apps/widviztime/widget.js
```diff
@@ -1,30 +1,20 @@
...
 
     function reveal() {
         if (!Bangle.isLCDOn() || !saved) return;
-        for (var wd of WIDGETS) {
-            var o = saved.shift();
-            wd.draw = o.d;
-            wd.area = o.a;
-        }
+        widget_utils.show();
         Bangle.drawWidgets();
```
gfwilliams, line 18: `Bangle.drawWidgets();` shouldn't be needed here, and I think you need `saved=null`

apps/widviztime/widget.js
```diff
@@ -1,30 +1,20 @@
 (() => {
+    let widget_utils = require('widget_utils');
 
     var saved = null;
 
 
     function hide() {
         if (!Bangle.isLCDOn() || saved) return;
         saved = [];
```
gfwilliams, line 9: I think you need saved=true here

apps/widviz/widget.js
```diff
@@ -1,26 +1,19 @@
 (() => {
+  let widget_utils = require('widget_utils');
 
   var saved = null;
 
   function hide(){
     if (!Bangle.isLCDOn() || saved) return;
     saved = [];
```
gfwilliams, line 8: saved=true here instead

apps/widviz/widget.js
```diff
@@ -1,26 +1,19 @@
...
 
   function reveal(){
     if (!Bangle.isLCDOn() || !saved) return;
-    for (var wd of WIDGETS) {
-      var o = saved.shift();
-      wd.draw = o.d;
-      wd.area = o.a;
-    }
+    widget_utils.show();
     Bangle.drawWidgets();
```
gfwilliams, line 17: You can remove drawWidgets but you need saved=false

apps/widviztime/widget.js
thyttan: null or false doesn't matter?

thyttan:
Updated per suggestions.

apps/widviz/widget.js
gfwilliams: It shouldn't, no... But using false everywhere is nicer

thyttan:
>Thanks, but I don't think you quite updated the widgets properly... See the comments - GitHub's message thing is annoying as I left the comment for saved=true here instead next to  saved = []; - but it looks from the section posted above like I meant it for var saved = null;

Yeah I see, if I were a little more observant I could have worked that out, sorry :P

gfwilliams:
Great - thanks!

PR #2493, opened 2023-01-12
agpsdata - Switch off GPS after use

halemmerich:
Previously GPS was left on and needed a reload to be switched off. When updating through the app this is not a problem. Automatic updates in the background while on clock however left GPS on until the next reload and caused substantial battery draw.

### DIFF
--- a/apps/agpsdata/ChangeLog
+++ b/apps/agpsdata/ChangeLog
@@ -5,3 +5,4 @@
 0.04: Write AGPS data chunks with delay to improve reliability
 0.05: Show last success date
       Do not start A-GPS update automatically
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks! I just spotted the AGPS app is the same too, so I'm fixing that now.

But maybe it's work turning GPS off using setTimeout of ~1sec - just in case it's turned off before all data has finished sending/it's finished writing it?

halemmerich:
Good idea, changed that for the good case. I think in case of errors we can switch off immediately.

gfwilliams:
Yes, totally - thanks!

PR #2495, opened 2023-01-13
Health - Fix heartrate sensor timeout

### LINKED ISSUE #2435: Minor bug in health app
Hi,

this timeout
https://github.com/espruino/BangleApps/blob/1f1b8a20d566eddedd397d06af9e578b1547ce81/apps/health/boot.js#L14
seems way too long. Shouldn't it be just 60k ms?

halemmerich:
Fixes #2435


gfwilliams:
Thanks!

Issue #2474, opened 2023-01-08
Hideable widget bar blank when first shown


jt-nti:
### Affected hardware version

Bangle 2

### Your firmware version

2v16

### The bug

When swiping down the hidden widget bar I see a blank bar with no widgets visible

I've tried with the MacWatch2 clock and the widget bar seems to stay blank until after the screen locks/unlocks. After that the widgets (lock widget, lato pedometer, and battery widget) are shown as expected when swiping to reveal the hidden widget bar.

### Installed apps

widbat (0.09), waypoints (0.01), cprassist (0.02), widpb (0.05), calendar (0.11), antonclk (0.11), gpstime (0.05), beer (0.02), quicklaunch (0.08), boot (0.55), messagegui (0.63), barcode (0.10), messages (0.58), messageicons (0.03), launch (0.20), widalarm (0.01), about (0.14), lato (0.01), medicalinfo (0.01), geminiclock (0.01), gpsinfo (0.11), alarm (0.37), widmessages (0.04), widlock (0.08), setting (0.56), sched (0.18), notify (0.12), health (0.19), android (0.20), widclkbttm (0.05), macwatch2 (0.06)

sir-indy:
The above commit seems to have fixed this issue now, Gordon found the solution far quicker than I could!

sir-indy:
Yes, sorry, this only fixes the issue of a blank widget bar when first shown. I haven't tried installing your widmeda, though it looks very good!

PR #2497, opened 2023-01-13
[mosaic][fuzzyw] Update for fast launch, and optionally hide widgets with widget_utils

sir-indy:
I've updated my two clocks to the latest and greatest standard, with all the bits in for fast launch and using `widget_utils` to hide the widgets (looks great on the mosaic clock).

I've got a problem when swiping to show the widgets though - for the first few minutes it just shows an empty widget bar. Leave it for a few minutes and the widgets show as expected. This happens on the pebble clock too, so this could be a issue with my B2.


gfwilliams:
> I've got a problem when swiping to show the widgets though - for the first few minutes it just shows an empty widget bar.

Do you have any widgets that show at the bottom? Someone filed an issue for what happens when they have a big widget at the bottom... It gets added to the line at the top and then pushes other widgets off.

sir-indy:
Nope, no bottom widgets. Just reset to factory settings on the watch, installed the Pebble clock (rather than one of the ones I've just tweaked), but widgets still show as a blank bar when swiped down. Fast load works, but widgets still show as blank in the launcher when entered from the Pebble clock.

Give it a few minutes and the widgets appear again correctly when swiped down. Now widgets show in the launcher, but touching the 'back' icon or pressing the button means widget bar is blank again on the Pebble clock. Switching between the launcher and Pebble clock, widgets show correctly in the launcher but always start as blank on the clock.

Behaviour seems the same from a straight factory reset, and after updating installed apps. No errors in the IDE
Installed versions:
alarm (0.33), sched (0.13), setting (0.54), widlock (0.06), health (0.15), about (0.13), launch (0.19), boot (0.53), antonclk (0.11), widbat (0.09), widbt (0.08), widid (0.03), welcome (0.14), pebble (0.11)

Actually, it looks like the widgets just aren't being drawn on the overlay to start with. After a little while swiping down, the battery appeared, but the BT and ID widget didn't show. Battery widget correctly placed with space for the other widgets though, so it knows about the BT and ID widgets, just not drawing them straight away.

This is a lot of text, sorry! Should I pop this in the forum and see if anyone else has seem something like this?

gfwilliams:
Thanks for all those tests - I think it's best to talk about this on https://github.com/espruino/BangleApps/issues/2474 - but if it can be reproduced after a factory reset hopefully it'll be pretty easy to track down.

Looking at the code on https://github.com/espruino/BangleApps/blob/master/modules/widget_utils.js it may be that as you say, it just needs a drawWidgets call at the end of `swipeOn`

sir-indy:
Yep, thanks for the suggestion Gordon, that seems to have fixed it!

gfwilliams:
Great, thanks!

PR #2498, opened 2023-01-13
espruinoctrl updates for Bangle.js 2

nxdefiant:
Disable not existing BTN3 on Bangle.js 2, set maximum transmit power

### DIFF
--- a/apps/espruinoctrl/custom.html
+++ b/apps/espruinoctrl/custom.html
@@ -164,6 +164,7 @@ function cmd(cmd,mac,rx) {
   if (mac) {
     E.showMessage("Connecting\\n"+mac);
     if (mac.length==17) mac+=" random";
+    NRF.setTxPower(process.env.HWVERSION == 2 ? 8 : 4);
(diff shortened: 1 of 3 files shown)


apps/espruinoctrl/custom.html
```diff
@@ -164,6 +164,7 @@
   if (mac) {
     E.showMessage("Connecting\\n"+mac);
     if (mac.length==17) mac+=" random";
+    NRF.setTxPower(process.env.HWVERSION == 2 ? 8 : 4);
```
nxdefiant, line 167: A MAX_POWER constant might be helpful.
also a way to read current NRF settings.

Is there any additional thing to to to increase
range?

gfwilliams: I guess it would be an idea - or maybe the function could just clip the TX power to the level it can do rather than erroring.

I don't think there is much extra to increase range when connecting. For the peripheral device if it's NRF52840 I think it can now specify to enable long range advertising though

gfwilliams:
Thanks! It's surprising as I thought I'd used that on Bangle.js 2 recently :)

PR #2499, opened 2023-01-14
weather: redraw clock_info on update and provide color field

glemco:
The color field is used by circlesclock if no other color is set


gfwilliams:
Thanks!

glemco:
I tried to use `g.blendColor` for gradients in circlesclock as you suggested, however it mostly produces dithered colours (instead of making the gradient from 5-6 known colours, it has a much broader spectrum).
Is there anything I'm missing? Like some function that converts an 8 bits colour to the closest that looks good on the B2?

gfwilliams:
Ahh, interesting - yes, it just produces the blended 16 bit color - it's not aware of what looks good on the display or not.

There's nothing like that in the Bangle - it could be convinced to choose the nearest color from the 8 bit Web Palette, but that's not guaranteed to look good.

glemco:
Got it, then I'd say it can stay as it is for now

PR #2500, opened 2023-01-14
2047pp - Workaround for the minifier removing the class definition

halemmerich:
Wrapping the whole app in a scope seems to be enough to fix the TwoK class missing after minification.


gfwilliams:
Thanks!

PR #2501, opened 2023-01-14
Berlin Clock - Adds fullscreen mode

halemmerich:
This uses sliding widgets on B2 and just plain hidden widgets on B1. Also respects the appRect for compatibility with bottom widgets and dynamically changes size on showing the date.

### DIFF
--- a/apps/berlinc/ChangeLog
+++ b/apps/berlinc/ChangeLog
@@ -5,3 +5,4 @@
       Now show widgets
       Make compatible with themes, and Bangle.js 2
 0.06: Enable fastloading
+0.07: Adds fullscreen mode setting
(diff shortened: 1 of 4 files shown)


gfwilliams:
The CI check shows:

> Clock berlinc file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)

Please can you move loadWidgets down under `setUI`? If you don't do that then some widgets that check Bangle.CLOCK may get confused. I know it's less of a deal with fast loading but it's still good if we can keep all clocks doing the same thing

halemmerich:
Is there some location where this is directly visible, without reading the CI log output? I have been bitten by that before and seem to forget it anyway ;)

gfwilliams:
Ahh - well, if you look at the diffs eg https://github.com/espruino/BangleApps/pull/2501/files it's there (gone now since your update). For example https://github.com/espruino/BangleApps/pull/2501/commits/2ce26f9f00c52ce1907467a5307e800cd79d7732

You just have to give it a minute or two to run the tests first. Thanks for the change!

PR #2503, opened 2023-01-15
GPS Info: Fix number of packets received

myxor:
As written in https://forum.espruino.com/conversations/382632/#comment16830755


gfwilliams:
Thanks!

PR #2504, opened 2023-01-15
android - Use course from GB if available

halemmerich:
Make use of [https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3026](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3026). Will mark this as ready when that PR is merged.


gfwilliams:
Thanks! I guess either way it won't break anything for this to be in - I'll tweak it to !isFinite though as I think that reads better :)

PR #2505, opened 2023-01-15
ContourClock update

xxDUxx:
Fixed a major bug that would leave old font files on the watch when fewer fonts were installed than last time.


gfwilliams:
Thanks!

PR #2506, opened 2023-01-16
New app: chargent

notEvil:
see
https://github.com/notEvil/BangleApps/tree/chargent/apps/chargent
https://notevil.github.io/BangleApps/index.html?q=gently

Fyi I'm no expert and don't know the hardware ins and outs, so it could create false impressions (best practice, correctness, safety, ...). If I should state this in the readme, let me know.


gfwilliams:
Hi - thanks! I wonder how much this will really improve battery life, but I guess it's worth having as an option. 

... also, this minifying is fine but I really wouldn't worry about minifying stuff too much in the future - the app loader does pretokenising automatically now, which will be almost as good, and it's easier to maintain :)

notEvil:
> Hi - thanks! I wonder how much this will really improve battery life, but I guess it's worth having as an option.

I've been using http://www.accubatteryapp.com for a while now and they list their sources here:

https://accubattery.zendesk.com/hc/en-us/articles/210224725-Charging-research-and-methodology

Especially the section "Effect of topping off your charge" is the main motivation. (Probably should have looked it up earlier)
So in theory it should make a difference compared to charging for an extended period of time.

> ... also, this minifying is fine but I really wouldn't worry about minifying stuff too much in the future - the app loader does pretokenising automatically now, which will be almost as good, and it's easier to maintain :)

True, in this case there is not too much to gain from minification. In general however I really like not having to worry about the length of variable/function names or expression efficiency :)

bobrippling:
Hi @notEvil - I believe this will also work on the Bangle2, except we want pin3 instead of pin30, according to [these specs](https://hackaday.io/project/175577-hackable-nrf52840-smart-watch).
If that lines up with what you'd expect, I'll push up a PR

notEvil:
Sure, why not. I have no v2 to test, but if the PIN works the same and its a Li-ion, I don't see any reason why it shouldn't?

PR #2509, opened 2023-01-18
Don't send app_stats on disconnect

### LINKED ISSUE #2405: appusage.json not updated
Is it possible that the appusage.json is no longer updated?

About 4 weeks ago I installed my own app on my 2 watches. It does not show up in appusage.json as a favorite or installed.

All the app loaders I use (browser and Gadgetbridge) submit the analytics data.

BartS23:
Hopefully fix #2405

`sendUsageStats` is also called in case of a disconnect.
But then no apps are available anymore


gfwilliams:
Thanks! Let's see if it helps!

I guess the chances of someone uninstalling all apps and leaving their Bangle that way are pretty remote!

PR #2510, opened 2023-01-18
Emit events on alarm snooze/dismiss and new app

Kedlub:
This adds events to alarm dismiss and snooze for other apps to use
It also includes a standalone app that uses this for quiet mode disabling after alarm dismiss, as an alternative way of doing #2429 
This should be a cleaner way, that doesn't bloat the menus for alarm editing
For now `alarmqm` runs for all alarms, but maybe in the future there could be a configurable time range, and it will run only for alarms in that range


gfwilliams:
Thanks! This looks great! Sorry for not replying properly in https://github.com/espruino/BangleApps/pull/2429 but things have been manic this month - I've got a bunch of people chasing me for different things

PR #2511, opened 2023-01-18
[entonclk] For now, just return the really needed HRM reading

siatwe:
Next week I will take time to implement clock_info into my watchface. For now, I have simply adjusted the data for HRM.


gfwilliams:
Thanks!

PR #2514, opened 2023-01-19
Fix potential NPE in loader.js

myxor:
Can potentially fix https://forum.espruino.com/conversations/383162/#comment16836383

But why this happens for the user Hank is not clear to me.


gfwilliams:
Thanks! I'd be pretty sure the HTML has changed somehow, but it's good to have these checks in

PR #2515, opened 2023-01-19
Updated Layout.js

nravanelli:
These changes provide an option for users to select the color of the buttons (both background and border).

### DIFF
--- a/modules/Layout.js
+++ b/modules/Layout.js
@@ -186,8 +186,12 @@ Layout.prototype.render = function (l) {
         x+4,y+h-1,
         x,y+h-5,
         x,y+4
-      ], bg = l.selected?gfx.theme.bgH:gfx.theme.bg2;
-      gfx.setColor(bg).fillPoly(poly).setColor(l.selected ? gfx.theme.fgH : gfx.theme.fg2).drawPoly(poly);
+      ], bg = l.selected?gfx.theme.bgH:gfx.theme.bg2, 
+      btnborder = l.selected?gfx.theme.fgH:gfx.theme.fg2;
+    //add button color option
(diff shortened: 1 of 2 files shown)


modules/Layout.js
```diff
@@ -186,8 +186,12 @@ Layout.prototype.render = function (l) {
         x+4,y+h-1,
         x,y+h-5,
         x,y+4
-      ], bg = l.selected?gfx.theme.bgH:gfx.theme.bg2;
-      gfx.setColor(bg).fillPoly(poly).setColor(l.selected ? gfx.theme.fgH : gfx.theme.fg2).drawPoly(poly);
+      ], bg = l.selected?gfx.theme.bgH:gfx.theme.bg2, 
+      btnborder = l.selected?gfx.theme.fgH:gfx.theme.fg2;
+    //add button color option
+    if(l.btnCol!==undefined) bg = l.btnCol;
+    if(l.btnBorder1!==undefined) btnborder = l.btnBorder;
```
gfwilliams, line 193: There's a type here: `btnBorder1` - so it seems this wasn't tested?

Also, if you were to set a color, it looks like the button is then unable to show when it is selected or not (which happens on Bangle.js 1 where there is no touchscreen and you must select using buttons)? It might be better to just set the colors to the defaults if the button is selected, but then allow them to be customised when it's not selected.

... also, do we want `l.btnCol`? It might be better just to use the pre-existing `l.bgCol` value which isn't used here anyway?

nravanelli:
Fixed my fat finger issue, and incorporated the bgCol as suggested. When selected on BangleJS1, it _should_ default to the normal `l.selected` button color. Can't test as I have a BangleJS2.

gfwilliams:
Looks great - thanks!

PR #2516, opened 2023-01-19
Sched: Fix crash in clock_info

myxor:
I saw "Uncaught Error: clearInterval(undefined) not allowed" when scheduler / alarm was set as clock info item and starting the launcher.

This fixes the crash.

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -21,3 +21,4 @@
 0.18: Update clock_info to avoid a redraw
 0.19: Update clock_info to refresh periodically on active alarms/timers
 0.20: Alarm dismiss and snooze events
+0.21: Fix crash in clock_info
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks!

Please can you keep an eye open for this happening in other clockinfos?

Generally we say that whatever displays them is responsible for calling 'show' and 'hide' the correct number of times. So if 'hide' gets called twice it's not the clockinfos problem, it's `addInteractive` or similar.

But it looks like in this case if there are no alarms the interval doesn't get added, and that's why you'd get this error?

myxor:
Yes i had no alarm configured and therefore the crash happened - i assume.

I will try to have a look at the other clockinfos.

gfwilliams:
Don't worry about actively looking - just if you see an error like that crop up again please let me know!

PR #2518, opened 2023-01-19
calendar: Mark dated events on a day

nxdefiant:


gfwilliams:
Nice idea - thanks! In general it'd be good to use the `sched` library but it looks like it doesn't add a great deal of useful functionality in this case anyway...

PR #2521, opened 2023-01-20
Some changes to the clkinfo

glemco:
May be questionable but i put a range for altitude to be displayed by clocks and let the focused item in `addInteractive` once the screen locks.

### DIFF
--- a/modules/clock_info.js
+++ b/modules/clock_info.js
@@ -54,7 +54,7 @@ let storage = require("Storage");
 let stepGoal = undefined;
 // Load step goal from health app and pedometer widget
 let d = storage.readJSON("health.json", true) || {};
-stepGoal = d != undefined && d.settings != undefined ? d.settings.stepGoal : undefined;
+stepGoal = d.stepGoal;
 if (stepGoal == undefined) {
   d = storage.readJSON("wpedom.json", true) || {};
   stepGoal = d != undefined && d.settings != undefined ? d.settings.goal : 10000;
--- a/modules/clock_info.js
+++ b/modules/clock_info.js
@@ -120,8 +120,10 @@ exports.load = function() {
 
   if (Bangle.getPressure){  // Altimeter may not exist
     bangleItems.push({ name : "Altitude",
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Instead of trying to read 'timeout' from the global settings file, do you think you could maybe just respond to `Bangle.on('lock'`?

You'd have to remove the handler in `.remove` though :)

gfwilliams:
Looks great - thanks!

peerdavid:
@glemco One question: You selected altitude max. with 3000 -- is there a problem with the sensor if the value is above 3000 i.e. measures are not precise or is this value random? I use my bangle also for hiking and climbing, therefore I'm asking because I usually reach more than 3000m :)

glemco:
Well, there's no limit I'm aware of, just I imagined not many bangles would go that far (maybe 4000 would be a better practical maximum, depending on the area where you live).
Basically we could put something like the highest place on earth, but then any linear indicator (like the one in circlesclock) would always show low values unless you are really high up.
I assume if you see on an altimeter something about 3000m (a full bar in this case) you would think you got quite high already, what would be some sensible ranges for you otherwise?

peerdavid:
Thanks for your quick answer!
Ah I got it -- so we not ensure that that the value x must be in min < x < max, its used by clocks like the circlesclock where a max exists in the UI -- then I think it really doesn't matter and I guess 3000 is totally fine :D Sry I thought nothing above 3000 will be shown :D

glemco:
Well, that's how I intended it, at least also on the weather temperature (using a maximum of 70 C would just mess up for most users).
But you are right as this should be documented/agreed somewhere, considering some other clock besides circlesclock may handle these values..

gfwilliams:
> so we not ensure that that the value x must be in min < x < max, its used by clocks like the circlesclock where a max exists in the UI

Yes, exactly. I think this is another thing that could be configurable when clock_info turns into an app. In the UK you could climb the highest peak and it still wouldn't even read 50% :)

PR #2522, opened 2023-01-22
Graphics utils

halemmerich:
* Modifies the default value for stepAngle to allow full circle arcs without hitting the 64 vertex limit of `fillPoly`
* Fixes the use of a wrong parameter name in `radiansToDegrees`


gfwilliams:
Thanks!

PR #2523, opened 2023-01-22
magnav: Outsource tilt compensation to library

nxdefiant:
and let waypointer use the lib

Notify @jeffmer @hughbarney

### DIFF
--- a/apps/magnav/ChangeLog
+++ b/apps/magnav/ChangeLog
@@ -2,6 +2,5 @@
 0.02: Course marker
 0.03: Tilt compensation and calibration
 0.04: Fix Font size
-0.05: Inital portable version 
-
-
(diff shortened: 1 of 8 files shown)


halemmerich:
Cool, could use this in GPS Trekking :)

nxdefiant:
@halemmerich FYI I plan to write a small script that runs at boot time and overwrites the course value in the gps event with the compass direction for lower speeds.

gfwilliams:
Sounds great, thanks!

If this doesn't specifically require 'magnav' we could just pull it out into something in the `modules` folder?

nxdefiant:
Right now starting magnav is the only way to calibrate the compass. So right now I would not put it in modules.

PR #2524, opened 2023-01-22
[backswipe] New app

Kedlub:
This boot service allows you to trigger the app's back action using a left-to-right swipe
It also includes a setting to blacklist (or whitelist) apps where it should be enabled or disabled, for apps like `sleeplog` which is operated using swipes while also having a back button


gfwilliams:
Looks great, thanks!

thyttan:
@Kedlub 

I really like this! 

I noticed one thing that I don't really know how to solve. If I start a keyboard that uses drag/swipe events in a whitelisted app, backswipe will trigger while it ideally shouldn't.

gfwilliams:
Maybe backswipe could check if there are any other event listeners for the swipe event, and if there are it could ignore the swipe?

Kedlub:
@thyttan I'm glad you like it!
And thanks I haven't thought of that, is there actually any clean way to get the registered swipe handlers? Or at least a way to detect if textinput is running. 
I was thinking about making a smart detection if the app has swipes while developing this, but I thought I couldn't get info about registered swipe handlers.
Maybe I could detect the swipe and drag handler in my overridden setUI, and if another app uses it then disable the `backswipe`.

gfwilliams:
You can do stuff like:

```
(Bangle["#onswipe"] instanceof Array)&&Bangle["#onswipe"].length>1
```

to detect if there's >1 handler. It's not documented but honestly at this point it's unlikely to change :)

PR #2529, opened 2023-01-25
[sched, alarm] allow dated Event repeat

lauzonhomeschool:
Dated Events can now be repeated every X number of days, weeks, months or years. (eventually we'd add a new "until" param, but I won't do that until we have a date picker)

e.g. to repeat the Event every 2 months, sched.json would have an entry with:
rp : { interval: "month", num: 2 }


nxdefiant:
Cool, I wonder how do you add the dates, do you use gadgetbridge for that? Do we need some kind of ics uploader in the webinterface?

lauzonhomeschool:
Currently, the Alarm app allows to input dates for the new "Event" alarm type.  Sched already supported dates, but not date repeats - the new part in this PR.  The current idea is to have a Standalone way to manage alarm + title + date, but I'm sure there would be a way to use the same with gadgetbridge, if it's not already done in another app (but your phone may have more "repeat" options than the ones added here, so you'd have to manage compatibility or expand the available repeat types in sched).

nxdefiant:
@gfwilliams I wonder, what is the reason why this PR is not approved?

gfwilliams:
This came in alongside a bunch of other work, and since it touched the internals of `sched` which have historically caused a bit of trouble, I wanted some time to check it out properly.

I've just taken a look, and I'm a bit concerned about what happens for existing alarms with dates (I think they will maybe just end up getting left on and set to the same day), so I might have to make some changes for that before merging.

apps/sched/sched.js
```diff
@@ -84,6 +84,9 @@ function showAlarm(alarm) {
     let date = new Date(alarm.date);
     let rp = alarm.rp;
     switch(rp.interval) {
+      case true:
```
lauzonhomeschool, line 87: Note that the switch is on "rp.interval", which is a new field (e.g. rp : { interval: "month", num: 2 } ).
So this line 87 will not account for rp: true (as far as I know, there was no behavior defined for rp: true for an alarm with a date), but it will account for rp : { interval: true }.
({rp: true}).interval == undefined, so the previous code would just have left the date as before for rp : true.

bobrippling: I've attempted to track this in some typings (#2640), am I right in thinking if we've got a date, we'll have `rp: { interval, num }`, and otherwise we've not got `.date`, so we'll have `rp: true / false` ?
Or could there be cases where we overlap still? (eg `{ date: "...", rp: true }`)

Unfortunately I've not been able to shoehorn `tsc` into typechecking JS successfully yet, so if we do get contributions later on we'll not be able to use the types to check these things just yet.

lauzonhomeschool: You're right, the Alarm UI supports the case without overlap, but the latest commit also supports that, if a given custom app puts { date: "...", rp: true }, the alarm will just be repeated everyday from that day on.
For the other cases, the [readme ](https://github.com/espruino/BangleApps/tree/master/apps/sched) of `sched` is up to date.

PR #2531, opened 2023-01-25
quicklaunch: do not react if clkinfo is focused

glemco:
Allows the swipe actions from the quicklaunch not to run together with actions in clkinfo, making the two work fine together.
I'm not sure this is the right way to do this (setting an attribute in the `Bangle` object), but it's the simplest way I could think of

### DIFF
--- a/apps/quicklaunch/ChangeLog
+++ b/apps/quicklaunch/ChangeLog
@@ -6,3 +6,4 @@
 0.06: Use Bangle.load() to allow 'fast switch' for apps where it's available.
 0.07: Revert version 0.06. This version is the same as 0.05.
 0.08: Respect appRect on touch events
+0.09: Do not react if clkinfo is focused
--- a/apps/quicklaunch/boot.js
+++ b/apps/quicklaunch/boot.js
@@ -10,6 +10,7 @@
 
   Bangle.on("touch", (_,e) => {
     if (!Bangle.CLOCK) return;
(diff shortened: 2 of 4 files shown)


gfwilliams:
Thanks! Yes, that looks like it works well.

I wonder if this should be documented in at minimum the docs at the top of `modules/clock_info.js` though?

peerdavid:
do you use the latest version of bw clock / did you re-install bwclock to ensure that the clkinfo module is updated?

thyttan:
No, will do that now!

thyttan:
@peerdavid Yes, after reinstalling it works. Thank you!

peerdavid:
Once clkinfo is a lib this should no more be necessary 🤗

nxdefiant:
May I recommend to bump version on all affected clocks?

gfwilliams:
> May I recommend to bump version on all affected clocks?

When I get a moment I'll make clockinfo an app, and then this will be a non-issue...

glemco:
> I wonder if this should be documented in at minimum the docs at the top of `modules/clock_info.js` though?

Well, probably yes, other apps may need to do the same thing. I guess that could be done while transforming the clockinfo in an app

gfwilliams:
Just done: de40f314c7037b5ff5cf01078c17d98552d31adf

Tutorial will update soon

peerdavid:
Ohh wow that great, thank you so much!

PR #2533, opened 2023-01-27
widalarmeta & sleepphasealarm fixes

nxdefiant:
widalarmeta: Fix Bell not appearing on alarms > 24h and redrawing interval
sleepphasealarm: Fix dated events alarm on wrong date


gfwilliams:
Great, thanks!

PR #2534, opened 2023-01-28
android - Handle GPS events from Gadgetbridge instead of connection events

halemmerich:
I had a go at using the GPS events from GadgetBridge as soon as they arrive. This removes the need to handle connection events and fixes GPS not working if for some reason (e.g. config) GadgetBridge never sends an GPS event or stops sending them.

### DIFF
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -24,3 +24,5 @@
 0.23: Handle 'act' Gadgetbridge messages for realtime activity monitoring
 0.24: Handle new 'nav' event for navigation
 0.25: Added option to 'ignore' an app from the message
+0.26: Change handling of GPS status to depend on GPS events instead of connection events
+
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -143,7 +143,7 @@
 
         // modify event for using it as Bangle GPS event
         delete event.t;
-        event.satellites = NaN;
+        if (!isFinite(event.satellites)) event.satellites = NaN;
         if (!isFinite(event.course)) event.course = NaN;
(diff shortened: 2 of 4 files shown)


halemmerich:
There seems to be a condition, where GB GPS stays active sending GPS positions, but the bangle side stops reacting, will have look at that before merging. Disconnecting from GB fixed it, so not that easy to diagnose. Logcat showed the Bangle not reacting at all which is a bit strange since I currently have some prints in there.
@gfwilliams Is it possible to send commands through GB to the connected Bangle? Maybe attach a debugger to GB?

gfwilliams:
Very strange... It is possible to send commands via Android intents - or yes you could debug and then call the relevant functions in BanglejsSupport.

Another option is you could ask it to use a custom App Loader which you could then add the ability to send/receive data into.

halemmerich:
Should be fixed in [https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3132](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3132). I'm waiting and testing until that is merged before marking this as ready.

halemmerich:
I have now used this successfully for a few days of hiking. With the current nightly version of GB containing
[https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3132](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3132) I could not get it to break using the metal box method.

gfwilliams:
That's great - thanks! I've just pushed a new Gadgetbridge version to the play store which has your PR in it, so this should be good now.

PR #2535, opened 2023-01-28
widalarmeta - Do not set thousands of timeouts

halemmerich:
The widget calculates the timeout as NaN if there is no next alarm and effectively sets timeouts as fast as it can.

FYI @nxdefiant

### DIFF
--- a/apps/widalarmeta/ChangeLog
+++ b/apps/widalarmeta/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Change font to 5x9 7 segment-style
       Add settings page
       Add option to show seconds
+0.03: Fix settings thousands of timeouts and slowing down everything if no next alarm found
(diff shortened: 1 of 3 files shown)


nxdefiant:
@halemmerich Sorry for that. I only began noticing it recently. Your solution looks similar to mine in #2533

halemmerich:
Ah, didn't even look for pull requests... had suspected the BTHRM at first and just poked my way around until I found widalarmeta :)
Your changes should work just fine. I have used this oneliner to find this issue: `global.setTimeout = ((o) => (a,b,c,d) => {print(a); o(a,b,c,d)})(global.setTimeout);` It just prints all registered callbacks for timeouts and if something misbehaves it becomes very obvious very fast ;)

nxdefiant:
aah good trick, I'll save it somewhere. And sorry for the time and work it caused you.

PR #2536, opened 2023-01-28
widbatpc remove jitter option

nxdefiant:
Add option 'Remove Jitter'='Drop only' to prevent percentage from getting up again when not charging


gfwilliams:
Thanks!

PR #2537, opened 2023-01-29
[TerminalClock] Add default HRM and altitude value

Stiralbios:
Add default HRM and altitude value (less precise but allow to show something quickly instead of waiting for a while)

There is also a lot of refactoring to add the fast loading in the future. However it's not enabled yet cause of a memory leak I am unable to fix. The refactor is here in case I get an eureka moment of if someone want to finish to implement it.

### DIFF
--- a/apps/terminalclock/ChangeLog
+++ b/apps/terminalclock/ChangeLog
@@ -7,3 +7,4 @@
 0.07: Use ClockFace module and rework the settings to be able to personnalize the order of the lines
 0.08: Hide widgets instead of not loading them at all
       Use Clockface_menu for widgets and power saving settings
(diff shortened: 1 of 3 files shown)


thyttan:
When I tried implemented fastload for BW Clock I overlooked removing a handler which then caused a RAM leak. Could that be what you're missing as well? https://github.com/espruino/BangleApps/pull/2421#issuecomment-1366268892 for reference.

gfwilliams:
Thanks! Do you think there's any possibility of using ClockInfo in the future (for example https://github.com/espruino/BangleApps/blob/master/apps/clock_info/lib.js#L340)? It would remove a lot of extra code, and would ensure that other apps could add new info to the clock.

PR #2539, opened 2023-01-29
[Happy Clock] New App

peerdavid:
Just a new happy clock that shows the time, battery, lock state and whether the step goal is reached or not:

screenshot_1

screenshot_2

And here an example when the battery level is low.
image


gfwilliams:
That looks great - thanks!

PR #2540, opened 2023-01-29
chargent: support BangleJS2

bobrippling:
This configures `chargent` to work with BangleJS2 and drops the minified JS, as mentioned in the [previous PR](https://github.com/espruino/BangleApps/pull/2506)


gfwilliams:
Thanks!

PR #2541, opened 2023-01-29
Iconlaunch - Use firmware E.showScroller

halemmerich:
This removes the copied and modified implementation of the scroller in favor of the one provided by the firmware since 2v16.

Do we have 2v16 in the wild for enough time, or should we wait for 2v17 before implementing this without a fallback for older firmwares?


gfwilliams:
Wow, yes. This looks like a good idea. Pretty sure 2v16 has been out long enough now

PR #2542, opened 2023-01-29
widalarmeta: don't show for hidden alarms

bobrippling:
This updates `widalarmeta` to match the default alarm widget, and not show itself when an alarm is hidden.


nxdefiant:
If this gets merged directly after my #2533 I think we can go without version update since I'll be the only one with the old 0.03 version.

gfwilliams:
Thanks, that looks good! I'll merge and add a note in the changelog

PR #2543, opened 2023-01-29
Lightswitch tapToLock

bobrippling:
This adds a "tap to lock" option for the lightswitch widget (and trims whitespace in affected files)


gfwilliams:
Thanks!

PR #2547, opened 2023-01-31
Adding animation to fuzzyw

sir-indy:
Just a little change to add a bit of visual interest when the time changes.

Following the discussion here, I was originally trying to get a general solution to use in multiple places, but all the good ideas were just too slow, so this simply clears the screen and redraws the text.


gfwilliams:
Thanks!

PR #2549, opened 2023-01-31
Settings.log, off,display,log, both

hughbarney:
I have attempted to modify settings app and bootupdate.js to handle debug settings
so that you can have 0=off, 1=display, 2=log, 3=both
I have tested but still get stuff drawn on the screen using log when there should be no output to the screen
so not 100% that I have covered all the bases.
Pull request so you have have a look at my code.

### DIFF
--- a/apps/boot/ChangeLog
+++ b/apps/boot/ChangeLog
@@ -64,3 +64,4 @@
 0.55: Add toLocalISOString polyfill for pre-2v15 firmwares
       Only add boot info comments if settings.bootDebug was set
       If settings.bootDebug is set, output timing for each section of .boot0
+0.56: Settings.log = 0,1,2,3 for off,display, log, both
--- a/apps/boot/bootupdate.js
+++ b/apps/boot/bootupdate.js
@@ -33,11 +33,8 @@ if (s.ble!==false) {
   }
 }
 // settings.log 0-off, 1-display, 2-log, 3-both
-if (s.log>=2) { // logging to file
-    boot += `_DBGLOG=require("Storage").open("log.txt","a");
-`;
-} if (s.blerepl===false) { // If not programmable, force terminal off Bluetooth
-  if (s.log>=2) { boot += `_DBGLOG=require("Storage").open("log.txt","a");`;
+if (s.blerepl===false) { // If not programmable, force terminal off Bluetooth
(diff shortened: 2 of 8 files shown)


hughbarney:
When set to log (2)  the boot0 code has 

_DBGLOG=require("Storage").open("log.txt","a");
_DBGLOG=require("Storage").open("log.txt","a");
LoopbackB.on('data',function(d) {_DBGLOG.write(d);Terminal.write(d);});

I suspect the logging to the console is happening due to the Terminal.write(d);

gfwilliams:
Hi - yes, I think this is broken.

> I suspect the logging to the console is happening due to the Terminal.write(d);

Yes, it will be. To log, you'll need the console on Loopback. So you want:

* Loopback.setConsole for log = 2/3
* Terminal.write(d) only for log = 3

Also:

```
if (s.log>=2) ...
else if (s.log==1||s.log==3) 
```

doesn't work anyway because of log==3 the second bit is never hit. I'll make a change to the PR and maybe you can test it?

gfwilliams:
Sorry - made a mess of that. I think it's ok now. The original code isn't ideal as it's quite duplicated - you've basically got 8 modes - the 4 log modes, but then when programmable=true and =false

hughbarney:
I was not entirely sure what you did here, you seem to adjust my pull request.
To test I decided to download the raw file of bootupdate.js  and upload it to my Bangle through the IDE.
I set the debug setting to log (ie =2) and loaded simplest.app.js with a console.log("Simplest") in the draw function.
Disconnected from the IDE and then reloaded simplest etc.

For 3 minutes I watched the watch face and could not see any debuig messages printed.
To download log.txt I had to reconnect to the IDE I then saw a debug message displayed on the bottom of the screen.
The only way to clear that is to long button reset , load a different watch,  then I reselected simplest clock.
After that no other debug appears.

CONCLUSION:  It appears to be working, but connecting to the IDE means you get some debug printed to the watch screen on connection until you reload an app.

Is this what you meant by caveats ?

hughbarney:
Gordon - its your call.  I would remove it on the basis that if you are debugging something then more than likely you will have switch programmable on.  I never switch programmable off on mine, I'm just not that paranoid that someone on a train is going to hack my watch :)

PR #2550, opened 2023-02-01
chargent: added threshold

notEvil:
Hi,

this changes the method so it would notify early instead of waiting for the peak to materialize. All charge curves I recorded so far show fast convergence at the peak, justifying this approach. *powermanager* does the same but relies on user supplied threshold, so I think this is not a duplicate.


gfwilliams:
Thanks!

PR #2552, opened 2023-02-02
sched: add interface.html

nxdefiant:
- Edit or delete existing events, add new
- Only date, time & message
- Add events from iCalendar file


gfwilliams:
Thanks!

PR #2553, opened 2023-02-02
[Power manager] back button for settings main menu

thyttan:
Back button for the main menu is added with this PR.


gfwilliams:
Thanks!

PR #2555, opened 2023-02-03
'Temperature Graph' Bangle 2 app

CarlR9:
Hi,

This is a Bangle 2 app for creating graphs from readings of its temperature sensor. Hope you like it.

Carl.


gfwilliams:
This looks great - thanks!

CarlR9:
Thanks Gordon.
I've only realised in the last week that the Bangle 1 has a temperature sensor too, (it's not mentioned among its features: https://www.espruino.com/Bangle.js), so I'll be looking into how to get this running on both watches.

CarlR9:
I thought of adding a calibration option to the app until I realised it gave quite accurate temperature readings - after the watch had cooled down. May add it for Bangle 1 support though, if I get that working.

I've added a couple of minor fixes to the current version. I'm new to github, but I assume I need to push those to you for the app to be updated on your app loader?

gfwilliams:
Yes, the Bangle.js 2 temperature sensor is pretty accurate I think. On the Bangle.js 1 the sensor is quite accurate at temperature *difference* - it's just the offset that needs calibration.

> I'm new to github, but I assume I need to push those to you for the app to be updated on your app loader?

Yes, just do another pull request in the same way you did for this one.

If you make changes *before* the PR is merged then the PR gets changed automatically, but after I've merged it your changes stay on your own copy.

CarlR9:
Oh yes - pull request - not push!
Just done that. Thanks.

PR #2556, opened 2023-02-03
qcenter - Cleanup timeout

halemmerich:
This prevents loading the clock after the timeout when an app was fastloaded from qcenter.


gfwilliams:
Great - thanks!

PR #2557, opened 2023-02-04
[backswipe] check for existing swipe/drag handlers

thyttan:
@Kedlub

As per:

> You can do stuff like:
> 
> ```
> (Bangle["#onswipe"] instanceof Array)&&Bangle["#onswipe"].length>1
> ```
> 
> to detect if there's >1 handler. It's not documented but honestly at this point it's unlikely to change :)

_Originally posted by @gfwilliams in https://github.com/espruino/BangleApps/issues/2524#issuecomment-1406230564_

I'll test it in just a minute.

### DIFF
--- a/apps/backswipe/ChangeLog
+++ b/apps/backswipe/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New App!
+0.02: Don't fire if the app uses swipes already.
--- a/apps/backswipe/README.md
+++ b/apps/backswipe/README.md
@@ -1,0 +1,16 @@
+Service that allows you to use an app's back button using left to right swipe gesture.
+
+## Settings
+
+Mode: Blacklist/Whitelist/Always On/Disabled
+App List: Black-/whitelisted apps
+standard # of swipe handlers: 0-10
+standard # of drag handlers: 0-10
+
+Standard # of handlers settings are used to fine tune when backswipe should trigger the back function. E.g. when using a keyboard that works on drags, we don't want the backswipe to trigger when we just wanted to select a letter. This might not be able to cover all cases however.
+
+## Creator
+Kedlub
+
+## Contributors
+thyttan
--- a/apps/backswipe/README.md
+++ b/apps/backswipe/README.md
@@ -4,8 +4,9 @@ Service that allows you to use an app's back button using left to right swipe ge
 
 Mode: Blacklist/Whitelist/Always On/Disabled
 App List: Black-/whitelisted apps
-Standard # of swipe handlers: 0-10
-Standard # of drag handlers: 0-10
+Standard # of swipe handlers: 0-10 (Default: 0, must be changed for backswipe to work at all)
+Standard # of drag handlers: 0-10 (Default: 0, must be changed for backswipe to work at all)
+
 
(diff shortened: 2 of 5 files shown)


thyttan:
I find it's a little hard to decide when backswipe should be active. For example on my setup there is 2-3 swipe handlers active at a time, and sometimes a drag handler.

Right now I think the most feasible solution is to have several "tiers" of black- and whitelists: 
- Always on, 
- On if more than x # of swipe handlers, 
- On if more than y # of drag handlers. 
An app would have to satisfy both the swipe whitelist and drag whitelist in order for backswipe to be active. This way the user can fine tune to match their setup.

But ideally backswipe could do these judgments, could it index whether swipe handlers use the LR-direction?

gfwilliams:
> I assume there is a way to check the first position parameter to see if it's an underscore or a named one...

There is, but it's *super* hacky. I wouldn't be very happy doing that at all.

Maybe the best bet is to add a third argument to `swipe` in the firmware, which passed an object: ` { ud, lr }`. Then, for the main users of swipe (I guess the clock_info is one?) we could set `e.handled=1` and that could then be checked.

We can do that with other event types right now but because swipe just passes int they're passed by value so it's not possible.

thyttan:
>Maybe the best bet is to add a third argument to swipe in the firmware, which passed an object:  { ud, lr }. Then, for the main users of swipe (I guess the clock_info is one?) we could set e.handled=1 and that could then be checked.
>
>We can do that with other event types right now but because swipe just passes int they're passed by value so it's not possible.

Would that noticeably impact the responsiveness of swipes? 
I guess we would also need to establish a hierarchy of swipe/drag handlers as well? (maybe that's what your saying with "main users")

I'm not qualified to say if it's the best way or not :P

thyttan:
I've now got a working solution where backswipe looks to see if there are extra swipe or drag handlers and so will not trigger the `goBack` function. This will probably not cover all cases yet. To do that, per app settings for number of handlers can be added.

But it may still be better to go forward with @gfwilliams solution. I just wanted to get to a working alternative for now.

gfwilliams:
Thanks - this looks good. At least it doesn't need firmware changes... You think it's ready to merge?

Does backswipe work on clocks? Because if so maybe it makes sense for 'clock_info' to be modified such that swipe handlers are only added *when the clockinfo is focussed*. I think right now they are on regardless and that would confuse this.

> Would that noticeably impact the responsiveness of swipes?

I don't think so, no...

thyttan:
>Does backswipe work on clocks? 

It doesn't seem to trigger on BW Clock for me now at least. But I think I remember it did before..

@Kedlub, does/should backswipe work on clocks? My first intuition would be no, but maybe there's a justification for it?

(I think clocks generally don't have back functions set? But there could be I guess, and then maybe backswipe should work.. I don't know)

>Because if so maybe it makes sense for 'clock_info' to be modified such that swipe handlers are only added when the clockinfo is focussed. I think right now they are on regardless and that would confuse this.

I guess that would make sense if backswipe should be active on clocks.

thyttan:
>You think it's ready to merge?

I added a readme with info regarding # handlers I felt needed to be there. Ok by me to merge now. :+1:

thyttan:
Then we seem to be in agreement 🙂 backswipe shouldn't trigger on clocks. If we notice it does somehow that will be a bug to fix.

>but thanks to you it works perfectly.

Thanks! But I suspect this solution will show to not be absolutely perfect in one situation or another 😛😅😉

gfwilliams:
Great - thanks!

Issue #2457, opened 2023-01-06
[Drag event] Hard/impossible to access all edge pixels in some situations


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v16.45

### The bug

When using Dragboard or Touch Keyboard (with the one-to-one input setting enabled) I have noticed I will pretty often have a hard time accessing characters closest to the edge of the screen when draging.

On Dragboard for example I sometimes find it's hard to access the bottom red rectangle to select letters if draging from further up the screen. This doesn't happen if I haven't calibrated the screen, but if I have it can become impossible to drag back down to the red rectangle to select letters. (calibrated like this while testing now: `touchX1: -2, touchY1: 9, touchX2: 195, touchY2: 177`)

I think I remember my problem with Touch Keyboard was present even on uncalibrated screen. I'll try again and come back with an update.

I've written this script below for seeing how the drag event acts close to the edges.

```
// Script for documenting the first/last pixel reported as touched on a drag from outside a screen edge, in on the screen and exiting the screen via the same edge.
//Upload to RAM in Web IDE and see the output in the console window

var lastB = 0;
var enterX, enterY, exitX, exitY, deltaX, deltaY;

function dragHandler(e) {

  // EDIT: Added lines to get direct visual feedback.
  g.clear().setColor(1,e.b,e.b).drawLine(0,e.y,g.getWidth(),e.y).drawLine(e.x,0,e.x,g.getHeight());

  if (lastB==0){
    print('enter via edge:');
    print(e);
    enterX = e.x;
    enterY = e.y;
    // g.dump();  // Uncomment to get graphics dumps to console field automatically.
  }

  if (e.b==0) {
    print('exit via same edge:'); 
    print(e);
    exitX = e.x;
    exitY = e.y;
    
    deltaX = exitX - enterX;
    deltaY = exitY - enterY;
    print('Delta x:', deltaX);
    print('Delta y:', deltaY);
    // g.dump();  // Uncomment to get graphics dumps to console field automatically.
    print('\n');
  }
  
  lastB=e.b;
}

Bangle.on('drag', dragHandler);
```

My findings are roughly (deltas given without sign):
- Top edge seems to not have too much problem, delta close to zero.
- Bottom edge I get a delta between 10-17 most of the time
- Left edge I get a delta between 6-10 most of the time
- Right edge I get a delta between 6-15 most of the time

I did the measurements with the touch screen uncalibrated.

I don't know exactly what solution I'm after here. But:
[...]

PR #2558, opened 2023-02-05
[dragboard] disregard touch screen calibration

thyttan:
>  The solution then I guess is to have sufficiently large borders where input is not needed for the apps functionality. These could be hardcoded or depend on the users touch screen calibration.
> 
> Another solution can be to remove the touch screen calibration on apps where access to the edges is needed and restore it when the app is closed.

_Originally posted by @thyttan in https://github.com/espruino/BangleApps/issues/2457#issuecomment-1418022051_

### DIFF
--- a/apps/dragboard/ChangeLog
+++ b/apps/dragboard/ChangeLog
@@ -5,3 +5,5 @@
 0.05: Now scrolls text when string gets longer than screen width.
 0.06: The code is now more reliable and the input snappier. Widgets will be drawn if present.
 0.07: Settings for display colors
+0.08: Disregard touch screen calibration as to not risk being unable to access
+some characters.
--- a/apps/dragboard/lib.js
+++ b/apps/dragboard/lib.js
@@ -2,7 +2,11 @@ exports.input = function(options) {
   options = options||{};
   var text = options.text;
   if ("string"!=typeof text) text="";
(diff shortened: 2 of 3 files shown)


gfwilliams:
Sorry, but this and #2559 seems like a really bad idea. I'm open to suggestions but the whole point of calibration is to ensure that where you press lines up with what's on the screen **for exactly these circumstances** and it was added at the request of several other users.

**Is it possible that on your watch you *also* have the touchscreen calibration app as well as the OS-provided one, and so it's double-correcting and so messing things up??**

If tapping on part of the screen isn't registering the tap in that area, maybe it's the calibration menu in settings that needs tweaking to ensure it works ok? One option for you is to just not use that touchscreen calibration app? But disabling it for all users just because you have problems doesn't seem great because I think the lack of similar reports shows that maybe it's not an issue for others?

thyttan:
Here's some more info where the last register point is given, and also the calibration data from Bangle.getOptions():
image

(I recalibrated once more via settings before doing this)

gfwilliams:
Ok, thanks! Right - I think that definitely designing to make them more accessible if we can't reach the edge *or* trying to fix calibration is better than this, so I'll close these two - and I'll post on #2457 in a sec

PR #2559, opened 2023-02-05
[Kbtouch] Disregard touch screen calibration

thyttan:
... as to not risk having some characters be inaccessible.

>  The solution then I guess is to have sufficiently large borders where input is not needed for the apps functionality. These could be hardcoded or depend on the users touch screen calibration.
> 
> Another solution can be to remove the touch screen calibration on apps where access to the edges is needed and restore it when the app is closed.

_Originally posted by @thyttan in https://github.com/espruino/BangleApps/issues/2457#issuecomment-1418022051_


gfwilliams:
Let's discuss this on https://github.com/espruino/BangleApps/pull/2558

gfwilliams:
Closing this and #2558 - discussing on #2457

PR #2560, opened 2023-02-05
powermanager - Add logging of power heavy things

halemmerich:
This tries to record information to help diagnose abnormal power use. ~~The values for the activated sensors still seem a bit off and there are some unintended effects when fastloading apps~~.

- [x] More testing
- [ ] Add recording for BT scanning
- [ ] Maybe refactor into a lib to minimize boot time influence when inactive?

grafik

grafik

### DIFF
--- a/apps/powermanager/ChangeLog
+++ b/apps/powermanager/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Remove calibration with current voltage (Calibrate->Auto) as it is now handled by settings app
       Allow automatic calibration on every charge longer than 3 hours
 0.05: Add back button to settings menu.
+0.06: Allow logging of some things using power
--- a/apps/powermanager/ChangeLog
+++ b/apps/powermanager/ChangeLog
@@ -5,3 +5,4 @@
       Allow automatic calibration on every charge longer than 3 hours
 0.05: Add back button to settings menu.
 0.06: Allow logging of some things using power
+      Add widget for live monitoring of power use
--- a/apps/powermanager/README.md
+++ b/apps/powermanager/README.md
@@ -7,6 +7,11 @@ Features:
 * Force monotonic battery percentage or voltage
 * Automatic calibration on charging uninterrupted longer than 3 hours (reloads of the watch reset the timer).
 
+
+## Widget
+The widget shows an approximate current power use. There is a power gauge showing the estimation of the currently used power and the currently active sensor with the biggest power draw.
+G for GPS, H for pulse sensor and C for compass.
+
 ## Internals
 
 Battery calibration offset is set by writing `batFullVoltage` in setting.json 
--- a/apps/powermanager/README.md
+++ b/apps/powermanager/README.md
@@ -9,8 +9,15 @@ Features:
 
 
 ## Widget
+
 The widget shows an approximate current power use. There is a power gauge showing the estimation of the currently used power and the currently active sensor with the biggest power draw.
 G for GPS, H for pulse sensor and C for compass.
+
+## Logging
+
+You can switch on logging in the options to diagnose unexpected power use. Currently the logging can capture the code running from timeouts and intervals and the power up and down of some devices. The captured times are probably not perfect but should be good enough to indicate problems.
+
+Do not use trace logging for extended time, it uses a lot of storage and can fill up the flash quite quick.
 
 ## Internals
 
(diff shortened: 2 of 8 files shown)


thyttan:
Awesome when done. Big if true!

I've had a little shaky battery life the last week or two (I suspect Message List+Fastload Utils). It will be nice to be able to evaluate that after this PR!

gfwilliams:
Wow, this looks very cool, thanks! It could be really handy to help diagnose power usage!

Some kind of README or info on using this would be really handy, and it might be worth adding that the logging will itself impact battery life/performance a little, but it's going to be a great tool!

gfwilliams:
I should add that I'm not 100% certain but it might be worth adding a handler for the `lock` event... The backlight can be quite a big user of power, but it's the OS that turns it on an off to overridding the setLCDpower calls won't help.

`NRF.setScan/findDevices` can also add significant power draw although not much stuff (except BTHRM/similar?) will use that.

You could also use some of the power stats from http://www.espruino.com/Bangle.js2#power-consumption to estimate the power consumption of each peripheral - for instance leaving the compass on doesn't hurt nearly as much as GPS.

halemmerich:
Handling lock events to take a guess at backlight on time is a good idea. Since apps using BT probably often do that in the background as well BT scanning is worth a look as well.

nxdefiant:
This should probe useful. What do you think about adding Bangle.buzz() to the list? Maybe with number of seconds running or something? While the buzzer is noticed by the user it is easy to forget that the it drains the battery when running often.

gfwilliams:
> What do you think about adding Bangle.buzz() to the list? 

To be honest, I'd say it's unlikely to be a huge source of power usage - but I guess if you're doing all this then you could keep track of it...

How are you feeling about merging now?

halemmerich:
Not yet, there are still some unwanted side effects. I get errors in the weather lib while logging is active. Something to do with how the update method is called, there is suddenly a number as parameter instead of the wanted weather object. Needs a bit more debugging on how that can happen, it should have no influence at all on that code.

halemmerich:
I think this now works well enough to be merged. Buzzer and BT tracking and other optimizations for performance or boot times can be done later depending on feedback I reckon.

gfwilliams:
Great - thanks! This'll be really handy

thyttan:
@halemmerich the widget will draw to the screen also when widget_utils.hide() is in use.

thyttan:
@halemmerich The widget draws over the Run+ karvonen screen, which calls widget_utils.hide()

halemmerich:
Hm, I can not reproduce this on my watch. Do you see this immediately, or does it happen sporadically?

thyttan:
Both sporadically and when I press the hw button to wake the watch.

halemmerich:
That should be fixed by the last commit in this PR, maybe try to rewrite settings in case .boot0 is not up to date?

thyttan:
I tried in this order:
Rewrite of settings didn't help.
Remove app and all corresponding files and then install again didn't help.
Reinstall all apps option via app loader didn't help.
**BUT, factory reset, update installed apps, install power manager, enable the widget, install Run+  - solved the issue.** 

So I guess I'll set my watch up manually and see if it keeps working 🙃

thyttan:
I think I've found the culprit in the Light Switch Widget. I believe it influences how the Power Manager widget updates to draw even on when `.hide()`-ing widgets.

The Light Switch Widget will also draw over the karvonen screen if I touch it's area in the top right corner, immediately triggering a redraw of the power manager widget.

So I'll see if I can understand the Light Switch Widget code and tweak it. If I can't manage that, maybe the Power Manager widget can make sure it sets the correct behavior before drawing in some way?

halemmerich:
Widgets are allowed to draw whenever they want. `.hide` changes their draw method to prevent overdrawing. It might be that Light Switch uses a direct reference to it's draw method on touch and as such does not use the changed draw method. Power Manager widget did do this too, but should not since the last commit.

halemmerich:
It is power managers reaction to `Bangle.setLCDBrightness` called by Light Switch which causes it to be drawn. Will have a look at that. Light switch drawing itself is a different problem.

halemmerich:
Fixed in https://github.com/espruino/BangleApps/pull/2601

thyttan:
Thanks!

bobrippling:
If anyone's interested in other power saving, I've opened a (draft) PR for turning off HRM when the watch is stationary and would value your input - #2644

PR #2561, opened 2023-02-06
[mtnclock] Added support for reading weather.json

sabrsorensen:
My apologies if I'm not following a preferred PR process, please let me know if there's a more appropriate way to request submitting these changes.

`weather.json` seems to be the standard way of storing weather data for access by other apps, this change allows the use of weather, owmweather, or other `weather.json` updaters to provide weather updates to mtnclock.


gfwilliams:
This looks great, thanks!

PR #2563, opened 2023-02-07
sleepphasealarm 0.14

nxdefiant:
- Reduce update interval of current time when seconds are not shown
- Limit logging on Bangle.js 1 to one day due to low memory
- Add plot logged data to settings


gfwilliams:
Thanks!

PR #2564, opened 2023-02-07
agenda: show day of the week in date

glemco:
Also fixed some layout and added a couple of Italian translations


gfwilliams:
Great - thanks!

PR #2566, opened 2023-02-08
Change colour

KungPhoo:
Changes the to yellow on < 50% and red on < 20% battery percentage.


gfwilliams:
Nice idea, thanks! Please can you put an entry in the changelog and bump the version number?

gfwilliams:
Thanks - for the future, it is possible to do everything in one PR rather than 3 separate ones

PR #2568, opened 2023-02-08
Update version

KungPhoo:


KungPhoo:
Sorry, I'm such a moron when it comes to using github. Now I changed the code, the json and the changelog.

PR #2569, opened 2023-02-09
[mtnclock] Addressed undefined weather timestamp

sabrsorensen:
Tested with different versions of weather.json that include or exclude the weather and time properties.


gfwilliams:
Thanks!

PR #2570, opened 2023-02-09
Regenerate typescript / main.d.ts

bobrippling:
This regenerates `main.d.ts` from [`build_types.js`] in the espruino repo, after [updates to optionals and other changes].

The main changes are new modules, and some older typings that might have not been regenerated since their jsdoc/comments changed.

New module typings:
- CC3000
- ESP8266
- Flash
- NetworkJS
- Storage
- TelnetServer
- WIZnet
- Wifi
- crypto
- dgram
- fs
- heatshrink
- http
- neopixel
- net
- tensorflow
- tls
- tv

[`build_types.js`]: https://github.com/espruino/Espruino/blob/149a7f6bace3a56c1892de47bd8baacfb3f496a5/scripts/build_types.js
[updates to optionals and other changes]: https://github.com/espruino/Espruino/pull/2322

### DIFF
--- a/typescript/types/main.d.ts
+++ b/typescript/types/main.d.ts
@@ -1,4 +1,5 @@
-// NOTE: This file has been automatically generated.
+// Type definitions for Espruino latest
+// Project: http://www.espruino.com/, https://github.com/espruino/espruinotools// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference path="other.d.ts" />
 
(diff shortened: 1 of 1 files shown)


bobrippling:
I realise it's a very large diff - I suspect things have been shuffled around since the last typings generation, but using `git diff --color-moved --color-moved-ws=ignore-space-change` shows that most lines have been moved around.

Sorting the before and after hunks and then diffing showed me a bunch of more "major" changes, such as the new modules mentioned above.

gfwilliams:
Thanks! I guess you might expect the diff - it's *possible* it's dependent on the order in the file system

PR #2571, opened 2023-02-10
calendar usage changes

nxdefiant:
- Change control to swipe left/right for next/previous month instead of touch left/right side
- Add up/down swipe for next/previous year
- Display scheduler events for current month on touch


gfwilliams:
Thanks!

PR #2573, opened 2023-02-12
taglaunch: add health, cleanup tool

nxdefiant:
Remove app from 'tool' when it has at least one other known tag
Add tag 'health' for apps like Heart Rate Monitor


gfwilliams:
Thanks!

PR #2574, opened 2023-02-12
New app: hasensors: Home Assistant Sensors

rigrig:
Creates battery level+state sensors in Home Assistant over HTTP

Inspired by https://forum.espruino.com/conversations/383470/ "Means of controlling the battery charge voltage limit on Bangle.js 2?"


gfwilliams:
Nice - thanks!

PR #2576, opened 2023-02-12
[quicklaunch] extension adding an extra screen to launch on swipes from

thyttan:
Swipe to the new screen with quicklaunch to get access to new apps with the same quicklaunch actions. Implemented as an app.js file that can be loaded via quicklaunch like any app.

Before changing from draft:

- [x]  check for RAM leaks when using Fastload Utils.

Potential future additions (Probably not in this PR):
- Hint the apps by writing out the name at the appRect borders.
- Access different groups of apps depending on if left, right, up, down swipe or tap was used to access the extension screen (How to best forward that information from quicklaunch.boot.js to quicklaunch.app.js?).
- Make it possible to 'stack' multiple extension screens? recursively add settings entries? Or have a setting for how many extension screens the user wants and then create that many settings entries by looping?


apps/quicklaunch/app.js
```diff
@@ -0,0 +1,43 @@
...
+  g.clearRect(Bangle.appRect);
+  Bangle.loadWidgets();
+  Bangle.drawWidgets();
+
+  // taken from Icon Launcher with some alterations
+  let timeoutToClock;
+  const updateTimeoutToClock = function(){
+    let time=2; // seconds
+    if (timeoutToClock) clearTimeout(timeoutToClock);
+    timeoutToClock = setTimeout(Bangle.showClock,time*1000);
```
thyttan, line 40: Should `Bangle.showClock` be changed to a regular `load()` call here?

thyttan: Changed it.

gfwilliams:
Neat, thanks!

PR #2579, opened 2023-02-15
sensortools: Fix time in gps

nxdefiant:
@halemmerich FyI


halemmerich:
Documentation says the type is `Date`, maybe that needs to be updated as well.

nxdefiant:
The documentation looks correct to me
```
new Date() instanceof Date
true
Date.now() instanceof Date
false
```

halemmerich:
Ah... I got that whole thing backwards 🤓

gfwilliams:
Thanks!

PR #2581, opened 2023-02-16
Adding gpsmagcourse service

nxdefiant:
Replace GPS heading with compass heading when speed is slow or standing still to avoid the heading from jumping around randomly


hughbarney:
This is an excellent idea.  I originally did this in my WayPointer app and was dissapointed to see that someone removed it that when they did a Bangle 2 version.  But the GPS heading is pretty useless unless you are in a car or a boat moving.  it is a shame we cant get the GPS into pedestrian mode as it might improve it.

To be honest when walking I would want the heading to be set from the compass all the time , so I would want the speed to use the compass to be set high to force that to happen all the time.

The only thing missing now is a small app called compasscal that will insert a calibrate compass menu into settings.
This would operate in the same way that MyLocation enables the setting of the the lat and lon for sunrise / sunset calculations etc.

Looking forward to testing this out soon.

nxdefiant:
Thanks.

As I did some work on BangleJs2 support on waypointer I hope it wasn't me that removed that feature from it, not intentionally at least. Also I already thought of adding a setting to waypointer to switch between GPS and compass course..

Currently the compass/GPS threshold can be adjusted between 1 and 20 km/h, default is 6. Should cover any walking speed ;)

I use the "compass"-app to calibrate it.

nxdefiant:
Status: Wait with further testing for espruino/Espruino#2332

nxdefiant:
Should be good to go.

gfwilliams:
Sorry for the delay merging this - not sure how this one fell off my radar

PR #2583, opened 2023-02-19
New widget `widdevst`

notEvil:
See https://notevil.github.io/BangleApps/?q=widdevst


thyttan:
I noticed it would redraw even when `require("widget_utils").hide()` had been called. If that can be fixed I will definitively start using this :)

notEvil:
Should behave now :)

thyttan:
Installed again, it behaved! 🙂

thyttan:
Could the bars be slightly more pronounced in some way? either tweaking the colors or making them wider maybe?

notEvil:
This is probably a matter of taste. Imo they shouldn't distract as long as there is no issue (running out of Flash/RAM space). Though, I wouldn't mind if there is an option to switch to a more pronounced representation (maybe fill the rectangle?).

thyttan:
I'll use it for a while then and see if I still feel I'd want them to be more pronounced. So no need to add options now.

thyttan:
`widdevst` seem indicates significant fluctuations in ram usage while `RAM Widget` does not. Also  calling `process.memory().usage` in the Web IDE console field I don't manage to catch those fluctuations either.

notEvil:
You probably see this because
- https://github.com/espruino/Espruino/commit/cf97ac90419ef6a011e28e861823ea54c98746c1 does fix an issue that causes blocks of memory to remain "used" even though they aren't actually referenced anymore. They get freed on next garbage collection.
- this widget, unlike `widram`, calls `process.memory(false)` which skips garbage collection.

You could call `process.memory()` in regular intervals with `setInterval` to "fix" the issue for now.

gfwilliams:
Nice, thanks! Do you think you could make this available for Bangle.js 2 as well?

It might be worth noting that this will impact battery usage a bit on Bangle.js 2 though. On a simple clock face that updates once a minute it could almost halve the battery life I imagine since it updates every 2 secs? 

On the Bangle.js 1 it's not a big deal since you're updating only when the LCD is on, but on B2 the LCD is always on... So I guess you could update every 2 secs when unlocked, or every 60 when locked which would fix this?

notEvil:
See

- https://github.com/notEvil/BangleApps/commit/fc330c69e8ff6b2bf6585572cd9d1eedab6c5203
- https://notevil.github.io/BangleApps/?q=widdevst

I don't own one so I tested using the emulator

bobrippling:
I've tested on Bangle.js 2 - all works for me

PR #2586, opened 2023-02-19
Update submodule

### LINKED ISSUE #2585: After "Install Default Apps", install of Boot fails with "undefined"
### Affected hardware version

Bangle 2

### Your firmware version

2v16

### App Loader: Install Default Apps fails during "Uploading boot" phase with error message "app install failed, undefined."
Screenshot 2023-02-19 at 15 50 13

### Expected Behavior
Installs default apps

I decided to do this after updating my watch which seemed to hang during the update of "boot".  In the editor there was some strange error message about something wrong with the file for boot which I sadly did not screenshot.  So I went to production app loader page, erased all apps, and then chose install default apps, thinking this would likely address any problems.

Tried install favorite apps as well, but get the same error during installation of boot, so the rest of the apps are not installed.

I am able to install each app individually, and get my watch running again. 

Thanks for all you do!

### Installed apps

boot (0.56)

BartS23:
fixes #2585


gfwilliams:
Argh - thanks!

PR #2587, opened 2023-02-19
New app: `btadv` (and typescript infrastructure)

bobrippling:
This app broadcasts the various sensors over bluetooth, usable from apps such as Strava (hrm). Each sensor can be toggled on/off and seen on the main screen on the app.

It's written in typescript and so includes typings for a few modules too - clock face, clock info, layout.


gfwilliams:
Great, thanks! Nice to see the typescript stuff used by more than one app!

Out of interest I see what looks like an `Object.assign` polyfill in the generated code? Since Bangle.js has that I wonder if we can convince the ts compiler not to include it?

bobrippling:
It was my pleasure!

I'll take a look at the `Object.assign` polyfill, I expect we can convince `tsc` to use the native function - I seem to remember some issue with needing a polyfill, but I think that was just with destructuring. I'll investigate either way 🔍

gfwilliams:
Great - thanks!

PR #2588, opened 2023-02-20
app widdevst: added support for Bangle.js 2

notEvil:
see https://github.com/espruino/BangleApps/pull/2583#issuecomment-1436800917


gfwilliams:
Thanks!

PR #2589, opened 2023-02-20
New clock info: stopw

bobrippling:
A small, simple stopwatch (written in typescript)

This PR also contains regenerated typescript types, from [this upstream espruino PR](https://github.com/espruino/Espruino/pull/2330)


gfwilliams:
Great, thanks!

Do you think it's worth adding some kind of comment to these TypeScript apps in the README (or in the generated JS) like "this app is written in typescript, see `typescript/README.md` for more info"?

bobrippling:
Good idea, sounds like a plan - I'll add that for the existing TS apps in a follow-up

PR #2590, opened 2023-02-20
add dependabot config for submodules

BartS23:
This PR adds a dependabot config file for checking the submodules :)

According to the documentation you have to enable dependabot in the settings.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


gfwilliams:
Ok, thanks - it might be worth just adding it to the submodules themselves though?

I feel like this isn't such a big deal though because it's not like we actually distribute anything that uses the submodules (apart from the CLI) - they're basically only used when developing so security bugs are less of a big deal.

PR #2594, opened 2023-02-22
[runplus] New app based on the "Run" app adding Karvonen UI

thyttan:
Reference the forum thread: https://forum.espruino.com/conversations/383175/

I haven't changed the icon. Just adding a plus sign would make sense, but I won't do it now 🙃.

~~I'm a little worried that I probably have broken git history between run and runplus, so that we won't be able to rebase runplus on top of changes to run if we wanted to. It feels like there should be a better way to do it than what I did.~~ In this new PR I've tried to create a better history. Not sure it will help with rebasing on the Run app, but it should be a little easier to follow if reading through it at least.

If it looks good it's ok to merge by me.


gfwilliams:
This looks great - thanks!

PR #2595, opened 2023-02-22
Add TypeScript details to READMEs

bobrippling:
And in some cases, add READMEs

(this is a follow-up to #2589)


gfwilliams:
Thanks!

PR #2596, opened 2023-02-22
btadv: manually setup TS helpers - use Object.assign()

bobrippling:
Currently we have `tsc` [generate code for es5], because we don't yet have es6 destructuring support in JSV (used [here], for example).                                            
                                                                                                                                                                                         
As a follow on from #2587, this disables the TS poiyfill and points it at our `Object.assign`, so we benefit from these new features, while still keeping the target at es5. 
                                                                                                                                                                                         
[generate code for es5]: https://github.com/espruino/BangleApps/blob/380af9c600121f50b940b83cdb83a5f83914b8a6/tsconfig.json#L4-L4                                                  
[here]: https://github.com/espruino/BangleApps/blob/380af9c600121f50b940b83cdb83a5f83914b8a6/apps/btadv/app.ts#L235

### DIFF
--- a/apps/btadv/app.js
+++ b/apps/btadv/app.js
@@ -1,15 +1,5 @@
 "use strict";
-var __assign = (this && this.__assign) || function () {
-    __assign = Object.assign || function(t) {
-        for (var s, i = 1, n = arguments.length; i < n; i++) {
-            s = arguments[i];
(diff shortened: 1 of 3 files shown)


bobrippling:
I didn't think it was worth bumping the version number/changelog for a no-op change, but happy to do so if you'd prefer

gfwilliams:
Great - thanks!

PR #2597, opened 2023-02-23
ActivityReminder: Add setting to unlock screen

myxor:


apps/activityreminder/settings.js
```diff
@@ -75,7 +75,14 @@
           settings.tempThreshold = v;
           activityreminder.writeSettings(settings);
         }
-      }
+      },
+      'Unlock on alarm': {
+        value: settings.unlock,
```
gfwilliams, line 80: You might want `value: !!settings.unlock,` here?

If `unlock` wasn't previously set (an old install?) this won't know it's meant to be a bool so won't be editable

apps/activityreminder/alert.js
```diff
@@ -26,6 +26,13 @@
     if (!(storage.readJSON('setting.json', 1) || {}).quiet) {
       Bangle.buzz(400);
     }
+
+    if ((storage.readJSON('activityreminder.s.json', 1) || {}).unlock) {
+      console.log("unlock");
```
gfwilliams, line 31: was the print here intentional, or just for debugging?

myxor:
I fixed your remarks @gfwilliams . Thanks!

gfwilliams:
Thanks!

PR #2598, opened 2023-02-23
sleepphasealarm: Fix inverted setting

nxdefiant:
because it was
`format: v => v?"No":"Yes",`


gfwilliams:
Argh, sorry about that! I must have missed that when I put those changes in :(

PR #2601, opened 2023-02-24
Powermanager - Fix wrapping

halemmerich:
* Fixes of by one error in indexing the arguments for the wrapped function. This causes the timeout to be given as first argument to functions instead of the actual first argument.
* Fixes widget drawing even if hidden

### DIFF
--- a/apps/powermanager/ChangeLog
+++ b/apps/powermanager/ChangeLog
@@ -8,3 +8,4 @@
       Add widget for live monitoring of power use
 0.07: Convert Yes/No On/Off in settings to checkboxes
 0.08: Fix the wrapping of intervals/timeouts with parameters
+      Fix the widget drawing if widgets are hidden and Bangle.setLCDBrightness is called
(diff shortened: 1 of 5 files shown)


nxdefiant:
Does that fix I can no longer edit alarms when logging is active?
> Uncaught TypeError: Expecting Object, got Number
 at line 26 col 213 in alarm.app.js
....assign(alarm,selectedAlarm);}var time=require("time_utils").de...
                                 ^
in function called from line 17 col 1948 in .boot0
....apply(undefined,arguments.slice(1));let end=Date.now()-start;let f=a.toSt...
                                            ^

Also I think that powermanager.default.json is missing in metadata.json.

halemmerich:
The error message certainly matches the problem caused by the fix. So maybe to slightly probably :)

nxdefiant:
Pasted the wrong file, it is "powermanager.json" that is missing in metadata.json.

halemmerich:
@nxdefiant Found an additional problem with alarms: `Multi Timer` gives a string instead of a function to setTimeout when the timer has finished. This caused the timer to overflow and not raising an alarm. This is now worked around by not really wrapping those string functions.

gfwilliams:
Thanks!

PR #2606, opened 2023-02-25
widlockunlock: tap-to-lock

bobrippling:
`widlockunlock` now has tap-to-lock functionality, and has been converted to typescript (happy to revert this part if prefered though). There's also now a setting to configure where the padlock widget appears.

This also updates the `Widget` type, to reflect changes in espruino/Espruino#2333.


gfwilliams:
Thanks - but yes, please can you revert the typescript change?

I'd rather not change existing apps to TypeScript as it doesn't seem to add much value - it means any potential contributor has to not only also know Typescript, but has to install the relevant tools from NPM and rebuild. It's a bit different to just hitting 'edit' on GitHub.

I don't think having a setting for widget position is a good idea either. We do already have an app that allows you to change the positions of your widgets from one place (https://banglejs.com/apps/?id=wid_edit) so it seems like the last thing we want is a bunch of PRs that reimplement that same functionality separately in every widget? If each widget checks its own settings file at startup it'll really slow things down too.

gfwilliams:
That's great - thanks for sorting that!

PR #2607, opened 2023-02-25
settings app: preserve whitelist when disabled

bobrippling:
Potentially controversial change - remember the BLE whitelist even when it's disabled. If we wanted to do this, we'd have to have some way of migrating user's `settings.json`. I had this originally as I kept toggling the whitelist, although I've not toggled it in a while now, and perhaps the ease of adding a new MAC to the whitelist is ok as it is.

Any thoughts from anyone?


gfwilliams:
Actually, looking at this, why can't we just keep whitelist as an array of MACs and then just check if the whitelist includes a field called `disabled`? Either that or add a field called `s.whitelist_disabled`?

That way no migration is needed.

**However** do you actually use the whitelist at all? Since this got added it seems basically every new Android/iOS device (and maybe even Windows/MacOS too?) use address randomization, which changes the host device's address all the time and makes the options basically useless. It seems like we should almost be considering removing it

bobrippling:
Interesting - I've had the whitelist on permanently for a while now and it does work fine with both my phone (Android) and various laptops (Windows & macOS). So I feel it's a useful feature to have still.

I was in two minds about the whitelist preservation but the address randomisation makes me feel like I would turn it off temporarily, now and again - I've made the `s.whitelist_disabled` change but I'm happy to discuss further

gfwilliams:
Looks great! Thanks for that!

Please could you bump the bootloader version as well? Otherwise when folks update they'll get the setting but it won't actually take effect

bobrippling:
Certainly can - done!

gfwilliams:
Great - thanks!

PR #2608, opened 2023-02-25
widbtstates

bobrippling:
This app is very similar to [widbt](https://github.com/bobrippling/BangleApps/blob/widbt2/apps/widbt/widget.js), except I wanted to know whether I had bluetooth off, on, or connected.

I'm able to know if it's connected or not, and I can tell if it's enabled by hooking `NRF.wake`/`NRF.sleep`, but I can't know this when the widget starts, since these won't have been called. Is it possible to know this, or would it require a firmware change?

Also, I made this a separate app, but if you think it's useful, I'm happy to do the work to merge it into `widbt`.


gfwilliams:
Hi - I think a separate app is fine given this adds quite a bit of extra bulk, but as you say I don't think there is a way to tell if Bangle.js is advertising right now or not.

I guess the easy fix would be to add an `advertising` field to `NRF.getSecurityStatus()` - that would do it.

But maybe it's enough that you just check in the global settings.json file for the bluetooth state? Right now that appears to be the only thing that would mess with bluetooth before the widget starts?

bobrippling:
Interesting, I'd quite like to add `advertising` to `getSecurityStatus()`, if that's alright by you?

bobrippling:
I'm happy with this now - I'd debated asking about adding a `"wake"` event to `NRF` but I think hooking the functions is much simpler

gfwilliams:
thanks! and it looks like this will work ok on older firmwares too

PR #2609, opened 2023-02-26
widalarmeta: use less cpu time each minute

nxdefiant:
Remember next alarm to reduce calculation amount
Redraw only every hour when no alarm in next 24h


gfwilliams:
Neat - thanks!

PR #2610, opened 2023-02-26
widbatpc: Only redraw when values change

nxdefiant:


gfwilliams:
Thanks - what is the reasoning behind this PR though? Is it intended to reduce battery usage, or is it for something else?

Because when off charge it's running once a minute - I wouldn't have thought not doing the draw would have a noticeable impact on battery life. It might be better to just keep things simple instead?

nxdefiant:
Yes, the intention is to reduce cpu time. widbatpc did show up pretty high in the [powermanager](https://banglejs.com/apps/?id=powermanager) logging of "accumulated execution times"

gfwilliams:
Ok, thanks. I just wanted to check this wasn't done just for the sake of it - merging now.

> widbatpc did show up pretty high in the [powermanager](https://banglejs.com/apps/?id=powermanager) logging of "accumulated execution times"

Just to check, could that have been when the Bangle is on charge? When it's charging the widget updates a lot more often as it then doesn't have to care about power usage.

PR #2614, opened 2023-02-27
tempmonitor v3

dapgo:
Visual changes and added Stop/start recording, change BG color, filesize info
by default var v_mode_debug=0


gfwilliams:
Thanks!

PR #2615, opened 2023-02-27
messagesoverlay - Fix handlers being restored incorrectly if message removed

halemmerich:
When GB removes a message the swipe/drag/touch handlers of the currently running app were not restored correctly.


gfwilliams:
Thanks!

PR #2617, opened 2023-02-28
weatherClock major update

lunctis-viribus:
I wanted to remove the wind speed from the weatherClock screen because for some reason quickWeather/Gadgetbridge always transmits wind:0 speed. I ended up making a whole settings file, integrating lib.js of the Weather app and more stuff. The default settings are such that the main screen is not changed when updating the app to v0.06. Only the weather icon reflects the weather condition better now.

README.md:
As of Weather Clock v0.06 the date, day of week, temperature, weather icon and/or wind speed can be hidden in Settings. The icons can be changed to those of the Weather app.
Changelog:
0.06: WeatherClock icons now reflect weather conditions better. Add settings menu to hide elements and to use weather icons of Weather app. Images placed into functions for performance.

Feedback is appreciated.


gfwilliams:
Thanks! Looks good, but for some reason you've got some changes to `core` in this commit which is meaning it conflicts.

Is there a way of doing this PR without that change to `core` in it?

lunctis-viribus:
Hi, I don't think anything is changed outside of weatherClock. I did change weather/lib.js in one of my commits for debugging purposes but have reverted that change. So nothing is changed as far as I know. Do you have details of the core conflict?

gfwilliams:
Check out what it says at: https://github.com/espruino/BangleApps/pull/2617/files

It seems like the `core` submodule has been updated and then committed somehow

lunctis-viribus:
Core says:
Showing 0 changed files with 0 additions and 0 deletions.
2 changes: 1 addition & 1 deletion (I guess this is the weather/lib.js file reverted change)

If you really want I can try to make a new PR later.

lunctis-viribus:
Hi @glemco,

`weatherClock` now has an option to use icons from your `weather` app. However there is an issue which could be solved easily.

In your `drawIcon()` function in `weather/lib.js` you set `monochrome` to `1` as default value (line 161). Can you explain why?

I propose changing the default value to `0` or change it to something like `var monochrome = B2 ? 0 : 1;`. Then `weatherClock` could use your icons. At this moment the `weather` icons in `weatherClock` are invisible because of this.

Thanks!

lunctis-viribus:
Thanks for your response. Yes, setting the `ovr` bpp at 1 will indeed give monochrome images. What is your use case for drawing monochrome images?

For now I will change the `ovr` bpp to 1 but it would of course be nicest to get the coloured images.

glemco:
It's used in the clkinfo (by specification they should be monochrome, I reckon), clocks are supposed to either give the desired colour (see circlesclock) or keep it in the foreground colour (bwclock). 
Anyway it makes sense to have it configurable from the weather lib, I'll send a pull request.

lunctis-viribus:
I just checked it. Looks great! Thanks for your help.

PR #2618, opened 2023-02-28
[Messages light] implement new events management

Rarder44:
The gui is called from the event of the "Message" Library.
Removed the code for the music (it will be managed by a separate app or by the "Messagegui")


gfwilliams:
Thanks!

PR #2622, opened 2023-02-28
[Not Analog] Respect system time format setting

jketterer:
Quick update to the Not Analog clock to respect the system time format setting for 12h time. Utilizing the ClockFace module is probably a better solution, but a much bigger lift than this quick fix :shrug: 

@peerdavid


gfwilliams:
Thanks! Other thing you can do is use the `locale` module but this is fine

PR #2623, opened 2023-03-01
multiple dice roller app

Le-Kat:
This app allows you to select up to 8 different dice to roll at the same time. Tap on one on the spaces to select the dice (d4 -> d6 -> d8 -> d10 -> d12 -> d20 -> d% -> none), and then shake the watch or press BTN to roll.


gfwilliams:
Great - thanks!

PR #2625, opened 2023-03-01
openstmap: display waypoints

pavelmachek:
If waypoints module is available, display the waypoints on the map. Also update the documentation and fix some english in the documentation.

### DIFF
--- a/apps/openstmap/ChangeLog
+++ b/apps/openstmap/ChangeLog
@@ -16,3 +16,4 @@
       Support for zooming in on map
       Satellite count moved to widget bar to leave more room for the map
 0.15: Make track drawing an option (default off)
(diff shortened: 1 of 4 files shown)


gfwilliams:
This is a great idea! Have you tested this without waypoints installed though? I think `var waypoints = require("waypoints").load();` will throw an exception and then break everything?

You'd just need to wrap it in a `try...catch`.

Also, please could you bump the version number and add a line to the ChangeLog too?

pavelmachek:
Thanks for review. Here is fixed version. I guess you want to squash commits while merging...?

gfwilliams:
No, no problem about squashing commits - thanks!

PR #2629, opened 2023-03-03
altimeter: Display sea-level pressure, too, and allow calibration

pavelmachek:
Add display of sea-level pressure and temperature, and allow calibration by setting the altitude. That way, you can get precise altitude if you already know sea-level pressure (from example from forecast), or you can calibrate using known altitude and have precise readings for a while.


gfwilliams:
Nice, thanks!

PR #2631, opened 2023-03-04
[draguboard] Create draguboard app from dragboard

lauzonhomeschool:
By using a bigger part of the touch area (vs the original dragboard), we can have bigger and more distanced characters to improve accuracy.


thyttan:
Cool!

thyttan:
The UI is drawn like this for me:

image

lauzonhomeschool:
That's weird.  What's your firmware version?  Mine is 2v16.7.  Do you see the original dragboard normally?

lauzonhomeschool:
Seems appRect is not initialized at the same time depending on the params, I'll rework it.  Thanks for testing.

PR #2632, opened 2023-03-05
[draguboard] Create draguboard app from dragboard

lauzonhomeschool:
By using a bigger part of the touch area (vs the original dragboard), we can have bigger and more distanced characters to improve accuracy.
(hopefully fixed the location of a few widgets vs previous PR for some configs - at least now it displays correctly also on the web emulator, which didn't have a display as strange as @thyttan, though...)


thyttan:
Sorry, but:

When uploading to ram:
image

When uploading to textInput:
image

lauzonhomeschool:
I presume the host program or some override is changing the fontAlign once the textInput is started.  I don't have that problem with the textInput in the Alarm App.  I just pushed a new commit to this PR to setFontAlign just before drawing the string, as it was in the original dragboard, rather than only once at the opening.  Let me know if that fixes the your problem.

gfwilliams:
Thanks - let's wait until we hear back from @thyttan - in general if you've got some code that's drawing it's best to call `g.reset()` and then set everything up exactly as you want it before you draw

gfwilliams:
Thanks!

thyttan:
@lauzonhomeschool Is there a particular reason for not starting the alphabet with A in the top left of the U-shape? I find the current layout with beginning of the alphabet in the middle a little confusing.

lauzonhomeschool:
Yes, there is a reason: it puts the "A" and "E" as the most accessible vowels, and the goal of the App is to allow to type faster on the Bangle.  I though about putting all vowels on the bottom row, but kept the alphabetical order for now.  If I had more time, I'd make a custom.html UI to make the characters on each row customizable on installation of the App.

PR #2633, opened 2023-03-06
[clockcal] fixed typo in settings.js for DRAGDOWN

MM2forever:
fixed typo in settings.js for DRAGDOWN that did cause the setting to not be saved correctly and hence wasn't settable


MM2forever:
> Thanks! Please could you increase the version number in metadata.json and add a line to the ChangeLog? Then folks will know they can update to fix this.

Hey yes sorry, this is my first ever commit on github and the first for a Bangle.JS app. thanks for the hints :)

gfwilliams:
No problem at all - thanks for contributing, and for getting that all sorted so quickly! Merging now :)

PR #2637, opened 2023-03-07
[message_light] fix and overlay 

Rarder44:
Fix graphic bug;
View via popup while there are other open apps


gfwilliams:
Thanks!

PR #2638, opened 2023-03-08
fix dtlauncher

dapgo:
fixed issue and added touch screen events


gfwilliams:
Hi - it looks like you also changed `helloworld` - but I think you have renamed the app file for some reason and then left the old `app.js`?

dapgo:
> Hi - it looks like you also changed `helloworld` - but I think you have renamed the app file for some reason and then left the old `app.js`?

Yes i forgot to remove the app.js
I  prefer to use a final name than a generic app.js name, at least when there is common app file.

gfwilliams:
> I prefer to use a final name than a generic app.js name, at least when there is common app file.

I'm not a big fan of that, it's much easier to look in a directory and see `icon.js/app.js`/etc and know what's there, and it's also what we do in the example app and the tutorials.

It's not a problem with this one, but please don't start renaming all the files in existing apps.

PR #2639, opened 2023-03-09
[quicklaunch] add hints to extension app

thyttan:


gfwilliams:
Thanks! Do you think you could try not to reformat the JSON files if you don't need to though?

Every other app uses a one-line format for the storage files as I feel like it's a little bit more compact.

thyttan:
Yes, sorry 🫡🙂

It's never my intention to change it, but when I save the file it autoformats and then I forget to fix it here on github. I'll try and change some setting so it doesn't do this again.

PR #2640, opened 2023-03-09
Add typings for `sched`, `locale` and expose via `require`

bobrippling:
I'd like to eventually make this opt-in, so if a user wants to, they can check JS files by popping a little `"apps/myapp/*.js"` into `tsconfig.json`. Currently this doesn't pick up any of the typescript bindings - something I'd like to address.

### DIFF
--- a/apps/alarm/app.js
+++ b/apps/alarm/app.js
@@ -279,7 +279,7 @@ function showEditRepeatMenu(repeat, day, dowChangeCallback) {
       }
     };
   } else {
-    var date = day; // eventually: detect day of date and configure a repeat e.g. 3rd Monday of Month
+    // var date = day; // eventually: detect day of date and configure a repeat e.g. 3rd Monday of Month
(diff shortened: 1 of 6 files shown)


gfwilliams:
Great, thanks!

> I'd like to eventually make this opt-in, so if a user wants to, they can check JS files

That would be amazing! If it could be configured in such a way that it wasn't too over-enthusiastic, and really did just warn/error when there were real problems then I feel like it's the kind of thing that could be enabled by default - or at the very least enabled for all the default apps.

I wasn't aware that type-checking JS was functionality that the TypeScript compiler supported, but that makes the TypeScript decls suddenly look extremely appealing!

bobrippling:
Great! I'll see what I can do. I expect there may be issues with TS assuming variables are `any` (since we won't have type annotations in JS), and so on, but perhaps `tsconfig` can be tweaked for that and it'll be useful

PR #2644, opened 2023-03-10
Pause health HRM readings when stationary

bobrippling:
I'm not sure about how useful this will be, hence the draft state - originally I'd thought about pausing HRM overnight, but doing it based on movement feels more reliable, for example if someone leaves their watch on a bedside table.

A few unknowns:
- [ ] Perhaps I'm being too stingy with the `'accel'` readings and could leave them on for longer - I don't think it draws too much battery
- [ ] The `Math.round(x * 100)` seems to be about right, changes in that by more than a single digit seem to indicate movement for me - I "calibrated" using the [BLE streaming example](https://www.espruino.com/Bangle.js+Data+Streaming)
- [ ] Is saving on HRM overnight (~8 hours) worth having this feature?

### DIFF
--- a/apps/health/boot.js
+++ b/apps/health/boot.js
@@ -1,8 +1,20 @@
 (function(){
   var settings = require("Storage").readJSON("health.json",1)||{};
   var hrm = 0|settings.hrm;
+  let lastSuccess = 0;
+  let paused = false;
+
   if (hrm == 1 || hrm == 2) {
    function onHealth() {
+     if (paused) {
+       // paused - 6am or earlier, remain paused
+       const now = new Date();
+       if (now.getHours() <= 6)
+         return;
+
+       paused = false;
+     }
+
      Bangle.setHRMPower(1, "health");
      setTimeout(()=>Bangle.setHRMPower(0, "health"),hrm*60000); // give it 1 minute detection time for 3 min setting and 2 minutes for 10 min setting
      if (hrm == 1){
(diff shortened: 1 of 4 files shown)


apps/health/boot.js
```diff
@@ -23,6 +26,35 @@
...
+    const lookForMovement = () => {
+      hadMovement = false;
+      x = null;
+      Bangle.on('accel', onAccel);
+      setTimeout(
+        () => {
+          Bangle.removeListener('accel', onAccel);
+          paused = !hadMovement;
+          // check again in 15 or 5 minutes
+          setTimeout(lookForMovement, hadMovement ? 900000 : 300000);
```
bobrippling, line 51: Perhaps `changeInterval` would be more effective here, to avoid burning through timeout ids

halemmerich:
I suspect the acceleration listener wakes the CPU very often. CPU alone uses about 4mA while active HRM uses about 5.5mA if I remember correctly. Maybe using the 10 minute movement average that the health event contains would be more efficient. Health takes heartrate by default every 10 minutes and activates the HRM for a few seconds (until first reading is there). So roundabout 500s active HRM which should come out to under 1mAh per night?

bobrippling:
Yes that makes sense, the acceleration listener I believe is always on (although throttled if the [poll interval](https://www.espruino.com/Reference#l_Bangle_setPollInterval)'s not been set) but yes, perhaps 1mAh isn't worth the cost of firing up JS to run.

Even more so, I think it would be `8 * 60 / 10` HRM checks per night (for 8 hours), so ~100 seconds worth of HRM (2 secs to run like you say), if the user's wearing the watch.

Although if they've not got the watch on, it's ~5.5k seconds worth of HRM (2 mins trying to detect)

gfwilliams:
As-is, I'd say this probably isn't much use - I bet the overhead of running JS on accelerometer events uses a bunch more power than the HRM.

However:

* the 'health' event contains a 'movement' fields which is basically the sum of all the accelerometer events, so at the risk of missing the first 10 mins of HRM data, you could check the 'movement' field to see if the watch is being worn or not with basically zero overhead
* As you noted the accelerometer events automatically slow down when the watch isn't moving around, so I think you could just use Bangle.getOptions and check if the accelerometer poll rate is more than 0.5 second or not.

I should add though - the HRM automatically detects if the sensor is against something or not, and if it isn't it goes into a low power mode which doesn't draw much at all.

So if you just take the Bangle off and put it on its side (rather than back-down on a surface) the HRM won't turn on anyway and will save a bunch of power

bobrippling:
Thanks for clarifying, that's really cool about the HRM's proximity detection - I think that negates any code changes for this kind of thing, so thanks again!

PR #2645, opened 2023-03-11
[quicklaunch] Implement stackable extension screens

thyttan:
- [x] Tweak `settings.js` to facilitate stacking
- [x] Investigate if current approach with the global variable `qlTrace` only works when fastload utils is present. -It did.
   - [x] leave a trace in quicklaunch.json


apps/quicklaunch/app.js
```diff
@@ -8,44 +8,58 @@
...
     mode: "custom",
     touch: touchHandler,
     swipe : swipeHandler,
-    remove: removeUI // Compatability with Fastload Utils.
+    remove: ()=>{if (timeoutToClock) clearTimeout(timeoutToClock);} // Compatibility with Fastload Utils.
   });
 
   g.clearRect(Bangle.appRect);
-  Bangle.loadWidgets(); // Compatability with Fastload Utils.
+  let loadWidgets = "Bangle.loadWidgets();"; // Hack: Fool Fastload Utils that we call Bangle.loadWidgets(). This way we get the fastest possibe loading in whichever environment we find ourselvs.
```
thyttan, line 45: @halemmerich is this smart or dumb?

halemmerich: I suspect in case of slow loading you don't want to load the widgets at all and just use the full screen to save on loading? In this case faking widget handling will probably work fine. You could just use a string without storing it in a variable though, just like "ram" for example.

thyttan: Yes that's precisely it. Ok, thanks!

gfwilliams:
Thanks!

PR #2652, opened 2023-03-16
widdevst v0.03

notEvil:
Hi,

this version adds GC because memory usage would otherwise grow continuously, albeit slow. For instance, function definitions inside function calls create reference cycles that prevent objects from being freed immediately. Reporting memory usage without GC is therefore misleading.

https://notevil.github.io/BangleApps/?q=widdevst

It still reports current memory usage every 2s because quickly increasing memory usage is an indicator for app/system (design) issues. If you disagree, please let me know.

Additionally, storage usage is updated less frequently, and element positioning is slightly improved.

### DIFF
--- a/apps/widdevst/ChangeLog
+++ b/apps/widdevst/ChangeLog
@@ -1,2 +1,3 @@
 0.01: First version
 0.02: Support for Bangle.js 2
+0.03: Update storage usage and perform GC every minute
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - yes, that makes sense. Doing GC every 2s might create minor pauses in scrolling/etc sometimes, and will hurt battery life - but since you're only doing that when unlocked and then just doing every minute otherwise, it seems like a good trade-off

PR #2655, opened 2023-03-18
Fix a menu crash in alarms app

blm768:
The alarm app menu was "crashing" when trying to create a new alarm that didn't yet have a message. This seems to fix that issue.

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -43,3 +43,4 @@
       Display alarm label in delete prompt
 0.39: Dated event repeat option
 0.40: Use substring of message when it's longer than fits the designated menu entry.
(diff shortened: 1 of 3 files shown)


thyttan:
Thanks for fixing that, I introduced that bug it seems.

I think your changes need to be applied to the timer menu following [line 386](https://github.com/espruino/BangleApps/blob/468fff9ad7a36552211a75e12e2cc39bd05f20dc/apps/alarm/app.js#L386) as well.

blm768:
Ah, so I do. Updated.

gfwilliams:
Great - thanks for getting this fix in so quickly!

@thyttan please could you try test out your PRs on core functionality a bit more? It seems like even adding a single alarm via the Alarms app would have shown this wasn't working?

PR #2656, opened 2023-03-18
[quicklaunch] Don't go down extensions path if nothing waits at the end, ...

thyttan:
... fixing/removing an unresponsive app state.


gfwilliams:
Thanks!

PR #2658, opened 2023-03-22
Improved Japanese translation.

tonykakuuu:
Improved the Japanese translation. Some strings are still a bit too long for the UI, which could be fixed by using more kanji and fewer loanwords (which tend to be used on similar devices).


gfwilliams:
Great - thank you!

PR #2659, opened 2023-03-25
[Fastload Utils] Add app history functionality

thyttan:
WIP

@halemmerich 

The app history is used to navigate back to previous apps when going back from the current app.

To do before converting PR from draft:
- [x] Run it at home for a while.
- [x] remove widget draws inside timeouts.
- [x] Clean up after testing.
- [x] Clear the app history if the user chosen clock app is loaded? - if `.bootcde` is loaded history is reset.
- [x] fix the situation: 1) Clock face is active, the app history isn't cleared. 2) Message arrives and auto-opens. 3) Message auto-closes. 4) The app to load is pulled from the history, when ideally the clock would be displayed again instead.

Can be loaded here in the meantime: https://thyttan.github.io/BangleApps/?q=fastload

Should this be a separate new app, does the added code contribute too much to loading times? (Not in my experience so far. No timings done)

### DIFF
--- a/apps/fastload/boot.js
+++ b/apps/fastload/boot.js
@@ -1,5 +1,6 @@
 {
-const SETTINGS = require("Storage").readJSON("fastload.json") || {};
+const s = require("Storage");
+const SETTINGS = s.readJSON("fastload.json") || {};
 
 let loadingScreen = function(){
   g.reset();
@@ -16,26 +17,26 @@ let loadingScreen = function(){
   g.flip(true);
 };
 
-let cache = require("Storage").readJSON("fastload.cache") || {};
+let cache = s.readJSON("fastload.cache") || {};
 
 let checkApp = function(n){
   // no widgets, no problem
   if (!global.WIDGETS) return true;
-  let app = require("Storage").read(n);
+  let app = s.read(n);
   if (cache[n] && E.CRC32(app) == cache[n].crc)
-    return cache[n].fast
+    return cache[n].fast;
   cache[n] = {};
   cache[n].fast = app.includes("Bangle.loadWidgets");
   cache[n].crc = E.CRC32(app);
-  require("Storage").writeJSON("fastload.cache", cache);
+  s.writeJSON("fastload.cache", cache);
   return cache[n].fast;
-}
+};
 
 global._load = load;
 
 let slowload = function(n){
   global._load(n);
-}
+};
 
 let fastload = function(n){
   if (!n || checkApp(n)){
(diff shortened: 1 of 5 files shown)


halemmerich:
Cool idea :+1:

apps/fastload/boot.js
```diff
@@ -50,17 +51,55 @@ let fastload = function(n){
...
+let recordHistory = function() {
+  print("record: ", appHistory);
+  s.writeJSON("fastload.history.json", appHistory);
+};
+
+let initBootcdeRun;
+let setInitBootcdeRunCheck = function() {
+  initBootcdeRun = true;
+  print("init: ", initBootcdeRun);
+  load(".bootcde"); // fastload will not act on the initial load of ".bootcde" it seems (e.g. when exiting settings), at least not in the same way as subsequent calls. To work around this ".bootcde" is loaded once more on init. This makes the logic involving the bool `initBootdeRun` work as intended. // However, this seems to be somewhat unstable, causing errors when trying to write changes to "fastload.5.boot.js" ("Uncaught ReferenceError: "e" is not defined", "Uncaught Error: File already written with different data", "Uncaught Error: Too much data for file size"). // Why does exiting the settings generate the Espruino logo in the Web IDE console field while a long press of HW button does not? This difference seems to be a key to making this logic work without this extra load of ".bootcde" on init.
```
thyttan, line 70: @halemmerich do you have an idea regarding the comment on line 70 (after `load(".bootcde");` ?

apps/fastload/boot.js
```diff
@@ -50,17 +51,46 @@ let fastload = function(n){
...
+let recordHistory = function() {
+  print("record: ", appHistory);
+  s.writeJSON("fastload.history.json", appHistory);
+};
+
 Bangle.load = (o => (name) => {
   if (Bangle.uiRemove && !SETTINGS.hideLoading) loadingScreen();
+  print(name);
+  if (SETTINGS.useAppHistory){
+    if (name && name!="fastload.5.boot.js" && name!=".bootcde" && name!="quicklaunch.app.js") {
```
thyttan, line 70: I'm unsure about `name!="fastload.5.boot.js"`, need to test with and without.

apps/fastload/boot.js
thyttan: I've reverted this.

thyttan:
removed commits that introduced `setTimeout(Bangle.drawWidgets,0);` in `onchange` functions.

gfwilliams:
How are you feeling about this now? It's been a while.

Battery usage could have been that issue with the `buzz` implementation?

thyttan:
> How are you feeling about this now? It's been a while.
> 
> Battery usage could have been that issue with the `buzz` implementation?

I'll make a judgement sunday night :+1:

thyttan:
I think it's ok to merge this. I've lost about 50 % of the battery in about four days. A little much maybe, but not very bad. And I suspect message handling more then app history for slurping the juice (I don't really have any data to back this up, just a hunch).

gfwilliams:
Great, thanks!

PR #2660, opened 2023-03-25
Add new clock: NES clock

eyecreate:
Adding new clock that has a few themes based on NES game title screens.

I was hoping to modularize which data structures(fonts and images) were being pulled in, but couldn't find something like that(dynamic include). Luckily it all fits in fine for now.


gfwilliams:
Wow, nice!

PR #2664, opened 2023-03-27
Fix DST's settings.js IIFE

bobrippling:
When loading DST, it presents an [invalid settings] error.
The trailing `;` prevents the DST settings function from being returned to the settings app. Removing this resolves being unable to view DST settings.

[invalid settings]: https://github.com/espruino/BangleApps/blob/0a2f308d2bda3b1562e008c4f0da37914372a960/apps/setting/settings.js#L818-L820

### DIFF
--- a/apps/widdst/settings.js
+++ b/apps/widdst/settings.js
@@ -179,4 +179,4 @@
 
   E.showMenu(dstMenu);
 
-});
+})


bobrippling:
Perhaps I'm missing something - a lot of apps seem to have a trailing semi-colon, yet this fixed DST for me

<details>
<summary>

```bash
$ for f in apps/*/*setting*.js ; do x=$(tail -1 "$f" | grep ') *;') && echo "$f: $x"; done   
```

</summary>

```bash
apps/activepedom/settings.js: });
apps/agpsdata/settings.js: });
apps/antonclkplus/settings.js: });
apps/banglexercise/settings.js: });
apps/barclock/settings.js: });
apps/berlinc/settings.js: });
apps/bikespeedo/settings.js: });
apps/bthrm/settings.js: });
apps/carcrazy/settings.js: });
apps/chimer/settings.js: });
apps/circlesclock/settings.js: });
apps/clicompleteclk/settings.js: });
apps/clockcal/settings.js: });
apps/cogclock/settings.js: });
apps/cprassist/settings.js: });
apps/dane_tcr/settings.js: });
apps/dragboard/settings.js: });
apps/draguboard/settings.js: });
apps/drinkcounter/settings.js: });
apps/dtlaunch/settings-b2.js: });
apps/ffcniftyb/settings.js: });
apps/gbmusic/settings.js: });
apps/getup/settings.js: });
apps/gipy/settings.js: });
apps/gpsrec/settings.js: })();
apps/gpssetup/settings.js: })();
apps/hworldclock/settings.js: });
apps/iconlaunch/settings.js: });
apps/infoclk/settings.js: });
apps/largeclock/settings.js: });
apps/launch/settings.js: });
apps/messagelist/settings.js: });
apps/messages/settings.js: });
apps/metronome/settings.js: });
apps/owmweather/settings.js: });
apps/pastel/test-settings.js: console.log(settings);
apps/pebble/pebble.settings.js: });
apps/pebbled/pebbled.settings.js: });
apps/qcenter/settings.js: });
apps/saclock/settings.js: });
apps/sched/settings.js: });
apps/score/score.settings.js: })();
apps/sensortools/settings.js: });
apps/setting/settings.js: showMainMenu();
apps/slomoclock/settings.js: });
apps/slopeclockpp/settings.js: });
apps/speedalt/settings.js: });
apps/speedalt2/settings.js: });
apps/swp2clk/settings.js: });
apps/taglaunch/settings.js: });
apps/timecal/timecal.settings.js: });
apps/timerclk/settings.js: });
apps/toucher/settings.js: });
apps/touchtimer/settings.js: });
apps/trex/settings.js: });
apps/weatherClock/settings.js: });
apps/wid_edit/settings.js: });
apps/widalarmeta/settings.js: });
apps/widbaroalarm/settings.js: });
apps/widbatwarn/settings.js: });
apps/widbgjs/settings.js: });
apps/widgps/settings.js: });
apps/widhrm/settings.js: });
apps/widmp/settings.js: });
apps/widsleepstatus/settings.js: });
```

</details>

bobrippling:
> My guess is that you have enabled:
> 
> ```
> Minify apps before upload (BETA, not recommended. Uploads smaller, faster apps but this may cause some apps to stop working)
> ```
> 
> Is that the case? because if so I don't think we should be changing the settings file (and actually I'm not sure this change would fix the problem anyway if you re-uploaded the file via the app loader).

Ah yes - that'll be it. Thinking about it, JS has a "last evaluated value" (I can't remember the exact terminology) which perhaps explains why we still get the function despite the semi-colon.

So perhaps it's a minifier bug? Oddly I can't seem to reproduce it now (loading from the app loader) - I'd guess we'd want to make some changes [around here](https://github.com/espruino/EspruinoTools/blob/master/plugins/minify.js#L208-L213) if I can get it to crop up again?

gfwilliams:
It's not really a minifier bug - the minifier is seeing a bunch of code that has no effect and is removing it.

I think if it happens again we don't really want to change the Espruino tools, but there is already a hack around settings here:

https://github.com/espruino/EspruinoAppLoaderCore/blob/425c4a98aed7c4d9b640e37463b534a45a20def7/js/appinfo.js#L115-L122

And actually looking at that regex it doesn't check the match is at the beginning/end of the file, so it should apply even when the file ends in `});` so I'm not really sure how it could have happened in your case?

I think I'll close this now, but maybe reopen if you find a way to reproduce.

It's possible you were using your own fork of the app loader that had an older AppLoaderCore submodule that didn't have that workaround in it?

Personally I think we should just get rid of the minify option. It's a relatively small speed improvement and I feel like it's causing instability/brokenness for a lot of people, not to mention a big support headache for me.

bobrippling:
Yeah that makes sense - I believe I was on the official app loader, but perhaps I'd had the tab open before a recent AppLoaderCore change, so was effectively on a version without the workaround? But agreed, I'll close this - I suppose also the pre-tokenise option covers most of what we gain from minification anyway?

PR #2672, opened 2023-04-01
[Long Press, Buzz!] Add longpressbuzz app (boot code)

thyttan:
Buzz on long press (assumed and hardcoded to 1500ms) to indicate the watch will be reinitiated per your command.

Would there be a 'better' way of doing something like this? Starting out I tried implementing it by using the init event and buzzing once the watch was reinitiated. But that involved some manipulation of storage which I can leave out when doing it this way instead.

EDIT: Logic changed per suggestions below.


gfwilliams:
If you just want to buzz when the watch reboots, maybe you could try `if (BTN.read()) Bangle.buzz()` in boot code?

Because generally if the watch reboots the button will still be pressed?

But otherwise what you're doing seems good. You could maybe define vars with `let` in `{` so they're not global, and could always have just one `setWatch` on `both` edges and use `e.state` instead?

gfwilliams:
Thanks! Did you intend to swap it to `if (BTN.read()) Bangle.buzz()`? It seems like the underlying code hasn't changed?

thyttan:
Yes, thanks for noticing. I thought I did that. Will fix it tonight.

thyttan:
Did a fixup of commit https://github.com/espruino/BangleApps/pull/2672/commits/e34728849f0fba853ab27abf229e4d1bb6ec3c5c ("change the logic to buzz at boot instead of before reboot")

thyttan:
Ready for merge.

gfwilliams:
Thanks!

PR #2675, opened 2023-04-03
grocery interface: Automatically scroll down on add

nxdefiant:
no version update since app is unchanged.


apps/grocery/interface.html
```diff
@@ -46,7 +46,6 @@ <h4>Add a new product</h4>
 
       var $name = document.getElementById('add_product_name')
       var $form = document.getElementById('add_product_form')
-      var $button = document.getElementById('add_product_button')
```
nxdefiant, line 49: unused

gfwilliams:
Thanks!

PR #2676, opened 2023-04-03
[encourageclk] - draw update

elykittytee:
fixed the order watchface calls draw


gfwilliams:
Thanks! Please can you bump up the version number and add a line to the Changelog?

elykittytee:
Okay! bumped and added

gfwilliams:
Looks good, thanks!

PR #2678, opened 2023-04-05
[Swipe Menus] pull in 'Ensure that changing a menu item when half-scrolled off scre…

thyttan:
…en doesn't overwrite widgets' from the standard scroller


gfwilliams:
Thanks!

Issue #2677, opened 2023-04-05
Sleeplog App creates hundreds of empty files


Moini:
### Affected hardware version

Bangle 2

### Your firmware version

2v17

### Issue

The sleep log app created hundreds of empty log files. Not all of them appear in the webinterface for the app, where logs are listed and can be deleted from.

I've just deleted files starting from 2007 (I think) up to 2019, in two week steps per file, one by one, and this took ... eternally. When I check the files in the Files app on the bangle, or in the storage view in the IDE, I can see that there are hundreds more that do not appear in the web interface, and that are empty (Storage app shows `[]`) - I see no reasonable way to delete them. The 'Files' app apparently crashed when trying to remove one of them, and the IDE Storage view reloads after deleting every single file, which takes about 20 seconds or longer for one file.

### What should have happened

* in the web interface:
  * all sleeplog log files should show up, not only those with numbers > 1000
  * multiselect to quickly remove multiple log files at once
  * on mobile:
    * buttons must be larger, it's near impossible to hit them in gadgetbridge bangle.js
    * screen rotation mustn't make the app return to the start screen
 * in the IDE:
   * multiselect for files in storage to delete them
   * maybe consider not doing a full reload?
* in the sleeplog app:
  * only logs for existing dates should be created, not for 10 years back into the past
  * empty logs could be autodeleted...     
  
### Additional questions

How do I delete the sleeplog files 0 to 999 via js from the webIDE? What does the term 'trash files' mean (I apparently have 6500kb of trash files) and can that be emptied? I looked in the Software reference, but it seems it is only possible to list them, not to get rid of them?

### Installed apps

Probably too many:

[...]

PR #2679, opened 2023-04-05
[sleeplog] Improving triggers and web interface

storm64:
to 0.13:
- Prevent to stay in consecutive sleep if not worn.
- Correct calling triggers where last time is earlier then first time.
- Add the trigger object itself as argument to the fn function to be able to store data or functions inside the trigger object.

to 0.14:
improvements initiated by #2677
- Add a option to delete all logfiles before a specific date from the watch
- Correct regex file search from `\d\d\d\d` / `\d{4}` to `\d+`


gfwilliams:
Looks great - thanks!

PR #2680, opened 2023-04-06
chargent: show alert when charge peak reached

bobrippling:
This causes `chargent` to display an alert when charged.

I will occasionally be away from my desk and hear what I think is my watch buzzing while charging. Sometimes it's other clattering - I'd like for `chargent` to display a little alert (`E.showAlert()`) so I don't have to wait 30 seconds to check. @notEvil, is this something you wouldn't mind me adding in?

Currently chargent is "zero-config" although there is [a comment](https://github.com/espruino/BangleApps/blob/master/apps/chargent/boot.js#L27-L29) about making the buzz customisable. I feel adding an option just for showing the alert might not be worth it, but equally don't want to jar people's use-cases.

### DIFF
--- a/apps/chargent/boot.js
+++ b/apps/chargent/boot.js
@@ -27,6 +27,7 @@
             // TODO ? customizable
             Bangle.buzz(500);
             setTimeout(() => Bangle.buzz(500), 1000);
+            E.showAlert("Charge peak reached").then(() => load());
           }
         }, 30*1000);
       }


apps/chargent/boot.js
```diff
@@ -27,6 +27,7 @@
             // TODO ? customizable
             Bangle.buzz(500);
             setTimeout(() => Bangle.buzz(500), 1000);
+            E.showAlert("Charge peak reached").then(() => load());
```
bobrippling, line 30: I believe `load()` is needed, as otherwise the alert draws over a clock face and when dismissed, leaves a blank screen - although this will bump the user out of any app they're in

notEvil:
Thanks, I like the idea! How about, instead of using a fullscreen alert, show a notification

see https://github.com/notEvil/BangleApps/commit/b0b10592cd915c816d0d365f05eb33a636963db2

bobrippling:
That looks ideal in fact! Is that something you plan to PR back to this repo? If so, I'll close this in favour of yours

PR #2681, opened 2023-04-07
chargent: notification

notEvil:
See https://github.com/espruino/BangleApps/pull/2680 for the use case and original idea.

Thanks @bobrippling


gfwilliams:
Neat - thanks!

PR #2682, opened 2023-04-07
[Fast Reset] add app 'Fast Reset' (boot code)

thyttan:
WIP

Reset the watch by holding the hardware button for one second. If 'Fastload Utils' is installed this will typically be done with fastloading. A buzz acts as indicator.

Fast Reset was developed with the app history feature of 'Fastload Utils' in mind. If many apps are in the history stack, the user may want a fast way to exit directly to the clock face without using the firmwares reset function.


gfwilliams:
I'm not quite sure I see how this works? Also doing loadwidgets/drawwidgets in boot code seems like a really bad idea. I would have thought it was likely to break some widgets that may have expected that their boot code was already run.

thyttan:
>I'd use just load() rather than load(".bootcde") though - they should have the same effect.

I suppose you're right. I'll just double check that behaves the same with the WIP logic for app history in #2659.

thyttan:
The way things are implemented now in #2659, if app history is active a `load()` call should go back one step in history while a `load(".bootcde")` call should load the clock face and reset the history. 

See line 67-68 and line 71-75 in `apps/fastload/boot.js` in [#2659.](https://github.com/espruino/BangleApps/pull/2659/files#)

gfwilliams:
>  I emulate a Bangle.showClock() call, so maybe I should change to that?

Ahh, cool - actually that would be great. Then I think we're good to merge?

thyttan:
I changed to `Bangle.showClock()`. But I'd like to wait with merging until I'm confident about the changes to "Fastload Utils".

My Bangle 2 is currently quite battery hungry, ideally I should try and find out if my changes to "Fastlod Utils" could be to blame.

thyttan:
Ready for merge I think.

gfwilliams:
Thanks!

PR #2683, opened 2023-04-07
[Long Press, Buzz!] add icons8 link

thyttan:


gfwilliams:
Thanks! Although I wouldn't worry about this - I pay a relatively pricey icons8 subscription to support them and avoid the need for this (and I believe I still link to them in the repo anyway).

Unless you really feel very strongly about this, I'd say maybe we close this and don't add the acknowledgement - pretty much every other app uses an icons8 icon :)

thyttan:
Ok, I'll close it 👍

PR #2684, opened 2023-04-07
Drained: Low power mode app

### LINKED ISSUE #2502: Feature request: low power mode
When the battery goes below a low battery level, enter low power mode only showing the time, not allowing any other features and no backlight. I imagine this would be baked into the firmware, but I guess an app could accomplish most of these things as well. Inspiration for the request: https://youtu.be/VXSUQSpBYTo?t=327

bobrippling:
This adds an app which (if installed) will automatically load at 5% battery and attempt to put the watch into a low power mode.

(The `setUI` and `removeAllListeners` changes from espruino/Espruino#2348 are included for the build)

Checklist:
- [X] Loads on low battery
- [X] Power cycling restores `drained` app
- [X] Charging reaching a threshold restores original watch app
  - [X] Regardless of whether charging occurs before or while `drained` is running
- [X] Permit user cancellation of low power mode

Closes #2502


pavelmachek:
Looks good to me. I'd enlarge time (as there's enough space on display, and time is important) -- setting font to 55:

.clearRect(Bangle.appRect)
    .setFont("Vector", 55)
    .setFontAlign(0, 0)
    .drawString(timeStr, x, y)

bobrippling:
Sounds good!

thyttan:
Tried it as well, seems to work as intended. Well done and thanks! :)

Will report back if I notice a problem.

thyttan:
Tested some more and have two suggestions:

1. Maybe call [clearWatch](https://www.espruino.com/Reference#l__global_clearWatch) in addition to Bangle.removeAllListeners? I can press the hardware button to enter my launcher. Expected behaviour for me would be that I shouldn't be able to do that.

2. I'd like it if bluetooth was turned off as well.


(I changed the min battery level to 90 and the interval to 6 seconds while testing. I have 'Fastload Utils' installed.)

myxor:
I would like to have the 5% threehold configurable.

Apart from that it looks good.

bobrippling:
Nice ideas - they're now all configurable and bluetooth should turn off as well as clearing the setWatches

thyttan:
When my bangle is connected to the Web IDE, it will advertise some info at some interval, among which I think is battery level. Could this be used with a 'Bangle.on(...)' instead of the 'setInterval(...)' used currently? In that case, would it be better in terms of not having an extra interval running.

thyttan:
When my bangle is connected to the Web IDE, it will advertise some info at some interval, among which I think is battery level. Could this be used with a 'Bangle.on(...)' instead of the 'setInterval(...)' used currently? In that case, would it be better in terms of not having an extra interval running.

bobrippling:
Yes I have the same - I believe it'll be the [bootgattbat](https://github.com/espruino/BangleApps/tree/master/apps/bootgattbat) app. You're right, it uses `setInterval`, I'm not sure what kind of power impact it'll have though.

I did look into a battery event (for this app in fact), but didn't find one, so I think for the battery we'd need a `setInterval` though.

Having said that, we could have a way for boot code to put itself into a "low power" mode though - perhaps this app could [emit](https://www.espruino.com/Reference#l_Object_emit) an event on the `Bangle` object, and boot code can listen to that event to tweak their intervals (and other potentially power-intensive things).

apps/btadv/app.js
```diff
@@ -379,12 +379,12 @@ var setIntervals = function (locked, connected) {
     if (connected === void 0) { connected = NRF.getSecurityStatus().connected; }
     changeInterval(redrawInterval, locked ? 15000 : 5000);
     if (connected) {
-        var interval = btnsShown ? 5000 : 1000;
+        var interval_1 = btnsShown ? 5000 : 1000;
```
gfwilliams, line 382: Was there a reason for this change? It seems like it might be accidental? I'd be pretty sure it won't have any effect anyway

thyttan:
> I believe it'll be the bootgattbat app.

I have not installed that as far as I know. Will look on the app loader.

typescript/types/main.d.ts
```diff
@@ -90,6 +90,8 @@ type Widget = {
 };
 declare const WIDGETS: { [key: string]: Widget };
 
+type ShortBoolean = boolean | 0 | 1;
```
KTibow, line 93: Why not just use `true` or `false` if it's known beforehand, and cast to Boolean (by `Boolean(1)` or `!!1` or `1!=0`) if it's not?

bobrippling:
> This is the message that I see advertised when connected to Web IDE: `{"t":"status","bat":7,"chg":0}`
> 
> I'd guess it's either from firmware or from the Android Integration app.

@thyttan sorry - I'd totally misread your message and thought you were talking about advertising over BLE. Yeah, I expect it'll be [gbridge](https://github.com/espruino/BangleApps/blob/master/apps/gbridge/widget.js#L258-L260) or [android](https://github.com/espruino/BangleApps/blob/master/apps/android/boot.js#L215-L215), like you say.

---

@gfwilliams, about the `bootupdate.js` / `.boot0` handling - is this as simple as `Storage.erase(".boot0")` followed by `load(<self>);` to "apply" the fresh boot? If so, I'll then run `bootupdate` after a `"charging"` event, if that sounds reasonable?

typescript/types/main.d.ts
bobrippling: So here the intention is to enable this kind of usage, where anything truthy/falsy is expected:

```javascript
Bangle.setHRMPower(1, "myapp"); // equivalent to below - we tend to use 1 (or 0) instead of true/false
```

rather than:
```javascript
Bangle.setHRMPower(true, "myapp"); // longer to type, uses more storage, not the convention
```

And the advantage of typing this as `boolean | 0 | 1` means we don't open the door to any number being permitted, but we retain all the flexibility of javascript, with no runtime penalty (i.e. `Boolean(...)`, `!!1`, etc).

thyttan:
>sorry - I'd totally misread your message

No worries whatsoever. I wasn't very clear to begin with.

apps/btadv/app.js
bobrippling: Ah yes - I've backed this out, `tsc` seems to add it in if it finds a conflicting variable, although it doesn't seem too smart about it.

typescript/types/main.d.ts
KTibow: Gordon probably agrees with you but IMO it's still better to write true, and I think tokenization might help with the memory problems.

bobrippling: Agreed - the main rationale for the change is to support a long-term goal is to typecheck the existing JS, which makes use of this pattern

KTibow: Ah, good point

gfwilliams:
> @gfwilliams, about the bootupdate.js / .boot0 handling - is this as simple as Storage.erase(".boot0") followed by load(<self>);

Yes,  that's right - I think you figured this out though!

**However** I can imagine there's an issue where maybe you leave the low power mode app and start the clock again (either with a long press *or maybe because the battery ran out and then you restarted*), but then the `.boot0` is erased so none of your code (or any of the boot code) runs?

It might be better to write the following to `.boot0` rather than erasing it:

```
if (__FILE__!="drained.app.js") setTimeout(load, 100, "drained.app.js");
```

So now, if it ever gets in that state, the boot code will force it to reload to the drained app again, which will be able to restore the boot code correctly when you exit?

gfwilliams:
> Could it be the case that the user's watch has run out of battery, then they charge it up (without turning it on), then disconnect the charger and turn it on?

Yes, absolutely!

> what would you think of directly loading bootupdate there?

That would be an option, yes - I think you might have to force a `load()` after though, it'd need testing.

... but in a way, if you just checked the `Bangle.isCharging()` when you load your app (so you don't risk missing the charge event) then that might make sense to add anyway.

Also maybe you don't want to switch out of low power mode immediately when the charger is applied? Maybe it should wait until the battery is over 20% or so?

bobrippling:
Yeah that's a good point - I'll update the charging-check / bootupdate logic (along with keeping your [`.boot0` idea](https://github.com/espruino/BangleApps/pull/2684#issuecomment-1511044642)) and give it a test - thanks!

bobrippling:
Thanks - I believe I have everything working now (see the PR description for scenarios I've tested). There's just a small niggle I'd like to clarify with you.

When I eval `bootupdate.js`, I receive an error (that originates in the [bootupdate code](https://github.com/espruino/BangleApps/blob/9cb21ceb0f313e9b353495dba91108dc446464c4/apps/boot/bootupdate.js#L32-32) code), but only happens when `drained.app.js` is active.

It doesn't block updating `.boot0`, and I believe what's happening is that we successfully write `.boot0`, then we [eval it](https://github.com/espruino/BangleApps/blob/9cb21ceb0f313e9b353495dba91108dc446464c4/apps/boot/bootupdate.js#L172), but for some reason `bleServiceOptions` isn't present as a global, so we hit an error.

If I [catch this error and then `load()`](https://github.com/bobrippling/BangleApps/blob/96c8b9e49dde8a9fdb10bb944b2ffe3b998a288f/apps/drained/app.ts#L108-L112), the watch functions perfectly, but I'd quite like to avoid this workaround. I don't understand how it can happen, as the initial `.boot0` reads as:
```javascript
if (E.CRC32(require('Storage').read('setting.json'))+require('Storage').hash(/\.boot\.js/)+E.CRC32(process.env.GIT_COMMIT)!=...) { eval(require('Storage').read('bootupdate.js')); throw "Storage Updated!"}
E.setFlags({pretokenise:1});
var bleServices = {}, bleServiceOptions = { uart : true};
//                    ^~~~~~~~~~~~~~~~~ declared here
Bangle.HID = E.toUint8Array(atob("..."));bleServiceOptions.hid=Bangle.HID;
//                                       ^~~~~~~~~~~~~~~~~~~~~ used here
```

<summary>
The generated exception
<details>

```
Uncaught ReferenceError: "bleServiceOptions" is not defined
 at line 4 col 228 in .boot0
...ECwA=="));bleServiceOptions.hid=Bangle.HID;
                              ^
at line 68 col 50 in drained.app.js
eval(require('Storage').read('bootupdate.js'));
                                             ^
```

</details></summary>

gfwilliams:
Thanks! I think this is fine to merge - I don't think that error isn't related to your changes.

What firmware version are you on at the moment? I know there was an error I'd seen in the past where there was this kind of `is not defined` error if there was a var statement in an eval, but it should be fixed for 2v17 (if not before).

Perhaps the easiest solution for now is to do:

```
setTimeout(load,0);
eval(require('Storage').read('bootupdate.js'));
```

So then even if bootupdate does fail with an error, `load` will still end up getting called.

bobrippling:
Thanks - I'm on 2v16 which would probably explain it. I'll double check when I upgrade the firmware.

But yes, for the moment I found the [try-catch](https://github.com/espruino/BangleApps/blob/bce9def441eef2b717e8fbd25cc086f3665d55ad/apps/drained/app.ts#L106-L110) handled it, I guess effectively the same as the `setTimeout` fix. And thanks for merging!

pavelmachek:
According to these messages Drained should be merged, and I'd like to try it, but I don't see it in https://banglejs.com/apps/?q= . Am I doing something wrong?

thyttan:
@pavelmachek 

As soon as something is merged here it will show up in the development app loader, afaik. It's accessible via the 'More...' tab on the official app loader. Or follow this link: https://espruino.github.io/BangleApps

@gfwilliams updates the official app loader with some delay. I believe to ensure a higher degree of stability. It's also not automatic if I understand correctly, so it saves time to do it when more changes have been made than just one.

PR #2685, opened 2023-04-08
Update/new futures on rings app

pinq-:
Hi all.

This is my first app edition and I wanted to add/edit some futures in rings app. The app is great, but wanted to to add some more futures. I would like to hear some comments from @lightnin.
The  changes what I made:
- Added inner circle for battery charge
- Added function "drawCircle" that takes whole object in
- Added settings options
- In settings you can select bubble or square marking
- In settings you can select show numbers or not
- Added "drawArc" function for battery view


lightnin:
Hey - this is looking really cool! My focus is mostly on aesthetics here, as I'm not much of a coder - or just enough to recognize that my original code needed some optimizations. Here are some thoughts on this pull request:
1. I didn't expect to like having the battery state indicator present all the time, as I generally like to keep the information offered minimal, to just what's necessary. But the way you did this with another ring and no numbers just looks really nice. 
2. For the same minimalist reasons, I like to limit the number of settings offered. Having the option to show numbers with or without "bubble" backgrounds per ring seems complex - especially when we allow different states for different rings. I prefer non-bubbled numbers. I'd suggest you consider making a single global setting for each value that affects all rings (rather than per-ring). Having different rings in different states with regard to numbers / bubbles doesn't seem so useful or aesthetically nice.  So the settings menu would offer: ( )Bubbles ( ) Numbers.   And it might be worth adding a third menu item for:  ( ) Battery.
Overall this looks really nice - thanks for making these improvements!

lightnin:
One other note - it's worth updating the screenshots with how it looks now (in both states). There's a screen dump capability for the Bangle Js2 documented somewhere - that was the best way of getting images I found with the earlier version.

pinq-:
> Hey - this is looking really cool! My focus is mostly on aesthetics here, as I'm not much of a coder - or just enough to recognize that my original code needed some optimizations. Here are some thoughts on this pull request:
> 
>     1. I didn't expect to like having the battery state indicator present all the time, as I generally like to keep the information offered minimal, to just what's necessary. But the way you did this with another ring and no numbers just looks really nice.
> 
>     2. For the same minimalist reasons, I like to limit the number of settings offered. Having the option to show numbers with or without "bubble" backgrounds per ring seems complex - especially when we allow different states for different rings. I prefer non-bubbled numbers. I'd suggest you consider making a single global setting for each value that affects all rings (rather than per-ring). Having different rings in different states with regard to numbers / bubbles doesn't seem so useful or aesthetically nice.  So the settings menu would offer: ( )Bubbles ( ) Numbers.   And it might be worth adding a third menu item for:  ( ) Battery.
>        Overall this looks really nice - thanks for making these improvements!

Great to here, that you liked :).
1. Yeah, this was added for practical reasons: when I was using the original app, a couple of times I did run out of battery, because I didn't check how much there is left. So I added that. 
2. The "bubble" background I added just for fun and try out. The options I added because I liked how it looks without numbers, but with month day I wanted to see what day it is today. And if I added number to battery level, that didn't look good add all :D so I removed that option. What if I remove hour and minute option and make them have common option? 

That's good idea to update the screenshots. I completely forgot about it.

lightnin:
> What if I remove hour and minute option and make them have common option?
Ah - I see. Having a  date ring without the date represented as a number would be confusing. So if we had a global "show numbers" setting, and if it was off, it would be tough to guess the date. I agree - that would be a problem. 
What you suggest could work, but still has a lot of menu complexity that I'd like to try and avoid. I think it would be easier to just say: There's no case where we don't want the date number displayed.
Let me propose something a little different, which I was thinking of doing anyway.  Let's assume we have a single global menu setting for show numbers. When selected, the number is shown in the notch in all the time circles (including the date) as it is now. When not selected, the notch still appears, but the number of the date becomes part of the Month / Year text in the outermost ring. Currently on my watch that text says: "April 2023", and the "8" is shown in the notch cut into the circle. But if the global "Show numbers" was un-selected, then the text would say "April 8 2023", and the notch would be empty (but still in the 8/31 rotated position.) This would allow for a simpler settings menu, while still showing the date number at all times (one way or another). What do you think?

lightnin:
Also - regarding battery level. I agree it's much better without numbers. I also think that it behaves differently from the other rings - which I like - because instead of indicating amount by "notch" position, it indicates by fullness of circle. For this reason, I think it is understandable that when people select or de-select the global option "numbers" in the settings menu, it does not affect the battery ring. The battery ring never shows numbers no matter what, and that's ok.

pinq-:
I think it's not good idea to add the number to the text, because you can't really reed the text. 
What If I change the option page like this:

- Show number
-->
    - [ ] Date
    - [ ] TIme ( or minute and hour)

- Bubble background
-->
    - [ ] Date
    - [ ] Time

I like that I can see the date number so I would keep it. What to you think?

lightnin:
Yeah, the text on the outer ring is a little hard to read - I guess as a result of the angling operation. That said, I think I can read it well enough to make out the date if we added it there.
I like that the menu options are getting shorter, and less complicated. Great! But I still feel like they are designed to accommodate a use case that I'm not sure I believe in. 
As you pointed out earlier, you always want to see the date as a number on the month ring. I can't really imagine someone wanting to only see the "notch" position in the date ring, without a number to represent the date exactly. With the hour and minute rings, it's easy and intuitive to estimate what the number is based on the position of the notch.  But it's hard for me to imagine that anyone wants only a rough estimate of which day of the month it is - they want to know (and see) the exact number. So - I don't think there should ever be an option to not show what day of the month it is in the outermost month ring (one way or another - inside the notch or somewhere else). Do you agree, or should I try to look at it another way?

But it's far from a terrible solution to use the menu design you suggest, and perhaps you would like to complete this PR and get rev. 2 up there. If so, that's ok with me - please go ahead! Already this PR has some really nice improvements. But if so, maybe we can keep this little problem in mind, and see if we can come up with a better solution in a future revision. 

Thank you for making this PR, and for inviting me to this interesting design discussion.  :)

pinq-:
Made some changes:

- I added the date to text. 
- Changed the text font to vector, because it is more clear. 
- Made bubble option global
- Added some screenshots 

I think it is now better and ready to publish. What to you think?

pinq-:
Done.

gfwilliams:
So are we happy this is all good to merge now?

lightnin:
Yep! One of these days I might try to play with the fonts a bit - but that will have to wait until after my dissertation is finished. As far as I'm concerned this is ready to merge. Thanks for our collaboration @pinq- and thanks, as always, for the awesome ecosystem you've made around this, @gfwilliams . :)

gfwilliams:
Ok, great! I'll get this merged. Looks like it's failing because screenshot2 doesn't exist, but I'll rename the files when I merge

pinq-:
Cool. Thank! I was just about to fix it. Don't know why it wants screenshot2 when there is no reference to it. 
My first merge to the apps!

thyttan:
>Don't know why it wants screenshot2 when there is no reference to it.

It was specified in the metadata as a screenshot to be shown on the app loader, but since the png-file itself was removed it complained that there was no file with that name. :)

image

image

pinq-:
Ou yeah, I forgot the whole metadata. Now it the readme is referring to wrong image. Ou well, maybe it irritates me enough to develop it further :)

PR #2688, opened 2023-04-10
Initial release of widtemp temperature widget

gellnerm:
I wrote a small widget showing the internal temperature. I hope it will be useful to others.


gfwilliams:
Really nice - thanks!

Did you know that you can produce screenshots direct from the Bangle? All you need to do is type `g.dump()` in the left-hand side of the IDE and it'll copy the screen contents to an image for you (which you can then right-click on and save)

PR #2691, opened 2023-04-13
Fix some HRM logic in the health app

KTibow:
I haven't tested this yet, but it should work fine. I noticed some odd formatting and (possibly unintended) logic in the HRM section of the health app, and this PR fixes it.


KTibow:
I made the PR because as it is, if:
- the hrm reading has non-zero confidence at the time of running: it won't turn on the HRM sensor
- the hrm reading has 80+ confidence, but not more than the stored reading: it'll turn off the HRM sensor early

I'll go ahead and try to test it in the IDE.

KTibow:
I don't see any errors, and with some added logging, things seem to be working as they should.

KTibow:
Looks like I should update boot.min.js too, what minifier are you using? Is https://skalman.github.io/UglifyJS-online/ okay?

gfwilliams:
Ahh, yes - sorry.

It's best to clone https://github.com/espruino/EspruinoDocs and use `bin/minify.js` which uses Closure with some special settings. That's what I use.

gfwilliams:
How are you feeling about this now? Still seems to work well, and it's producing better HRM results?

gfwilliams:
Ok, thanks!

PR #2692, opened 2023-04-14
clkinfostopw - set format to h:mm:ss to shorten string

hughbarney:


bobrippling:
The JS that's changed is generated from [the typescript file](https://github.com/bobrippling/BangleApps/blob/low-power-mode/apps/clkinfostopw/clkinfo.ts), that's where we'll want the changes :)

Perhaps this should be an option though, I prefer seeing the "h" or "m" so I can tell at a glance rather than counting the number of digit groups

apps/clkinfostopw/clkinfo.js
```diff
@@ -21,10 +21,10 @@
         var mins = seconds / 60;
         seconds %= 60;
         if (mins < 60)
-            return "".concat(pad2(mins), "m").concat(pad2(seconds), "s");
+            return "".concat(mins.toFixed(0)+":").concat(pad2(seconds));
         var hours = mins / 60;
         mins %= 60;
-        return "".concat(Math.round(hours), "h").concat(pad2(mins), "m").concat(pad2(seconds), "s");
+        return "".concat(hours.toFixed(0)+":").concat(pad2(mins)).concat(pad2(seconds));
```
bobrippling, line 27: Here we've removed the separator between minutes and seconds, so they'll be read as a single long number

hughbarney:
I'll close the PR as it now clashes with other changes I have done.
Personallly I think 01:05:02 is much more intuitive than 010502, maybe the seperator could be a setting.

hughbarney:
@bobrippling, whats your username on the forum, I will add you into a thread I started about stopwtch clock_info which I like a lot.

bobrippling:
> Personallly I think 01:05:02 is much more intuitive than 010502, maybe the seperator could be a setting.

Yeah without the colons is difficult - if you don't mind, I'll remix your original PR into one that adds a setting for the stopwatch to toggle between `01h05m02s` and `01:05:02`?

I've just seen the forum thread - I'll see what I can do about the battery usage too. My username's also [`bobrippling`](https://forum.espruino.com/profiles/199966/), just created an account :)

PR #2693, opened 2023-04-16
widalarmeta: Selectable font. Optional hour padding.

c-mauderer:
Allow to select between bitmap and teletext font. Allow to disable the hour padding.

### DIFF
--- a/apps/widalarmeta/widget.js
+++ b/apps/widalarmeta/widget.js
@@ -1,9 +1,13 @@
 (() => {
   require("Font5x9Numeric7Seg").add(Graphics);
+  require("FontTeletext5x9Ascii").add(Graphics);
+  require("Font4x5").add(Graphics);
   const config = Object.assign({
     maxhours: 24,
     drawBell: false,
+    padHours: true,
     showSeconds: 0, // 0=never, 1=only when display is unlocked, 2=for less than a minute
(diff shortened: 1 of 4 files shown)


apps/widalarmeta/widget.js
```diff
@@ -49,17 +52,34 @@
...
+      var extraPixels = 2;
+      var nrDigits = 5;
+
+      var text = "";
+      if (config.padHours) {
+        text += hours.padStart(2, '0');
+      } else {
+        text += hours;
+        if (hours < 10) {
+          nrDigits -= 1;
```
nxdefiant, line 65: Wouldn't it be easier to calculate the width
from the string length?

c-mauderer: You are most likely right. I clearly haven't thought enough about it. It was a fixed value before and I just adapted it. I'll change that.

c-mauderer: @nxdefiant: I used the `g.stringWidth(...)` function instead. With that it is even easier to add further fonts. As an example I added another 4x5 font option.

nxdefiant: cool and a lot easier to read

gfwilliams:
Thanks! merging now.

c-mauderer:
> Thanks! merging now.

Thank you.

Issue #2686, opened 2023-04-09
LCDclock wont display stop watch clock info, SlopeClock++ does


hughbarney:
### Affected hardware version

Bangle 2

### Your firmware version

2v16.93

### The bug

Install LCDclock, slopeClock++ and stop watch clock_info.
One slopeClock++ you can scroll through the clock_infos and display the stopwatch clock_info.
On LCDclock you cant get the stop watch to appear.

### Installed apps

Installed | clkinfosunrise (0.03), widbata (0.04), widpb (0.05), mylocation (0.09), widlock (0.08), lato (0.02), clock_info (0.02), simplestpp (0.03), slopeclockpp (0.09), widalarm (0.01), tictactoe (0.01), clkinfostopw (0.01), lcdclock (0.02), setting (0.59), boot (0.57), sched (0.22), alarm (0.41), widgps (0.09), gpstouch (0.02), nesclock (0.03), iconlaunch (0.17), messageicons (0.05), widmessages (0.04), android (0.23), gpsinfo (0.12), messages_light (1.5)
-- | --

hughbarney:
not a bug, see thread:
https://forum.espruino.com/conversations/384497/#comment16942919
closing

PR #2695, opened 2023-04-16
Improvements to the stopwatch clockinfo

bobrippling:
Several improvements to the stopwatch clockinfo, inspired by #2692 and #2686:

- Format option added, to reduce space used (see [forum thread on various clocks](https://forum.espruino.com/conversations/384497/))
- Reduced interval to 1s, no point redrawing the same thing 10x per second
- Don't queue the redraw when shown, if we're paused
  - This was a large drain on the battery, shown in the [`powermanager`](https://github.com/espruino/BangleApps/tree/master/apps/powermanager) app (`this.emit(...)`)

### DIFF
--- a/apps/btadv/app.js
+++ b/apps/btadv/app.js
@@ -1,4 +1,3 @@
-"use strict";
 var __assign = Object.assign;
 var Layout = require("Layout");
 Bangle.loadWidgets();
--- a/apps/clkinfostopw/ChangeLog
+++ b/apps/clkinfostopw/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New clkinfo!
+0.02: changed format to h:mm:ss, to reduce size of text string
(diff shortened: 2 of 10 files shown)


gfwilliams:
Thanks! Just to check, are you sure the settings page works ok?

[apps/clkinfostopw/settings.js](https://github.com/espruino/BangleApps/pull/2695/files#diff-73ef0327b0be654435a7b8511c6e1ffa1401e8690c20eb03cbb872431939b095) should really not have `"use strict"` in it, since the code is passed through an `eval(...)` statement in order to load the menu (it's not a separate app)

bobrippling:
Yeah, it did work with `"use strict"`, but I can drop it if you like?
A few other generated files have it (`widChargingStatus`, `widbtstates`, `btadv`), I can drop it from those too? (via a `tsconfig` change)

gfwilliams:
Thanks! Yes, for Espruino the "use strict" has no effect - in apps it's not a big deal (but not adding it in the first place is good). However in the case of the settings file since it's effectively an expression that should be evaluated, having it really isn't great at all - the sanity checker should really have a check in there for stuff like that.

PR #2697, opened 2023-04-19
stopw: only show milliseconds for the first second

bobrippling:
(Currently in draft while we [decide on usability tweaks](https://forum.espruino.com/conversations/384497/#comment16947612))


gfwilliams:
Thanks, that's great!

PR #2701, opened 2023-04-20
Update typescript & remove "use strict"s

bobrippling:
Including using typescript's [`satisfies` feature](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) to avoid having to manually type settings or clockinfo functions.


gfwilliams:
Looks great - thanks!

PR #2702, opened 2023-04-20
New app: xxlmessage

KungPhoo:
The xxlmessage is a replacement for the messageui app and scrolls the incoming message with a very large font across the screen.
If the messageui app is installed, it will be used.

Note:
I did not make any changes to the message app. I don't know, why there are changes. Please ignore these. Github is very difficult to use for me.


gfwilliams:
Ok, thanks - I'm not sure why those other changes have showed up then.

So am I right in thinking that the *only* files that should be in this commit are in `apps/xxlmessage`?

KungPhoo:
Yes, just that folder.

21 Apr 2023 09:26:56 Gordon Williams ***@***.***>:

gfwilliams:
Ok, just pulled this in - I ended up copying the files as the Git commit history has got messed up here.

I'd advise maybe deleting your current fork and then re-forking again? It seems the easiest way to ensure you're back to a clean slate

PR #2706, opened 2023-04-24
Fix `wid_edit`'s `draw()` - pass `WIDGET` as first argument

bobrippling:
When attempting to edit a widget whose `draw` function takes the widget as an argument, we get an exception from that `draw` function. This fixes that by catering for both APIs.

The back button widget expects this, as does [`widsleepstatus`], [`widlock`] and [`widlockunlock`].

However, there are other widgets ([`widmessages`], [`widbatpc`] and [`widalarmeta`]) which use something else for the first parameter. So I'm thinking we change the behaviour of the first three (to use `this` instead of an argument, and revert this `wid_edit` change), but we'd also need to change the back widget - but I'm not sure where this comes from.

Before I dig into that, does this sound like a good approach?

[`widsleepstatus`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widsleepstatus/widget.js#L20
[`widlock`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widlock/widget.js#L5
[`widlockunlock`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widlockunlock/widget.js#L27

[`widmessages`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widmessages/widget.js#L14
[`widbatpc`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widbatpc/widget.js#L89
[`widalarmeta`]: https://github.com/espruino/BangleApps/blob/2f6862024d6b92a11fd05a5183a1dd9a53dc5e48/apps/widalarmeta/widget.js#L28

### DIFF
--- a/apps/wid_edit/ChangeLog
+++ b/apps/wid_edit/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Wrap loadWidgets instead of replacing to keep original functionality intact
       Change back entry to menu option
       Allow changing widgets into all areas, including bottom widget bar
+0.03: Fix editing widgets whose draw method takes the widget
--- a/apps/wid_edit/metadata.json
+++ b/apps/wid_edit/metadata.json
@@ -1,6 +1,6 @@
 {
   "id": "wid_edit",
-  "version": "0.02",
+  "version": "0.03",
   "name": "Widget Editor",
   "icon": "icon.png",
   "description": "Customize widget locations",
--- a/apps/wid_edit/settings.js
+++ b/apps/wid_edit/settings.js
@@ -67,7 +67,7 @@
     function highlight() {
       if (WIDGET.width > 0) {
         // draw widget, then draw a highlighted border on top
-        WIDGET.draw();
+        WIDGET.draw(WIDGET);
         g.setColor(g.theme.fgH)
           .drawRect(WIDGET.x, WIDGET.y, WIDGET.x+WIDGET.width-1, WIDGET.y+23);
       } else {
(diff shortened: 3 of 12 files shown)


nxdefiant:
widbatpc and widalarmeta both can handle this as first argument because its type its checked. Afaik Bangle.drawWidgets() does also pass this to the draw function.

bobrippling:
Ah I see what you mean, so perhaps this fix is ok as-is in that case

gfwilliams:
Thanks - yes, this fix is fine. Good spot about widmessages, widbatpc and widalarmeta - those really shouldn't rely on passing a different first argument (it's not like it's a big deal making that the second argument) - as @nxdefiant  pointed out, drawWidgets automatically passes the first argument in, and that is built into the firmware so I wouldn't change it.

Are you up for including changes to fix widmessages, widbatpc and widalarmeta as well?

bobrippling:
Of course - for the moment I've chosen to pass in the widget as the second argument, but we could equally pass `null` there, since the same file defines that function and so can trust that it'll be unused - do you think that approach wins much in terms of performance/battery over the current one?

gfwilliams:
This looks great - thanks! Please can you bump the versions and I'll merge?

bobrippling:
Ah yep - done

gfwilliams:
Thanks!

Issue #2562, opened 2023-02-06
Exception "Uncaught Error: Cannot read property 'reset' of undefined" 


myxor:
### Affected hardware version

Bangle 2

### Your firmware version

2v16

### The bug

I am not quiet sure where this error happens but it happens on my side when i try to edit an existing alarm via the alarms app:

```
Uncaught Error: Cannot read property 'reset' of undefined
 at line 2 col 171
...elete Bangle.uiRemove;a()}g.reset();if(c){if("updown"==c){va...
                              ^
in function called from line 2 col 291
...:c.x,y:c.y+f-b.y-a*d.h}))}});var b=Bangle.appRect,h=0|d.scro...
                              ^
at line 2 col 291
...:c.x,y:c.y+f-b.y-a*d.h}))}});var b=Bangle.appRect,h=0|d.scro...
                              ^
in function called from line 1 col 28
m.scroller=E.showScroller(r)
                           ^
at line 1 col 28
m.scroller=E.showScroller(r)
                           ^
in function "q" called from line 7 col 422
....scroller.drawItem(a)}}};q();return m
                              ^
in function called from line 1 col 2677 in alarm.app.js
...10,a,d,g))})}),E.showMenu(e)
                              ^
at line 1 col 2677 in alarm.app.js
...10,a,d,g))})}),E.showMenu(e)
```

i currently got no time to investigate this but maybe someone else can have a look :)

dashavoo:
Try deselecting all the minification options etc. in the app loader and then reinstall.

nxdefiant:
Neither alarm nor sched.js do a g.reset(). Can you please post your installed apps with versions?

gfwilliams:
I think the code that's causing the problem will be in E.showScroller by the look of it? But the issue is really that *something* has overwritten/deleted the global variable `g` which is used for the graphics.

I just tried and I can't reproduce here by editing a pre-existing alarm - maybe you could post some more specific steps? Also see if it happens in the emulator or with a fresh install?

It might also be worth logging in with the IDE and checking `g`. Maybe keep checking while you're doing different things and find out at which point it gets altered. That might help?

You could also try pasting this code into the REPL:

```
var oldG = g;
delete g;
Object.defineProperty(global, "g", { get : () => oldG, set : () => {
  throw new Error("Don't set g");
}});
```

[...]

PR #2709, opened 2023-04-25
Move minify setting into expandable area, and warn

bobrippling:
This moves the minify option to an expandable advanced section, and changes the warning ("may" -> "will"). See #2562 for more details.

Deployed to my [gh pages](https://bobrippling.github.io/BangleApps/) if you want a preview.

### DIFF
--- a/index.html
+++ b/index.html
@@ -147,10 +147,6 @@
             <i class="form-icon"></i> Pretokenise apps before upload (smaller, faster apps)
           </label>
           <label class="form-switch">
-            <input type="checkbox" id="settings-minify">
-            <i class="form-icon"></i> Minify apps before upload (BETA, not recommended. Uploads smaller, faster apps but this may cause some apps to stop working)
-          </label>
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Looks good, but maybe you could move the 'translations' setting above the 'Advanced Options'? it's a bit odd having it in the middle.

Also, I really appreciate the link for testing - it makes life so much easier!

bobrippling:
Yeah that looks much nicer - [gh pages](https://bobrippling.github.io/BangleApps/) updated too if you wanted to re-check (maybe a bit of a gap (`<br>`?) between "advanced options" and the "default settings" button?)

gfwilliams:
Looks great, thanks! Actually I think I'll move the 'Default Settings' button into 'Advanced' then it's less likely to get accidentally clicked!

bobrippling:
Ah nice idea, and thanks for merging!

PR #2711, opened 2023-04-27
cogclock: Support Bangle.js 2

rigrig:
Very tight, but I think it looks acceptable with hidden widgets: 
download

(I thought about defaulting the widget setting to "swipe", but figured it would add too much clutter.)


gfwilliams:
Looks great, thanks! Is it possible to take the change to `core` out of the PR though?

gfwilliams:
Thanks!

PR #2713, opened 2023-04-29
clkinfogps: disable debug, README updated, Lato: save a fast and slow version

hughbarney:


gfwilliams:
Thanks! Is there a reason we need to have a fast and slow version? Generally the fastload version should also be capable of running without fast load if required.

Also, those files aren't mentioned in the metadata, so right now they will have no effect. Was that intentional

hughbarney:
Yes it was intentional to have 2 files fast and slow.  The app.js is a copy of the fast version.  This is more for reference and useful for debugging than anything else.  I've been trying to pin point a hang in Lato that I see every now and again, that seems to be something to do with timers and more likely to occur when I have the sunrise clock_info installed.  Hence why I have revisited the debug/log setting discussion in the forum.   The two versions were so I could eliminate fast load as a potential issue.  I decided to check in the code so I did not have to code it again if I ever needed it later.

gfwilliams:
Ok - although I'm not super happy with having two identical files in there.

Also, in Git every change you ever made is stored, so if you did overwrite slow with fast at some point that information is still there.

PR #2717, opened 2023-05-02
Update to nesclock to improve memory usage

eyecreate:
Put bitmaps into object set based on clock settings instead of all bitmaps into variables all the time. Also tried doing what was mentioned here:
https://forum.espruino.com/conversations/383678/#comment16863053
to see if it will fix issue with time disappearing on occasion.


gfwilliams:
Thanks!

PR #2720, opened 2023-05-03
New launcher/boot app: popcon

bobrippling:
When installed, this app will cause launchers to sort their apps by most popular, going off launch counts within the last month, and then the most recently launched app.


apps/popconlaunch/boot.js
```diff
@@ -0,0 +1,78 @@
...
+            cache = {};
+    }
+    return cache;
+};
+var saveCache = function () {
+    require("Storage").writeJSON("popcon.cache.json", cache);
+    if (orderchanged) {
+        require("Storage")
+            .list(/launch.*cache/)
+            .forEach(function (f) { return require("Storage").erase(f); });
```
bobrippling, line 18: This is a little hacky, but I feel reasonably safe - the only other approach I could think of would be to iterate through `*.info` files, look for launchers, then look for their data files which contain the word "cache" (and delete those)

gfwilliams: This looks fine - but just so I know, what examples of filenames are you trying to get here? Is it worth using `$` or `^` to restrict the match at the beginning or end?

bobrippling: Mainly aiming for the [launcher cache](https://github.com/espruino/BangleApps/blob/master/apps/launch/app.js#L23-L23) - `launch.cache.json` (reused in [`dtlaunch`](https://github.com/espruino/BangleApps/blob/master/apps/dtlaunch/app-b2.js#L17-L17)), and [`iconlaunch`](https://github.com/espruino/BangleApps/blob/master/apps/iconlaunch/app.js#L19-L19) (`iconlaunch.cache.json`) and [`taglaunch`'s caches](https://github.com/espruino/BangleApps/blob/master/apps/taglaunch/app.js#L44-L44) (`taglaunch.cache.json`) - I realise now my explanation wasn't great, it wasn't about the `*.info` files at all.

I guess `launch\.cache\.json$` would work, although I've realised I could just bump `popcon`'s info to trigger a rehash, which I think would be less coupled. How does this change seem?

gfwilliams: Yes, if you could just tweak the info, I'd have thought that might be less error-prone? Thanks!

gfwilliams:
Just noticed, is there any way of not defining global variables like `oldLoad`? Maybe just:

```
{
  let oldLoad = ....
  ...
}
```

Also I'm not sure if this should come with a warning that it might slow down other things since it overwrites readJSON?

bobrippling:
Sounds good to me - I'd originally made this app a launcher so forgot about it being in a boot context. I'll add the warning too, nice idea

gfwilliams:
Great - thanks!

PR #2722, opened 2023-05-05
widalarmeta: font 4x5 -> 6x8:1x2

nxdefiant:
- Change 4x5 font to more readable 6x8 2xheight
- teletext is now default font
- reload widget on settings change


gfwilliams:
Thanks!

PR #2724, opened 2023-05-05
Calendar: Add support for holidays

nxdefiant:
Holidays have same color as weekends and can be edited with the 'Download'-interface, e.g. by uploading an iCalendar file.


gfwilliams:
Great - thank you!

PR #2726, opened 2023-05-06
stlap - reduce battery consumption

bobrippling:
@bruceblore I've made a little tweak to `stlap`, so we drop the redraw interval to a second, unless the user's looking at, or interacting with the watch, what do you think?

### DIFF
--- a/apps/stlap/ChangeLog
+++ b/apps/stlap/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02: Bug fixes
 0.03: Submitted to the app loader
+0.04: Reduce battery consumption when the user's not looking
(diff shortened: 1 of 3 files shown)


bruceblore:
I like that. Updating 100 times per second (which I don't know if JS can even keep up with on this watch) is indeed only really necessary if the user is actively looking at the watch. And it appears that you've done a good job making it seamless. I especially like the use of `Bangle.on("twist")`, and the call to `setupTimerIntervalFast()` at the top of the touch handler. I also appreciate the basic quality stuff like not throwing exceptions with corrupted JSON, and adding the data files to metadata.json. 

It appears that you forgot to update the version number and note the changes in the ChangeLog. Those are the only problems I can see.

bobrippling:
Great thanks! And good catch, I'll sort the changelog too

gfwilliams:
Looks great - thanks!

PR #2728, opened 2023-05-07
`drained`: align clock and permit boot exceptions

bobrippling:
This aligns drain's clock so it looks more balanced:

image

and permits boot exceptions, with [`widdst`](https://github.com/espruino/BangleApps/tree/master/apps/widdst) being a default, so the `drained` clock isn't out by an hour when it loads.


bobrippling:
I've noticed when I toggle between menus on the watch I get some rendering artifacts, and the screen ends up hard to navigate until I force a redraw. Am I [using `E.showMenu()`](https://github.com/bobrippling/BangleApps/blob/e489b26834dda62f9fc9fa1c2068e55ebdce88b6/apps/drained/settings.ts#L75-L84) incorrectly?

## Initial settings screen
image


## On tapping "Keep startup code"
image

(note there shouldn't be anything below "Keep startup code", but "Restore watch at %" appears)

## If I reload the settings menu, and then make "Keep startup code" off
image

(again, there should be "Keep startup code" and "Startup exceptions" but "Keep startup code" appears twice)

bobrippling:
I'd say this is ready to merge, but I'm unsure about the menu bug - I think I've seen it before, so I don't think it's specific to this app

gfwilliams:
Ahh, sorry I missed this at the time. I think the issue is:

```
"Keep startup code": {
      value: settings.keepStartup as boolean,
      format: formatBool,
      onchange: (b: boolean) => {
        settings.keepStartup = b;
        save();
        updateMenu();
        E.showMenu(menu);
      },
    },
```

What happens is when you click a menu item, it calls `onchange`, and *then* redraws the menu. But you're changing the menu, redrawing a new menu and *then* it redraws the old one.

So you could see if it works just by deleting `E.showMenu(menu);` - since you seem to be changing the menu object that the menu is working from it *may* work. But otherwise just wrap those in a timeout so they get called after the menu redraws itelf:

```
"Keep startup code": {
      value: settings.keepStartup as boolean,
      format: formatBool,
      onchange: (b: boolean) => {
        settings.keepStartup = b;
        save();
        setTimeout(()=> {
          updateMenu();
          E.showMenu(menu);
        }, 10);
      },
    },
```

bobrippling:
No worries, and thanks! I see what you mean now, if there's a submenu we redraw the entire menu, otherwise we redraw that item, so it looks like your timeout idea should do the trick - I've given it a try and it's done the trick, thank you

gfwilliams:
Great!

PR #2732, opened 2023-05-08
Disallow emulator

gellnerm:
Disallow emulator as the emu link links to a non-existent resource


gfwilliams:
Thanks! Yes, I think right now the emulator doesn't handle just widgets

PR #2733, opened 2023-05-08
bikespeedo - tweaks

bobrippling:
I've made a few additions to the bikespeedo app, if they look ok to you @HilmarSt ?

- Add option to display speed in user's locale
- Guard against showing `NaN` as a speed
- Add option to record a ride (like the run app) and stop the recording when exiting the app


HilmarSt:
I'm currently not using the Bangle.js - but by having a short look on it, the changes seem to be ok.
Please feel free to continue with the PR.

bobrippling:
Great - thanks for checking over the changes too

PR #2734, opened 2023-05-08
New widget: widhid

bobrippling:
This app's similar to [Swipe Bluetooth Music Controls](https://github.com/espruino/BangleApps/tree/master/apps/hidmsicswipe), except it operates as a widget, so you can control music without leaving your current app (e.g. on a run, bike ride or just watching the clock).

The app also disables all other drag and tap handlers while this widget is "active" (in a similar manner to [`backswipe`](https://github.com/espruino/BangleApps/pull/2524#issuecomment-1406230564) and [`lightswitch`](https://github.com/espruino/Espruino/issues/2151#issuecomment-1042423211)).


typescript/types/main.d.ts
```diff
@@ -1584,7 +1584,7 @@ declare class NRF {
    * @param {any} callback - A callback function to be called when the data is sent
    * @url http://www.espruino.com/Reference#l_NRF_sendHIDReport
    */
-  static sendHIDReport(data: any, callback: any): void;
+  static sendHIDReport(data: number[], callback?: () => void): void
```
bobrippling, line 1587: Hope you don't mind - took the liberty of regenerating `main.d.ts` from my [PR in Espruino](https://github.com/espruino/Espruino/pull/2357)

apps/widhid/wid.ts
```diff
@@ -0,0 +1,195 @@
...
+			Bangle.removeListener("drag", onDrag);
+			resumeOthers();
+
+			redraw();
+		}, 3000);
+	};
+
+	const redraw = () => setTimeout(Bangle.drawWidgets, 50);
+
+	const connected = NRF.getSecurityStatus().connected;
```
bobrippling, line 101: Out of interest, I've seen some scripts deleting variables - should I be deleting variables like `connected` to save memory, even if they're not captured by any closures? Or is there a different rationale behind that?

gfwilliams: In this case `connected` will hang around, so yes you could delete it or try and avoid using it (if not for memory usage, just to avoid having something called `connected` kicking around which doesn't reflect the current state!). 

If you've got a function defined which could possibly access a variable then Espruino will have to keep it - it's not smart enough to remove variables that aren't used in referencing functions.

bobrippling: Gotya, makes sense - I'll do that what with this being a widget then, I reckon - thanks for the explanation

gfwilliams:
Thanks!

PR #2735, opened 2023-05-08
RAM clock_info - a clock info that shows % ram used

hughbarney:


gfwilliams:
Great - thanks! Just to add, using `process.memory(false);` there won't be GC so you may see reported RAM usage rise over time even though the memory is actually not used as such (and a GC would free it).

hughbarney:
> `process.memory(false);` there won't be GC so you may see reported RAM usage rise over time

Tricky one.  Both options have their pros and cons I guess.  On balance I think I prefer false.

PR #2737, opened 2023-05-09
Update interface.html for shadowclk

stweedo:
Lots of fixes, fairly reliable now.

GadgetBridge throws a "Malformed JSON from Bangle.js: No value for t"

### DIFF
--- a/apps/shadowclk/interface.html
+++ b/apps/shadowclk/interface.html
@@ -2,6 +2,7 @@
 <html>
 
 <head>
+  <meta charset="UTF-8">
   <link rel="stylesheet" href="../../css/spectre.min.css">
   <link href="https://fonts.googleapis.com/css2?family=Londrina+Solid&display=swap" rel="stylesheet">
   <link href="https://fonts.googleapis.com/css2?family=Londrina+Shadow&display=swap" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - if you're connected with the IDE and just using the watch as-is, do you see any unexpected messages? It might be your clock code is printing something to the console and that is confusing Gadgetbridge

stweedo:
I didn't notice anything odd, but sure enough there's some weird data at the end of the id for my clock when the app loader connects to it
```
{"id":"shadowclk","version":"0.02","files":"shadowclk.info,shadowclk.app.js,shadowclk.settings.js,shadowclk.img","data":"shadowclk.json","type":"clock"},"BANGLEJS2","2v17.74",1683704307,86432740]
```

gfwilliams:
Ahh, right - so that's the problem. It's a hard one really - Gadgetbridge expects JSON-formatted objects to be commands for it from the watch, and it's always looking for data - so when you read the settings and get  `{"color":"#000","theme":"dark"}` it sees JSON with `{...}` it thinks it is a command for itself, but it doesn't see a `t` so it complains.

It's probably easiest to do:

```
Puck.eval('[require("Storage").readJSON("shadowclk.json", true)]', (data) => {
          if (data) data = data[0];
          if (data) { ...          
```

So then it's wrapped in an array which won't be parsed by Gadgetbridge

stweedo:
That indeed was the problem with the malformed JSON errors. Thank you. 

Now I just need to figure out the communication issue that's still there sometimes

PR #2738, opened 2023-05-10
popcon: trim old entries from the cache

bobrippling:
Just to stop the storage/cache potentially filling up with old unused apps


gfwilliams:
Thanks!

PR #2740, opened 2023-05-10
Update interface.html - shadowclk

stweedo:
Add comments
Reduce timeout to 50ms
Still has some comms issues sometimes in which case the default values are used for the preview

### DIFF
--- a/apps/shadowclk/interface.html
+++ b/apps/shadowclk/interface.html
@@ -82,7 +82,6 @@
 
     #message-container {
       height: 40px;
-      /* adjust the height based on your desired fixed height */
       margin-top: 20px;
       text-align: center;
     }
@@ -106,6 +105,7 @@
   </div>
   <script>
     const messageDiv = document.getElementById('message');
+    // Create and add color buttons to the UI
     let colors = ['#000', '#f00', '#0f0', '#ff0', '#00f', '#f0f', '#0ff', '#fff'];
     let colorButtonsContainer = document.getElementById('color-buttons-container');
     colors.forEach((color, i) => {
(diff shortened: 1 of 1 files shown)


gfwilliams:
I'd argue that 50ms is almost certainly far too short - if nothing has been sent over the connection for a while the connection interval drops to 200ms, and there would be a round-trip so in that case you'd be looking at minimum 400ms.

stweedo:
I had originally tried 500 and that added a noticable delay to the preview loading so that's why I tried 50 which still does work around half of the time. I'll do some more testing. I appreciate your feedback

gfwilliams:
Ok, let me know when you think you're happy with it.

`eval` should have its own timeout internally, so I don't really think there's any need to add your own?

stweedo:
> Ok, let me know when you think you're happy with it.

It's fully functional now with the new settings now as far as I can tell.

> `eval` should have its own timeout internally, so I don't really think there's any need to add your own?

The timeout is only there to wait for `eval` and if it doesn't return anything then it moves onto using the defaults for the preview. I'm still having some connection issues when opening and closing it repeatedly but It seems really reliable if it's the first few times you try loading the interface back up.

gfwilliams:
Ok, great -mergning then.

> The timeout is only there to wait for eval and if it doesn't return anything then it moves on to using the defaults for the preview. 

That's kind of my point - if `Puck.eval` never returns something is very broken deeper down and we shouldn't be hacking around it at a higher level.

stweedo:
Ahh okay, gotcha. That definitely makes sense and I'll keep that in mind.

PR #2741, opened 2023-05-10
Buzzing behavior in message app

targor:
The intention of the message app was, to set a buzzing interval and a buzzing timeout. The problem is, that the buzzing only occours twice on new messages and then stops. This is due to the fact that settimeout was used and not setinterval. This pull request changes it and makes the message app work as initially intended.


targor:
> Thanks - will merge now but it looks like we did also want to change `if (exports.buzzTimeout) clearTimeout(exports.buzzTimeout);` to `clearInterval` - but i'll do that now

I missed that, thank you for your help 👍 .

PR #2743, opened 2023-05-10
widhid: avoid clashing with other swipe/drag handlers

bobrippling:
Particularly in menus, the widget's swipe handler would activate, preventing the menu from working. Tested now with a launcher active, on a clock, on a menu and on a few apps.

### DIFF
--- a/apps/widhid/ChangeLog
+++ b/apps/widhid/ChangeLog
@@ -1,0 +1,3 @@
+0.01: New widget - music control via a swipe
+0.02: Improve interactivity - avoid responding to swipes when a menu or
+      launcher is active.
--- a/apps/widhid/README.md
+++ b/apps/widhid/README.md
@@ -13,6 +13,7 @@ Swipe down to enable - note the icon changes from blue to orange, indicating it'
 
 All other watch interaction is disabled for 3 seconds, to prevent clashing taps/drags - this period is extended as you continue to alter the volume, play/pause and jump between tracks.
 
+Requires espruino firmware > 2v17 to avoid event handler clashes.
 
 # Setup / Technical details
 
--- a/apps/widhid/metadata.json
+++ b/apps/widhid/metadata.json
@@ -2,7 +2,7 @@
   "id": "widhid",
   "name": "Bluetooth Music Swipe Control Widget",
   "shortName": "BLE Swipe Widget",
-  "version": "0.01",
+  "version": "0.02",
   "description": "Based on Swipe Bluetooth Music Controls (based on Bluetooth Music Controls). Swipe down to enable, then swipe up/down for volume, left/right for previous and next and tap for play/pause. Enable HID in settings, pair with your phone/computer, then use this widget to control music from your watch!",
   "icon": "icon.png",
   "readme": "README.md",
--- a/apps/widhid/wid.js
+++ b/apps/widhid/wid.js
@@ -19,8 +19,6 @@
         }
     });
     var onDrag = (function (e) {
-        if (Bangle.CLKINFO_FOCUS)
-            return;
         if (e.b === 0) {
             var wasDragging = dragging;
             dragging = false;
--- a/apps/widhid/wid.js
+++ b/apps/widhid/wid.js
@@ -10,34 +10,14 @@
     var dragging = false;
     var activeTimeout;
     var waitForRelease = true;
-    var mayInterceptSwipe = function () {
-        if (Bangle.CLKINFO_FOCUS)
-            return 0;
-        if (Bangle.CLOCK)
-            return 1;
-        var swipes = Bangle["#onswipe"];
-        if (typeof swipes === "function") {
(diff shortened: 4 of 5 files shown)


nxdefiant:
Did you check compatibility with quicklaunch?

nxdefiant:
I think if everything works if downward swipe of quicklauch is not used it would be fine.

gfwilliams:
You can order the bootcode by putting a number in the filename - that could help if yours goes first/last - or just execute with a timeout?

This does all look like a lot of overkill though - I'd be amazed if overriding the event handers didn't break something for someone at some point. For some events (where an object is passed) we did try adding a `handled=true` field, and checking that in any affected apps.

But part of me wonders if it would be worth having some way of stopping propogation of events (there is a hack where just adding+removing a handler for the event during that event will stop others getting fired). Not sure if anyone has any ideas? On the web we'd have `stopPropagation` but we're not always passing an object for our events so we can't do that.

Stuff that comes to mind:

* Check the return value - maybe if `return E.HANDLED;` is returned we don't call any other events in the list?
* Have a new `.stop()` function on event-emitting objects that can be called during event execution
* Have `E.stopEventPropagation()` which does the same if called during event execution

bobrippling:
Yeah, I did feel like I was going down a rabbit hole with trying to detect conflicting handlers.

Number 3 feels most web-js-like, although I do like 1 for avoiding the need for tracking/checking more global state. I'll have a play around and come back with perhaps some diffs that we can consider

gfwilliams:
Actually no need - I've just implemented option 3 in 35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b in the Espruino repo.

So you just need:

```
E.stopEventPropagation&&E.stopEventPropagation();
```

And that should do it I hope, while being backwards compatible.

bobrippling:
Cool, thanks! I'll have popcon use that - it'll leave its "activation" handler in whatever order the user's wanted to sort the boot apps by, and override its order for the temporary music-control drag handler, which I think is what a user would expect

bobrippling:
Thinking about this more (apologies!), if a menu comes along and renders, it'll have its scroll handler added to the end of the handler list, so when popcon activates (after a little downward scroll), the menu won't respond.

I was thinking about this before and considered overriding `E.showMenu()` to set a `E.MENU_VISIBLE` property, but there's no way to know when the menu is dismissed, since taps on menu items are obviously user-code which might leave the menu open.

Maybe popcon could try to ensure it's always last in the swipe handlers? But then we'd still need a way to know if a previous handler had used the event or not.


I think, like mentioned above, this approach might be asking for trouble either way. I'm thinking that migrating popcon's activation handler to a gesture that's not going to be used by any other app/widget/launcher would be a better approach.

thyttan:
Thinking out loud:
When starting a listener with `Bangle.on()` or in `Bangle.setUI()` maybe there could be an argument passed that specified priority/where in the order of handlers it should run? The argument could have values reflecting first/last position or a number of "priority groups". Maybe nesting would be possible so you could say "this handler should run last in priority group 1" or "this handler should be run in subgroup 2 in priority group 3".

thyttan:
>if that's the kind of thing you're getting at with the priority groups?

That could surely be one way to use it! I have't thought about it very deeply myself.

gfwilliams:
Just to say I think this looks like a big improvement as-is, so merging for now (although having to mess directly with `#on` for ordering is a shame I guess the only other option is having the `boot` file ordered to run early and leaving the handler in all the time).

We could always change it in another PR.

bobrippling:
Thanks - yes, my concern with the `dragHandler` was that once a menu's shown we're disabled for good, since nothing removes the `dragHandler` (apart from another menu!). But perhaps that's ok, since menus tend to `load()` other apps one way or another - in which case cheers for the merge!

thyttan:
> Actually no need - I've just implemented option 3 in 35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b in the Espruino repo.
> 
> So you just need:
> 
> ```
> E.stopEventPropagation&&E.stopEventPropagation();
> ```
> 
> And that should do it I hope, while being backwards compatible.

How does `E.stopEventPropagation&&E.stopEventPropagation();` work?

From some testing in the Web IDE I get the sense it will just return `undefined` if the function isn't defined (i.e. older FWs) and will otherwise execute the function. Have I understood correctly?

bobrippling:
> From some testing in the Web IDE I get the sense it will just return `undefined` if the function isn't defined (i.e. older FWs) and will otherwise execute the function. Have I understood correctly?

Exactly - it'll only call the function if it exists, so you're right - for older firmwares it's a no-op

gfwilliams:
> How does E.stopEventPropagation&&E.stopEventPropagation(); work?

Yes, that's correct. For `a && b` it works as follows:

* If `a==false` return a
* Otherwise execute and return  `b`

PR #2745, opened 2023-05-11
+settings for UI, less IO r/w, rm timeout from eval

stweedo:
Still has comm issues sometimes. I haven't looked into it too much yet since it still mostly works. Last I saw it was getting stuck between comms.js and puck.js

### DIFF
--- a/apps/shadowclk/app.js
+++ b/apps/shadowclk/app.js
@@ -1,6 +1,6 @@
 // Clock with large colored digits using the "Londrina" font and a slightly larger "Londrina Shadow" font on top
 
-Graphics.prototype.setFontLondrinaSolid = function() {
+Graphics.prototype.setFontLondrinaSolid = function () {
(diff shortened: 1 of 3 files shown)


stweedo:
I have lots of new improvements made, just doing some final touches on them. I also removed the timeout for `Puck.eval` but now I need to handle switching to defaults differently. I think besides the communications issues I might have everything full functional (latest bug was leading zero state not reading correctly because of some state save logic that wasn't being set).

I don't see how to convert this pull request to a draft.

gfwilliams:
Top right under 'Reviewers'? At least i see it - I'll do it for you now anyway

stweedo:
Just committed the final touches. Seems nice and stable and fast. Let me know what you think. Thanks!

gfwilliams:
Don't have time to look into it too deeply, but it looks good - merging now!

PR #2746, opened 2023-05-11
Powermanager - Tweaks

halemmerich:
* Use charging state on boot
* Tables for trace log and deferred calls are now horizontally scrollable
* Tracing logs timestamp
* Better formatting for tracing table


gfwilliams:
Great - thanks!

PR #2748, opened 2023-05-12
bug fix in menu and optimization

stweedo:


gfwilliams:
Thanks! Normally we'd want a version bump but I think you changed this already very recently?

stweedo:
Yes, sorry. This was a quick hotfix I suppose

stweedo:
I found another small bug in the menu. So just one more small commit and everything should be all good now.

gfwilliams:
Great - thanks!

PR #2750, opened 2023-05-12
Updated infoclk with multiple new features and bugfixes

bruceblore:
Updated infoclk to version 0.09, providing the following bug fixes and features:

* Added weather dependency, because the clock has an already-existing integrated gadgetbridge weather feature
* Up and down swipes can now be configured separately
* The settings menu no longer crashes if the user removes an app they had a shortcut to
* Default notification app is now messageui rather than messages
* Optional "dual stage unlock", rendering the watch face as though it was locked until a certain number of taps. Primarily useful for accelerometer-based unlocking and showing the time to other people in the dark.
* Optional calendar bar, which shows the next few hours of events as a slowly scrolling bar in the middle
* Power efficiency operations: Update the bottom row only on change and everything else except the seconds every minute
* Option to show the seconds when unlocked, even when otherwise hidden by other settings (low battery or time range)
* Broke out config loading into separate file to avoid duplicating hundreds of lines in both app and settings
* Added support for fast loading, both for the launcher and optionally for the shortcuts.


bruceblore:
Is there an easy way to run these checks locally so next time I can get them right the first time?

gfwilliams:
Great - thanks!

PR #2751, opened 2023-05-12
Updated keytimer to use sched library

bruceblore:
When I initially added keytimer in #2217, there was a suggestion to use the scheduler library. I am now implementing that recommendation.


bruceblore:
That's definitely a good idea. Done!

gfwilliams:
Thanks!

PR #2752, opened 2023-05-12
New app: folderlaunch: Folder-based launcher

bruceblore:
In my opinion, scrolling through a lot of apps is inconvenient. This app addresses that by allowing the user to place apps into folders, configurable directly on the watch itself. Folders can be nested within other folders too. It also allows fast-loading apps that the user indicates are able to be fast loaded.

(Do you think allowing apps to specify in their metadata whether they can be fast loaded would be a good idea? I find that fast loading most apps greatly improves the usability of the watch, but asking the user if every app can be fast loaded can be annoying and error prone.)


thyttan:
> but asking the user if every app can be fast loaded can be annoying and error prone

If you tried the 'Fastload Utils' app, don't you think it's automatic judgement of what apps can be fastloaded is good enough? Maybe you could suggest to users to use Fastload Utils along with your app in the readme?

apps/folderlaunch/types.d.ts
```diff
@@ -0,0 +1,48 @@
+type AppInfoFile = {                // Contents of a .info file
```
bobrippling, line 1: If you like, we could reuse `info.d.ts`'s [`AppInfo` type](https://github.com/espruino/BangleApps/blob/master/typescript/types/info.d.ts), I suppose we'd want to add `id`, `data`, `files`, `tags` and `version` to it (all as optional)

apps/folderlaunch/settings.ts
```diff
@@ -0,0 +1,272 @@
+(function (back: Function) {
```
bobrippling, line 1: We could use the [`SettingsFunc` type](https://github.com/espruino/BangleApps/blob/master/typescript/types/funcs.d.ts#L1-L1) for more accurate types (`() => void` instead of `any => any`), so it'd go from something like:

```typescript
(function (back: Function) { ... })
```

to

```typescript
(function (back) { ... }) satisfies SettingsFunc
```

bobrippling:
Looks cool! How did you find using typescript for this? Any stumbling points that could be improved?

bruceblore:
@thyttan I was not aware of the existence of the fastload utils app. I will give it a look.

@bobrippling I found using typescript to be reasonably easy. Maybe the documentation around the build system could be improved or made easy to find. I was just using vscode's "build" function to perform the build. I'm not sure how it would be done manually if I happened to not be using vscode. I couldn't find documentation on this.

apps/folderlaunch/settings.ts
bruceblore: I like that. I think it's always good to use more specific types when possible.

(Github claims I can accept this and have it automatically committed with you getting credit. I'm trying to figure out how to do that.)

apps/folderlaunch/types.d.ts
bruceblore: I like that idea. Can we just do that, or is there a chance of breaking apps that rely on the AppInfo type? I don't *think* so because we're just adding stuff to it, but I just want to be sure.

bruceblore: Seems like popconlaunch is the only other app that uses this?

bruceblore:
I gave fastload utils a try, and I'm mostly happy with it. It's certainly nicer than manually configuring it for every app (plus it works on other launchers too), so I removed that functionality from my app. The only problem is that now I get two buzzes when I launch an app: both before and after. I'll have to look into that.

I have not yet implemented the suggestion to use the AppInfo type rather than my own AppInfoFile type, because I'm still only about 90% sure that it's OK to make the changes to the AppInfo type.

apps/folderlaunch/types.d.ts
bobrippling: Great, thanks! Yeah that sounds right to me, and adding shouldn't be a problem in this case either

apps/folderlaunch/settings.ts
bobrippling: Cool - I'm not sure if I have to have put it in as a suggestion, does this work?
```suggestion
(function (back) {
```

But then I'd also need another suggestion for the `satisfies` bit at the end, so maybe not!

bobrippling:
About typescript, thanks for the feedback - glad to hear it was mostly straightforward. I'll see what I can do to make things more clear in the docs too

apps/folderlaunch/settings.ts
bruceblore: There are obvious buttons now (which are grayed out because things have changed since then), but yeah, I think you did it right.

apps/folderlaunch/types.d.ts
bruceblore: OK, I have added to the type.

bruceblore:
Any other concerns, or do you think I can mark this as ready?

apps/folderlaunch/types.d.ts
bobrippling: Great, thanks! I've just checked and going off [`createAppJSON`](https://github.com/espruino/EspruinoAppLoaderCore/blob/770b1b71f16399eb3eb86a4c3198e1fbed979110/js/appinfo.js#L243-L280), I think we can make `tags`, `type` and `version` optional too, wdyt?

apps/folderlaunch/settings.ts
bobrippling: Gotya, sounds good :)

bobrippling:
Oh yeah, no need to hold it back on my part!

apps/folderlaunch/types.d.ts
bruceblore: I'm surprised that the app loader allows apps to not have a type and version, but I agree that if it's possible, the type should reflect it.

bruceblore:
Awesome! I think it's ready now.

apps/folderlaunch/types.d.ts
bobrippling: Yeah the version especially, although perhaps it's for local development or custom apps

gfwilliams:
This looks really neat - thanks!

PR #2753, opened 2023-05-14
Kanagawa watchface update to v0.3

targor:
Fixes a bug where the watch still showed seconds when locked screen is on, after a message from the messageui was displayed.


gfwilliams:
Thanks!

PR #2755, opened 2023-05-14
fixed issue on tempmonitor and minor fiixes

dapgo:
fixed issue on tempmonitor and minor fiixes

### DIFF
--- a/apps/mixdiganclock/metadata.json
+++ b/apps/mixdiganclock/metadata.json
@@ -7,7 +7,7 @@
   "type": "clock",
   "tags": "clock",
   "screenshots": [{"url":"ss_mixdigan_ems.png"}],
-  "supports": ["BANGLEJS","BANGLEJS"],
+  "supports": ["BANGLEJS","BANGLEJS2"],
   "readme": "README.md",
   "allow_emulator": true,
   "storage": [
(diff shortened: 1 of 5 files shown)


gfwilliams:
Thanks, but I'm not sure it's a good solution to just not use the barometer at all (although by the look of this it never worked?)... What firmware version are you on? Plenty of things seem to be able to use the barometer just fine.

It might be all you need it `Bangle.getPressure().then( ... ).catch(() => {})` to ignore the error (although it shouldn't actually cause you any problems - it'd just mean that there wasn't an update that one time).

dapgo:
> Thanks, but I'm not sure it's a good solution to just not use the barometer at all (although by the look of this it never worked?)... What firmware version are you on? Plenty of things seem to be able to use the barometer just fine.
> 
> It might be all you need it `Bangle.getPressure().then( ... ).catch(() => {})` to ignore the error (although it shouldn't actually cause you any problems - it'd just mean that there wasn't an update that one time).
Thanks for your attention to the detail. there is a reason. Originally I tested on emulator and yesterday I realized that real BJS2 was crashing. I tried to fix but as i had not enough time to learn and fix, I went for a quick workaround. But I just  commented lines and I set an entry on my TODO list to go back to it.

gfwilliams:
Ok, thanks! I guess this is definitely an improvement over it not working at all on Bangle.js 2!

Issue #2754, opened 2023-05-14
[Light Switch Widget / Quick Launcher] Swipe to change brightness triggers quicklaunch as well


thyttan:
### Affected hardware version

Bangle 2

### Your firmware version

2v17.106

### The bug

Tapping the Light Switch Widget and then draging changes the brightness like it should, but the swipe handler of quick launcher is also triggered which ideally wouldn't happen.

To fix this I've tried to make use of the new function `E.stopEventPropagation()` introduced in https://github.com/espruino/Espruino/commit/35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b but I can't manage to make Light Switch Widget stop Quick Launcher from acting also, or vice versa. 

I added `E.stopEventPropagation();` inside the `dragListener` function 
in Light Switch Widget [here](https://github.com/espruino/BangleApps/blob/master/apps/lightswitch/widget.js#L144) and inside the swipe handler in Quick Launcher [here](https://github.com/espruino/BangleApps/blob/master/apps/quicklaunch/boot.js#L34). But neither try stopped the other app from executing. 

From the [documentation](https://github.com/espruino/Espruino/blob/35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b/src/jswrap_espruino.c#L2307):
> When using events with `X.on('foo', function() { ... })`
> and then `X.emit('foo')` you might want to stop subsequent
> event handlers from being executed.
> Calling this function doing the execution of events will
> ensure that no subsequent event handlers are executed.
> ```
> var X = {}; // in Espruino all objects are EventEmitters
> X.on('foo', function() { print("A"); })
> X.on('foo', function() { print("B"); E.stopEventPropagation(); })
> X.on('foo', function() { print("C"); })
> X.emit('foo');
> // prints A,B but not C
> ```

First mentioned in #2743:

> Actually no need - I've just implemented option 3 in [35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b](https://github.com/espruino/Espruino/commit/35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b) in the Espruino repo.
> 
> So you just need:
> 
> ```
> E.stopEventPropagation&&E.stopEventPropagation();
> ```
> 
> And that should do it I hope, while being backwards compatible.

_Originally posted by @gfwilliams in https://github.com/espruino/BangleApps/issues/2743#issuecomment-1545357023_

### Installed apps

notify (0.12)
podadrem (0.07)
[...]

PR #2756, opened 2023-05-14
[Light Switch Widget] handle and intercept swipe event

thyttan:
I've tried to use the same kind of logic present for the drag event. 

Don't know if it would be better to do some refactor to use the new `E.stopEventPropagation()` function?

Edit:

And to add, I don't really understand why this works, but from testing it it seems to at least.

Can be tested on my app loader: https://thyttan.github.io/BangleApps/?q=lightswitch

### DIFF
--- a/apps/lightswitch/ChangeLog
+++ b/apps/lightswitch/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Add masking widget input to other apps (using espruino/Espruino#2151), add a oversize option to increase the touch area.
 0.05: Prevent drawing into app area.
 0.06: Fix issue where .draw was being called by reference (not allowing widgets to be hidden)
+0.07: Handle the swipe event that is generated when draging to change light intensity, so it doesn't trigger some other swipe handler.
--- a/apps/lightswitch/metadata.json
+++ b/apps/lightswitch/metadata.json
@@ -2,7 +2,7 @@
   "id": "lightswitch",
   "name": "Light Switch Widget",
   "shortName": "Light Switch",
-  "version": "0.06",
+  "version": "0.07",
   "description": "A fast way to switch LCD backlight on/off, change the brightness and show the lock status. All in one widget.",
   "icon": "images/app.png",
   "screenshots": [
--- a/apps/lightswitch/widget.js
+++ b/apps/lightswitch/widget.js
@@ -165,13 +165,12 @@
       w.changeValue(value, event.b);
 
       // masks this drag event by messing up the event handler
-      // see https://github.com/espruino/Espruino/issues/2151
-      Bangle.removeListener("drag", w.dragListener);
-      Bangle["#ondrag"] = [w.dragListener].concat(Bangle["#ondrag"]);
+      E.stopEventPropagation&&E.stopEventPropagation();
 
       // on touch release remove drag listener and reset drag status to indicate stopped drag action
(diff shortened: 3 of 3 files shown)


gfwilliams:
I'd say use `E.stopEventPropagation();` if you can - it's a shame to be adding hacks when the underlying problem is already solved. As long as you check that  `E.stopEventPropagation();` exists first, on old firmware it'll continue as before, and on new firmware it'll be fixed

gfwilliams:
Also, I didn't see this when I replied to the bug you filed and closed: https://github.com/espruino/BangleApps/issues/2754#issuecomment-1547357991

But it looks like it's basically what you're doing here

thyttan:
I think my changes as they are now will sometimes mess up swipe handlers e.g. for clock infos and quicklaunch so they are not responsive. I'll see when I get back to this.

thyttan:
>I'd say use E.stopEventPropagation(); if you can - it's a shame to be adding hacks when the underlying problem is already solved. 

Don't Light Switch's handlers need to be placed first, with the concat-lines, in order to stop the propagation before the event reaches other handlers?

gfwilliams:
Yes, but it looks like they are?

thyttan:
Yes, ok, I thought that was part of the hack we wanted to avoid. Thanks

gfwilliams:
> Yes, ok, I thought that was part of the hack we wanted to avoid. Thanks

Ideally yes, but I don't think there's a way around it at the moment (unless I add extra options to `on` like `on(event, callback, {first:true})`)

thyttan:
>Ideally yes, but I don't think there's a way around it at the moment

Ah, I see 👍

thyttan:
This has been working for me for a week now. Ok to merge by me. I still don't really understand HOW it works though...

gfwilliams:
Please can you use `E.stopEventPropagatio&&E.stopEventPropagation();`  like I'd suggested above?

thyttan:
> Please can you use `E.stopEventPropagatio&&E.stopEventPropagation();` like I'd suggested above?

Added it in. I hope I did it correctly - I tried it on the watch and it seems to work.

gfwilliams:
Sorry, you actually need it instead of the fiddling with handlers, so:

```
swipeListener: function(_,__) {
      // setup shortcut to this widget
      var w = WIDGETS.lightswitch;

      // masks this swipe event by messing up the event handler
      // see https://github.com/espruino/Espruino/issues/2151
      Bangle.removeListener("swipe", w.swipeListener);
      Bangle["#onswipe"] = [w.swipeListener].concat(Bangle["#onswipe"]);
      E.stopEventPropagation&&E.stopEventPropagation();
    },
```

should be:

```
swipeListener: function(_,__) {
      // masks this swipe event
      E.stopEventPropagation&&E.stopEventPropagation();
    },
```

and the same where the other call to stopEventPropagation was too - it won't work on 2v17 and before but should work on cutting edge

gfwilliams:
> ... Will break the masking behaviour on 2v17?

Yes, that's right. Ok - probably a good idea then. Hopefully I'll be able to get a 2v18 release out soon

kamilkrz:
@thyttan @gfwilliams 
I believe 2v18 is already out. ;)

gfwilliams:
Thanks!

PR #2757, opened 2023-05-15
Rescalc - New app

stweedo:
Just whipped this up and looking for feedback for improvements or suggestions to the code. I still need to get the raw data for the image.js files, but the .img files should also work if placed directly on the bangle storage.


gfwilliams:
Not sure I can add much about the code - except to say the it looks like the resistor image might have loads of whitespace - you could crop that off and it'd make it smaller and much faster to render.

It's a really nice idea though!

stweedo:
I will definitely do that, thanks! I drew that up in inkscape and it was originally smack dab in the center. I just shifted it down. Probably could be roughly 1/3 of the size it is now with the whitespace.

PR #2759, opened 2023-05-16
[Home Assistant Sensors] inline config, add icons, quotes unification

rigrig:
* Instead of uploading a separate config file, adjust the code directly during customization.   
* Add icons depending on level+state: Home Assistant already uses a level-dependent icon, but doesn't know the sensors belong together, so doesn't add a charging indicator.
* Cleanup: use double quotes everywhere, instead of randomly mixed `'` and `"`


gfwilliams:
Nice - thanks!

PR #2760, opened 2023-05-16
Small stlapview menu fixes

bobrippling:
This fixes referencing a variable that's not in scope for `stlapview`, letting the user go back from a lap view.

It also removes the `eval` previously mentioned in a comment, and adds a back button to the main menu - @bruceblore hope these are ok, happy to revert them or change if you'd prefer

### DIFF
--- a/apps/stlapview/app.js
+++ b/apps/stlapview/app.js
@@ -93,11 +93,8 @@ function showMainMenu() {
     }
   };
 
-  //I know eval is evil, but I can't think of any other way to do this.
   for (let lapFile of LAP_FILES) {
-    mainMenu[fileNameToDateString(lapFile)] = eval(`(function() {
-      view('${lapFile}');
-    })`);
+    mainMenu[fileNameToDateString(lapFile)] = () => view(lapFile);
   }
 
   if (LAP_FILES.length == 0) {
--- a/apps/stlapview/app.js
+++ b/apps/stlapview/app.js
@@ -53,7 +53,7 @@ function view(fileName) {
   let lapMenu = {
     '': {
       'title': fileNameToDateString(fileName),
-      'back': () => { E.showMenu(mainMenu); }
+      'back': showMainMenu
     },
   };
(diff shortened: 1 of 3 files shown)


apps/stlapview/app.js
```diff
@@ -53,7 +53,7 @@ function view(fileName) {
   let lapMenu = {
     '': {
       'title': fileNameToDateString(fileName),
-      'back': () => { E.showMenu(mainMenu); }
```
gfwilliams, line 56: Please can you revert this?

The issue is that the callbacks are given the current menu as an argument - even if it's not used, that usually sticks around in the closure if it directly calls any function that declares a function inside and causes a memory leak.

bobrippling: Ah certainly can - would you be ok if I changed it to `() => { showMainMenu(); }` ? The original issue is that `mainMenu` isn't declared here, so this code always throws an exception

gfwilliams: Oh yes, absolutely, sorry. It's just calling functions directly that causes problems.

I should really just avoid passing the menu as an argument in the showMenu implementation, but I'm a bit concerned about the possibility of regressions

bobrippling: Ah gotya, I see now, like what you mentioned on the [other PR](https://github.com/espruino/BangleApps/pull/2734#discussion_r1188433030)

About passing the menu in, is it [this call](https://github.com/espruino/Espruino/blob/d7e760bc7896eb15a875d5ea2c43dfa8900643e6/libs/js/banglejs/E_showMenu_F18.js#L143-L143) you're referring to?
I had a quick scan through but by no means exhaustive, and it does seem like most apps don't use the menu argument, but there's a common pattern with the back function being used like this:

```javascript
(function(back) {
  let menu = {
    "": {"title": "Title"},
    "< Back": back, // <-----------
  };

  ...
})
```

I suppose one option would be to get an AST checker to delve through `apps/`, look for calls to `E.showMenu()`, find those menus and so on, although it's a question of if you think that's worth doing?

PR #2761, opened 2023-05-16
Fix `widhid`'s HID sending errors and handling of #ondrag

bobrippling:
- `sendHIDReport` sometimes errors if it can't send a HID - we now handle this to stop it bailing out the whole event handler chain.
- We also now handle the special case of `Bangle["#ondrag"]` being a function, catching this error with typings
- This also updates `main.d.ts` from the espruino repo (for `E.stopEventPropagation()`)

### DIFF
--- a/apps/widhid/ChangeLog
+++ b/apps/widhid/ChangeLog
@@ -1,3 +1,5 @@
 0.01: New widget - music control via a swipe
 0.02: Improve interactivity - avoid responding to swipes when a menu or
       launcher is active.
+0.03: Handle errors when sending input over BLE and the special-case of
+      replacing a single handler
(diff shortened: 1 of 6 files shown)


gfwilliams:
Thanks - what happens with the handlers is to save a bit of memory, if there's only one handler then it's put in as a function, otherwise if there are more it's an array.

It looks like your code may just ignore it if it's a function? Maybe what you need is just `if ("function"==typeof Bangle["#ondrag"]) Bangle["#ondrag"] = [Bangle["#ondrag"]];`?

However when you're doing `dragHandlers.filter`, is there a reason you can't just use `.removeListener` which does all that automatically?

bobrippling:
Makes sense to me - I'd gone with `any` in my typings and overlooked that case, but sorted now.

The piece of code in question is just reordering the `#ondrag` array rather than removing the listener - I wanted widhid's `onDrag` to be handled first (so it can temporarily override the others with `E.stopEventPropagation()` during its time interval).

Then in the case that `Bangle["#ondrag"]` is just a function, we know that it must be `onDrag`, so we don't need to do any reordering.

Issue #2395, opened 2022-12-14
[lcars/advcasio/cassioWatch/entonclk] don't report heart rate properly


gfwilliams:
It seems many clocks report HRM and use this pattern:

```
Math.round(Bangle.getHealthStatus("last").bpm)
```

But this returns the best HRM in the last 10 minutes - so it won't update for between 0-10 minutes.

Probably what's needed is really:

```
Math.round(Bangle.getHealthStatus().bpm||Bangle.getHealthStatus("last").bpm)
```

To get the 'best' HRM reading from this 10 minutes, or the reading from the last 10 minutes, or 0... and maybe even updating based on the HRM event.

But the best way would be to use the new `clock_info` module instead as this handles it properly now, and we can keep it updated if we spot issues like this in the future.

Affected:

- [ ] `lcars`
- [x] `advcasio`
- [ ] `cassioWatch`
- [x] `entonclk`

amunizp:
I have lcars in bangejs 2. HRM is stuck at 0.

amunizp:
@peerdavid and @Ronin0000 I don't know how to do clockinfo but copy and pasted patch.

PR #2765, opened 2023-05-19
Update lcars.app.js

amunizp:
Changed HRM as patch proposed in  #2395 I could not find the clockinfo module.

This change has not been tested.


gfwilliams:
Thanks - please could you update the ChangeLog and version number and I'll merge this? It'd be great if you could do the other affected clocks too.

Clock Info details are on http://www.espruino.com/Bangle.js+Clock+Info - but adding it is a much bigger deal than just changing this one line

gfwilliams:
Just applied this change to all 3 clocks and updated changelog myself

PR #2771, opened 2023-05-24
popcon: add reset ability and don't advertise as a launcher

bobrippling:
Some little usability tweaks to the tool


gfwilliams:
Thanks!

PR #2773, opened 2023-05-25
[chargerot] initial commit

kamilkrz:
Simple bootloader code to rotate screen during charging. As this is my first attempt, please check it correctly.
Actions turns ok so its ready to be checked if it's ready for release.
Two things to be aware of:

- Is there nicer method to get settings strings for rotation name to keep it in sync with settings app?
- Is there necessity to put settings app as requirement as I gather settings for screen rotation in normal mode?


apps/chargerot/settings.js
```diff
@@ -0,0 +1,28 @@
...
+    rotate: 0,
+  }, require('Storage').readJSON(FILE, true) || {});
+
+  function writeSettings() {
+    require('Storage').writeJSON(FILE, appSettings);
+  }
+
+
+  E.showMenu({
+    "" : { "title" : "Carging rotation" },
```
gfwilliams, line 14: Slight typo here :)

gfwilliams:
Hi! This looks good! I think I'd be a little concerned that when Bangle.js loads an app the screen orientation is changed automatically by the boot code (if it's rotated - otherwise there isn't anything). So if you set orientation and do `Bangle.showClock()` then in some cases (where the currently loaded app doesn't support fast load) it'll load the app fresh and change the orientation back.

... So I think just having these lines:

```
if (Bangle.isCharging()) g.setRotation(chargingRotation&3,chargingRotation>>2).clear();
```

In the boot code just before `Bangle.on('charging'` should fix it for you in that case

> Is there nicer method to get settings strings for rotation name to keep it in sync with settings app?

No, I think that's fine though - the values are based on what's fed to the firmware, and that won't change now.

> Is there necessity to put settings app as requirement as I gather settings for screen rotation in normal mode?

I wouldn't bother - pretty much everything needs settings so if it's removed you have more issues :)

kamilkrz:
> Hi! This looks good! I think I'd be a little concerned that when Bangle.js loads an app the screen orientation is changed automatically by the boot code (if it's rotated - otherwise there isn't anything). So if you set orientation and do `Bangle.showClock()` then in some cases (where the currently loaded app doesn't support fast load) it'll load the app fresh and change the orientation back.

Ideal solution for this (as I imagined it) should be ability to simple force to redraw screen. Then no need for Bangle.showClock() as we can choose app (simply by open it and dock watch) that we are interested in at the moment (i.e. timers) but I have no luck finding good option for refreshing screen with new settings.

For now I added this line and this should do the trick.

PR #2775, opened 2023-05-25
bikespeedo recorder / show max functions

bobrippling:
This lets the user toggle the "show max" function in bikespeedo (previously inaccesible through the UI), and if the recorder app is present, will allow toggling of recording via the hardware button, like the run app.

This also adds `isRecording()` to the recorder widget, so we can reuse its tracking of whether we're recording (and avoid potentially getting out of sync).

(also removed some never-executed code)

### DIFF
--- a/apps/bikespeedo/app.js
+++ b/apps/bikespeedo/app.js
@@ -12,7 +12,7 @@ const fontFactorB2 = 2/3;
 const colfg=g.theme.fg, colbg=g.theme.bg;
 const col1=colfg, colUncertain="#88f"; // if (lf.fix) g.setColor(col1); else g.setColor(colUncertain);
 
-var altiGPS=0, altiBaro=0;
+var altiBaro=0;
 var hdngGPS=0, hdngCompass=0, calibrateCompass=false;
 
 /*kalmanjs, Wouter Bulten, MIT, https://github.com/wouterbulten/kalmanjs */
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks good - thanks! Please can you bump the versions?

Just so you know, there's also a cycling speed sensor app that can use BLE speed sensors. I guess ideally we could find some way of making that a library that could be used by other apps (or at minimum a 'recorder' plugin)

bobrippling:
Ah yes - done!

And that's cool - are you thinking a library that handles the toggling of the recorder, along with initial prompts and exit handling?

gfwilliams:
Please can you bump recorder too? Otherwise the Speedo will update but will break because recorder won't have is isRecording in.

Well, right now that is handled by the recorder widget, so while not ideal it works ok.

What I mean is making the Bluetooth sensor a library. By recorder plugin I mean making it so you can add a column in recorded data for the Bluetooth sensor values

bobrippling:
Doh! Sure - updated recorder.

And I see what you mean now, the [recorder plugin](https://github.com/espruino/BangleApps/blob/f609e3e051e6f789f2b8378a8cf47b392cb7f3f7/apps/recorder/widget.js#L158-L159), like how [bthrm](https://github.com/espruino/BangleApps/blob/f609e3e051e6f789f2b8378a8cf47b392cb7f3f7/apps/bthrm/recorder.js) or [coretemp](https://github.com/espruino/BangleApps/blob/f609e3e051e6f789f2b8378a8cf47b392cb7f3f7/apps/coretemp/recorder.js) have one? Gotya

gfwilliams:
Great - thanks!

PR #2776, opened 2023-05-25
settings: permit temporarily allowing a BLE connection

bobrippling:
I find my phone (Android) changes its mac address most times I toggle bluetooth, whereas my desktop, laptop and raspberry pi are fine.

[As mentioned on a previous PR](https://github.com/espruino/BangleApps/pull/2607#issuecomment-1446047423), this is a potential issue - I'd like to keep my whitelist but be able to allow one-shot connections (e.g. if I'm doing a [GPS update](https://espruino.github.io/BangleApps/?q=agps) from my phone).

This PR adds a setting menu for bluetooth that temporarily disables the whitelist check on NRF connection, allowing me to connect temporarily, without having to remember to re-enable the whitelist afterwards. Happy to hear other's thoughts on this of course.

---

Seem to have some trouble testing this - I can't upload the JS to my watch:
```
Error: Unreachable point. logically broken. in widget_utils
```

### DIFF
--- a/apps/setting/settings.js
+++ b/apps/setting/settings.js
@@ -363,17 +363,6 @@ function showWhitelistMenu() {
     };
   }
 
-  if (settings.whitelist) settings.whitelist.forEach(function(d){
-    menu[d.substr(0,17)] = function() {
-      E.showPrompt(/*LANG*/'Remove\n'+d).then((v) => {
-        if (v) {
-          settings.whitelist.splice(settings.whitelist.indexOf(d),1);
-          updateSettings();
-        }
-        setTimeout(showWhitelistMenu, 50);
-      });
-    }
-  });
   menu[/*LANG*/'Add Device']=function() {
     E.showAlert(/*LANG*/"Connect device\nto add to\nwhitelist",/*LANG*/"Whitelist").then(function() {
       NRF.removeAllListeners('connect');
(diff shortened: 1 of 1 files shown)


gfwilliams:
Does it make sense to just add this to the `Make Connectable` menu option? It feels like that's the idea of that - to make the device connectable no matter what the settings are...

Removing all connect handlers is a bit of a pain as folks might be looking at the Bluetooth symbol in the top right to see if they have a connection?

You could do this I think (remove any handlers that have 'whitelist' in the function code):

```
var whitelist = ((NRF["#onconnect"] instanceof Array)?NRF["#onconnect"]:[NRF["#onconnect"]]).find(f=>f.toString().includes("whitelist"));
if (whitelist) NRF.removeListener("connect",whitelist);
```

But it's a bit hacky. Although I guess if it ever breaks it should just not remove the handler rather than causing problems.

 I guess the alternative is modifying bootupdate.js to make the whitelist handler check before applying the whitelist.

bobrippling:
Yes, the `Make Connectable` menu makes more sense actually - shall move.

And I see what you mean about the handlers, too - I'll have look at tweaking `bootupdate.js` and push an update.

bobrippling:
I didn't want to add a global variable and picked `NRF` as the "host", does this seem reasonable?

gfwilliams:
Yes - looks good to me! Merging...

PR #2778, opened 2023-05-26
feat: localisation pebbled

RomanistHere:
- added option for miles and AM/PM time as per request


gfwilliams:
Thanks! Merging, although did you know there is something in Espruino already that will help you display distances/times in *whatever* locale you want? http://www.espruino.com/Bangle.js+Locale

PR #2780, opened 2023-05-29
kbmulti - auto lowercase

bobrippling:
@sir-indy @thyttan what do you think of something like this in kbmulti? When the setting's on, it'll auto-lowercase the next character, when the user's in caps mode

(I've also made a little tweak to two `displayText()` calls, hope that's ok)

### DIFF
--- a/apps/kbmulti/ChangeLog
+++ b/apps/kbmulti/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Use default Bangle formatter for booleans
 0.04: Allow moving the cursor
 0.05: Switch swipe directions for Caps Lock and moving cursor.
+0.06: Add ability to auto-lowercase after a capital letter insertion.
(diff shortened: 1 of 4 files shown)


thyttan:
Sounds like it could be useful - don't know if I'll use it or not. 👍

I seem to remember someone have a stance on auto case. I think that time they were against it. But I don't even know which one of the keyboards it was.

bobrippling:
Thanks! Yeah thought I'd better make it an option just in case, although I don't type all that often on my watch to be fair

thyttan:
If you think having it enabled is the better default maybe you should do it like that 🤔🙂

bobrippling:
Yeah that's a good point, we can always flip the default back too if a lot of people prefer it that way - thanks

bobrippling:
Great, cheers for your response and glad to hear too. I've popped my name in the readme as well - thanks!

gfwilliams:
Thanks!

PR #2782, opened 2023-05-29
Add timers and non-date alarms to sched interface

bobrippling:
This lets users add, edit and delete timers and non-date alarms in the sched interface. They can also turn alarms/timers on and off.

It also reloads the watch on upload, to ensure alarms are then queued.

Deployed on [my github branches](https://bobrippling.github.io/BangleApps/?q=sched) if you want to give it a whirl.

To test:
- [X] Adding a new event
- [X] Adding a new alarm
- [X] Adding a new timer
  - Doesn't seem to show up in active alarms, but still sounds. Might appear at midnight - probably need to set `.t`
  - Solved - set `.t` on upload
- [x] Modifying an event
  - doesn't seem to queue, despite `.on === true`
  - seems to appear as enabled (e.g. it's shown by [widalarmeta]) but never fires
  - [fixed - `t.last`](https://github.com/espruino/BangleApps/pull/2782/commits/aa8ecc7e2c4e3fc345ae3b1ba29ca8fabfd03bf0)
- [X] Modifying an alarm
- [X] Modifying a timer
  - [X] slightly offset - clock drift between host & watch

---

- [X] Side/non-blocker bug: new timers show up in [alarms], but not in [multitimer]
  - [Intentional behaviour in multitimer](https://github.com/bobrippling/BangleApps/blob/855f196fd360a002b430523a36a6d7f4c8d2c3bd/apps/multitimer/app.js#L61-L61)

[alarms]: https://github.com/bobrippling/BangleApps/tree/master/apps/alarm
[multitimer]: https://github.com/bobrippling/BangleApps/tree/master/apps/multitimer
[widalarmeta]: https://github.com/espruino/BangleApps/tree/master/apps/widalarmeta

### DIFF
--- a/apps/sched/interface.html
+++ b/apps/sched/interface.html
@@ -86,6 +86,16 @@ function eventToAlarm(event, offsetMs) {
 }
 
 function upload() {
+  // kick off all the (active) timers
+  const now = new Date();
+  const currentTime = now.getHours()*3600000
+    + now.getMinutes()*60000
+    + now.getSeconds()*1000;
+
+  for (const alarm of alarms)
+    if (alarm.timer != undefined && alarm.on)
+      alarm.t = currentTime + alarm.timer;
+
   Util.showModal("Saving...");
   Util.writeStorage("sched.json", JSON.stringify(alarms), () => {
     Puck.write(`\x10require("sched").reload();\n`, () => {
@@ -125,9 +135,7 @@ function renderAlarm(alarm, exists) {
       tdType.textContent = "Timer";
       inputTime.onchange = e => {
         alarm.timer = hmsToMs(inputTime.value);
-        const now = new Date();
-        const currentTime = (now.getHours()*3600000)+(now.getMinutes()*60000)+(now.getSeconds()*1000);
-        alarm.t = currentTime + alarm.timer;
+        // alarm.t is set on upload
       };
     } else {
       tdType.textContent = "Alarm";


apps/sched/interface.html
```diff
@@ -88,29 +88,62 @@
 function upload() {
   Util.showModal("Saving...");
   Util.writeStorage("sched.json", JSON.stringify(alarms), () => {
-    location.reload(); // reload so we see current data
+    Puck.write(`\x10load()\n`, () => { // reload watch to load alarms
+      setTimeout(() => {
```
bobrippling, line 92: I'm not a fan of this, but the introduction of `Puck.write` / `load()` means that when we reload the interface and [re-read `sched.json`](https://github.com/espruino/BangleApps/blob/825594dc9fa0ba0af5803805d0e2ff3e3fcf15f7/apps/sched/interface.html#L252-L253), I get this data back:

```
×]<[...]
   ^^^^^ sched.json contents
^^^ unknown
```

Unless I wait for a little bit. Is there something I might be missing?

thyttan:
Looks super nice! I can definitely see this being useful 🙂

apps/sched/interface.html
```diff
@@ -185,15 +285,26 @@ <h4>Manage dated events</h4>
...
 
     <table class="table">
       <thead>
         <tr>
-          <th>Date</th>
+          <th>Type</th>
+          <th>Date/Time</th>
           <th>Summary</th>
           <th></th>
+          <th></th>
```
nxdefiant, line 307: I think this extra column could use a caption.
Maybe simply "On?"

apps/sched/interface.html
```diff
@@ -150,12 +200,53 @@
...
+  delete alarm.date;
+  renderAlarm(alarm);
+  alarms.push(alarm);
+}
+
+function addTimer() {
+  const alarmDefaults = getAlarmDefaults();
+  const timer = {
+    timer: hmsToMs("00:00:30"),
+    on: false,
```
nxdefiant, line 238: Is there a reason why it defaults to disabled?

apps/sched/interface.html
```diff
@@ -185,15 +285,26 @@ <h4>Manage dated events</h4>
     <div class="float-right">
       <button class="btn" onclick="addAlarm()">
         <i class="icon icon-plus"></i>
+        Alarm (with date)
```
nxdefiant, line 288: This is named "Event" in the watch.
I think we should have the same name for it.

apps/sched/interface.html
```diff
@@ -88,29 +88,62 @@
...
+    if (alarm.timer) {
+      tdType.textContent = "Timer";
+      inputTime.onchange = e => {
+        alarm.timer = hmsToMs(inputTime.value);
+        const now = new Date();
+        const currentTime = (now.getHours()*3600000)+(now.getMinutes()*60000)+(now.getSeconds()*1000);
+        alarm.t = currentTime + alarm.timer;
+      };
+    } else {
+      tdType.textContent = "Alarm";
```
nxdefiant, line 135: Can you also differentiate beween "Alarm" and
"Event" please? Something like
`alarm.date ? "Event" : "Alarm"`

apps/sched/interface.html
bobrippling: Sounds good to me 👍

apps/sched/interface.html
bobrippling: No strong opinion - I was thinking a user might want to populate with a few default timers but thinking about it, defaulting to on makes more sense.

apps/sched/interface.html
bobrippling: Thank you - I was struggling to come up with a name to differentiate, as you can perhaps tell

apps/sched/interface.html
bobrippling: Sounds good - I had been using the different input (datetime vs time) to hint to the user but this is much more clear

apps/sched/interface.html
gfwilliams: Ahh - it's possible that when you reload you see the data that's meant to be sent to gadgetbridge? Or maybe just the Espruino header text? Doing `Puck.debug=3` might help you see more of what's happening but as you say about all you can really do is delay...

Is there a reason to force reload? In most cases we reset the Bangle and then just display that 'press the button to reload' message which at least helps us avoid this sort of thing

bobrippling: Yes perhaps that's it - I see the below output with `Puck.debug` set to 3, not sure if there's much of a clue there, maybe the `NaN` size? But yes - I prefer matching the existing "Press button to reload" approach, I'll adjust it to do that. Thanks!

<details>
<summary>

console output with `Puck.debug=3`

</summary>

```
<COMMS> writeFile "sched.json"
comms.js:29 <COMMS> getProgressCmd START
comms.js:29 <COMMS> getProgressCmd 100%
comms.js:83 <COMMS> Response:  "OK\r"
index.js:214 IFRAME loaded
interface.js:104 <INTERFACE> init message received {id: 'BANGLEJS2', version: '2v17', info: {…}, connected: true, appsInstalled: Array(47), …}
comms.js:480 <COMMS> readFile "sched.json"
comms.js:473 <COMMS> readTextBlock read started...
comms.js:459 <COMMS> size is NaN
VM2336:1 Uncaught SyntaxError: Unexpected token '×', "×]8[{"t":3"... is not valid JSON
    at JSON.parse (<anonymous>)
    at interface.html:257:19
    at interface.js:116:13
(anonymous) @ interface.html:257
(anonymous) @ interface.js:116
postMessage (async)
(anonymous) @ index.js:248
Promise.then (async)
(anonymous) @ index.js:247
postMessage (async)
readStorage @ interface.js:48
getData @ interface.html:256
onInit @ interface.html:282
(anonymous) @ interface.js:108
postMessage (async)
(anonymous) @ index.js:266
load (async)
iframeSetup @ index.js:213
(anonymous) @ index.js:353
handleAppInterface @ index.js:337
(anonymous) @ index.js:895
```

</details>

gfwilliams: Ahh - try Puck.debug=3 when you don't have interface.html open - I think you might have been running it in interface.html's JS instance

bobrippling: Nice call, that's done it - looks like it's the powermanager and popcon apps I have perhaps delaying the response.

There's no more exception although it does hang now - but I guess we can sidestep this with the "Hold button to reload" message?

```
<COMMS> readFile "sched.json"
<BLE> Sending "\u0003\u0010(function() {\nvar s = require(\"Storage\").read(\"sched.json\");\nBluetooth.println(((s.length+2)/3)<<2); // estimate file size\nfor"
<BLE> Sent
<BLE> Sending " (var i=0;i<s.length;i+=384) Bluetooth.print(btoa(s.substr(i,384)));\nBluetooth.print(\"ÿ\");\n})()\n"
<BLE> Sent
<COMMS> readTextBlock read started...
<BLE> Received "\r\u001b[JExecution Interrupted\r\nat line 25 col 1295 in .boot0\r\n...ldRead(fname,skipExceptions);if(j&&/\\.info$/.test(fname)){va...\r\n  "
<BLE> Received "                            ^\r\nin function \"readJSON\" called from line 26 col 278 in .boot0\r\n...N(\"powermanager.def.json\",true)|"
<BLE> Received "|{};if(!def.start)def.start=Date....\r\n                                   ^\r\nin function called from line 26 col 3603 in .boot0\r\n"
<BLE> Received "...rging)chargeStart=undefined;});}})();;\r\n                                      ^\r\n>1104\r\nW3sidCI6MzAxMjAwMDAsImRlbCI6ZmFsc2UsI"

# there's then a bunch of atob()'d data after this, but no unable-to-parse-json error now
```

bobrippling:
Thanks for testing - sorted that `Comms` problem now.
I've matched the current behaviour of the app loader (popping up a "press button to reload" message) for now, which preserves the existing behaviour of sched's `interface.html`, except now we also get a prompt displayed on the watch for the user, but happy to discuss further

nxdefiant:
I really don't like the "hold button" message, its too confusing because when uploading an app a short press is enough, so the user is tempted to do also a short button press here - I was..

Did you test the "Add Timer functionality"?

bobrippling:
That's a fair point, although I would agree with Gordon that we maintain consistency with the rest of the app loader - perhaps that's a global option in the app loader that you'd find useful, to reload the watch when you've updated a collections of apps too, for example?

I've tested adding timers - I've not checked how the code behaves when we modify/enable a pre-existing timer (i.e. one that already has both a `.timer` and `.t` properties) - I've added this to my checklist (on holiday at the moment so it may be a little while)

nxdefiant:
With the small difference that in your version a small button press gets me back to the cloak without a reload of the scheduler. A small button press after the installation of an app does nothing.

Timer: It looks like if I just add a timer and upload it without a change (default 30 seconds)  it is set to midnight.

bobrippling:
About the button press, yeah that's true and I see the same - in my case my clock redraws when I unlock (removing the reload message) and I imagine yours does something similar, or a kind of draw-on-tap.

Had a thought this morning though - we don't need to reload the watch, but we can still apply the alarms. Let me know what you think of the [latest version](https://github.com/espruino/BangleApps/pull/2782/commits/4a48511fe933b2cb1e5c70f539eb89401ed6ab87).

(for some reason I see a strange little bar at the bottom, can't see how it's something I've done but you never know)
image

---

And thanks for checking the timer too - I seem to get a timer firing ~30s after I upload it (on creating a new one), although it doesn't seem to appear in `require("sched").getActiveAlarms(...)`, so something's up. Shall check.

nxdefiant:
nice idea with the reload()

bobrippling:
Thanks again for checking it out (and for the progress bar confirmation!). All good for a re-review

gfwilliams:
Looks great to me - thank you!

PR #888, opened 2021-11-18
Cube Scramble

nlisgo:
Simple app to display random scrambles for a 3x3 Rubik's cube.

PR #2783, opened 2023-05-29
notify: improve title bg colour for dark theme

bobrippling:
This permits `0x0` as a `titleBgColor` for notifications, and sets the notification's default title bg-colour depending on the user's theme.

## More details

Previously, a dark theme would display the below notification on a bangle2:

image

Which is readable here, but on a real device, looks like this:

image

This tweaks the background colour for the dark theme to almost-black, making the notification easier to read.

### DIFF
--- a/apps/notify/notify_bjs2.js
+++ b/apps/notify/notify_bjs2.js
@@ -100,7 +100,7 @@ exports.show = function(options) {
   gg.clearRect(x,y, r,b);
   // title bar
   if (options.title || options.src) {
-    gg.setColor(options.titleBgColor||0x39C7).fillRect(x,y, r,y+20);
+    gg.setColor("titleBgColor" in options ? options.titleBgColor : g.theme.dark ? 0x1 : 0x39C7).fillRect(x,y, r,y+20);
     const title = options.title||options.src;
     gg.setColor(g.theme.fg).setFontAlign(-1, -1, 0).setFont("6x8", 2);
     gg.drawString(title.trim().substring(0, 13), x+25,y+3);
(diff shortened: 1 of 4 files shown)


apps/notify/notify_bjs2.js
```diff
@@ -100,7 +100,7 @@ exports.show = function(options) {
   gg.clearRect(x,y, r,b);
   // title bar
   if (options.title || options.src) {
-    gg.setColor(options.titleBgColor||0x39C7).fillRect(x,y, r,y+20);
+    gg.setColor("titleBgColor" in options ? options.titleBgColor : g.theme.dark ? 0x1 : 0x39C7).fillRect(x,y, r,y+20);
```
bobrippling, line 103: Happy to undo the `titleBgColor in options` here if desired, as a user could pass in `"#000"` to get around the falsy-check

gfwilliams: No, this looks good to me - thanks!

Personally I'd really like to replace the hard-coded 16 bit color values with `#888`/etc as it's far more maintainable too :)

bobrippling: Gotya, sounds good - I could have a zip through and switch them over if you like? To check - in this case it'd be `"#0039C7"` right?

gfwilliams: Thanks! Nope, you'd have to use:

```
(c=>"#"+[(c&0xF800)>>8,(c&0x7E0)>>3,(c&0x1F)<<3].map(c=>((c+8)>>4).toString(16)).join(""))(0x39C7)
```

So `#444` - the color values are 16 bit RGB565 (why it's such a nightmare to work with them!)

Issue #2718, opened 2023-05-03
[Sleep Log Alarm] Sleep Log Alarm hides Contour Clock 


elfehr:
### Affected hardware version

Bangle 2

### Your firmware version

2v17

### Describe the bug
After installing Sleep Log Alarm and tweaking some of its settings, Contour Clock becomes invisible (only widgets are visible on the clock screen). I suspect it might have been linked to the setting to hide the widget, because the "Zzz" are not hidden when activating it. The clock only reappears when Sleep Log Alarm is disabled or uninstalled, and disappears again after reinstalling / enabling it. I had a bunch of other clocks installed to try them out, among them the default Anton clock, and they were not affected.

### Expected behavior
- Contour Clock never disappears
- Sleep Log Alarm's setting to hide the widget makes it disappear from the clock screen

### Steps to reproduce
1) Have Contour Clock installed.
2) Install Sleep Log Alarm for the first time. At this point all was good: I remember noticing its widget with the time of the next alarm, and nothing was wrong with the clock.
3) Play with Sleep Log Alarm's settings. The clock was still ok after changing the widget color. I suspect the "widget > hide" setting, but I can't be certain because I touched a lot of things before going back to the clock again.
4) Go back to the clock: Contour Clock is invisible. It's not a font problem: in Contour Clock's settings, the text appears as normal in the menu to choose the font.
5) The setting to hide Sleep Log Alarm's widget seems to have no effect.
6) When launching clocks from the menu: Anton clock and Time Traveller's Chronometer are not affected, Contour Clock is. 
7) Reinstall Contour Clock: no effect.
8) Consistently, disabling Sleep Log Alarm from its settings or uninstalling it makes Contour Clock reappears. Enabling / installing it makes Contour Clock disappears again. I can't go back to the initial state when both were working fine after the first installation.

### Installed apps

boot (0.57), health (0.22), setting (0.59), stopwatch (0.04), hrm (0.11), iconlaunch (0.17), widcloselaunch (0.02), scicalc (0.01), sched (0.22), notify (0.12), chargent (0.04), sleeplog (0.14), alarm (0.41), widpedom (0.22), widbthide (0.01), widminbat (0.02), contourclock (0.30), gallifr (0.03), swiperclocklaunch (0.04)

storm64:
I just tried to reproduce this issue and couldn't hide the widget, too. While having a look into the code I realized, that I might have wanted to implement this feature, but never added to check for this option on drawing the widget. 😅

[...]

PR #2785, opened 2023-05-29
[sleeplogalarm] Correct hide + replace all `var`

storm64:
Correct hide function + replace all `var` with `let`.
Might correct this issue: #2718


storm64:
I would suggest to merge this PR after #2756.

gfwilliams:
Great - thanks!

PR #2786, opened 2023-05-29
Update messagegui.app.js

stweedo:
To include "toward" as well for better compatibility when splitting up street names.

Also added support for "ft" in the en_US locale.


gfwilliams:
Thanks! Normally I'd say bump the version of the message gui but I think we can leave that for now

PR #2788, opened 2023-05-31
Update locale.html to remove leading zero in 12 hour mode

stweedo:
After a factory reset the bangle.js 12 hour mode does not have a leading zero in front of the hour. After updating the locale using languages from the app loader it will always have a leading zero in front of the hour. This fixes that and ensures 24 hour mode still has a leading zero.


gfwilliams:
Thanks!

PR #2789, opened 2023-05-31
[shadowclk] - Better settings and hour formatting

stweedo:
Updated the app the work with changes made in https://github.com/espruino/BangleApps/pull/2788


gfwilliams:
Thanks!

PR #2790, opened 2023-05-31
Switch Caller number and name

RandGH:
Would have been nicer to identify if we actually have a name and if not then display number instead in the body (only or too).

### DIFF
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -61,7 +61,7 @@
           t:event.cmd=="incoming"?"add":"remove",
           id:"call", src:"Phone",
           positive:true, negative:true,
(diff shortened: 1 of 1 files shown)


RandGH:
Sorry, simple solution only. 
Wouldn't be an issue to code the compare in a regular function but I don't know what value event.name has when Caller name is not found in Phone [or access is not granted] (event.number or undefined etc), else just doing a comparison would have been simple.
Just throwing this out in case you want to make it more elegant;)

gfwilliams:
Thanks - I think we need to look into this a bit more and do some tests.

Pretty sure this will say `From undefined` or something like that for unknown numbers which really isn't ideal

thyttan:
I'll close this - please open again if you want @RandGH .

PR #2797, opened 2023-06-05
Rescalc - update to v0.02

stweedo:
Fixes colors not matching user input from color menu in some cases, 3 bands are now shown larger, various code improvements.


gfwilliams:
Great!

PR #2798, opened 2023-06-06
"run" App: Expose HRM via BLE and Minor Formatting/Linting Changes

AnotherStranger:
### Description of Changes
This pull request introduces new functionality to the Banglejs "run" app. The following changes have been made:

1. Added functionality to expose Heart Rate Monitor (HRM) via Bluetooth Low Energy (BLE). This enables the app to broadcast HRM data to compatible devices.
2. The HRM functionality is disabled by default to avoid unnecessary battery drain. Users can enable it through the app's settings.
3. Made minor formatting and linting changes to improve code readability and maintain consistency. This includes adding missing semicolons, fixing whitespace issues, and other minor code style improvements.

### Proposed Changes Summary
1. Added HRM functionality via BLE.
2. Disabled HRM functionality by default.
3. Minor formatting and linting changes.

### Checklist
- [x] Added HRM functionality via BLE.
- [x] Implemented setting to enable/disable HRM.
- [x] Performed code formatting and linting changes.

### Reviewer Request
Please review the changes made in this pull request and provide your valuable feedback. Kindly pay particular attention to the new HRM functionality, ensuring it is implemented correctly and the user can easily enable/disable it. 

Thank you for your time and assistance in reviewing these changes.

### Some Images
FitoTrack using BangleJS as HRM device:
fito_track_using_hrm

Settings Page:
settings

Unchanged App Interface:
run_app

### DIFF
--- a/apps/run/ChangeLog
+++ b/apps/run/ChangeLog
@@ -15,3 +15,4 @@
 0.14: Fix Bangle.js 1 issue where after the 'overwrite track' menu, the start/stop button stopped working
 0.15: Keep run state between runs (allowing you to exit and restart the app)
 0.16: Added ability to resume a run that was stopped previously (fix #1907)
+0.17: Added functionality to act as a BLE HRM device while running
--- a/apps/run/app.js
+++ b/apps/run/app.js
@@ -1,5 +1,5 @@
 var ExStats = require("exstats");
-var B2 = process.env.HWVERSION===2;
+var B2 = process.env.HWVERSION === 2;
 var Layout = require("Layout");
 var locale = require("locale");
 var fontHeading = "6x8:2";
@@ -7,12 +7,16 @@ var fontValue = B2 ? "6x15:2" : "6x8:3";
 var headingCol = "#888";
 var fixCount = 0;
 var isMenuDisplayed = false;
+let isHRMAdvertisingSetUp = false;
+
 
 g.clear();
 Bangle.loadWidgets();
 Bangle.drawWidgets();
 
-// ---------------------------
+/*********************************************************************
+ * READ SETTINGS AND INITIALIZE EXSTATS WITH CONFIGURED MEASUREMENTS *
+ *********************************************************************/
 let settings = Object.assign({
   record: true,
   B1: "dist",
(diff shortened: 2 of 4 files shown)


AnotherStranger:
@gfwilliams I implemented this feature based on the forum conversation here [https://forum.espruino.com/conversations/387132/#comment17001076](https://forum.espruino.com/conversations/387132/#comment17001076). I figured this feature may be useful for others!

gfwilliams:
Thanks! However I'm really trying to make the `run` app itself as lightweight as possible.

Is there a reason this couldn't be a separate app in its own right? Maybe just 'boot code' - for instance there is one here that advertises BLE gatt data: https://banglejs.com/apps/?id=bootgattbat

So then you could advertise heart rate whatever app you were running - it wouldn't have to be `Run` but could be OpenStreetMap/etc.

Also, I'm not sure the initial `setAdvertising` is entirely right, since you're not supposed to advertise HRM that way, and you never update the value from `10` anyway.

I think what you need is to not have setAdvertising but to do:

```
NRF.setServices({
    0x180D: { // heart_rate
      0x2A37: { // heart_rate_measurement
        notify: true,
        value: [0x06, 0],
      }
    }
  }, advertise: [ '180D']);
```

Which would cause the heart rate service to be advertised as a service without data.

bobrippling:
I'd be a fan of this as boot code, currently when running I use [`btadv`](https://github.com/espruino/BangleApps/tree/master/apps/btadv) to transmit my heart rate, but I can't use any other apps.

Perhaps we have this also just fire up the HRM when BLE is connected, or if possible, only when the central/bluetooth host requests data from the HRM service?

AnotherStranger:
> Thanks! However I'm really trying to make the `run` app itself as lightweight as possible.
> 
> Is there a reason this couldn't be a separate app in its own right? Maybe just 'boot code' - for instance there is one here that advertises BLE gatt data: https://banglejs.com/apps/?id=bootgattbat
> 
> So then you could advertise heart rate whatever app you were running - it wouldn't have to be `Run` but could be OpenStreetMap/etc.
> 
> Also, I'm not sure the initial `setAdvertising` is entirely right, since you're not supposed to advertise HRM that way, and you never update the value from `10` anyway.
> 
> I think what you need is to not have setAdvertising but to do:
> 
> ```
> NRF.setServices({
>     0x180D: { // heart_rate
>       0x2A37: { // heart_rate_measurement
>         notify: true,
>         value: [0x06, 0],
>       }
>     }
>   }, advertise: [ '180D']);
> ```
> 
> Which would cause the heart rate service to be advertised as a service without data.

Thank you for your suggestion. I'm new to working with the Bangle so I appreciate your feedback!
I will create a new pullrequest with just the bootcode!

gfwilliams:
Great, thanks!

PR #2799, opened 2023-06-07
chimer: use async for chime delay, generalise next-chime logic

bobrippling:
@aaronrolls would you mind if I published an MR like this to tweak the busy-loop in chimer? I've also generalised the logic for deciding how long to wait for the next hourly/half-hourly/etc chime, if that looks ok to you?


aaronrolls:
That looks fine to me. Thanks for trimming down the code.

bobrippling:
Thank you, and thanks for the chime idea in the first place, takes me back to a watch I had decades ago!

gfwilliams:
Thanks!

PR #2800, opened 2023-06-09
[lightswitch] - Minor tidying up

stweedo:
What do you think @storm64? Does this look okay to you?


stweedo:
Could either go like this :
```
} else {
        // backlight off
        g.setColor(g.getBgColor());
      }
```

Or like this:
```
} else {
        // backlight off
        let innerColorWhenOff = g.getBgColor();
        g.setColor(innerColorWhenOff);
      }
```

Either one will make it look more consistent when off between themes where the inner bulb color matches the background.

stweedo:
Example here where the background color was changed, now it matches when the light is off instead of being just a black bulb shape. 

download

stweedo:
> ```
> g.setColor(g.theme.bg);
> ```

Thanks! I forgot about that one.

storm64:
The problem with using the bg color is on the light themes. In my opinion it is counterintuitive if a white bg should display an off status.

Here the default theme examples from the reedme:

stweedo:
I think we can leave that as is if that's the intended behavior.

gfwilliams:
Ok, I see what you mean - so I guess we pull out the `setBgColor(g.theme.bg)` change?

stweedo:
> Ok, I see what you mean - so I guess we pull out the `setBgColor(g.theme.bg)` change?

I think so, I'm not sure how best to implement a new setting right now. Should I still go through with the PR for the rest?

stweedo:
Sounds good. I changed it back.

gfwilliams:
Looks good - thanks!

PR #2801, opened 2023-06-09
[widbt] - fixes widget not showing on blue background

stweedo:


stweedo:
I didn't even think about that for screens that have more than 3bit color. What color format is the theme system using?

gfwilliams:
It uses 16bit RGB565. Even if you're not using 3 bit sometimes you might specify something non-3 bit in order to dither though. It's a hard one as I don't think I have a way of getting back to the `"#RGB"` format - I think your solution is ok for now.

stweedo:
Sounds like we need to check against a range of values close to the widget colors and if it's in that range make it use a dissimilar color. 

The battery widget would also need something similar for green backgrounds.

gfwilliams:
Yes, although for the battery widget I was wondering if we should just make the BG color white/black always - so even on green it's easily visible?

stweedo:
>Yes, although for the battery widget I was wondering if we should just make the BG color white/black always - so even on green it's easily visible?

I think that would be okay.

How about this for a concept? - A module that checks if a RGB565 color is predominantly green or blue and returns true if it is:
```
/*
This module provides utility functions for color analysis. Specifically, it can determine 
whether a given color is predominantly green or blue.
*/

// The ColorUtils constructor function
function ColorUtils() {
}

// The isBlueHue method checks if the blue component of a color is greater than both the green and red components
ColorUtils.prototype.isBlueHue = function(color) {
  // Gets the blue component by masking the color with 0x1F
  const blueComponent = color & 0x1F;
  // Shifts 5 bits to the right (discarding blue bits) and masks with 0x3F to get the green component
  const greenComponent = (color >> 5) & 0x3F;
  // Shifts 11 bits to the right (discarding blue and green bits) and masks with 0x1F to get the red component
  const redComponent = (color >> 11) & 0x1F;
  // Checks if the color is predominantly blue
  return blueComponent > greenComponent && blueComponent > redComponent;
};

// The isGreenHue method checks if the green component of a color is greater than both the blue and red components
ColorUtils.prototype.isGreenHue = function(color) {
  // Gets the blue component by masking the color with 0x1F
  const blueComponent = color & 0x1F;
  // Shifts 5 bits to the right (discarding blue bits) and masks with 0x3F to get the green component
  const greenComponent = (color >> 5) & 0x3F;
  // Shifts 11 bits to the right (discarding blue and green bits) and masks with 0x1F to get the red component
  const redComponent = (color >> 11) & 0x1F;
  // Checks if the color is predominantly green
  return greenComponent > blueComponent && greenComponent > redComponent;
};

/** This is 'exported' so it can be used with `require('ColorUtils.js')` */
exports = ColorUtils;
```

And then the widbt code could look something like this:
```
var ColorUtils = require('ColorUtils.js');

WIDGETS["bluetooth"]={area:"tr",width:15,draw:function() {
  g.reset();
  var colorUtils = new ColorUtils();
  if (NRF.getSecurityStatus().connected) {
    if (colorUtils.isBlueHue(g.theme.bg)) {
      g.setColor("#0ff");
    } else {
      g.setColor((g.getBPP()>8) ? "#07f" : (g.theme.dark ? "#0ff" : "#00f"));
    }
  } else {
    g.setColor(g.theme.dark ? "#666" : "#999");
  }
  g.drawImage(atob("CxQBBgDgFgJgR4jZMawfAcA4D4NYybEYIwTAsBwDAA=="),2+this.x,2+this.y);
},changed:function() {
  WIDGETS["bluetooth"].draw();
}};
NRF.on('connect',WIDGETS["bluetooth"].changed);
NRF.on('disconnect',WIDGETS["bluetooth"].changed);
```

gfwilliams:
I think that's just too much I'm afraid.

We're still supporting Bangle.js 1 where RAM is at a real premium, and that would at least double, maybe triple the memory consumption of the default widget... And I'm not sure there are even any default apps that use a blue background?

I feel like you'd be better just having a new bluetooth widget that was a different color - for instance https://banglejs.com/apps/?id=widminbt but that was just the FG color - and then anyone who had issues could install that in a few seconds

stweedo:
Okay, no problem. Thanks for the explanation. I was just just having some fun experimenting

stweedo:
That reminds me of another thing with the Pebble++ clock, I like the way it looks with the light theme better, but I also want the menus to use the dark theme at the same time. Should there be an option for inverting one or the other? Or should I simply use the customize menu to accomplish this? 

I'll be honest I haven't messed with the custom theme stuff much, but I'm going to now that I see it adds a ton of customizability to the whole experience and even my existing clock app (I'll for sure have to incorporate the custom theme settings into my clock app).

gfwilliams:
Well, you should use the normal system-wide theme settings to set what you want the menus and things to be.

But it looks like if you go `Settings -> Apps -> Pebble++` you can set a custom theme just for Pebble?

stweedo:
I tried using the system theme option in Pebble++ but that still doesn't seem to let me set the menus dark and the clock light.

It's not a big deal, just something I noticed and thought it might be nice to have as an option. I will probably look at finding a solution.

gfwilliams:
yeah, in pebble++

* system -> use whatever the system's theme is (set via Settings->System->Theme)
* dark/light -> force it

stweedo:
Ahh! Now I understand. For some reason I was confusing myself with what those settings did 🤦‍♂️

That indeed works exactly how I wanted it to. Thank you

PR #2802, opened 2023-06-09
[bootgatthrm]: Extract HRM Sensor Data into Separate Bootloader App

AnotherStranger:
This pull request aims to extract the features introduced in pull request #2798 into a new bootloader app. The primary focus of this PR is the creation of a boot code that exposes the HRM sensor data over BLE. By separating this functionality into a standalone service, users will have the flexibility to install and utilize it independently.

Changes Made:

- Extracted the relevant features from pull request #2798 into a new bootloader app.
-  Developed a boot code specifically designed to expose HRM sensor data over BLE.
- Ensured the boot code can be installed as a separate service.

@gfwilliams thank you for your guidance and valuable Feedback!!


gfwilliams:
This looks great, thanks! 

This won't actually turn the HRM on, but if the HRM is on for any other app (run, or the health monitoring for instance) then it'll send the current HRM over - but I think that's probably good behaviour?

AnotherStranger:
> This looks great, thanks!
> 
> This won't actually turn the HRM on, but if the HRM is on for any other app (run, or the health monitoring for instance) then it'll send the current HRM over - but I think that's probably good behaviour?

@gfwilliams 
I didn't turn on HRM intentionally. I tried to minimize battery drain caused by the bootcode.
However, I could implement a settings page which allows the user to control this behaviour. If there is demand for this feature.

I have one question remaining: When does the bangle call boot.js files? On every reload? 
Sorry, if this is an obvious thing I might have missed...

gfwilliams:
Yes, I think that's the right idea at the moment re: HRM. Just thought I'd mention in case you hadn't thought of it.

> When does the bangle call boot.js files? On every reload?

Pretty much, yes. When an app/clock/etc is loaded - it's needed because generally we tear everything down and put it back to ensure the watch doesn't get unreliable if one app wasn't behaving.

But we do now have 'fast load' especially for clocks, and then those don't call boot code again (but nothing gets torn down either)

PR #2803, opened 2023-06-09
Initial import of the sunrise watchface

trufae:
This app mimics the Apple Watch watchface that shows the sunrise and sunset time.

screenshot


gfwilliams:
Great! Do you think you could maybe use lat/lon from mylocation? https://github.com/espruino/BangleApps/tree/master/apps/mylocation

That way there's already an app that allows you to set up location (and it works with other apps that are looking for location too)

gfwilliams:
Also, it looks like the JS icon isn't right - maybe you used 'image object' and not 'image string'?

trufae:
i used image string, 4 bit mac palette, compression with the app.png (48x48 image) i wish there was a way to generate the images from the terminal in an automated way :?

good idea about the mylocation, i didnt knew this app. i'll add the lines to handle it!

trufae:
i have:

* indented the code
* added support for mylocation
* fixed the icon :? (just followed all the steps one by one)
* force pushed

gfwilliams:
Awesome, thanks! Merging now.

Right now there is no command-line app, but the tools themselves are available at https://github.com/espruino/EspruinoWebTools

I guess maybe we could add a command-line app to that repository, so then you could grab it from NPM and use it.

Either that or it could be added to https://www.npmjs.com/package/espruino - but I guess it'd need to be a second command-line app as it doesn't make much sense to add it to the main CLI app.

PR #2808, opened 2023-06-10
[kbmatry] Implement Matryoshka Keyboard

philipandresen:
First app, feedback welcome. I think I am doing this PR properly. Let me know if not!

# Matryoshka Keyboard
The idea behind this keyboard was to maximize screen space usage, optimize for number of inputs per character, and allow full key customization. Keyboards can be laid out to have all characters or subsets, and the characters themselves can be laid out into tiers, such that no letter is more than two touches away.

Multi-tier customizable keyboard library. Complies with the existing keyboard contact for all apps, but also includes a few additional features to help improve performance if people want to integrate this into their apps directly.

help
screenshot6

See README.md for more detailed information, but this keyboard library exposes the following:

1. input: Standard keyboard library entrypoint. 
2. generateKeyboard: Allows developers to pre-layout their keyboards to improve load speed on the keyboard if there are a lot of characters to lay out. Passed to input function as an optional part of the first argument object.
3. createCharSet: Allows developers to create custom keyboard layouts automatically given a string.
4. defaultCharSet: A pre-laid out character set corresponding to the default layout. Used with "generateKeyboard" to pre-load assets and improve keyboard load speed at runtime.
5. defaultCharSetShift: Similar to 4 but with alternate keys (caps / specials) for when "shift" mode is entered.

## Enhancements that I would like to do but haven't done:

* Swipe left for space. Swipe right for backspace. (wanted to figure out a good way to communicate to the user that this is possible, bit of a UX issue rather than an implementation one, same as long-press for shift)
* Option to pre-load subkeys versus deferring sub-key load to the first time a multi-character key is pressed. Right now there's a small delay when you hit a multi-letter key for the first time as it generates the subkeys. This improves time to first render DRAMATICALLY for large layouts but slightly degrades user experience. If you are already pre-generating your keyboard before it is needed you might as well generate all the sub-keys.
* Storage in flash memory of fully generated keyboards, so a keyboard could be generated once per install and never needs to be generated again. Would just need a content-specific ID (like a hash of the character set) to help determine if it needs to be re-generated or if an existing one can be used.
* More flexible character / font / space determinations. I've hard coded a few fonts into the "get best font" function which is fine, but not perfect.
* Support for unlimited nesting and special key nesting. It might be desirable to have "OK / cancel / delete / capslock" all grouped together in their own multi-key. Also you might want to have bigger keys with more nesting for accessibility purposes. Shouldn't be too hard but haven't gotten to it yet.
* Support for widget-respectful sizing.
* Support for "User doesn't need to see what they are typing" mode.
* Support for "Submit on first button press" mode.
* Side button press for cancel

## Known issues

* https://github.com/espruino/Espruino/issues/2377 Prevents this keyboard from running on the emulator, due to the use of `Promise.all`, but it seems like a fix is in the pipeline!
* Since the layout takes up the full screen, if you use it inside an app that loads widgets, those widgets may draw themselves over the output display occasionally. They're immediately covered over again but they do leave a few pixels hanging out here and there in the margins. Haven't worked out the best way to address that yet. I actually made a version of this that respects widget space but I felt it compromised the intent of "This needs to be as big as possible".
* Sometimes the user will accidentally hit the "OK" button and submit their text prematurely. Would like to find a good UX solution to that problem. Nesting special keys would be one way to solve indirectly (thus two presses required to confirm) or a simple confirmation dialogue - or requiring a long press of "OK". A Few options out there.
* If your app implements pre-generation code for this keyboard, you'll need to hide it behind some kind of logic to determine if that functionality exists, otherwise it will not be compatible with other keyboards.

## Feedback - Default layout

I would love suggestions for a good default layout. I put a pretty simple one together that covers all the keys on my keyboard, but there are some awkward bibbly bobs to it, like the fact that > and ? get their own dedicated key, and the fact that space is a character instead of a dedicated key. Do people really need curly braces in their bangle apps? I dunno! 

I WILL iterate on this, but if someone wants to take time to put together a thoughtful all-purpose reasonable default that would be swell too.


apps/kbmatry/lib.js
```diff
@@ -0,0 +1,489 @@
...
+
+
+/**
+ * This is probably the most intense part of this keyboard library. If you don't do it ahead of time, it will happen
+ * when you call the keyboard, and it can take up to 1.5 seconds for a full keyboard. Not a super great user experience
+ * SO if you have a tiny keyset, don't worry about it so much, but if you want to maximize performance, generate
+ * a keyboard ahead of time and pass it into the "keyboard" argument of the object in the "input" function.
+ * @param charSets
+ * @returns {Promise<unknown>}
+ */
```
philipandresen, line 178: Hmm, actually, this comment is out of date. I improved performance significantly since writing this. Will update with more recent performance data.

philipandresen: Fixed!

bobrippling:
> if you use it inside an app that loads widgets, those widgets may draw themselves over the output display occasionally

For this, I suspect `require("widget_utils").hide()` will work - it should stash all the widgets while your keyboard's active (of course you'd have to restore the widgets afterwards with a `.show()`)

gfwilliams:
This looks awesome - thanks! And I really appreciate the effort put into documenting this so it's easy to use too!

 As mentioned above I think widget_utils will sort out the widgets for you - just remeber to call `show` when the keyboard exits :)

PR #2809, opened 2023-06-10
dtlaunch: remove listeners when back is pressed

bobrippling:
At the moment, if I use the back button (or the physical bangle 2 button) on dtlaunch, I'll be presented with the clock, but it seems dtlaunch's listeners are still present. So if I go back to dtlaunch, I then get two instances of the launcher.

This change adds a `Bangle.setUI()` call, but while it fixes the double-launcher issue, I now see a loading screen when going between the launcher and the clock. Am I doing something wrong with fastload?


halemmerich:
I think by explicitly calling `setUI()` you remove the app and during the actual load there is no app left to cleanup for fastloading. So the Bangle decides to do a normal load and with that your double launcher issue is "fixed". I think to leave the launcher you should only call `load` or `showClock` as needed. Those will remove the UI for you.

thyttan:
I haven't looked into this on my watch since you opened this PR, but I don't notice any ux problems at least.

I thought removing the listeners was solved when defining them with let inside the app scope.

But if you want to introduce a Bangle.setUI() call to be extra sure, I think it should go inside the remove function.

halemmerich:
Listeners set by `setUI` should be correctly cleaned up when leaving the app. No need to do that manually.

bobrippling:
Found the problem - I have `swiperclocklaunch` installed, which adds its own handlers on at the [end of `setUI`](https://github.com/espruino/BangleApps/blob/d620ff25be2566d88ae1b0346f57dacc455b46ab/apps/swiperclocklaunch/boot.js#L7-L17). This means subsequent `setUI()` calls will clear its handlers, but not any it's replaced, such as `dtlaunch`'s.
(thanks for the prompt about not seeing the UX issues yourself @thyttan - that led me to the issue).

Both your comments make sense now - I'll adjust swiperclocklaunch to not drop the old listeners / handle this case.

thyttan:
Just a app tip: I personally prefer backswipe over swiperclocklaunch.

bobrippling:
Ah nice, thanks for the tip!

gfwilliams:
Thanks!

PR #2811, opened 2023-06-11
Battery Widget - Hanks Mod, add setting to hide high-mark

bobrippling:
## Description

The hank mod battery widget shows a high mark:

image

This is intentional behaviour, going off the code, and this PR adds an option to turn it off, so we get just the left-hand bar instead:

image

@pebl-hank - are these changes ok with you?

## Filename length

I realise I've added a settings page with an inconsistent name due to the [length limits](https://github.com/espruino/BangleApps/pull/1947#issuecomment-1154010055) - perhaps we should rename the app?


pebl-hank:
If I recall this correctly, actually the high mark is not intentional so removing it would be totally fine. Also I still get ghost pixels when the "100" is shown, but no time yet to fix this. Thanks for the PR!

bobrippling:
Cool, thanks! I'll drop it and the setting in that case. The ghost pixels for 100 I can't reproduce, but suspect it might be how the old text is cleared. If you're interested, I've [altered the clearing code](https://github.com/espruino/BangleApps/blob/37236b254e88f06742d82b16da7351d9a70c70d4/apps/hwid_a_battery_widget/widget.js) a little, which might fix it

PR #2812, opened 2023-06-11
Sensortools - Fix GPS interpolation

halemmerich:
This changes the interpolation to get a speed of roundabout 20km/h instead of hundreds and calculates useful course changes during interpolation. README has improved formatting and gpx files of the simulated routes.


gfwilliams:
Thanks!

Issue #2816, opened 2023-06-12
[E.showMenu] Title of selected item in options array appears in subsequent list


philipandresen:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### Description
When using the array version of menus along with E.showMenu (as described in the docs), while navigating from one menu to another, the selected item on the first menu appears in the second menu.

### Expected Behaviour
The second menu renders normally and the first menu can no longer be seen.

### Reproduction steps
1. Open web IDE, write the following code. Note that only the first menu needs to be the "array" type to reproduce.
```
let menu = [
  { title:"A"},
  { title:"B"},
  { title:"C", onchange:function() { E.showMenu(submenu); } },
];
menu[""] = { title:"Hello" };

let submenu = {
  "" : { title : "World", back : function() { E.showMenu(menu); } },
  "One" : undefined,
  "Two" : undefined,
  "Three" : undefined,
  "Four" : undefined
};

E.showMenu(menu);
```
2. When the code loads, select item "C".
3. Note that the second menu renders normally but with "C" superimposed (this follows scrolling motions too).
4. Note that the menu item that lingers from one to the other is the one you pressed to navigate.

The lingering button cannot be interacted with, but it can be cleared when an element underneath it is pressed such that it refreshes its render.

### Root cause hypothesis:
Relevant file: https://github.com/espruino/Espruino/blob/bc52028cb9271c4901259c6d399835715c2868a8/libs/js/banglejs/E_showMenu_Q3.js#L4
The issue I think becomes pretty clear when you compare the code for handling "onchange" versus an object-type-menu that just has items that are functions. In the case where we are seeing the bug we have:
```
if (item.onchange) item.onchange(item.value);
scr.scroll = l.scroller.scroll; // set scroll to prev position
show(); // redraw original menu
```
And in the case where we have no bug we haveL
```
if ("function" == typeof item) item(l);
else if ("object" == typeof item) {
```

#### Proposed solution:
Add an "onSelect" parameter that would be distinct from "onchange" in that it would just execute the function and not do anything else afterward. This would allow the array type menu to be closer in functionality to the object type menu without any backwards compatibility issues. I will see about getting a PR together if that seems like a good solution.

#### Alternate proposal:
Take a value returned by "onchange" and decide whether or not to render again based on that. Upside to this would be that it's close to the same programmatic contract with the users. Downside would be there's a greater likelihood of regressions for users that may be returning values that are ignored currently.

### Alternate hypothesis
[...]

PR #2818, opened 2023-06-13
[wrkmem] A working memory prosthesis

philipandresen:
# Working Memory Helper
screenshot screenshot2 screenshot5 screenshot6

## Description
Detailed information in the readme.

This app is why I bought the BangleJs in the first place. The idea is to have a device on my wrist that can remind me every 30 seconds or so of what I am doing, so that if I get distracted I can quickly get back on task. 

I created a custom menu system for certain parts of this app which I might like to abstract away into its own project eventually, but for now they are a large portion of the submitted code. Tried to ensure I was relying exclusively on "setUI" to manage user interface so that there wouldn't be any lingering listeners that could interfere with other apps or menus.

One feature I'm particularly excited about is the sorta-modular swipe control affordances. Each menu can be accompanied by up to four swipe hints giving users a little descriptive text and a swipe direction so they know that a given swipe control is available, and what it will do when they swipe that way.

## Known issues
The only real issues with this code right now is the clock is not updated once per minute - it may be updated more or less depending on how often the user gets a reminder. Ultimately the clock wasn't the focus, and updating "roughly once every time the user looks" was good enough for what I needed.

This app also supports keyboard pre-generation as implemented in [kbmatry] but should still work fine with any other installed keyboard library.


philipandresen:
This code actually contains a few instances of a workaround for https://github.com/espruino/BangleApps/issues/2816 which was resolved with https://github.com/espruino/Espruino/commit/a0e2d9231df709849f81abf572a742b0fceab85b . 

I wans't expecting it to be fixed so quickly! I will add a commit to remove those workarounds in anticipation of that change.

gfwilliams:
Looks great, thanks!

Out of interest, was there a reason that the Layout library didn't work for you? http://www.espruino.com/Bangle.js+Layout 

I really like the swipe indicators - I added something to the Messages library just a week or so ago but having something built-in and easy to use would be really neat.

philipandresen:
> Out of interest, was there a reason that the Layout library didn't work for you? http://www.espruino.com/Bangle.js+Layout

I actually did a first implementation of this app with that lib, but it ended up not being flexible enough for everything I needed. In retrospect that was more my fault than the library's fault, but I think it helped me to do it all myself the second time.

Maybe next commit will adapt it to the layout library, now that I know what I want and can think just in terms of custom render functions in that framework.

gfwilliams:
> Maybe next commit will adapt it to the layout library

I wouldn't worry about it. Layout can actually end up being quite slow to lay everything out, and what you have there doesn't look like it takes up much space and would be faster as well.

It has been suggested that actually what's needed is a simple (scrollable) grid layout system (maybe 3x3 which feels like the smallest most folks can reliably touch), which I guess may help with this kind of thing - but with your particular layout I'm not sure it would help

PR #2821, opened 2023-06-13
Catch additional errors from readJSON (via typescript)

bobrippling:
This change alters a few things typescript-related, but the actual JS changes come down to two apps - popcon and folderlaunch (hence only those having version bumps).

In summary:
- `main.d.ts` has been regenerated from the espruino repo
- Typescript apps which use `storage.readJSON()` now must declare the type they expect for the JSON, and also must handle `undefined`, in the `noExceptions` case (to avoid a crash)
- It's now harder for typescript apps to mix up `setTimeout` and `setInterval` ids
- popcon's `readJSON` override now handles the case where we get back `undefined`

### DIFF
--- a/apps/folderlaunch/types.d.ts
+++ b/apps/folderlaunch/types.d.ts
@@ -21,7 +21,6 @@ type Config = {
     apps: {                         // Saved info for each app
         [key: string]: {
             folder: FolderList,     // Folder path
-            fast: boolean,          // Whether the app should be fast launched
             nagged: boolean         // Whether the app's fast launch setting was configured
         }
     },
--- a/apps/btadv/app.ts
+++ b/apps/btadv/app.ts
@@ -686,7 +686,7 @@ const setIntervals = (
 const redrawInterval = setInterval(redraw, /*replaced*/1000);
 Bangle.on("lock", locked => setIntervals(locked));
 
-let bleInterval: undefined | number;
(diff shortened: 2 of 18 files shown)


apps/folderlaunch/types.d.ts
```diff
@@ -21,7 +21,6 @@ type Config = {
     apps: {                         // Saved info for each app
         [key: string]: {
             folder: FolderList,     // Folder path
-            fast: boolean,          // Whether the app should be fast launched
```
bobrippling, line 24: @bruceblore hope these changes are ok with you - to explain this one: with the typings added to `readJSON`, typescript now points out that when we [create an app config](https://github.com/espruino/BangleApps/blob/57f01d295141cf5d6e6c085315526ae21580c5d6/apps/folderlaunch/configLoad.ts#L128-L131) we weren't adding the `fast` property. I checked and it doesn't seem used, so I've taken it out here

bruceblore: Yes, that's OK to remove. (I should have removed it myself a while ago.) Originally, I intended folderlaunch to allow for manually setting apps to be fast loaded. However, someone pointed out that fastload utils existed, and therefore that my own manually-specified fast-loading feature did not make sense. So I removed the feature, but I guess I just forgot to remove this reference to it from the types.

PR #2823, opened 2023-06-14
Revert "remove workaround fixed by a0e2d92"

philipandresen:
This reverts commit f947105677ea2bcdece8317bfdbaee6fae0d9a0e. 
I didn't realize this wouldn't be fixed until 2v19 was available, which isn't quite yet! Adding the workaround back in until 2v19 is out.


gfwilliams:
Ok, thanks!

PR #2824, opened 2023-06-14
Check apps register their settings files

bobrippling:
Before I put time into this, I'm interested in whether you think it's a good idea.

I was wondering how many settings JSON files are left on the watch after an app's been uninstalled - we only [remove files in `metadata.json`'s `data`](https://github.com/espruino/EspruinoAppLoaderCore/blob/127c90aaa9e3d23f8853807e1ad17451a37dc3c1/js/comms.js#L324-L328) and so we could have a warning if it looks like an app's got a settings page but no `data` files.

(we could try to be clever and guess the `settings.json` name, and look for that within `data`, or keep it simple and assume if `data` is present then the app author's probably tracked the settings file)


gfwilliams:
I'm all for extra checks like this, although I'd search for `name.endsWith("setting.js")`, not just includes.

There's also the possibility that an app may write settings but not have a settings file, but this should catch most cases I guess.

bobrippling:
Great, thanks - I've a few tweaks to make yet, and will update the apps that error

bobrippling:
All good for review, fixes applied to all apps except three - two use csvs and one calls its file `.loading`

PR #2825, opened 2023-06-14
[folderlaunch] Add setting to disable vibration

kevdliu:
This PR adds a new option in the settings menu of Folder Launcher to disable vibrations in the launcher. Some users (myself included) might find the vibrations annoying so it would be nice to have an option to disable it. 

I wasn't sure whether to open this PR against the official repo or the app's fork so @bruceblore if you would rather have me open this PR against your repo instead I can do that too. 

Thank you!

### DIFF
--- a/apps/folderlaunch/app.ts
+++ b/apps/folderlaunch/app.ts
@@ -255,12 +255,7 @@
   /**
    * Vibrate the watch if vibration is enabled
    */
-  let buzz = (length: number | undefined = undefined, strength: number | undefined = undefined) => {
-    if (!config.disableVibration) {
-      Bangle.buzz(length, strength);
-    }
-  }
-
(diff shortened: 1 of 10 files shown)


apps/folderlaunch/app.ts
```diff
@@ -252,6 +252,15 @@
     }
   }
 
+  /**
+   * Vibrate the watch if vibration is enabled
+   */
+  let buzz = (length: number | undefined = undefined, strength: number | undefined = undefined) => {
```
bobrippling, line 258: If you like, we could avoid check inside the call:

```typescript
let buzz = config.disableVibration ? () => {} : Bangle.buzz;
```

kevdliu: I did not know you could do that with javascript, thanks!

gfwilliams:
Looks great to me! Thanks for tweaking it according to those comments!

PR #2827, opened 2023-06-16
Boxclk - New clock app

stweedo:
I put this together and it seems to work well for positioning the text around by dragging and dropping it after tapping to select and then you can tap again away from any box to deselect it. Currently it's drawing the background img during every redraw while dragging text so it's a tad bit sluggish (it's faster in the emulator).

It's definitely far from being finished and has lots of room for improvement. I just found out it screws up the default launcher and causes a reboot. Maybe I'm doing something wrong with `Bangle.setUI`? The swipe widgets work though! So that's good.

It would be nice to have a settings page to cycle through saved configs, and currently there is no way to save the adjusted positions from the clock app, only by editing the json file.


stweedo:
I fixed the launcher, I needed to move the custom font into the scope where `setUI` was being called and also restore the original `g.drawString` function

apps/boxclk/app.js
```diff
@@ -0,0 +1,223 @@
...
+    });
+    draw(boxes);
+  };
+
+  Bangle.on('touch', touchHandler);
+  Bangle.on('drag', dragHandler);
+
+  Bangle.setUI({
+    mode : "clock",
+    remove : function() {
```
bobrippling, line 197: Nice clock! Just thinking, I believe for the fastloading you'd also want to put the variables of this file inside a scope, so they're removed when the clock's unloaded (or technically once the code's run and any closures over them are gc'd)

apps/boxclk/metadata.json
```diff
@@ -0,0 +1,22 @@
...
+    {"url":"screenshot-1.png"}
+  ],
+  "type": "clock",
+  "tags": "clock",
+  "supports": ["BANGLEJS2"],
+  "readme": "README.md",
+  "allow_emulator": true,
+  "storage": [
+    {"name":"boxclk.app.js","url":"app.js"},
+    {"name":"boxclk.json","url":"boxclk.json"},
```
gfwilliams, line 18: you probably want to move this into:

```
"data" : [
{"name":"boxclk.json","url":"boxclk.json"},
]
```

Because then upgrades won't overwrite any settings stored in this file?

apps/boxclk/app.js
gfwilliams: Yes - check out http://www.espruino.com/Bangle.js+Fast+Load for more info - all you need to do is put a `{}` around everything and then define your global functions as `let functionName = function() { ... }`

gfwilliams:
Looks great! I just suggested a few changes - before merging, ensuring Fast Load works is important though (we don't want it causing a memory leak or interfering with other apps).

Also, please can you rename images like `beachhouse.img` to `boxclk.beachhouse.img`? Since we don't have directories it helps to figure out which apps files belong to

apps/boxclk/app.js
stweedo: Thanks! It passes the memory leak test after wrapping the entire program in an IIFE. I think this is sufficient, right?

```
>Bangle.loadWidgets();Bangle.drawWidgets();
=undefined
=2242
=undefined
>process.memory().usage
=3549
=undefined
=2242
```

gfwilliams: Yeah, it's ok - but because of the way Espruino works it ends up parsing the same code twice - once to load the function and once to run it, so it's quite a bit slower to load than my suggestion.

But if you're fine with this I'm ok to merge it

stweedo:
Should be okay to add now.

The next update will hopefully include a way to cycle between multiple config files and be able to save the adjusted positions to the current config.

stweedo:
I'm not sure why, but double tapping on the screen to save doesn't work very reliably outside of the emulator. Other than that and not having an easy modular way of adding custom fonts it's ready to go.

stweedo:
I busted out the trusty console.log() commands and found it was issuing a (drag, touch, drag) from the touchscreen taps. Adding a small movement threshold for drags fixed the save issue when double tapping outside of the emulator.

Only issue left that's readily apparent is the widgets show along with the save icon.

gfwilliams:
Great - thanks!

PR #2830, opened 2023-06-20
[boxclk] - v0.02 update

stweedo:
Adds more options for configs such as steps, meridian, short and long formats, prefixes and suffixes. The settings menu now sorts by ascending numerical order.


stweedo:
Yes, I would like to add that as well. I'm not entirely sure the best way at the moment, but I can look into it some more.

PR #2833, opened 2023-06-21
[boxclk] - Update screenshots and README to include example configs

stweedo:


stweedo:
I seem to have broken the remove function in setUI with my new font module. It works to enter the launcher and then comes back to the clock, but at that point the module is no longer loaded and i get `"unloadCustomBoxClkFonts" is not defined`.

stweedo:
The thought process for creating the font module was to store all the custom fonts there along with a function that intelligently loads only the ones that will be used to keep RAM usage down (only loading the fonts needed) and keep the main app light and slim without loading up or containing unused fonts.

I realize now it was still would have loaded all of the fonts in "lib.js" into RAM with the way I had it. Is something like this doable? When I initially tried to selectively export the fonts, it was telling me `setCustomFont` (what the Graphics.prototype returns) is undefined within my module.

gfwilliams:
Yes, unfortunately putting all the fonts in one module would be just as slow (if not more!) than putting them in the app.

If you had a separate library for each font that'd be better - however in the cutting edge builds I've added support for loading font files direct from storage (using Pebble's PBF font format) so really, that'd be the way to go.

PR #2837, opened 2023-06-25
[boxclk] - v0.04 Update

stweedo:
Improves responsiveness of background apps that also use the touchscreen (such as Pattern Launcher) by only using the drag handler if a box is actively selected.

### DIFF
--- a/apps/boxclk/ChangeLog
+++ b/apps/boxclk/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New App!
 0.02: New config options such as step, meridian, short/long formats, custom prefix/suffix
 0.03: Allows showing the month in short or long format by setting `"shortMonth"` to true or false
+0.04: Improves touchscreen drag handling for background apps such as Pattern Launcher
--- a/apps/boxclk/app.js
+++ b/apps/boxclk/app.js
@@ -4,6 +4,7 @@
   * 1. Module dependencies and initial configurations
   * ---------------------------------------------------------------
   */
+
   let storage = require("Storage");
   let locale = require("locale");
   let widgets = require("widget_utils");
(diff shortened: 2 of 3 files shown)


stweedo:
I made another [branch](https://github.com/stweedo/BangleApps/tree/boxclk-clkinfo/apps/boxclk) with some code I've been testing to add clock info compatibility.

A couple of things I'm still trying to figure out with the clock info menus:

1. Getting them to redraw properly
2. How to make them moveable like the other boxes.

I've only been focusing on issue 1 so far. I've tried merging the two draw functions (main and clock info), I've tried calling the clock info draw function both at the end and after of the main one. The module handles when to redraw, right? How do I make sure it redraws after every screen refresh? 

I think 2 should be doable by switching over to the pixel coordinates in boxPos and using those after loading the initial clock info menus. Currently the generated clock info menu data from the module is stored in the `clockInfoMenus` object. Maybe it's just a matter of having the module use the boxPos in the `boxes` for the menus just like the text boxes?

gfwilliams:
Thanks! I don't know if it helps in this case but 2v18 has http://www.espruino.com/Reference#l_E_stopEventPropagation which will allow you to stop events being passed down the list of `on(...` handlers.

For redraw, I'm pretty sure if you've got `let clockInfoMenu = require("clock_info").addInteractive(...)` all you need to do is call `clockInfoMenu.redraw()`

And hopefully for moving,you just have to change `clockInfoMenu.x/y` and it should be ok too... I think the only gotcha will be making sure you set `clockInfoMenu.focus=false` when dragging to stop the clockinfo picking up swipe events while you drag

stweedo:
Awesome! Thanks for looking into it. I'll try your suggestions later.

PR #2838, opened 2023-06-25
a half-life alyx style clock

devsnd:
a half-life alyx gravity glove style clock!


gfwilliams:
Looks great, thanks!

One thing though - I bet it's a bit slow drawing the hearts/stuff with the `icon` function. You may be better off either declaring the image inline with [`createImage`](http://www.espruino.com/Reference#l_Graphics_createImage) including the gaps between pixels (you can do `g.drawImage(img,x,y,{scale:2})` to double the size if you need, or actually just  including the image inline as base64 with https://www.espruino.com/Image+Converter

devsnd:
Thanks for the tip! I wanted to do some animations in the heart symbols, but then I ran out of time, so I thought better ship it as is, otherwise I'll never get it out the door :)

I would have had an easier time using images than rendering everything manually in any case...

Thanks for the great software and hardware, I'm really enjoying it!

PR #2839, opened 2023-06-25
mylocation: Add option to set location from waypoint

nxdefiant:
Notify @hughbarney 

Added dependency for waypoint. I could change that to "only show menu if waypoint is installed".


gfwilliams:
Thanks! Do you think you could change it to 'only show menu if waypoints installed'?

It's just that we have a few apps that depend on `mylocation` to get a location, and they don't really *need* waypoint installed as well - it's a nice feature if they have it though :)

nxdefiant:
will do

nxdefiant:
done.

PR #2842, opened 2023-06-27
[boxclk] - Update v0.05: Fix step count not resetting

stweedo:
Use `Bangle.getHealthStatus("day").steps)` instead of `Bangle.getStepCount())`


gfwilliams:
Great - thanks!

PR #2844, opened 2023-06-28
openstmap & recorder changes

nxdefiant:
openstmap: sort maps by scale desc
This results in the map with the highest resolution (lowest scale) to be drawn last

recorder: set scale on OpenStMap plot
When plotting with OpenStMap scale map to track width & height


apps/openstmap/interface.html
```diff
@@ -183,7 +183,7 @@
                     }).addTo(map);
                     let marker = new L.marker(latlon).addTo(map);
                     map.fitBounds(latlon.toBounds(2000/*meters*/), {animation: false});
-                  }, 100);
+                  }, 500);
```
nxdefiant, line 186: osm interface was always loading for me with
the previous 100ms timeout, no idea why.

Since there already was a timeout I assume
that I was not the first with this problem..

gfwilliams: That's odd - but glad this fixes it for you!

gfwilliams:
Thanks! Looks great!

If you're interested, something I found a bit tricky was when you had multiple maps, then in the OSM app you centered on GPS location you didn't always get the correct zoom level

nxdefiant:
@gfwilliams 
Looked at it, I find it ok that "center on GPS" does not change the zoom level: The user might already have chosen a zoom level that is appropriate. 

I think I would prefer openstmap to remember the position & zoom level on exit and load it on start.

gfwilliams:
> remember the position & zoom level on exit and load it on start.

That's a good plan, yes - I guess maybe do that unless there is no map under the current position (eg the maps were changed)

nxdefiant:
I guess map changed could be simply detected by saving & checking `require("Storage").hash(/openstmap\.\d+\.json/);` but that reduces load time. A faster way I found is `btoa(require('crypto').SHA1(JSON.stringify(m.maps)))`

Or I could have exports.draw() in openstmap.js to return the number of maps it painted and check that on first load.

@gfwilliams Any preference?

gfwilliams:
Both sound good, but:

> exports.draw() in openstmap.js return the number of maps it painted and check that on first draw

That sounds great - backward compatible and probably quite useful.

Personally I guess if I had the map somewhere, then updated maps, it'd be nice to still show that same location which that would do for us

PR #2845, opened 2023-06-28
podadrem: Send GB wake command only once

thyttan:
...instead of twice like we did before. It should not be needed now we issue a newline before GB commands.


gfwilliams:
Thanks!

PR #2848, opened 2023-06-29
[gassist] - New app for voice control

stweedo:


apps/gassist/settings.js
```diff
@@ -0,0 +1,32 @@
+let storage = require('Storage');
```
gfwilliams, line 1: Please can you put this inside the function? The way these settings menus are evaluated it's a bit iffy not having *just* the function declaration itself in the file

gfwilliams:
Great - thanks!

PR #2849, opened 2023-06-30
chimer: sleep between buzzes, instead of during

bobrippling:
A small fix to #2799, we now sleep between buzzes instead of kicking off a buzz and instantly sleeping


gfwilliams:
Thanks!

PR #2851, opened 2023-06-30
New app: rep

bobrippling:
This app is intended to help time running reps and can be programmed via the app loader.
I've made some small tweaks to `btadv`, but no version bump as there's no behaviour change.

I'd quite like the icon to be centered, is this possible with the [image conversion tool](https://www.espruino.com/Image+Converter), or maybe some manual tweaks to the image string? (see below screenshot)

image


gfwilliams:
It should be possible to center the icon - just don't enable 'crop' so you have the full 48x48px icon.

Of course the launcher itself could center any images by looking at imageMetrics to get the width/height, or just doing `{rotate:0}` to get drawImage to render the image centered on the x/y position.

bobrippling:
Ah I hadn't thought about the launcher not centering it. That gives me an idea..

bobrippling:
I thought my image was square, sorting that resolves this, all good to go

bobrippling:
I've also added in some recorder functionality

gfwilliams:
Great, thanks!

PR #2853, opened 2023-07-01
chargent: fixed boot

notEvil:
Fixes boot so resetting the watch doesn't disable chargent.

https://notevil.github.io/BangleApps/?q=chargent


gfwilliams:
Thanks!

PR #2855, opened 2023-07-01
battery voltage estimate

d3nd3:
I think this is the intended value the author wanted to show.
Very useful for people who want to keep their battery voltage within a certain range.


d3nd3:
I left the VREF in there just in case anyone is enjoying that option. Now new functional one is named `BattVolt` in settings and `BATV` for short.

gfwilliams:
I think vref can be useful (maybe) since if it starts to drop, it shows the battery voltage is low enough the regulator can't keep the full voltage up.

While it's fine for a display, the *actual* voltage may be a bit different depending on the device and we have a calibration value that could be used if needed. Actual code for it is https://github.com/espruino/Espruino/blob/master/libs/banglejs/jswrap_bangle.c#L2599-L2604

Issue #2856, opened 2023-07-02
[agenda] Wrong label for Today in allday events


glemco:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

The agenda app with `useToday` enabled can show labels like today or tomorrow instead of the current date.
On events lasting the whole day, this label is sometimes wrong: in an event scheduled for tomorrow, I see written tomorrow in the event's page (correct) but today in the list view.
Rough guess is it's related to having the start date (at midnight for all day events) wrongly reporting the previous day if changed to UTC.

PR #2857, opened 2023-07-02
agenda: Fixed "Today" and "Yesterday" wrongly displayed for allDay events on some time zones

glemco:
The way we were comparing the date for Today and Yesterday was to compute the ISOstring, extract the date and compare that with today's. Such date is however in UTC and allDay events start at midnight, which makes them belong to the previous day on half of the time zones (the same problem could appear on other events earlier than the time zone's offset).
With this PR we compare day and month, which is safer as it should take into consideration also  time zones.

Fixes: https://github.com/espruino/BangleApps/issues/2856

### DIFF
--- a/apps/agenda/ChangeLog
+++ b/apps/agenda/ChangeLog
@@ -13,3 +13,4 @@
       Added dynamic, short and range fields to clkinfo
 0.12: Added color field and updating clkinfo periodically (running events)
 0.13: Show day of the week in date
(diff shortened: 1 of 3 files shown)


gfwilliams:
Nice - thanks! That's a pain though - I wonder if that hits us in other places too where we use ISOstring

glemco:
I would say so, if it's trimmed to check the date, it can definitely give +1 or -1 days depending on the time and timezone..

PR #2858, opened 2023-07-03
Add chess game

nxdefiant:
Human with white and computer with black.

Based on p4wn engine[1]. Computations are rather slow but it works.

Usage: Drag on the touchscreen to move the green cursor onto a piece, select it with a single touch and drag the now red cursor around. Release the piece with another touch to finish the move. The button opens a menu.

[1] https://p4wn.sourceforge.net/


gfwilliams:
Wow, that's awesome!

PR #2862, opened 2023-07-04
support fastload

d3nd3:


gfwilliams:
Looks good - just checking, you did test this like on http://www.espruino.com/Bangle.js+Fast+Load to check for memory leaks?

d3nd3:
> you did test this like on http://www.espruino.com/Bangle.js+Fast+Load to check for memory leaks?

Yes, the numbers were the same.

gfwilliams:
Great - thanks!

PR #2864, opened 2023-07-05
openstmap: Remember latitude, longitude & scale

nxdefiant:
As discussed in #2844

Also disables the internal widget if widgps is installed. Kind of troubled me seeing two GPS widgets.


gfwilliams:
Looks great, thanks!

PR #2868, opened 2023-07-08
Add guitar-songs app

devsnd:
This is an app that lets you upload lyrics and chords for playing songs on the guitar.


gfwilliams:
Wow, that's really neat - thanks!

PR #2872, opened 2023-07-12
messagesoverlay - Use event based mechanism

halemmerich:
This uses the `messages` events instead of overloading the require function to get the messages.


gfwilliams:
Thanks! That looks mile better.

You happy with this or are you waiting to test it a bit more

halemmerich:
It works fine in the sense that everything seems to work as before, but currently the unhandled messages (tested nav and music) seem to reach the original message ui code, but do not seem to be handled there. I do not know yet if the issue lies in `messagesoverlay` or `messagegui`.

halemmerich:
Some further testing showed that handling the messages works fine, they just do not auto open on clock when using test messages, selecting them in the message list works fine. Testing with "real" navigation using OsmAnd seems to work fine, including auto opening on the clock.

gfwilliams:
Great - thanks!

PR #2875, opened 2023-07-14
Gpstrek - Handle GPX 1.0 in interface.html

### LINKED ISSUE #2854: gpstrek: Can not parse gpx file
The following test.gpx created with gpsprune can not be parsed from the gpstrek interface:
[testgpx.zip](https://github.com/espruino/BangleApps/files/11925834/testgpx.zip)

After clicking on "Upload" all I get is

> File could not be converted correctly.
error

Notify @halemmerich

When I replace "GPX/1/0" with "GPX/1/1" in the gpx it seems to work.

halemmerich:
No version bump since no code changes for the Bangle side of things. Fixes #2854


gfwilliams:
Thanks!

PR #2876, opened 2023-07-14
multitimer: fix hardmode in new alarms/timers

bobrippling:
A small bugfix leftover from #2781 - the previous PR changed how hardmode worked, but didn't update how new alarms/timers were created.


gfwilliams:
Thanks!

PR #2877, opened 2023-07-14
dtlaunch: add interaction-buzz setting

bobrippling:
@thyttan @hughbarney @halemmerich @jeffmer would you mind me adding a buzz setting to `dtlaunch`? I've mimicked `E.showMenu`'s buzz behaviour


thyttan:
I will not use it myself but don't mind the option being there.

jeffmer:
No objection whatsoever.

gfwilliams:
Thanks!

PR #2879, opened 2023-07-14
multitimer: fastload

bobrippling:
I find I regularly want to set timers/alarms, and this speeds up that interaction.

There is a 1-byte memory leak however - have we seen this kind of pattern before?

```javascript
>Bangle.loadWidgets()
=undefined
>Bangle.drawWidgets()
=undefined
>process.memory().usage
=3540                       // base

>eval(require("Storage").read("multitimer.app.js"))
=undefined
>process.memory().usage
=4023                       // initial app load

>Bangle.setUI()
=undefined
>process.memory().usage
=3800                       // app loads modules, memory remains bumped

>eval(require("Storage").read("multitimer.app.js"))
=undefined
>process.memory().usage
=4024                       // second app load - 1 byte more in use
>Bangle.setUI()
=undefined
>process.memory().usage
=3801                       // app removal - memory leak of 1 byte
```


gfwilliams:
> There is a 1-byte memory leak however - have we seen this kind of pattern before?

Could be a variable is written to in a function without `var` before? So it gets added to `global`? You could unload the app and then check `Object.keys(global)` to see if anything fishy is there?

bobrippling:
Thanks! I found a few accidentally global variables which I've resolved.

That didn't change memory usage but your `global` keys did the trick - the extra entry is `ReferenceError`. I imagine `multitimer` is misbehaving somewhere, or possibly a widget with the influence of an app having loaded.

I can't find any undeclared variables, and I imagine the error must be being caused in the callback of a `setTimeout` (as otherwise I'd expect to see an error & backtrace). With this in mind, does this seem acceptable, or would you prefer the `ReferenceError` to be tracked down?

gfwilliams:
Thanks! That's good to know.

If you get a chance it might be nice to know where the error is coming from, but it's not a big deal - I'm happy to merge this anyway.

It might just be that some code is testing if something exists using try...catch

gfwilliams:
Just to add, I've started using the timer clockinfo to set timers for myself - it's really cool :)

bobrippling:
> It might just be that some code is testing if something exists using try...catch

Yeah I suspect it might be unrelated boot code I have installed on my watch, which I imagine will be loaded even after a `reset()` ?


And thanks! Is that [this clockinfo](https://github.com/espruino/BangleApps/tree/master/apps/clkinfostopw) you mean?

gfwilliams:
yes, that's probably it.

For the timer, I use https://banglejs.com/apps/?id=smpltmr

bobrippling:
That's a great use of clockinfo, I'd never thought of different "lines" for different inputs. Thanks!

Issue #2847, opened 2023-06-29
Keeping forks up to date


gfwilliams:
There are a bunch of apps that end up being copies of other apps/code but with some changes applied.

Often that original code will have some changes/bugfixes later down the line and it's always a pain to ensure that those get propagated.

Discussed:

* https://github.com/espruino/BangleApps/commit/625f31c8d44258bd977ce1a757a720e7da5f8ad0
* https://github.com/espruino/BangleApps/pull/2843#issuecomment-1612147866

The apps might be copies of other apps in the repo, or of built-in functionality that is in other places like https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showMenu_Q3.js

So how do we keep these up to date? 

As a simple, fast solution, I'm coming down on something like an entry in `metadata.json`:

```
{
  // ...
  patches : [
   { ours: "app.js", theirs : "https://raw.githubusercontent.com/espruino/Espruino/master/libs/js/banglejs/E_showMenu_Q3.js", patch: "appchanges.patch" }  
  ]
}
```

So then (maybe even just in `sanitycheck.js`) we just grab the file from the URL given, attempt to patch it and compare the result, then flag up a warning if the two differ or the patch fails to apply?

bobrippling:
Yeah I like the sound of that, since our changes in `runplus` are pretty slim or I imagine similarly with `swscroll`.

With `theirs: "app/run/app.js"`, I don't see a problem, since we can only break `sanitycheck.js` in the commit that changes `run/app.js`. Whereas if we have `theirs: "https://..."`, then the build could break at an arbitrary time. So perhaps we make the former hard errors, and the latter warnings?

thyttan:
I'm in favor generally. I don't have too much input regarding technicalities - sounds like a good solution @gfwilliams !

How would a patch file look?

nxdefiant:
I think the list of possible URLs should be whitelistet to avoid a repository not under control to inject malicious code.

gfwilliams:
Ok, sounds good - non-http links where possible is a much better idea. And as you say, only warn for stuff that's not in the repo.

> whitelist

I think for now we only make it load from `https://raw.githubusercontent.com/espruino/...` URLs - or maybe even don't do HTTP at all and only support `app/...` and `Espruino/...`

bobrippling:
[...]

PR #2881, opened 2023-07-15
openstmap: current position overlay

nxdefiant:
- Draw a current position marker as mentioned in #2847 (Bangle.js 2 only)
- Enable/Disable previous position marker in new setting "Draw cont. position"
-  Add exit to menu

### DIFF
--- a/apps/openstmap/app.js
+++ b/apps/openstmap/app.js
@@ -82,13 +82,17 @@ function drawLocation() {
   }
 
   if (!fix.fix || !mapVisible) {
-    Bangle.setLCDOverlay(); // clear if map is not visible or no fix
+    if (this.hasOverlay) {
+      Bangle.setLCDOverlay(); // clear if map is not visible or no fix
+      this.hasOverlay = false;
+    }
     return;
   }
 
   const icon = require("heatshrink").decompress(atob("jEYwYPMyVJkgHEkgICyAHCgIIDyQIChIIEoAIDC4IIEBwOAgEEyVIBAY4DBD4sGHxBQIMRAIIPpAyCHAYILUJEAiVJkAIFgVJXo5fCABQA==")); // 24x24px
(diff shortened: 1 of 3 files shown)


apps/openstmap/app.js
```diff
@@ -67,16 +69,35 @@ function drawPOI() {
...
 }
 
+// Draw current location with LCD Overlay (Bangle.js 2 only)
+function drawLocation() {
+  if (!Bangle.setLCDOverlay) {
+    return; // Overlay not supported
+  }
+
+  if (!fix.fix || !mapVisible) {
+    Bangle.setLCDOverlay(); // clear if map is not visible or no fix
```
nxdefiant, line 85: This seems undocumented but works on my
machine..

gfwilliams: Yes, that's fine but I think it may force a full redraw of the screen each time it's called, so ideally do some checking first to avoid calling it repeatedly if you know you had't added an overlay?

nxdefiant: done

gfwilliams:
Thanks! It'd be nice if the menu item for it only appeared on Bangle.js 2 though?

Also, now, what would be really cool is if you could actually draw a pointer using the compass heading ;)

nxdefiant:
> Also, now, what would be really cool is if you could actually draw a pointer using the compass heading

Interesting idea

nxdefiant:
ok I'm convinced to replace the location icon with a direction one. But I prefer to use the course from the GPS instead the compass heading and leave it to gpsmagcourse to drop to the compass heading for lower speeds.

nxdefiant:
...or I make it configurable wither to use compass heading or gps course..

gfwilliams:
That's great - thanks! Always nice to make it configurable - I think when walking the compass might be handy, but driving/etc GPS is going to be preferable

PR #2882, opened 2023-07-15
qmsched - Allow to use two theme files for switching themes

halemmerich:
This allows using the same theme files as in the settings app to switch between two custom themes instead of being fixed onto the two default light and dark themes.
This works by copying the themes into the qmsched settings and from there into system settings to make persistent. Themes are marked with an additional `quiet` value to make `dark==false` themes usable in quiet mode. Not setting any theme in the options menu still switches between the default themes.

@rigrig Do you have an opinion on these changes?


apps/qmsched/lib.js
```diff
@@ -3,23 +3,36 @@
  * @param {int} mode Quiet Mode
  */
 function switchTheme(mode) {
-  if (!!mode === g.theme.dark) return; // nothing to do
   let s = require("Storage").readJSON("setting.json", 1) || {};
+  print("Theme is ", s.theme);
```
rigrig, line 7: Forgotten debug statement?

halemmerich: Yes, exactly.

halemmerich:
The idea was to save on checking at those points by setting the `quiet` property. `switchTheme` should only happen because of the scheduled changes or when in the menu, correct? I think the time needed by the in depth check of the theme does not really matter for those few times. Maybe it would make sense to not even store the theme but only the filename and apply it directly from the file instead of making multiple copies?

apps/qmsched/lib.js
halemmerich: Interesting, I pushed a new commit to remove this, but it does not (yet?) show up here. But it does here: https://github.com/halemmerich/BangleApps/commit/d60b2da468f280b0fbbb822a6520711eb81a2719

rigrig:
> `switchTheme` should only happen because of the scheduled changes or when in the menu, correct?

Yes, running a complete comparison shouldn't be a problem.
(I'm not even sure we even need the check, but it seems like the safe thing to do)

> Maybe it would make sense to not even store the theme but only the filename and apply it directly from the file instead of making multiple copies?

Good idea.

halemmerich:
Seems to work fine. Manual and automatic mode switches work and themes change as expected.

gfwilliams:
Looks good - thanks!

PR #2886, opened 2023-07-17
Test branching

kugelm:


kugelm:
First sync just for test porpose. Updates in gpstrek e.g.

PR #2888, opened 2023-07-17
Fix git submodule init command

joserebelo:
Fix typo in the git submodule init command - `git submodule --init` is not valid.


gfwilliams:
Thanks!

PR #2889, opened 2023-07-18
Implement API for activity fetching

### LINKED ISSUE #718: gadget bridge - offline activity
This is a feature request which is I think a bit difficult to implement. It is related to #716 

At the moment, activity is sent only when BT connection between smartphone and bangle is established. As step counter is cumulative it doesn't work if during a period the connection wasn't established.

I think having the same behaviour as MiBand2 with Gadget Bridge could be interesting. MiBand2 stores data locally and regularly GB get the data with possibility to let data locally or delete it. That way only GB get data and locally on the bangle activity are stored doesn't matter if link is established or not.

So there is two feature : 

* change how activity is handle in banglejs gadget bridge. it should store every 3min (or another period) in file values of steps, hrm timestamped. It should be interesting having other data like sleep detection and movement score.
* change in GB android app which scrapes regularly the banglejs and requests data it has not yet collected. I think there is already an interface existing for this. Maybe it is related to this not yet implemented function [here](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/banglejs/BangleJSDeviceSupport.java#L484)

joserebelo:
Add a simple API for ativify fetching from Gadgetbridge. The API receives the start timestamp, and returns activity up until the current timestamp.

PR on Gadgetbridge: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3212

Not incredibly happy with the code, but I am trying to avoid iterating through entire months needlessly, since sync might happen periodically.

Still very much a WIP.

Fixes #718

### DIFF
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -193,9 +193,36 @@
         Bangle.on('HRM',actHRMHandler);
         actInterval = setInterval(function() {
           var steps = Bangle.getStepCount();
-          gbSend({ t: "act", stp: steps-lastSteps, hrm: lastBPM });
+          gbSend({ t: "act", stp: steps-lastSteps, hrm: lastBPM, rt:1 });
           lastSteps = steps;
         }, event.int*1000);
+      },
+      // {t:"actfetch", ts:long}
+      "actfetch": function() {
+        gbSend({t: "actfetch", state: "start"});
+        var actCount = 0;
+        var actCb = function(r) {
+          // The health lib saves the samples at the start of the 10-minute block
+          // However, GB expects them at the end of the block, so let's offset them
+          // here to keep a consistent API in the health lib
+          var sampleTs = r.date.getTime() + 600000;
+          if (sampleTs >= event.ts) {
+            gbSend({
+              t: "act",
+              ts: sampleTs,
+              stp: r.steps,
+              hrm: r.bpm,
+              mov: r.movement
+            });
+            actCount++;
+          }
+        }
+        if (event.ts != 0) {
+          require("health").readAllRecordsSince(new Date(event.ts - 600000), actCb);
+        } else {
+          require("health").readFullDatabase(actCb);
+        }
+        gbSend({t: "actfetch", state: "end", count: actCount});
       },
       "nav": function() {
         event.id="nav";
--- a/apps/gbridge/PROTOCOL.md
+++ b/apps/gbridge/PROTOCOL.md
@@ -59,6 +59,26 @@ Send a response to a notification from phone
 * n can be one of "dismiss", "dismiss all", "open", "mute", "reply"
 * id, tel and message are optional
 
+## activity data
+
+```json
+{
+  "t": "act",
+  "ts": 1692226800000,
+  "stp": 26,
+  "hrm": 70,
+  "mov": 10,
+  "rt": 1
+}
+```
+
+* `ts` is the sample timestamp, in milliseconds since the unix epoch
+* `stp` is the number of steps
+* `hrm` is heart rate, in bpm
+* `mov` is the movement intensity (todo: range?)
+* `rt` whether it is a real-time sample
+
+
 # Phone -> Watch
 
 ## show notification
@@ -177,3 +197,14 @@ n is the intensity
 * hum is the humidity
 * txt is the weather condition
 * loc is the location
+
+## fetch activity data
+
+```json
+{
+  "t": "actfetch",
+  "ts": 1692226800000
+}
+```
+
+* `ts` is the start timestamp, in milliseconds since the unix epoch
--- a/apps/health/lib.js
+++ b/apps/health/lib.js
@@ -37,7 +37,36 @@ exports.readAllRecords = function(d, cb) {
     }
     idx += DB_RECORD_LEN; // +1 because we have an extra record with totals for the end of the day
   }
-}
+};
+
+// Read the entire database. There is no guarantee that the months are read in order.
+exports.readFullDatabase = function(cb) {
+  require("Storage").list(/health-[0-9]+-[0-9]+.raw/).forEach(val => {
+    console.log(val);
+    var parts = val.split('-');
+    var y = parseInt(parts[1]);
+    var mo = parseInt(parts[2].replace('.raw', ''));
+
+    exports.readAllRecords(new Date(y, mo, 1), (r) => {
+      r.date = new Date(y, mo, r.day, r.hr, r.min);
+      cb(r);
+    });
+  });
+};
+
+// Read all records per day, until the current time.
+// There may be some records for the day of the timestamp previous to the timestamp
+exports.readAllRecordsSince = function(d, cb) {
+  var currentDate = new Date().getTime();
+  var di = d;
+  while (di.getTime() <= currentDate) {
+    exports.readDay(di, (r) => {
+      r.date = new Date(di.getFullYear(), di.getMonth(), di.getDate(), r.hr, r.min);
+      cb(r);
+    });
+    di.setDate(di.getDate() + 1);
+  }
+};
 
 // Read daily summaries from the given month
 exports.readDailySummaries = function(d, cb) {
--- a/apps/android/ChangeLog
+++ b/apps/android/ChangeLog
@@ -30,3 +30,4 @@
 0.29: Support for http request xpath return format
 0.30: Send firmware and hardware versions on connection
       Allow alarm enable/disable
+0.31: Implement API for activity fetching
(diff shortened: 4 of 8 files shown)


apps/android/boot.js
```diff
@@ -197,6 +197,21 @@
           lastSteps = steps;
         }, event.int*1000);
       },
+      // {t:"actFetch", ts:int}
+      "actFetch": function() {
```
joserebelo, line 201: Should maybe be actlast, as the existing code in Gadgetbridge that I missed.

apps/gbridge/PROTOCOL.md
```diff
@@ -59,6 +59,23 @@ Send a response to a notification from phone
 * n can be one of "dismiss", "dismiss all", "open", "mute", "reply"
 * id, tel and message are optional
 
+## activity data
```
joserebelo, line 62: To move to https://github.com/espruino/EspruinoDocs/blob/master/info/Gadgetbridge.md once API finalized.

apps/android/boot.js
gfwilliams: Thanks! Yes, I think `actlast` would be good as the code's already in Gadgetbridge for that now...

gfwilliams:
Thanks! Sorry about the duplication - I'd started the work on this and then didn't get around to doing the Bangle.js bit before my holiday :)

Do you think maybe it makes sense for `readAllRecordsSince` just to call `readDay` rather than duplicating the code? I guess `readDay` could accept a 'since' argument, although hopefully out of all the days, only one of them will need part of the day transmitted, so it could just be an idea to check `date.getTime() > since` before forwarding the health record?

gfwilliams:
Ahh, ok - I see, we don't create a `Date` object. It would be nice to use `getDay` if possible though, especially as I guess we'll now be de-duplicating in Gadgetbridge.

I don't think speed will be an issue here - the thing slowing us down will be bluetooth transmission speed :)

apps/gbridge/PROTOCOL.md
```diff
@@ -177,3 +196,14 @@ n is the intensity
 * hum is the humidity
 * txt is the weather condition
 * loc is the location
+
+## fetch activity data
+
+```json
+{
+  "t": "actFetch",
+  "ts": 1692226800
```
joserebelo, line 205: Should be millis.

apps/health/lib.js
```diff
@@ -8,10 +8,11 @@ const DB_FILE_LEN = DB_HEADER_LEN + DB_RECORDS_PER_MONTH*DB_RECORD_LEN;
 function getRecordFN(d) {
   return "health-"+d.getFullYear()+"-"+(d.getMonth()+1)+".raw";
 }
-function getRecordIdx(d) {
-  return (DB_RECORDS_PER_DAY*(d.getDate()-1)) +
-         (DB_RECORDS_PER_HR*d.getHours()) +
-         (0|(d.getMinutes()*DB_RECORDS_PER_HR/60));
+
+function getRecordIdx(d, h, m) {
```
joserebelo, line 12: Need to revert this.

apps/health/lib.js
```diff
@@ -37,7 +37,36 @@ exports.readAllRecords = function(d, cb) {
...
+};
+
+// Read all records per day, until the current time.
+// There may be some records for the day of the timestamp previous to the timestamp
+exports.readAllRecordsSince = function(d, cb) {
+  var currentDate = new Date().getTime();
+  var di = d;
+  while (di.getTime() < currentDate) {
+    exports.readDay(di, (r) => {
+      r.date = new Date(di.getFullYear(), di.getMonth(), di.getDate(), r.hr, r.min);
```
joserebelo, line 64: @gfwilliams the samples are reaching Gadgetbridge with a 10 minute offset:

- A real-time sample had timestamp of `1691245201` in the Gadgetbridge database (14:20:01 UTC)
- The same sample when sent by this function has a timestamp of `1691244600` (14:10:00 UTC)

It looks like the Bangle database persists the samples 10 minutes "before" the activity, so the realtime samples are now offset in the Gadgetbridge database by 10 minutes.

Should we just send them to Gadgetbridge with a -10 minute offset here? Searching for a potential duplicate 10 minutes before the current sample may lead to data inconsistencies.

apps/android/boot.js
```diff
@@ -193,10 +193,31 @@
...
+          gbSend({ t: "act", stp: steps-lastSteps, hrm: lastBPM, rt:1 });
           lastSteps = steps;
         }, event.int*1000);
       },
+      // {t:"actfetch", ts:long}
+      "actfetch": function() {
+        gbSend({t: "actfetch", state: "start"});
+        var actCount = 0;
+        var actCb = function(r) {
+          gbSend({
```
joserebelo, line 205: Forgot to check ts < event.ts before sending.

apps/health/lib.js
gfwilliams: Wow, good spot! Yes, it's tricky - I'd be tempted to offset in the Android lib to try and keep the API the same - having different functions returning +/- 10 min feels like it would be asking for trouble. Also looking at the code it's probably easier once you've got the Date object created anyway as it's just `d.getTime()-60000`

I guess one question is what does Gadgetbridge expect the timestamp to be? At the beginning or end of a time slot? I guess it's possible that we should actually have the live activity sample as -10 minutes? Or maybe even meet in the middle (although that doesn't seem like a great idea)

joserebelo: Checking Huami devices (they have per-minute samples), it looks like they actually send the value at the end of the slot (ie. at 19:47:30, if we sync the last sample will we get will have timestamp 19:47:00).

Which I guess makes sense (at time X we had Y steps).

Changed the logic to perform the conversion here.

apps/android/boot.js
```diff
@@ -258,7 +285,7 @@
   }, 2000));
   NRF.on("disconnect", () => {
     // disable HRM/activity monitoring ('act' message)
-    GB({t:"act",stp:0,hrm:0,int:0}); // just call the handler to save duplication
+    GB({t:"act",stp:0,hrm:0,int:0,rt:1}); // just call the handler to save duplication
```
gfwilliams, line 288: This one isn't needed I think? It just calls the same function in boot.js rather than sending anything to Gadgetbridge - it's basically just replicating Gadgetbridge telling Bangle.js to turn off realtime steps

apps/android/boot.js
```diff
@@ -268,7 +295,7 @@
   // Health tracking
   Bangle.on('health', health=>{
     if (actInterval===undefined) // if 'realtime' we do it differently
-      gbSend({ t: "act", stp: health.steps, hrm: health.bpm });
+      gbSend({ t: "act", stp: health.steps, hrm: health.bpm, rt:1 });
```
gfwilliams, line 298: Pretty sure this one *isn't* realtime? It only happens every 10 minutes when we get health data - so this is basically the health data that we're trying to sync above and we need to not have `rt` set

apps/android/boot.js
joserebelo: Ah - you're right.

gfwilliams:
Thanks for sorting that so quickly - merging now!

apps/gbridge/PROTOCOL.md
joserebelo: Should I revert this from here?

joserebelo: Nevermind :)

gfwilliams: Sorry - I was a bit hasty merging! I'll tweak it here - I realised we probably wanted to send movement health data too

thyttan:
Since metadata says `    {"name":"health","url":"lib.min.js"},` and `lib.min.s` wasn't updated with the new additions the fetching breaks.

Solution is either:
- pull from `lib.js` instead or
- update `lib.min.js`

gfwilliams:
just done I hope!

joserebelo:
Ooops, I thought the minified files were automatically generated when deploying this 😐

thyttan:
Now it works - super nice! :)

@joserebelo The fetching would never be stopped when `readAllRecordsSince` was missing - if that's interesting info.

joserebelo:
> The fetching would never be stopped when readAllRecordsSince was missing - if that's interesting info.

Yup - we only stop it once the Bangle signals that it finished. If that message never arrives, it never stops.

It's a more or less known issue, there used to be a similar problem in Huami devices.

I think we may want to eventually introduce a timeout on the Gadgetbridge side for operations that become stuck.

PR #2890, opened 2023-07-18
android: Send fw and hw on connection

joserebelo:
Add a simple command to get the current process.env.

This should allow us to display the firmware and hardware versions on Gadgetbridge, as well as toggle features that are only available in the Bangle.js 2 (eg. Screenshots).

Additionally, include whether the Bangle is currently programmable, as that might be require for some features that execute js directly.

Used in https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3213


apps/android/boot.js
```diff
@@ -197,6 +197,13 @@
           lastSteps = steps;
         }, event.int*1000);
       },
+      "env": function() {
+        gbSend({
+          t: "env",
+          env: process.env,
```
thyttan, line 203: If we are going to do this on every connection to gadgetbridge maybe we want to only send the info of process.env we know we want over to GB? To lessen the activity of bluetooth data transfer which I think draws more energy than manipulating the data on the bangle? Or maybe this is negligible?

thyttan: On the other hand we also want to keep bootcode as short as possible. I don't know :p

joserebelo: That is a good point. In the line below, it reads setting.json on every connection as well - not sure if desired, or if we have a simpler way of determining if the bangle is programmable or not.

thyttan: I guess we could do some kind of hack over on GB side? that tries to send over code to execute that will send an answer back if programmable is toggled on. If no response is given GB can assume that programmable is toggled off.

thyttan: > On the other hand we also want to keep bootcode as short as possible. I don't know :p

A tangent regarding this, would it help speed up the boot process if we moved much of boot files logic into libraries?

gfwilliams:
Maybe we could just make the Bangle send the info automatically on connection? We should totally filter the data we send as well to just what's required - I just checked and it's 300 bytes which would really impact the connection speed.... We already send battery+force_caledar_sync on every connection so adding a *little* more data wouldn't be a big deal. https://github.com/espruino/BangleApps/blob/master/apps/android/boot.js#L254-L257

I also spotted that we do already handle the version info being sent in Gadgetbridge by the look of it? https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/banglejs/BangleJSDeviceSupport.java#L522-L526

apps/android/boot.js
gfwilliams: > would it help speed up the boot process if we moved much of boot files logic into libraries?

If the code was code that gets called at boot time, it'd actually make it slower. It'd only help if it were code that only gets called occasionally, as then you can push the parsing overhead to the point when the code is first run.

joserebelo:
> Maybe we could just make the Bangle send the info automatically on connection?

I'm not used to being able to control the watch-side of the code :) this is definitely a better idea.

apps/android/boot.js
```diff
@@ -197,6 +197,13 @@
...
+          programmable: ((require('Storage').readJSON('setting.json',1)||{}).blerepl)!==false
```
joserebelo, line 204: @gfwilliams is there any more elegant way of determining this?

apps/android/boot.js
joserebelo: Reduced this to fw, hw and repl.

apps/android/boot.js
```diff
@@ -251,8 +251,17 @@
   // Battery monitor
   function sendBattery() { gbSend({ t: "status", bat: E.getBattery(), chg: Bangle.isCharging()?1:0 }); }
   Bangle.on("charging", sendBattery);
+  function sendEnv() {
+    gbSend({
```
joserebelo, line 255: Do we want to keep this in one line? My understanding is that we minify the js for "release", so it should only become slightly slower during development.

gfwilliams: I'm not quite sure I understand the question? But it'd be better to put it straight in the `connect` handler - not for speed but because I don't believe the minified will inline it, and then it'll leave the extra function stored in RAM. Not a big deal on Bangle.js 2, but on v1 we're pretty tight ram-wise.

sendBattery is only a function because we're calling it from more than one place

apps/android/boot.js
joserebelo: Removed this since I don't have any use-case for it anyway yet.

gfwilliams:
Great - thanks! Please can you bump the version in the metadata and put a line in the changelog and I'll merge it?

joserebelo:
Done.

@gfwilliams one last question before merging this - I do not currently have a use for the fw version on the Gadgetbridge side, other than displaying it. Do you think it's still worth it to send on every connection, or should we just get Gadgetbridge to query the hw version on-connection, if unknown? (should only happen once, and never change).

I am not too familiar with the limitations of the original Bangle, so I'd prefer to impact it as little as possible :)

If sending on every connection is not really a concern for you, I think this is fine to merge.

gfwilliams:
Yes, I think sending it is fine - it's probably maybe 10 chars extra or something like that, so not a big deal - and it's handy for users to be able to see anyway.

PR #2893, opened 2023-07-21
fastload - Monitor *.boot.js and setting.json and force real load on …

halemmerich:
…change

This allows things like `qmsched` to work correctly. When setting.json or one of the boot files is changed in the background and the app does a reload to apply the change, fastload will usually prevent this until another app was launched and that has triggered a real reload when leaving it. In the case of `qmsched` this means the theme could be applied maybe hours later than planned if the watch only changes between clock and launcher or overlay messages.

The method of monitoring writes is a bit hacky (and dangerous in case of errors, recovery requires deleting `fastboot.5.boot.js`), but I could not find better way to implement this without actually changing the `Storage` implementation.

### DIFF
--- a/apps/fastload/ChangeLog
+++ b/apps/fastload/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Allow redirection of loads to the launcher
 0.03: Allow hiding the fastloading info screen
 0.04: (WIP) Allow use of app history when going back (`load()` or `Bangle.load()` calls without specified app).
+0.05: Allow forcing full reload on changes of setting.json or *.boot.js files
--- a/apps/fastload/README.md
+++ b/apps/fastload/README.md
@@ -12,6 +12,7 @@ This allows fast loading of all apps with two conditions:
 * If Quick Launch is installed it can be excluded from app history
 * Allows to redirect all loads usually loading the clock to the launcher instead
 * The "Fastloading..." screen can be switched off
+* Enable monitoring for writes to `*.boot.js` and `setting.json` and force a complete load on writes to these files
 
 ## App history
 
--- a/apps/fastload/boot.js
+++ b/apps/fastload/boot.js
@@ -19,7 +19,41 @@ let loadingScreen = function(){
 
 let cache = s.readJSON("fastload.cache") || {};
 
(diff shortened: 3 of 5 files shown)


gfwilliams:
I'm not sure monitoring writes that way is such a great idea - having something that slows down all writes to storage seems at odds with the idea of "fast load" - the manual copy method will also cause problems if more methods ever get added to Storage.

How about using `require("Storage").hash(/setting\.json|boot\.js/)` - this will create an integer hash based on the addresses of all those files (in reality you probably only really care about `setting.json` since the boot files won't usually change at runtime). If a file changes its address (and so the hash) will change, so all you have to do is check if it's the same at the point you reload? Checking just setting.json with `hash` takes 70ms so it's not too bad

halemmerich:
Checking hashes for all boot files was too slow, which prompted this monitoring solution. But only hashing `setting.json` seems reasonable, at least it solves the problem which initially caused me thinking about this (`qmsched` themes not applied correctly).
Maybe a clean implementation for monitoring files in `Storage` would be an alternative to the current use of hashing in launchers and boot which costs some few hundred ms on each reload?
I will close this for now and come back with a new PR just hashing `setting.json` to solve my immediate problem with `qmsched`.

gfwilliams:
Thanks!

Yeah, I just wonder whether having apps monitoring other apps' writes is a bit over the top. We'll just end up with this massive interconnected web of stuff that's pretty much guaranteed to break at some point...

halemmerich:
I thought more on the lines if asking `Storage` to monitor specific files and just issuing an event or writing a marker file on changes. No contents or more specifics, just that some kind of change happened. Can we get a the file address and check faster than hashing for changes? Maybe that approximates actual change close enough for this use case.

gfwilliams:
> Can we get the file address and check faster than hashing for changes?

Yes - annoyingly because it's not an actual address in memory (storage isn't not memory-mapped) E.getAddressOf(..,1) won't do it, but you could access the var directly: `peek32(E.getAddressOf(require("Storage").read("setting.json")))` - I guess at some point that may break, but it's unlikely to ever change

PR #2895, opened 2023-07-22
iconlaunch - Better performance

halemmerich:
Update:
Firmware change in 2v185 got `drawImage` much faster so baseline is now 23ms per `drawItem` and 420ms startup time with populated cache.
* Reduces array access via index to prevent list searching
* Only clear item area if needed
* Real icon for apps not having an icon (no performance benefit, but better icon :)
* Presort apps into scroller items for faster access
* jit for `drawItem`

Overall result is 460ms startup time and 19ms per `drawItem`, so ~9% longer startup for ~20% faster drawing during scrolling.


gfwilliams:
That's great - thanks! You happy to merge?

I have to admit I'm very surprised about drawImages being faster though! it scans through every image for every pixel rendered - but I guess maybe because it's only rendering a slice of an image that is actually more efficient

gfwilliams:
It'd actually be interesting to do some benchmarks on drawImage - it looks like maybe by default we iterate over all pixels in the image regardless of how many are on screen - so it might be possible to improve the logic to at the very least skip the top + bottom rows that aren't needed

halemmerich:
This works fine for me but currently the precaching of icons trades some better smoothness during the initial scroll for a bit of additional startup time for the caching. I'm not yet sure if I like this tradeoff and might go back on that. Needs a bit more benchmarking to decide this.
The differences between `g.drawImage` and `g.drawImages` are certainly interesting. I had just blamed that on giving the whole work in form of the layers object to native code at once instead of doing it once at a time in a JS-loop. We might have some performance to gain here for all apps :)
At least for scrollers usually the top and bottom items together can be redrawn using the pixel count of only one item since both are cut off, effectively removing one item worth of pixels to draw.

halemmerich:
> it looks like maybe by default we iterate over all pixels in the image

It sure does:
```
drawImage                   21.59729003906
drawImages overlapping      41.20292663574
drawImages not overlapping  41.26739501953
drawImages half optioned    22.51243591308
drawImages full optioned    0.89416503906
drawImage half clipped      21.39587402343
drawImages half clipped     22.62573242187
drawImage full clipped      22.45178222656
drawImages full clipped     0.75035095214
```
Clipping via `setClipRect` and otherwise via `drawImages` options object. Times are averaged over 400 images drawn. `drawImages` used a layers object with 4 images.

So in conclusion `drawImages` is indeed slower as you said until enough pixels are clipped of to make up for the overhead of scanning all images. The additional improvement bringing it over the edge in this case probably came from restructuring the loop code to build the layer object.

gfwilliams:
Thanks for those benchmarks!

> Why does drawImages scan through all images

drawImages scans over the area, checking every image for each pixel and drawing the pixel *once* only.

It's done this way specifically to remove any chance of flicker on Bangle.js 1. Obviously if you can find a way to optimise it, PRs welcome, but the current solution is reasonably tidy.

Please can you try the latest cutting edge build now with a drawImage benchmark? It should be a lot faster - I think with this, you could probably revert the `drawImage->drawImages` changes and you'd see a decent speed boost?

halemmerich:
The benchmark shows good improvements:
```
drawImage                   21.60003662109
drawImages overlapping      41.23794555664
drawImages not overlapping  41.23626708984
drawImages half optioned    22.52113342285
drawImages full optioned    0.82275390625
drawImage half clipped      12.56256103515
drawImages half clipped     22.62023925781
drawImage full clipped      3.20739746093
drawImages full clipped     0.67245483398
```

For `drawItem` I now get about 2ms less than the last end result, so it should be 8ms all in all since the original step from `drawImage` to `drawImages` saved 6ms. So pretty great result overall I would say.

KTibow:
This PR might've broken some stuff, I can't elaborate more because I'm on mobile but before I updated the launcher worked, and now other icons get randomly hidden when I select an icon

halemmerich:
Do you use the direct launch setting? If not, it might be a problem that does no longer show up on the current nightly firmwares.

KTibow:
Okay, I'm fine waiting for the firmware to update then.
Question: I don't have enough icons to scroll. Do I need to worry about the launcher's performance going down because of this?

halemmerich:
If you don't have enough icons to scroll, scrolling will only redraw what is already there in the worst case. In that case it is actually pretty irrelevant how long the redraw takes :)
As for the different icon, yes that would be possible. Showing the first letter is a good idea for this. What app is shown that does not have an icon for you? Probably that app should just get an icon.

KTibow:
> If you don't have enough icons to scroll, scrolling will only redraw what is already there in the worst case

I was more talking about startup performance - but it probably doesn't matter that much given I can't tell if it's slower.

> What app is shown that does not have an icon for you? Probably that app should just get an icon.

Just some custom apps I coded.

PR #2898, opened 2023-07-22
Kineticscroll - Implement kinetic scrolling as replacement for E.showScroller

halemmerich:
Sneakpeek at [https://halemmerich.github.io/BangleApps/?id=kineticscroll](https://halemmerich.github.io/BangleApps/?id=kineticscroll)


thyttan:
Did you consider adding `"ram"` for the drag handling function to maybe improve performance?

halemmerich:
Yes, but I believe the benefits would be negligible, minified it is not that much code and the most expensive operations are drawing the icons.

thyttan:
Ok! 👍 Adding `"ram"` to dragboard made it much snappier in my experience.

gfwilliams:
Looks great - thanks!

PR #2899, opened 2023-07-23
Messagesoverlay - Fix low memory handling when overlay already visible

halemmerich:
The lock handler had kept a reference to the first created overlay buffer, which caused the first message to be shown again on unlock (and used memory). Additionally the switch from normal to low memory mode did not work correctly because the buffer was not recreated with 1 bit depth.


gfwilliams:
Thanks!

Issue #2865, opened 2023-07-05
[Banglejs 2] Filesystem corrupted


pavelmachek:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

It seems like something is wrong with the filesystem. My own application that writes log hangs on startup. "Run" using "recorder" hangs when I tell it to create new track. I can connect with ide, can get file listing, and can even retrieve my log file.

When I attemp to start my app from IDE, I get lot of:

FW addr 0x00448800 fail
Status 0
FW addr 0x00448900 fail
Status 0
FW addr 0x00448a00 fail
Status 0

Any ideas how to debug this? Should I try to preserve the filesystem somehow?

### Installed apps

widlock (0.08), mylocation (0.09), daisy (0.10), magnav (0.06), devstopwatch (0.04), scicalc (0.01), gpsnav (0.08), flow (0.01), f9lander (0.03), orloj (0.01), ac_ac (1.00), astrocalc (0.04), sixths (0.01), widbat (0.11), widclk (0.07), waypoints (0.01), widadjust (0.02), dtlaunch (0.23), altimeter (0.03), weather (0.25), sched (0.22), messageicons (0.05), waypointer (0.09), waypoint_editor (0.01), alarm (0.41), health (0.23), openstmap (0.17), widmessages (0.05), bordle (0.03), notify (0.13), run (0.16), infoclk (0.09), pebbled (0.04), clock_info (0.05), boot (0.58), setting (0.61), messagegui (0.71), messages (0.60), drained (0.03), recorder (0.27)

pavelmachek:
I can reproduce the hang by opening file from cmdline.

```
>let logfile = require("Storage").open("sixths.egt", "a");
=StorageFile: {
  name: "sixths.egt",
  chunk: 4, offset: 0, addr: 0, len: 4802139, mode: 97 }
>logfile.write("\n");
``

How big is the flash? Can the length be correct? Should I just delete the file?

pavelmachek:
When I receive the file using ide, I get:

-rw-r--r-- 1 pavel pavel 122784 Jul  5 21:31 /tmp/sixths.egt

I deleted it using ide, but the hang still remains.

pavelmachek:
I'm getting

weather.wid.js SyntaxError: SyntaxError: Got '*' expected ':'

when trying to launch code from IDE, then the "FW addr" flood described above. It ends with:

FW addr 0x007d9fc8 fail
Status 0
[...]

PR #2900, opened 2023-07-23
Recorder: add clockinfo

bobrippling:
This introduces a clockinfo for the recorder, to allow quick toggling of recording state.

Currently the pause icon is white (for a dark background) - I use `drawImage({ ..., palette })` to alter this based on the `g.theme`, but I'd like the icon to highlight when the clock info is selected, like the step icon.

I can't see how the step icon does it though - it's a [fixed image](https://github.com/espruino/BangleApps/blob/8980be400c6a67809d545ca18a2cb021a23221ce/apps/clock_info/lib.js#L81-L81), which apps then draw just by passing to `g.drawImage()` - does it work by something like the icon is 1-bit and so `drawImage` uses the background or foreground colour as its palette?


gfwilliams:
Great idea, thanks!

Does `Recorder` really need its own category though given it only has one icon? Maybe it's better just as a member of `Bangle`.

Also, maybe if there is no widget (eg the clock face isn't loading widgets) we should just not create a clockinfo at all? I think `return { name: "Bangle", items: [] }` should do it. Currently if there is no widget it'll return `Paused` but I think if it's tapped on, it'll cause an exception. The whole recorder widget isn't great, a library would be better - but it's a bit of a hangover from how things used to be done.

> I'd like the icon to highlight when the clock info is selected, like the step icon.

You need to make sure you create a 1bpp icon, ideally transparent, with white as the color of the icon itself. I think most clocks expect 24x24px too - that's what we do for other clockinfos and it should work fine... If there's any trouble post the image up and I can do it here

bobrippling:
> Does `Recorder` really need its own category though given it only has one icon? Maybe it's better just as a member of `Bangle`.
>
> Also, maybe if there is no widget (eg the clock face isn't loading widgets) we should just not create a clockinfo at all? I think `return { name: "Bangle", items: [] }` should do it. Currently if there is no widget it'll return `Paused` but I think if it's tapped on, it'll cause an exception. The whole recorder widget isn't great, a library would be better - but it's a bit of a hangover from how things used to be done.

Ah yes, it's much nicer like that, and likewise with handling a missing widget - thanks!

> a 1bpp icon, ideally transparent, with white as the color of the icon itself. I think most clocks expect 24x24px too - that's what we do for other clockinfos and it should work fine... If there's any trouble post the image up and I can do it here

Thanks - I believe I have it sorted but can't test on my watch - unable to write and I'm holding off resetting it in case you'd like to debug into the [`FW addr fail` error](https://github.com/espruino/BangleApps/issues/2865#issuecomment-1648495597)

gfwilliams:
Thanks - let's wait until you can test nicely then - thanks for the help debugging storage

bobrippling:
The 1-bit image did the trick, thanks again! I've made a small tweak to `clock_info` too, which helped debug a load issue (`WIDGETS` not being defined), happy to move to another PR if desired.

----

There's one problem: if a clock loads `clock_info` before it does `Bangle.loadWidgets()`, the user will never see the recorder clock info because [`WIDGETS` doesn't exist](https://github.com/espruino/BangleApps/blob/8a2a5851be0f7474f56d5f8b0b5cbb2b305c7b44/apps/recorder/clkinfo.js#L10) at that point.

Out of the clocks that use both `clock_info` and widgets, only [`lcdclock`](https://github.com/espruino/BangleApps/blob/ddc9ac34b1aa502ab5fb147fb05cf33293945dba/apps/lcdclock/app.js) loads widgets first, so all the other clocks won't see the recorder widget.

<details>
<summary>
Clocks using `clock_info` & widgets
</summary>

```sh
grep -l clock_info apps/*/*.js | xargs grep -El 'clock_info|Bangle\.loadWidgets'
```

</details>

To fix this I think would be difficult, we'd have to make it so clocks load `clock_info` after any widget loads, or somehow make the `clock_info` menu dynamic. What do you think?

bobrippling:
Sorted it - we always show the recorder clkinfo and if the widget's not present, the clock hasn't run `Bangle.loadWidgets()` but we just let the user scroll to another clkinfo.

gfwilliams:
Looks great, thanks! Is it worth putting in a PR for LCD clock to change it?

PR #2904, opened 2023-07-24
openstmap: Enable Compass if needed

nxdefiant:
Bugfix: In the last PR I forgot to power the compass. I have a script that powers on the compass in the background which is why I had no issue when testing. Just remembered that.

Sorry for the noise.


gfwilliams:
No problem - thanks for checking it!

PR #2908, opened 2023-07-25
miclock2 maintenance

nxdefiant:
- Register as clock (setui)
- Implement fast loading
- minor speed improvements


gfwilliams:
Thanks!

Issue #1555, opened 2022-03-10
Fix Alarms


gfwilliams:
The current Alarm system is stupid. We now have:

* alarm (the default)
* qalarm
* hardalarm
* Plus some other as yet uncontributed forks

They are all almost identical, but all of them have issues, none of them work together, and none sync with Gadgetbridge. 

Plenty of apps also implement their own timers, which stop when the app is exited:

* chronowid (chrono widget)
* countdowntimer

There are outstanding issues like:

* https://github.com/espruino/BangleApps/issues/1416
* https://github.com/espruino/BangleApps/issues/896
* qalarm: 'hard' is broken on Bangle.js 2
* hardalarm: doesn't work on Bangle.js 2
* alarm: timers don't activate until you leave the app

And I've lost track of the hours I've spent trying to debug and maintain the 3 different alarm apps, which I could have spent making real improvements to a single app.

Also it'd be nice to have an API where other apps can query the alarm state, like https://github.com/espruino/BangleApps/pull/1513

So the plan is to:

* Have a 'core alarm' library that provides an API and well tested, reliable and efficient code trigger alarms, but that doesn't contain any apps/widgets so folks can make their own that depend on it
* Try and merge features from other apps into the default alarm app, and remove the others.
* Add Gadgetbridge alarm syncing support

andrewgoz:
[TimerClock](https://banglejs.com/apps/?id=timerclk) is my current clock and it implements its own alarm system unrelated to anything else. It doesn't appear to work with [SleepPhaseAlarm](https://banglejs.com/apps/?id=sleepphasealarm), which I assume is designed to work with the default alarm app. A common API would be useful here!

I've been making some changes to TimerClock (no PRs, yet). If I have time I might look into making it compatible with the default alarm app (or whatever SleepPhaseAlarm is trying to do), unless something happens here first!

I think it would be a good idea if alarms required the ID of the app that creates them, so apps could query the global alarm list for 'their' alarms.

nxdefiant:
Yes, SleepPhaseAlarm checks only the config alarm.json of the default alarm for enabled buzzers.

gfwilliams:
> Yes, SleepPhaseAlarm checks only the config alarm.json of the default alarm for enabled buzzers.

Ok, great - please keep doing that as I'll use `alarm.json` for this thing.

[...]

PR #2909, opened 2023-07-25
timerclk: Improve swipe responsiveness on main clock screen

ticalc-travis:
On my Bangle 2 I noticed that the existing Timer Clock app seems to be unusually fickle with registering swipes to switch to the various screens compared to other apps, sometimes requiring me several attempted swipes to get to the desired mode. So I took a look at the code and managed to make a couple of small changes that make it a lot more reliable.

The first change significantly increases the sensitivity (that is, decreases the minimum swipe length) required to activate the switch to timer, stopwatch, or alarm modes. Since the main screen only cares about left vs. right swipes, decreasing the length to just 5 from 50 works well, as the Bangle 2 touchscreen already seems to have a fair bit of hysteresis in movement before registering a drag to begin with.

The other change fixes a coding error where the variables involved in summing up the swipe length were not initialized at the start, resulting in a sum of NaN. As a result, the first swipe upon waking the screen was pretty much guaranteed not to be recognized no matter what. With this change, the first swipe works reliably.

I fixed the similar variable initialization issue for the stopwatch/timer/alarm screens, but I did not tweak the sensitivity for those here because those do both vertical and horizontal swipe detection and I found that allowing too-short swipes makes it more prone to misreading the axis of the intended swipe. Ideally, I think this could improved by reworking the touch code in a different way, but I'm a bit new to this still so I'll leave it at that for now. :)

### DIFF
--- a/apps/timerclk/ChangeLog
+++ b/apps/timerclk/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Add sunrise/sunset. Fix timer bugs.
 0.03: Use default Bangle formatter for booleans
 0.04: Use 'modules/suncalc.js' to avoid it being copied 8 times for different apps
+0.05: Improve responsiveness and detection of swipes on main clock screen
--- a/apps/timerclk/app.js
+++ b/apps/timerclk/app.js
@@ -148,23 +148,23 @@ if (process.env.HWVERSION==1) {
   setWatch(()=>load("timerclk.alarm.js"), BTN3);
   setWatch(()=>load("timerclk.alarm.js"), BTN1);
 } else {
-  var absY, lastX, lastY;
+  var absY, lastX=0, lastY=0;
   Bangle.on('drag', e=>{
     if (!e.b) {
-      if (lastX > 50) { // right
+      if (lastX > 5) { // right
         if (absY < dragBorder) { // drag over time
           load("timerclk.timer.js");
         }else { // drag over date/dow
           load("timerclk.alarm.js");
         }
-      } else if (lastX < -50) { // left
+      } else if (lastX < -5) { // left
         if (absY < dragBorder) { // drag over time
           load("timerclk.stopwatch.js");
         }else { // drag over date/dow
           load("timerclk.alarm.js");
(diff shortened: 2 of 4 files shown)


thyttan:
I guess there might be a reason for swipes being implemented with the `"drag"` event here.

But maybe some or all swipe interactions could be handled with the `"swipe"` event instead? I find it's much more responsive, and it still allows for up/down and left/right swipes (I think it only supported left/right at first, maybe that is the reason for timerclk using `"drag"`?).

ticalc-travis:
I'm assuming it's because it wants to know where on the touch screen the swipe occurred (the alarm requires a horizontal swipe near the bottom, and this is also used on other screens to detect button presses and which digit to change). It seems that accessing the alarm could just as well have been done with a down-swipe instead; can't say why it was done the way it was.

thyttan:
Also if your up for it  in a future PR, it looks like timerclk is yet to be refactored into using [the standard scheduling library `sched`](https://github.com/espruino/BangleApps/tree/master/apps/sched).

Reference https://github.com/espruino/BangleApps/issues/1555#issuecomment-1080259472

gfwilliams:
Great! as @thyttan mentioned it's be great to use the sched library at some point as not having extra boot code will make things load faster.

Just to check, it wouldn't be a big problem if a tap got taken as a swipe would it? It's probably quite likely that with only 5px movement, pretty much any press would pass that test?

ticalc-travis:
The sched thing makes sense. I kind of got the idea that it was doing its own implementation when I tried to set an alarm/timer in this app when I had another one set using an app based on sched, and ended up with two separate alarm icons in the widget bar, which took up space and looked quite silly. :)

I did consider that tapping would probably be a better choice than swiping, as it's easier to do and this part of the app doesn't do anything with taps anyway. (It would also be nice when setting timers/alarms if a single swipe could change the number by multiple steps like the built-in number changer UI, as one step per swipe is rather tedious.) But I thought it might be considerate to try to contact the original author to see how they would feel about such changes.

PR #2912, opened 2023-07-26
btadv: sensor location & accelerometer adverts

### LINKED ISSUE #2907: [app/btadv] Acceleromter is not working
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

App works fine except for acceleromter advertising. Seems the part accl sensor is missing. Any plan to complete it ?

### Installed apps

_No response_

bobrippling:
Add accelerometer adverts and sensor location to `btadv`

Closes #2907

### DIFF
--- a/apps/btadv/ChangeLog
+++ b/apps/btadv/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New app!
+0.02: Advertise accelerometer data and sensor location
--- a/apps/btadv/app.js
+++ b/apps/btadv/app.js
@@ -352,13 +352,13 @@ var _a;
             case "E95D0753251D470AA062FA1922DFA9A8": {
                 var o = {};
                 if (acc_1 || initial) {
-                    o["0xE95DCA4B251D470AA062FA1922DFA9A8"] = {
+                    o["E95DCA4B251D470AA062FA1922DFA9A8"] = {
                         maxLen: encodeAcc_1.maxLen,
                         readable: true,
                         notify: true,
                     };
                     if (acc_1) {
-                        o["0xE95DCA4B251D470AA062FA1922DFA9A8"].value = encodeAcc_1(acc_1);
+                        o["E95DCA4B251D470AA062FA1922DFA9A8"].value = encodeAcc_1(acc_1);
                         acc_1 = undefined;
                     }
                 }
--- a/apps/btadv/app.js
+++ b/apps/btadv/app.js
@@ -221,15 +221,20 @@ var _a;
         ];
     };
     encodeGpsHeadingOnly_1.maxLen = 17;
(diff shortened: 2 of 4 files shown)


gfwilliams:
You think this is ready now?

bobrippling:
I think it'll be ready within the week, waiting on feedback in #2907 to confirm it works well

bobrippling:
This is ready now - there's [an issue](https://github.com/espruino/BangleApps/issues/2907#issuecomment-1659589783) with connecting over BLE when the app's running that I've not found the cause of yet, but I'll move that to a separate PR.

> 3. after turn on the sensors, ble connections from bangle.js web ide/apploader will always fail while the app is on, connections from phone has no problem.

bobrippling:
Could it be that changing the ble advert prevents the apploader & ide from connecting, as they're looking for a uart service? Although that should still be advertised.
I'll try to capture [uncaught exceptions](https://www.espruino.com/ReferenceBANGLEJS2#l_process_uncaughtException), save them to storage and do a post mortem on the app

gfwilliams:
Thanks! I'll merge then

> Could it be that changing the ble advert prevents the apploader & ide from connecting, as they're looking for a uart service? 

Yes... the apploader/IDE looks for UART service, *or* a name it recognises like Bangle.js - if both of those get removed then it won't see it... You could check with NRF Connect and see what *it* sees?

Maybe just add a `on('kill',...)` handler that resets advertising when the app exits?

bobrippling:
Carrying on from this, I've attempted to capture what's happening when `btadv` prevents BLE connections from being established to it. I suspect it's trying to cram too much into the BLE advertising packets which is resulting in garbage.

### Issue

Summarised in #2907:

> after turn on the sensors, ble connections from bangle.js web ide/apploader will always fail while the app is on, connections from phone has no problem

### Exception Logging

No error logs appear except those mentioned in #2907, [about not being able to update services until restart](https://github.com/espruino/BangleApps/issues/2907#issuecomment-1659589783):
```
Error: Can't update services until BLE restart
```

<details>
<summary>Stashing this code here for non-IDE collection of logs</summary>

```javascript
{
  let f = require("Storage").open("uncaught.log", "a");

  process.on(
    'uncaughtException',
    e => {
      console.log("uncaught: ", e);
      f.write((new Date).toISOString() + " " + e + "\n");
    });
}
```

</details>

### External Diganosis

I notice my watch advertises a HID service (`0x1812`), which isn't set in `btadv` or any app, for that matter:

image

Which leads me to the above mentioned idea that I've put the NRF chip into a broken state, seeing as a hard reset is required to be able to connect to my watch. I'll see about altering btadv / reverting some more recent changes to see if I can undo this and restore its advertising functionality.

PR #2916, opened 2023-07-27
fastload: fix caching

ssievert42:
broken since e2ba3ca64 (fastload - Replace E.CRC32 with Storage.hash, 2023-07-24)

Storage.hash() expects a regex (or apparently the name as a string) to match files that should be hashed, but was called with the contents of the file

### DIFF
--- a/apps/fastload/ChangeLog
+++ b/apps/fastload/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Allow hiding the fastloading info screen
 0.04: (WIP) Allow use of app history when going back (`load()` or `Bangle.load()` calls without specified app).
 0.05: Check for changes in setting.js and force real reload if needed
+0.06: Fix caching whether an app is fastloadable
--- a/apps/fastload/boot.js
+++ b/apps/fastload/boot.js
@@ -30,12 +30,13 @@ let appFastloadPossible = function(n){
 
   // no widgets, no problem
   if (!global.WIDGETS) return true;
-  let app = s.read(n);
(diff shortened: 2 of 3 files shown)


halemmerich:
Nice find! Can I ask how you noticed the bug? Did you have a look at the cache file or was there some odd behaviour?

ssievert42:
Thanks :)
I just got my Bangle (a truly wonderful device!) a few days ago and me trying to change two things kind of led to this:
First off, it seemed like some apps really like writing to storage (like for example activityreminder) which I wanted to avoid. So now I have hacky versions of both activityreminder and quicklaunch, that only write on `E.kill`, but still need some love before they can be considered merge request worthy.
Secondly, I wanted apps to load faster, and fastload provided me with a solution. But only for some apps, and closing apps still led to the "loading" message and `E.kill` being fired.
Thus I gobbled up all information I could find on fastloading and unloading the current app.
While grepping through my installed apps to look for storage writes, I stumbled upon fastload writing to a cache file.
To finally answer your question: No odd behaviour; I just kind of enjoy reading and understanding code, and passing the contents of a file to `Storage.hash()` seemed off :sweat_smile:

ssievert42:
Just noticed that I forgot to bump the version; should be fixed now :sweat_smile:

gfwilliams:
That's great! Thanks for finding and getting a fix in for this!

Once you're happy with them, PRs for the tweaks you had to activityreminder and other apps would be really helpful too

PR #2917, opened 2023-07-27
multitimer: ensure onDrag is registered after all E.showXYZ calls

bobrippling:
This fixes multitimer's fastload support, from #2879


gfwilliams:
Thanks!

PR #2918, opened 2023-07-27
widshipbell: Bump version as commit 2b6f776

nxdefiant:
Load new buzz.js - see 2b6f7765cc69db5e8dd34fa9ad86051914e26601


nxdefiant:
Yes, normally I thought I would be the only one using this app, but the app loader reports two installations.

PR #2920, opened 2023-07-27
widedit optimisation

bobrippling:
This removes a second sort from `widedit`


gfwilliams:
Thanks!

rigrig:
That sort was actually [added](https://github.com/bobrippling/BangleApps/commit/9a8ccf0f4f5a206c936ab0bcfa93364ed5c7b363) to keep widgets with the same `.sortorder` in alphabetical order. (Definitely could've used a comment though)

bobrippling:
Ah interesting, I've brought it back with a comment (#2938), thanks for letting me know!

PR #2923, opened 2023-07-28
gipy: new release : bugfix for negative coordinates + experimental powersaving

wagnerf42:
hi,

i took more time than expected but here is a new release.
it fixes the bug which was reported last time.
i also played a bit with powersaving ideas however since i met the hard reboots i mentioned on the forum i had to pinpoint the problems in a very manual way and it took me way more time than i expected.

i had no time to try a large trip but i used it a few times to go to work (15km) and it seems ok.
normally i would test a bit more before releasing but i'm going to be offline for a few weeks so i guess it's best to release now.


nxdefiant:
I'm not sure I understand the display thing: Does this means the gipy now "randomly" enables & disables the display backlight? Why don't just keep the backlight disabled like in all previous versions?

wagnerf42:
i'm not sure what you mean with the backlight being disabled in the previous versions. for me it was always enabled. isn't it enabled by default ? and then disabled when idle ?

wagnerf42:
note that there is a setting for the max backlight level that you can set to 0. also,there is another setting to turn the screen off completely between segments ends.

nxdefiant:
Interesting, I think I do understand motive a bit more. Let me tell you about my setup:
The only wakeup source I have enabled is the button, so no wakeup with twist or anything. So for me the default on the watch is backlight off until I enable it with the button.

Until the last gipy version I was just driving with the backlight off and could read the display just fine so for me an automatic turn on of the display is an extra power draw.

wagnerf42:
i see. i didn't know there was a global lcd settings in the options.
i'm not too sure what's the way to go here. since there is already a 'brightness' option to set the backlight to 0 isn't it good enough in its current state ? that's actually what i use. i also don't want the backlight. i defaulted to 0.5 because i was not sure what a good setting is for an end user.  also you might want to turn the light on at night so you still need a way to turn it on. 

i'm not too sure about a 'no changes to lcd' option. because it's kind of confusing to have both a 'no changes to lcd' and a 'brightness' setting.

nxdefiant:
One way would be to have brightness go from -0.1 to 1 instead of 0 to 1.
In settings < 0 would be formatted as "leave unchanged" and have it handled in the app accordingly.

wagnerf42:
if you are like me and your default lcd setting is 'no backlight when idle' you might think you are fine with the global setting but you are not. because of the road bumps this kind of  setting will behave badly and it's actually not soo easy to figure it out. and note that this is the default global setting.

that's why if we have a 'no changes to lcd' i would like it disabled by default (by default we change the lcd).

so, that's a possibility but is it worth it ? you'll still need to ajust the settings to your taste. and setting the brightness to 0 will have the same effect.

nxdefiant:
I'm ok with default to change the brightness as long as I have the option to disable it.

Another idea: We could read the global lcd settings and only do the gipy lcd changes if the wake on twist/face up are enabled.

PR #2925, opened 2023-07-28
widclkscrl: New widget

nxdefiant:
A widget that displays the current date & time after unlocking the watch when not showing a fullscreen clock. The information is scrolled by in a two digit field, so this widget is kept tight.


gfwilliams:
Thanks!

PR #2926, opened 2023-07-28
widanclk: add todays date behind clock hands.

thyttan:
Currently adds the days date behind the clock hands in the `g.theme.bgH` color.

Should this:
- just be added like this, always on?
- be configurable via settings?
- be a new separate widget?

### DIFF
--- a/apps/widanclk/ChangeLog
+++ b/apps/widanclk/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New app
 0.02: Clear between redraws
+0.03: Add todays date behind the clock hands.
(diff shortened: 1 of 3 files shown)


gfwilliams:
Well, I'm fine with your changes. I guess maybe if someone complains we could revert and split it out into a new widget or add a settings page for it?

But maybe you just could mention in the description the numbers are the date?

thyttan:
Ok! It's ready - I added to the description.

gfwilliams:
Thanks!

PR #2928, opened 2023-07-29
[rescalc] - v0.04 update

stweedo:
This update not only enhances code efficiency and readability, but it also rectifies previous inaccuracies in converting resistance values to color bands, specifically those involving decimal resistances.


gfwilliams:
Thanks!

PR #2929, opened 2023-07-30
Alarm: Fix date not getting saved when tapping Confirm in event menu

ticalc-travis:
This fixes a bug I discovered in the Alarms app. When editing an event, any changes to the date do not get saved to the alarm if the user taps Confirm on the menu… only when pressing back. This is because the code for saving the alarm when tapping back did not match the code for the Confirm menu button; the latter only saved the time and not the date.

This patch adjusts the two pieces of code to match each other. Probably an even better idea would be to wrap this “save and quit” code into a small function and reference it by name in both the back and Confirm button entries rather than copy/paste the code. This would ensure that both actions in the UI cannot accidentally diverge and cause bugs like this.


gfwilliams:
Great - thanks for finding and fixing this!

PR #2931, opened 2023-07-31
widbaroalarm: Compatibility with hideable Widgets

nxdefiant:
Adds Widget on initialization, before widget_utils replaces the draw method.


gfwilliams:
Great - thanks for spotting that!

Issue #2910, opened 2023-07-26
[backswipe] Feature Request: Disable on some keyboard libraries


thyttan:
There might be an app where it makes sense for backswipe to be active generally but the user uses a keyboard library that uses the right swipe for it's interface. In this situation it would be ideal if backswipe did not execute when the keyboard was active in the app, but was otherwise active.

Edit: An example app and keyboard combination is `podadrem` + `kbmulti`.

bobrippling:
I think this ties in with a few other event handler discussions that've been going on recently (e.g. #2743).
We have `E.stopEventPropagation()` to skip subsequent handlers, but there's not the opposite - a way to ignore the event in favour of a subsequent handler, although it's something we could do, e.g.

```javascript
handler = () => {
	var onSwipe = Bangle["#onswipe"];
	if (typeof onSwipe !== "function") {
		var i = onSwipe.indexOf(handler);
		if (i >= 0 && i < onSwipe.length - 1) {
			// there's another handler after us, defer to it
			return;
		}
	}

	// normal swipe handling code
};

Bangle.on("swipe", handler);
```

thyttan:
An example app and keyboard combination is `podadrem` + `kbmulti`.

thyttan:
>We have E.stopEventPropagation() to skip subsequent handlers, but there's not the opposite - a way to ignore the event in favour of a subsequent handler, although it's something we could do, e.g.

I haven't tried to see if `E.stopEventPropagation()` added to `kbmulti` solves the issue. I'll try that.

thyttan:
Adding `E.stopEventPropagation();` to `kbmulti` didn't help. 

I guess there are at least two other ways to go about this feature request besides your idea @bobrippling.

If there was a boolean telling whether or not `textinput` is active that condition could be added to the if-statement in `backswipe`-s `goBack` function.

Or something along these lines:

>Thinking out loud:
> When starting a listener with `Bangle.on()` or in `Bangle.setUI()` maybe there could be an argument passed that specified priority/where in the order of handlers it should run? The argument could have values reflecting first/last position or a number of "priority groups". Maybe nesting would be possible so you could say "this handler should run last in priority group 1" or "this handler should be run in subgroup 2 in priority group 3".

_Originally posted by @thyttan in https://github.com/espruino/BangleApps/issues/2743#issuecomment-1546910113_

gfwilliams:
> maybe there could be an argument passed that specified priority/where in the order of handlers it should run?

[...]

PR #2934, opened 2023-07-31
dragboard: Compat w backswipe on newer firmwares

thyttan:
Reference https://github.com/espruino/BangleApps/issues/2910.

This should work, does in the testing I've done on fw 2v18.115. Also on older firmwares if I did things correctly (not tested).

### DIFF
--- a/apps/dragboard/lib.js
+++ b/apps/dragboard/lib.js
@@ -103,26 +103,7 @@ exports.input = function(options) {
   initDraw();
   //setTimeout(initDraw, 0); // So Bangle.appRect reads the correct environment. It would draw off to the side sometimes otherwise.
 
-  function changeCase(abcHL) {
-    if (settings.uppercase) return;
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks!

apps/dragboard/lib.js
```diff
@@ -243,7 +224,34 @@ exports.input = function(options) {
             }
           }
         }
-      }
+      };
+
+  let catchSwipe = (_,__)=>{
```
thyttan, line 229: @gfwilliams, mostly out of curiosity, would it be smarter to just do `let catchSwipe = ()=>(` here, without the arguments? If I get to add this to other keyboards I can do whatever is better also.

gfwilliams: Yes, if you're not using the arguments it's better not to add them - it's not much but each argument you don't define saves you one variable's worth of memory

PR #2936, opened 2023-07-31
draguboard: Compat w backswipe on newer firmwares

thyttan:
Reference https://github.com/espruino/BangleApps/pull/2934 and https://github.com/espruino/BangleApps/issues/2910.

This should work, does in the testing I've done on fw 2v18.115. Also shouldn't break on older firmwares if I did things right (not tested).


gfwilliams:
Thanks!

PR #2939, opened 2023-08-01
Allow zooming map in/out by clicking.

pavelmachek:
By simply clicking on screen instead of going to menu to zoom in/out, navigation can be much quicker.


nxdefiant:
May I suggest to add a hint on where to click to zoom on/out in the readme or changelog?

pavelmachek:
nxdefiant: Yes, it makes sense to document it in the README. Done now.

gfwilliams:
Thanks!

PR #2940, opened 2023-08-02
Sixth sense -- outdoor watch with GPS support.

pavelmachek:
Add initial version of the application. I already use it a lot, as it displays just the data I need. It can log GPS in power-saving mode, which is useful on its own, and also makes GPS ready for use.


gfwilliams:
Looks great, thanks! You might find you get more folks trying it out if there's a screenshot for it though.

Just a note but if you're wanting to log GPS it might be worth trying to use `recorder` and just toggling it from the clock - it's not just the ability to log other data as well, but it also contains code to read the log back in the App Loader and convert it to GPX/KML files - and it would be a shame to end up duplicating that

thyttan:
Cool clock face!

One question: The input indication will be drawn in top left corner if i `drag` in the lower right corner and vice versa. Is this intended?

Two small bugs:
- The input indication will be drawn over the widgets field and not be cleared. I suggest only draw on the `appRect`, or draw the widgets again after the user finished the `drag`-ing.
- The app doesn't tell the bangle that it's a clock, so `widanclk` will draw in the widget field.

pavelmachek:
thyttan: Thanks for comments. Yes, reversed input indication was intentional debugging feature. I should probably disable it. I'm not currently using widanclk, will Bangle.setUI({
    mode : "clock"
}); do the trick?

thyttan:
That should do it, yes! 🙂

PR #2942, opened 2023-08-02
health: Correct calculation of daily summary for movement

ticalc-travis:
This corrects a couple of logic errors in the daily summary calculation that result in bad data being written for the movement field.

One was that the wrong byte of data was being summed (the heart rate rather than movement); the other was that the data was divided by 8 to fit into a byte for each record, but then the same routine was called for the daily average, which resulted in an erroneous second division by 8.

The simplest and least invasive way to fix the division issue I could think of was to move the division out of the function getRecordData() that wraps up the data into a binary record and instead put it in the main function just ahead of its initial call to write the normal DB records. The later summary code doesn't need to divide it again when calling getRecordData().

The line of code that copies the data object before modifying it to divide movement by 8 (`health = Object.assign({}, health);`) might not be strictly necessary. I added it for safety because I wasn't sure whether it's safe to mutate the actual object passed to the health event handler. (Several years of Python have taught me to be extremely wary of mutating objects that have a remote possibility of also being used by something else!)


ticalc-travis:
On second thought, I realized that the movement daily summary averages still don't quite mesh with the hourly values as shown in the bar graph in the Health UI. I didn't realize that the graph adds together the movement values for each hour while the summary code does a flat average of all the 10-minute records instead. This means that the daily summary in the graph will still not reflect an appropriate average (or sum, as the case may be) of the hourly data. 

I'm not sure if it's ever been established what the exact definition of the movement data summary should be or in what way it's intended to be used, so I'm not really sure what the right thing to do here would be.

ticalc-travis:
Yes, I'm referring to the graphs on the watch. What I mean is that for movement, these graphs seem to add up all the records for each hour and then display that, just as it does with step counts, whereas with heart rate it does an average. Of course, it makes sense to add up steps and average heart rate, but movement seems not as clear cut. It would probably be more consistent to have the hourly graph average movement rather than sum it since we're already doing that for the daily data.

(Personally, it seems a bit confusing also to divide movement by 8 in the first place trying to save space in the DB since that causes the data reported in this app to vary from what the system reports to other apps (for instance, it would have been nice to be able to relate the threshold settings in the sleep tracker app which doesn't divide by 8 directly to the movement values reported here), but unfortunately it seems too late to break compatibility and risk bugs to attempt to change that.)

I'd sort of been thinking about ways to improve the on-watch graphs myself. For instance, I often want to be able to view hourly data from before midnight, but currently there isn't a way to do that. (The same seems to be true of daily data when the month has changed.) Ideally you'd be able to drill down and view detailed data from any point that's available. Plus perhaps enhancing the input handling on Bangle v2 so that graphs are quicker and smoother to scroll through. Lots of ideas, just such limited time though! :)

gfwilliams:
For average vs sum, assuming there is a record for each it should make no difference to the graph apart from the scale being 6x as high? But you're right, average makes far more sense if any records are missing.

> divide movement by 8

It was just because the value is unscaled anyway, and it just didn't make sense to make the whole database 20% bigger for something that nobody was using at the time anyway. *But* the health lib `read*` functions should definitely multiply by 8 again to keep the scale the same.

I've just checked and I can't see any apps in the app loader using the 'movement' field reported by those functions, so I reckon we're safe to add a 'x8' to the fields reported back.

I've just committed those two changes (I hope!).

> I often want to be able to view hourly data from before midnight, but currently there isn't a way to do that in the app

Yes, that'd be a great feature to add.

It'd also be nice sometimes to include the current health data - for instance to see the steps for today in the daily graph as well as the historical data (these can be got from `Bangle.getHealthStatus`)

ticalc-travis:
Yeah, I don't think it would mathematically make much difference for the graph to average instead of sum. The case of missing records for average is a good point, though; I hadn't thought of that.

Multiplying the returned value by 8 to rescale it seems to make sense too. It technically means less precision than the full value, but maybe not enough to matter. (It's no worse than it already is, at least.)

ticalc-travis:
> It was just because the value is unscaled anyway, and it just didn't make sense to make the whole database 20% bigger for something that nobody was using at the time anyway. _But_ the health lib `read*` functions should definitely multiply by 8 again to keep the scale the same.
> 
> I've just checked and I can't see any apps in the app loader using the 'movement' field reported by those functions, so I reckon we're safe to add a 'x8' to the fields reported back.
> 
> I've just committed those two changes (I hope!).

Just got around to testing it. There was a copy/paste error which PR #2986 should fix. Other than that it seems to look reasonable so far. :)

PR #2946, opened 2023-08-03
kbtouch: Compat w backswipe on newer firmwares

thyttan:
Reference https://github.com/espruino/BangleApps/pull/2934 and https://github.com/espruino/BangleApps/issues/2910.


gfwilliams:
Thanks!

PR #2948, opened 2023-08-03
draguboard: remove arguments from `catchSwipe`

thyttan:
Reference https://github.com/espruino/BangleApps/pull/2934 and https://github.com/espruino/BangleApps/pull/2936.

I don't think it warrants a version bump.


gfwilliams:
Thanks!

PR #2949, opened 2023-08-03
chargerot: handle missing settings (e.g. first install)

bobrippling:
A small tweak to `chargerot` to avoid exceptions on boot


gfwilliams:
That's great - thanks!

PR #2950, opened 2023-08-05
Calendar: Edit holidays on device in settings

nxdefiant:
settings: Implement code to edit holidays on device
app: Only refactoring, no changed functionality


gfwilliams:
Great - thanks!

PR #2951, opened 2023-08-05
android: Allow alarm enable/disable

joserebelo:
Allow alarms to be enabled or disabled from Gadgetbridge, without deleting them from the watch.

Needs https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3240, but stays backwards compatible.


gfwilliams:
Great - thanks!

PR #2952, opened 2023-08-06
Gpstrek

halemmerich:
- Adds settings
- Draws direction arrows on route
- Turn off compass when GPS fix is available to save some power


gfwilliams:
Thanks!

PR #2955, opened 2023-08-08
lcdclock: load clockinfo before widgets

bobrippling:
This just rearranges code to match the behaviour of other clocks, meaning any dependencies between widgets and clock info is consistent

See also #2900 / [this comment](https://github.com/espruino/BangleApps/pull/2900#issuecomment-1663510899)


gfwilliams:
Thanks!

PR #2956, opened 2023-08-08
bootgattbat & bthometemp: handle bleAdvert array

bobrippling:
Following on from #2912, this handles `Bangle.bleAdvert` being an array instead of an object


gfwilliams:
Thanks!

PR #2957, opened 2023-08-08
Added USGS app

inhof009:
App will obtain USGS water data and display on screen when opened. Location and what types of data to be obtained are modifiable. Requires http to obtain data, and keyboard to modify location.


gfwilliams:
Looks great, thanks!

gfwilliams:
I just made some small changes to this - I'm pretty sure the settings page wouldn't have worked as-is (FILE/settings weren't dfined). I think it's ok now but it might be worth checking.

inhof009:
Thank you for catching that! That definition got lost when transitioning code between my testing environments. Tested it out, and works as intended!

PR #2958, opened 2023-08-08
Ensure progress toast is visible above modal dialogs

joserebelo:
The modal dialog has a z-index of 400, so the toasts would not be visible (eg. while downloading data from the health app).

Setting a z-index of 500 fixes this.


gfwilliams:
Thanks! I've been noticing that for a while but I never thought to change it :)

PR #2959, opened 2023-08-09
USGS v0.02

inhof009:
Edited menu items to reflect actual functionality, instead of "my first app" names that were present.

Added button interactions, 1 press to refresh data, 2 presses to return to launcher.


gfwilliams:
Great - thanks!

PR #2964, opened 2023-08-11
add edgeclk

tinxx:
Hey guys, I added my personal clock/watch face.

Let me know if something is not OK :)


gfwilliams:
it looks great - thanks!

thyttan:
Nice one!

I noticed if I switch to launcher (dtlaunch) and back not all widgets are drawn, just the 'light switch' one.

I have fastload utils installed.

tinxx:
Hey @thyttan, thanks for your comment!

I am using the same launcher and have tried quite some widgets during the last days, without any problems.

Is it possible that the issue has to do with your and @halemmerich's *fastload utils*? I haven't got them installed on my Bangle, and they are clearly marked as *EXPERIMENTAL*.

I am new to Bangle and Espruino, so I am happy if you could give me some hints if I did something not quite right.

Cheers
@tinxx

thyttan:
I tested some by removing, in order:
- Fastload utils
- Backswipe
- Light Switch Widget

and only after I removed Light Switch did the widgets draw correctly. So there's some incompatibility between edge clock and that. Light Switch Widget is often a little tricky.. 😅 

So probably something should change in Light switch widget. However I don't get the problem on e.g. BW Clock light.

tinxx:
Hey @thyttan, light switch looks like a rather complex app. I think it is possible that the problem lies there.

On the other hand, bwclock light does not implement fast loading in the same way as I do it. It uses function context instead of scope.

Does the issue occur also on bwclock as well?

Maybe it is best to open a ticket for the matter and document your findings there. Feel free to mention me so I get notified!

thyttan:
@tinxx my bad, using BW Clock Lite the problem also shows up.

I'll open a ticket or a PR 👍

PR #2965, opened 2023-08-12
edgeclk: fix bar reset and 100k+ steps

tinxx:
Last night at midnight I realized that I did not clear the progress bars before drawing new gauge. I fixed that and now handle display of too many steps ;)


gfwilliams:
Thanks!

PR #2966, opened 2023-08-13
Add prompt before shutdown

matijatosic:
Small edit of the Poweroff app to add a prompt and prevent accidental shutdown

### DIFF
--- a/apps/poweroff/app.js
+++ b/apps/poweroff/app.js
@@ -1,13 +1,23 @@
 g.clear();
 
-g.setFont("6x8",2).setFontAlign(0,0);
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! Although I think this PR might accidentally contain some slopeclockpp changes too?

If you do a PR from your master branch on GitHub and then make more changes, those get mirrored too. Is it possible to do a new PR with just the poweroff changes?

matijatosic:
Ah, sorry about that. Here's the correct one: https://github.com/espruino/BangleApps/pull/2969

PR #2967, opened 2023-08-13
Add missing gx namespace for kml extensions

StefanBruens:
According to https://developers.google.com/kml/documentation/kmlreference#kml-extension-namespace-and-the-gx-prefix

> This namespace URI must be added to the <kml> element in any KML file using gx-prefixed elements

As the exported kml tracks always contain such elements, add the namespace.


gfwilliams:
Great! Thanks for looking into this! Out of interest, was there a particular app that had some problem with this?

StefanBruens:
> Great! Thanks for looking into this! Out of interest, was there a particular app that had some problem with this?

KDE's Marble was choking on it.

PR #2968, opened 2023-08-13
Fix hwid_a_battery_widget background color

matijatosic:
In some cases, if another app/widget changes the background color, hwid_a_battery_widget is drawn incorrectly.


gfwilliams:
Great - thanks for spotting and fixing this!

PR #2969, opened 2023-08-14
Add prompt before shutdown

matijatosic:
Small edit of the Poweroff app to add a prompt and prevent accidental shutdown


gfwilliams:
Thanks!

PR #2971, opened 2023-08-15
lcdclock: adjust when appRect is read

### LINKED ISSUE #2970: [lcdclock] Rendering issue in version 0.04
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### Describe the bug
After upgrading to v0.04, the clockface looks strange. It seems the whole clockface moves upwards and collides with the widgets:
IMG_0300
Problem persists after I lock then unlock the screen:
IMG_0301

### Expected behavior
Clockface looks normal after I apply the workaround below:
IMG_0302

### Step to reproduce
1. Open Settting
2. Go back to clockface by long pressing the button.

### Workaround
Go to app launcher and then go back by touching the arrow on top left of the screen.

### Installed apps

_No response_

bobrippling:
Fixes #2970

### DIFF
--- a/apps/lcdclock/ChangeLog
+++ b/apps/lcdclock/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Use clock_info module as an app
 0.03: clock_info now uses app name to maintain settings specifically for this clock face
 0.04: clock_info is loaded before widgets to match other clocks
+0.05: fix appRect calculation for above clock_info, widgets & setUI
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks like we both went to fix this - i did it in https://github.com/espruino/BangleApps/commit/656b8a275c1d737c0a383cd3b6c45b6fa3c5ada5

appRect's value depends on whether widgets are loaded - which is what's causing the issue here

bobrippling:
Ah nice! Beat me to it this time lol

PR #2973, opened 2023-08-16
boot, setting: whitelist: try to resolve peer address

ssievert42:
This uses NRF.resolveAddress() on newer firmwares, to try to resolve "random private resolvable addresses" of peers that connect, before checking the whitelist.

Now we can finally whitelist smartphones that change their address :)


gfwilliams:
Thanks! Just wondering, do we really need the `(resolved)` bit? Do the whitelisted addresses render ok in the menu with that extra length? It looks a bit like we only use the first 17 chars so it might be ok, but if that's the case, is there much point adding  `(resolved)` to them anyway?

PR #2975, opened 2023-08-17
messagelist: setUI weirdness workaround

rigrig:
For some reason `uiRemove()` was called by `Bangle.setUI()`, even though it was unset at the exact time of calling.
Workaround: don't pass `"remove"` in options, but set it inside a timeout.
Also: add some `debug()` lines, enabled by setting `global.DEBUG_MESSAGELIST = true`

I'm not sure why this happens, is it some Espruino quick/bug?

Looks like this has been broken for a while, but with https://github.com/espruino/BangleApps/commit/f4740b149b5649bf0328210de1414ce56c4fb839 it happens more often (before that `setUI` often neglected to set `"remove"`, so this didn't happen but things were otherwise broken)

If I change the [line](https://github.com/espruino/BangleApps/compare/master...rigrig:messagelist-setui-fix?expand=1#diff-1f5021d500659e8f9efc3cf30c0a7355f8950dbf37cbb863cdb39472d8e3c67cR107)
```js
if (remove) setTimeout(() => Bangle.uiRemove = remove);
```
into
```js
if (remove) Bangle.uiRemove = remove;
```

I get this debug log (by setting `global.DEBUG_MESSAGELIST = true`) when a new `"music"` message arrives:

```
messagelist: setUI( undefined undefined
messagelist: Bangle.uiRemove: undefined
messagelist: call Bangle.setUI( {
  "mode": "custom",
  "remove": function () {return uiRemove()}
 } undefined
messagelist: Bangle.setUI() done!
messagelist: uiRemove()
```
(that last line should not happen)

### DIFF
--- a/apps/messagelist/ChangeLog
+++ b/apps/messagelist/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02: Fix music updates while app is already open
 0.03: Fix invalid use of Bangle.setUI
+0.04: setUI weirdness workaround
--- a/apps/messagelist/app.js
+++ b/apps/messagelist/app.js
@@ -59,8 +59,10 @@
       })
     );
   };
+  let noRemove = false;
   const uiRemove = function() {
     debug('uiRemove()');
+    if (noRemove) return;
     if (musicTimeout) clearTimeout(musicTimeout);
     layout = undefined;
     Bangle.removeListener("message", onMessage);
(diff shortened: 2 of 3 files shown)


gfwilliams:
I really don't like this at all - it looks like a hack that's going to come back and bite us later :(

> For some reason uiRemove() was called by Bangle.setUI(), even though it was unset at the exact time of calling.

Is there some way you can come up with a very minimal bit of code that shows this happening?

My guess would be that setUI actually gets called twice in quick succession? In which case the second call will cause the `remove` from the first call to be executed.

rigrig:
> I really don't like this at all - it looks like a hack that's going to come back and bite us later :(

You're absolutely right.
I got a bit excited after finally figuring out where things went wrong, but thinking about it some more: "*something* goes wrong, and this hack works around it" is not something I'm happy with either.

I'll look into reproducing/fixing it properly.

PR #2976, opened 2023-08-18
Add astronomical clock

pavelmachek:
The plan is to have an (analog) astronomical clock with a lot of information on single dial.


gfwilliams:
That sounds great! I've just added `type:"clock"` to the metadata to ensure it shows up as a clock in the app loader list (and also auto-loads as a clock).

When you're happy with the look it'd be great if you could make a screenshot (with `g.dump()`) as I think folks are more willing to install/try and clock when they can see what it looks like

PR #2977, opened 2023-08-18
Add loyalty cards app

glemco:
Simple app to display loyalty cards synced from Catima through GadgetBridge.
The app can display the cards' info (balance, expiration, note, etc.) and tapping on the appropriate field will display the code, if the type is supported.

This is pretty much a proof of concept, many codes may be too complex to be rendered by the bangle's screen or hardware (at least with the current logic).

Simple synchronization support is added in https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3268 .
We send all cards in one go after pressing the button in GB, just like it is implemented on ZeppOS. We may think of some way (like hard limiting the number of cards) to avoid loading too many and crash the device. Right now it's responsibility of the user not to overload the watch. Any synchronization is overwriting the file.

Any help or comment is highly appreciated.


glemco:
Thanks for trying it out!
Basically that's something I misunderstood with the definition of the `touch` and the `tap` events.
A double tap (the one used to go back from the code view) doesn't involve the screen at all, you need to tap on the device, let's say on the frame.
I'd probably replace it with a long touch on the screen (double touch isn't just as trivial), but for now I wanted to see how it looked like.

joserebelo:
Oh wow - it's possible to double-tap the actual device :exploding_head: 

Couldn't we make single-tap on the button go back?

glemco:
Yeah that's another idea, I could try it out :+1:

bobrippling:
Cool idea - I'll let Gordon merge this as it's modified the `android` app, but thought I'd ask, would you say it's ready to go?
I just spotted this in [`android/boot.js`](https://github.com/espruino/BangleApps/blob/12720c5622df9f14e49b2404fb6e33422245cb9e/apps/android/boot.js#L245):
```javascript
console.log("Wrong cards event structure " + event); //TODO remove this
```

glemco:
Thanks @bobrippling for spotting this! Fixed and I'd say it's quite ready to go (of course considering the limitations I mentioned in the README)

bobrippling:
Thanks! One more thing I've just thought - is the C qrcode under any particular license?

glemco:
Both QR and barcode libs are under MIT, I linked them in the README. Should I add the licenses to this repo too?

gfwilliams:
Wrt licenses, please could you add a comment at the top of each of the JS files (just what they're based on and what the license is) - I think that'd be ok for us with it in the README as well.

It looks good to me (although I guess in an ideal world the `android` app would just emit a `GB` event and this app would tag onto that with a `boot.js` file) - but that's not something that is done at the moment

glemco:
@gfwilliams thanks for the answer, I updated the files.

Yeah I was thinking it wasn't a good way to add yet another handler there but just went with the flow..

gfwilliams:
Great - thanks!

PR #2978, opened 2023-08-18
[edgeclk] Option to show weather

peerdavid:
First of all, thanks @tinxx for this great clock - I really really like it! I thought it would be nice to show the weather on the lower right as an additional option. So when showWeather is selected in the settings the clock changes as follows:

screenshot4

Note that I show steps with "K" only if weather is active (or > 100000 as before) to save some space... for all the other settings its exactly the same as before :)


tinxx:
Hey @peerdavid, thanks for the addition! Looks nice on the screenshot, but I currently have no weather provider for my watch, so I can't really test it…

Please change these two little things, if you would be so kind:
- Same default options in app and settings
- Update README with a *Contributing* section or similar (mentioning us both)
- (You might want to pad less in line 163 if weather is displayed but probably not relevant as weather is drawn after steps)

Thanks!

peerdavid:
Thank you so much for your comments!

> * Same default options in app and settings
I just ensured that settings.json and app.json both have the same default settings -- looks good IMO now :)

> * Update README with a _Contributing_ section or similar (mentioning us both)
I added a contributor section

> * (You might want to pad less in line 163 if weather is displayed but probably not relevant as weather is drawn after steps)
Yea exactly so I decided to simply draw it afterwards then its fine. Otherwise I need a check if its weather and if its in"K" mode and set the padding accordingly. So this is IMO the easier option :)

Thanks!

tinxx:
Thanks for the changes and for your contribution in general, @peerdavid!

I am really happy that there is not only one person to use my clock but even a contributor! :heart:

PR #2980, opened 2023-08-18
Messagelist setUI proper fix

rigrig:
`new Layout(...)` [calls `Bangle.setUI()`](https://github.com/espruino/BangleApps/blob/master/modules/Layout.js#L81), causing `uiRemove()` to be called whenever a new messages arrives, instead of only when the app exits.
This fixes that by unsetting `Bangle.uiRemove` before calling `Layout`, and restoring it afterwards.

Also: add some `debug()` lines, enabled by setting `global.DEBUG_MESSAGELIST = true`

(Proper fix instead of https://github.com/espruino/BangleApps/pull/2975)

### DIFF
--- a/apps/messagelist/app.js
+++ b/apps/messagelist/app.js
@@ -24,6 +24,9 @@
     RIGHT = 1, LEFT = -1, // swipe directions
     UP = -1, DOWN = 1;    // updown directions
   const Layout = require("Layout");
+  const debug = function() {
+    if (global.DEBUG_MESSAGELIST) console.log.apply(console, ['messagelist:'].concat(arguments));
+  }
 
   const settings = () => require("messagegui").settings();
   const fontTiny = "6x8"; // fixed size, don't use this for important things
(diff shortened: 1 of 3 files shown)


apps/messagelist/app.js
```diff
@@ -24,6 +24,9 @@
     RIGHT = 1, LEFT = -1, // swipe directions
     UP = -1, DOWN = 1;    // updown directions
   const Layout = require("Layout");
+  const debug = function() {
+    if (global.DEBUG_MESSAGELIST) console.log.apply(console.log, ['messagelist:'].concat(arguments));
```
bobrippling, line 28: `console.log` doesn't use its `this` parameter, so while this doesn't have an effect here, we want to apply the function to the object it lives on:
```suggestion
    if (global.DEBUG_MESSAGELIST) console.log.apply(console, ['messagelist:'].concat(arguments));
```

rigrig: Thanks!

gfwilliams:
Thanks!

PR #2981, opened 2023-08-18
Add Space Weaver -- vector map application.

pavelmachek:
This really needs more work (as documented in README), but this is already quite useful.

Please check app.js file -- I used library for conversion between xyz and lat/lon, so its license applies. It seems to be compatible with bangle apps license.


gfwilliams:
Thanks! What are the chances of having something that can be run from within the app loader to load data?

I know we have the beer compass/others that can grab data from openstreetmap - i'm not sure if that's enough to use as a base for this?

pavelmachek:
Thank you!

I'm not really a web programmer, so something running from app loader would be quite hard for me. I'm considering creating small "demo data" -- most likely world map -- so that app is not empty by default.

gfwilliams:
Thanks - yes, that sounds like a good idea!

PR #2983, opened 2023-08-19
Asteroids - increased ship, asteroids and font size

hughbarney:


bobrippling:
Looks good to me, will merge around this time tomorrow if the other maintainers don't mention anything :)

gfwilliams:
well, it'd be nice to get a response from @hughbarney about that change to how asteroids are drawn - I feel like as there are generally quite a few of them it could make the game quite a lot slower - and the change to transformVertices would be pretty easy

bobrippling:
Ah gotya - the poly drawing replacing the circles for example?

gfwilliams:
Yes - poly draw itself should be ok, but I imagine constructing the polygon each time for every asteroid might slow things down

hughbarney:
Yes it does make the game a bit slower.  It is a lot nicer to look at.

I am afraid I dont know what 'and the change to transformVertices would be pretty easy' would mean in practice.
I'm fairly basic in my understanding of drawing stuff.  I can do a bit of trigonometry but there are a lot of graphics
concepts that are not really in my vocab.    I guess there suggestion is that there might be a more efficient way of calculating the points.

```
    g.drawPoly([
      a.x           , a.y - 1.5 * a.rad,
      a.x + a.rad   , a.y              ,
      a.x + a.rad/2 , a.y              ,
      a.x + a.rad/2 , a.y + a.rad/2    ,
      a.x           , a.y + a.rad      ,
      a.x - a.rad   , a.y              ,
      a.x - a.rad   , a.y - a.rad
    ],true);
```

hughbarney:
Thank you Gordon.  It looks very nice.  Sorry I did not know about `g.transformVertices()`, did not realise you meant using a built in function.

PR #2984, opened 2023-08-19
widbat: Use flash, not fork to indicate charging

pavelmachek:
Green fork is not easily visible, which can be confusing.

### DIFF
--- a/apps/widbat/ChangeLog
+++ b/apps/widbat/ChangeLog
@@ -8,3 +8,4 @@
 0.09: Misc speed/memory tweaks
 0.10: Color changes due to the battery level
 0.11: Change level for medium charge (50% -> 40%), and darken color on light themes as yellow was almost invisible
+0.12: Use black flash instead of green fork to indicate charging
--- a/apps/widbat/metadata.json
+++ b/apps/widbat/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "widbat",
   "name": "Battery Level Widget",
-  "version": "0.11",
+  "version": "0.12",
   "description": "Show the current battery level and charging status in the top right of the clock",
   "icon": "widget.png",
   "type": "widget",
--- a/apps/widbat/widget.js
+++ b/apps/widbat/widget.js
@@ -1,10 +1,10 @@
-(function(){
+{
   Bangle.on('charging',function(charging) {
     if(charging) Bangle.buzz();
-    Bangle.drawWidgets(); // re-layout widgets
+    WIDGETS["bat"].draw();
     g.flip();
   });
-  var batteryInterval = Bangle.isLCDOn() ? setInterval(()=>WIDGETS["bat"].draw(), 60000) : undefined;
+  let batteryInterval = Bangle.isLCDOn() ? setInterval(()=>WIDGETS["bat"].draw(), 60000) : undefined;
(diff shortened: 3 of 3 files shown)


gfwilliams:
Am I right in thinking that the widget still changes length, but that extra length now isn't used for anything?

If so I think we'd need to change that - and also just using an image rather than rendering the polygon would be a lot faster

KTibow:
It's a plug (US-style) not a fork lol

pavelmachek:
Yeah, I guess it is a plug. Problem is it is green, so it is hard to see.

I fixed code to no longer change the size, thanks for noticing.

Not sure about changing it to image; I believe this is fast enough and smaller code.

pavelmachek:
I did some benchmarking, and filled polygons seem to be pretty fast (only 2x slower than lines, for example), so this should be good solution from performance standpoint, too.

gfwilliams:
@thyttan, @halemmerich, @bobrippling  do you have any strong feelings about this? This is what it looks like:

image

I quite like the idea of not changing the widget width when on charge, but personally I'm not a fan of the look of the lightning strike, plus it does make it hard to see how full the battery is. I guess it might be possible to make a 'thinner' bitmap of the plug (or find a more obvious lighting spark image) and use that.

Or honestly, if you don't like the green charge plug image, we could just make it black?

The other option is to just have a new widget with these changes in.

thyttan:
>do you have any strong feelings about this?

Can't say I do :P

However:

PRO keeping as is:
- It feels like the plug is somewhat of a standard on many apps/widgets/clock faces that I use.
- If the old style is faster that's a plus in my book. Since widgets load often it's nice if they draw as quickly as possible to make everything feel snappy (but I don't know how many milliseconds we're talking here..)

PRO changing
- Not changing the widget size is a better situation I think.
- My android phone uses a flash inside the battery to indicate charging. So that would harmonize well there.

Personally I use [widminbat](https://espruino.github.io/BangleApps/?q=widminbat)

bobrippling:
Yeah, I'll add my vote for keeping this (i.e. merging this change).

Main reasoning:
- **Widget width**: +1 for keeping the widget width the same throughout, I like that most widgets take up the same space which makes distinguishing each one a breeze.
- **Flash vs plug**: The flash looks good and seems to be a common theme, and perhaps fits in better on the battery image, than a plug on the side. I agree it does make it harder to see the charge level - perhaps one where the flash is vertical and maybe yellow? Like these:
image
image
- **Easy to read**: I use the [hwid widget](https://espruino.github.io/BangleApps/?id=hwid_a_battery_widget) myself but for similar reasons - easy to read at a glance, but having a hint about whether I've landed the charging cable correctly, like this, is nice.

pavelmachek:
Yellow flash would be hard to see, but feel free to change flash shape/size.

State of charge being harder to tell may not be so bad -- percentage is not really reliable during charge, anyway.

Creating yet another battery widget .. well is doable but we should do right thing by default :-).

bobrippling:
Yeah I agree, and that's a good point about the percentage being a little high during charging - perhaps all good as-is

gfwilliams:
Ok, just changed the flash a bit to make it a bit more flash-like, and added a few other optimisations

pavelmachek:
Thank you, looks okay to me in a brief test.

PR #2986, opened 2023-08-22
Fix health daily movement graph again :-)

ticalc-travis:
This just fixes a typo introduced by commit 31fb8f8d02 which causes the movement daily summary graph to freeze on the “Loading…” screen and print an error to the console due to trying to access the field name “hr” when it should be “day”.

While I was at it, I noticed what I thought might have been a potential out-of-bounds subscripting error in the *PerDay() functions. The array size is 31 (indeces 0…30) but the indeces actually returned by readDailySummaries are 1…31. However, it appears that the graph never displays a partial day's data before it rolls over, so maybe this would never be a problem because presumably the 31st day of the month would not display? It's technically inconsistent with the fact that the hourly funcs dimension the array to 24 rather than 23, though, but since I'm not sure, I decided to leave that alone for now.

### DIFF
--- a/apps/health/ChangeLog
+++ b/apps/health/ChangeLog
@@ -27,3 +27,4 @@
       movement graph in app is now an average, not sum
       fix 11pm slot for daily HRM
 0.26: Implement API for activity fetching
+0.27: Fix typo in daily summary graph code causing graph not to load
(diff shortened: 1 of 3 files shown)


gfwilliams:
Argh, Thanks for getting this fix in!

If it's really returning 1..31 then yes, I'd imagine we should be setting array size to 32 (and effectively ignoring index 0) - I can't think that it would break anything to do that...

ticalc-travis:
That seems to be a good idea, and yes, the graph code for per-day data does seem to expect 1…31 and ignore 0. I initially tried adjusting the code to write to indeces 0…30 instead, but that caused the day labels on the graph to be 1 less than they should, and entry 0 was drawn but couldn't be selected. Array size 32 looks like the safest option.

PR #2992, opened 2023-08-30
added a new app

nikomin:
Nightwatch watches sensor data, records min/max value and shows a graph of historic values. Currently for temp and pressure.


nikomin:
Thanks

The `NaN`s are the top and bottom of the axes. The screen shot is from the emulator and since it doesn't have the sensors it shows that. I probably should fill it with default values or get a screenshot from the watch since on the device this  doesn't happen.

nikomin:
> Thanks
> 
> The `NaN`s are the top and bottom of the axes. The screen shot is from the emulator and since it doesn't have the sensors it shows that. I probably should fill it with default values or get a screenshot from the watch since on the device this doesn't happen.

I'm on it

nikomin:
So I did upload the changed screenshots. Do I need to make another pull request for the latest revision?

gfwilliams:
No, that looks great, thanks! I'm afraid I don't get notified when a PR changes, only when someone comments - otherwise I'd have merged this right away

PR #2994, opened 2023-09-01
gipy: new release

wagnerf42:
new release for gipy. the kids asked me an elevation profile to know in advance when the climb will finally end.


gfwilliams:
Nice - thanks!

PR #2995, opened 2023-09-03
Add rebbleagenda, pebble-like animated agenda.

SarahAlroe:
Hi there, 
I've been working on an agenda viewer inspired by the pebble time interface.
This first version allows for swiping through calendar events with markers separating days.

New here, so do say if I've missed something :)


gfwilliams:
This looks great - thanks!

thyttan:
Yes, I think this is super cool as well!

I had some weirdness with hidden widgets that when hiding would draw background colour over the side bar.

Also, the App Loader should probably ask what to do if another agenda is already installed (Replace/Cancel/Keep both), like it does for e.g. keyboards. Especially since the ´shortName´ of ´rebbleagenda´ is the same as ´agenda´s ´name´ so the settings pages interfere.

Again, cool!

gfwilliams:
> the App Loader should probably ask what to do if another agenda is already installed

Yes - I think that's something we would need to figure out rather than @SarahAlroe, but it should be possible (even if a bit of a hack) by adding `"provides_widgets" : ["agenda"]`  to both of them

SarahAlroe:
> I had some weirdness with hidden widgets that when hiding would draw background colour over the side bar.  

I do get some weird widget stuff at the last frame of closing the app (which I believe is unavoidable), but otherwise not.  
@thyttan Do you have an application list or idea of possible co-culprit so I can replicate?  
I have a suspicion it's because I load widgets pretty late, but would be nice to verify :)  

> it should be possible (even if a bit of a hack) by adding `"provides_widgets" : ["agenda"]` to both of them

I think that could become pretty confusing for people who are not aware of the issue, especially if someone starts making agenda widgets, whatever that would be. Adding some sort of "mutually_exclusive_group" property might be less confusing, but would complicate the metadata further so maybe not a great idea either.
For a very easy short term fix, we could just change the shortname to something else?

thyttan:
>Do you have an application list or idea of possible co-culprit so I can replicate?
I have a suspicion it's because I load widgets pretty late, but would be nice to verify :)

Yes, the main problem I just realized is I have [Fastload Utils](https://espruino.github.io/BangleApps/?q=fastload) installed (which I should have thought of immediately - gotta be better at remembering that >.< ). So with that uninstalled the problem goes away. I'd imagine if you want to try installing Fastload Utils you'll see the weirdness, given you have some widgets that usually sit to the right in the widget field. But don't feel you have to accommodate it, Fastload Utils is explicitly experimental. I can post my installed apps if you still want.

SarahAlroe:
> I'd imagine if you want to try installing Fastload Utils you'll see the weirdness

Yup, I see it, thanks :)  
Changing the drawing order seems to fix it.
I will include the change next time I make a pull request.

Issue #2985, opened 2023-08-22
[sched] Alarms not firing and uneditable.


JAVAHTML1:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

Schedular App not firing alarms. Alarms, timers, and events uneditable in the App loader.

Expected to feel alarms going off. Expecting to be able to edit the alarms in the App loader.

Ways to reproduce:
Unknown.

### Installed apps

widbt_notify (0.17), calculator (0.07), waypoints (0.01), info (0.03), hrm (0.11), blackjack (0.03), ptlaunch (0.15), files (0.08), powersave (0.03), magnav (0.06), altimeter (0.03), henkinen (0.01), launch (0.20), gpssetup (0.02), gpsautotime (0.04), clicompleteclk (0.03), widdevst (0.03), gpsmagcourse (0.01), bledetect (0.04), widminbat (0.03), drained (0.03), lightswitch (0.07), rpnsci (0.03), shownearby (0.02), messageicons (0.05), messages (0.60), notify (0.13), mylocation (0.10), compass (0.09), boxclk (0.05), thermom (0.07), messagegui (0.73), folderlaunch (0.03), gpsinfo (0.12), promenu (0.02), runplus (0.21), astrocalc (0.05), rescalc (0.04), gipy (0.20), multitimer (0.07), fastload (0.06), recorder (0.30), clock_info (0.07), kbtouch (0.04), batchart (0.12), speedalt (0.13), qrcode (0.06), health (0.26), fileman (0.03), boot (0.59), poweroff (0.03), setting (0.63), sched (0.22), widalarm (0.01), alarm (0.42), openhaystack (0.01)

nxdefiant:
What do you mean by "uneditable"? Can you edit the alarms on the watch?

gfwilliams:
What do you mean by `uneditable in the App loader`? You mean you don't see an option to edit them from the app loader but there isn't one and you have to edit from the watch itself?

> Schedular App not firing alarms

If I create an alarm and go back to the clock, it works fine me here. Or is the issue that you're creating an alarm but then not exiting the scheduler app, and the alarm isn't firing?

JAVAHTML1:
By `uneditable in the App loader` I mean if I try to edit my alarms using the Scheduler app on the App loader, it doesn't load the editing menu. I can try to edit the alarms in the alarms app on the watch, but they never go off.

nxdefiant:
[...]

PR #2996, opened 2023-09-04
sched: Fix readStorage when files do not exist

nxdefiant:
see #2985


bobrippling:
I'd noticed a similar issue with [`rep`](https://github.com/bobrippling/BangleApps/blob/a84e0654617709d9f8efdace29a4464365e03158/apps/rep/interface.html#L104-L104) and forgot to fix it - what do you think of promoting this functionaliy to `Util` for other apps?

apps/sched/interface.html
```diff
@@ -263,12 +263,25 @@
   alarms.push(timer);
 }
 
+function readStorageIfExists(filename, cb) {
+  Puck.write(`\x10(function() {
+    Bluetooth.print(JSON.stringify(require("Storage").list("${filename}").sort()));
```
bobrippling, line 268: If the filename is a substring we'll get a match, e.g. `require("Storage").list("rep.j")` will match `rep.json`, which might not be what we want - perhaps we could do a `require("Storage").read("rep.j") != undefined` ?

Or if we're promoting this into `Util`, we could modify [reading of files](https://github.com/espruino/EspruinoAppLoaderCore/blob/86819223f43a892622a8d5f1f430a2a1fc57031d/js/comms.js#L478-L502) to permit returning `undefined` if the file doesn't exist? (maybe I'm over thinking here though)

nxdefiant: Should be fixed and yes, having that in util would be
prefered.

bobrippling: Nice - I could sort out a follow up to pop it into util and sort out the `rep` app

bobrippling:
Looks good to me - @thyttan, @halemmerich any comments before I merge this? (mainly asking it's a "core" app, and the first non-Gordon merge)

thyttan:
> Looks good to me - @thyttan, @halemmerich any comments before I merge this? (mainly asking it's a "core" app, and the first non-Gordon merge)

- Without this fix I can't access `interface.html` from App Loader ever (checked to see that I had `sched.json` on my watch which was there).
- With the fix I can access it on Chromium browser on my laptop and in Bromite browser on my phone and set alarms. :+1: :)
- But on the App Loader inside Gadgetbridge I can't use the interface before or after the fix. It gets stuck on "Plaese wait Loading...". If I am connected to the Web IDE I get (don't know if it's relevant):

```
>["sched.json"]692
W3sidCI6MTczNDAwMDAsImRlbCI6ZmFsc2UsIm9uIjp0cnVlLCJycCI6dHJ1ZSwiYXMiOnRydWUsImRvdyI6MTI3LCJsYXN0IjowLCJ2aWJyYXRlIjoiOjoiLCJkYXRlIjoiMjAyMy0wOC0wNCIsIm1zZyI6IkJva2Ega2Fub3Rlcm5hIiwiZGF0YSI6eyJobSI6ZmFsc2V9fSx7InQiOjI1MTQwMDAwLCJkZWwiOmZhbHNlLCJvbiI6ZmFsc2UsInJwIjpmYWxzZSwiYXMiOmZhbHNlLCJkb3ciOjEyNywidmlicmF0ZSI6Ijo6IiwibXNnIjoiU2FuZ2tsYWRlciIsImRhdGEiOnsiaG0iOmZhbHNlfSwibGFzdCI6MjF9LHsidCI6MzkzMDAwMDAsImRlbCI6ZmFsc2UsIm9uIjpmYWxzZSwicnAiOmZhbHNlLCJhcyI6ZmFsc2UsImRvdyI6MTI3LCJsYXN0IjoyMCwidmlicmF0ZSI6Ijo6IiwiZGF0YSI6eyJobSI6ZmFsc2V9LCJtc2ciOiIifSx7InQiOjQ3OTQwMDAwLCJkZWwiOmZhbHNlLCJvbiI6ZmFsc2UsInJwIjpmYWxzZSwiYXMiOmZhbHNlLCJkb3ciOjEyNywibGFzdCI6MjgsInZpYnJhdGUiOiI6OiIsIm1zZyI6IiJ9XQ==?
```
- It's debatable, but I would probably bump the version and ChangeLog so users that tried but couldn't access the feature before get notified that there was a fix done.
- @nxdefiant, @halemmerich and you are probably better at judging the technicalities.

nxdefiant:
> * I suggest bumping the version and ChangeLog so users that tried to but couldn't access the feature before get notified that there was a fix done.

I don't see the point in increasing the version when the app itself is unchanged.  Also have never done that, e.g. #2552 #2979

thyttan:
>I can't say anything about Gadgetbridge

It seems to me like the problem inside Gadgetbridge affects all similar interfaces. So lets not stop this PR from being merged because of it. I'll open a separate issue for that.

>I don't see the point in increasing the version when the app itself is unchanged. Also have never done that, e.g. https://github.com/espruino/BangleApps/pull/2552 https://github.com/espruino/BangleApps/pull/2979

Fair enough :) I still think it's a good idea for the reason I wrote. But OK with me to merge also without. :+1:

gfwilliams:
So the issue is that `Util.readStorage` actually fails to return if the file doesn't exist?

As @bobrippling  says I think we should make sure `Util.readStorage` works even if the file doesn't exist - basically making it behave in the way it does in this PR, but for everyone (just return `undefined`). 

It should be easy enough to modify  without adding `list()` since `.read(filename)` just returns undefined if the file doesn't exist. It's a bit odd though as it looks like it should work that way already: https://github.com/espruino/EspruinoAppLoaderCore/blob/431a3fb743da5c370729ab748cb2c177e70a345b/js/comms.js#L479

You can test on the console of the web browser with `Comms.readFile("foo", x => console.log("f",x))` - I think it's probably `readTextBlock` that is broken?

bobrippling:
Yes, I believe [commit b8813ab](https://github.com/espruino/EspruinoAppLoaderCore/commit/b8813ab92ceb70fb8ec6a7de6baaec88f6b5026f) solved the missing-file problem, and means we see a callback fired and receive an empty string for the file. When I run:

```javascript
Util.readStorage("hi.txt", x => console.log("cb", JSON.stringify(x)))
```

I get a callback with this log:
```
cb "\u0000\u0000\u0000"
```

And `JSON.parse("\x00\x00\x00")` throws, which is probably what's stopping [interface.html from parsing](https://github.com/espruino/BangleApps/blob/2708cabc5e1ae3521e7a50c557db63d57dfa612a/apps/sched/interface.html#L269).

---

The three bytes come through because in [`atobSafe` we have this do-while](https://github.com/espruino/EspruinoAppLoaderCore/blob/431a3fb743da5c370729ab748cb2c177e70a345b/js/utils.js#L354-L361), which I think would be safe to change to a `while(i < input.length){...}` ?

With the fix from [b8813ab](https://github.com/espruino/EspruinoAppLoaderCore/commit/b8813ab92ceb70fb8ec6a7de6baaec88f6b5026f) and the `atobSafe` fix, I believe that'll let us receive an empty string when the file doesn't exist, and `interface.html`s will then be ok with missing files.

gfwilliams:
That sounds great to me! I've just added that change - also the code was in the IDE too so I've now changed that as well

bobrippling:
Ah great, thank you!

PR #2997, opened 2023-09-06
Add repeatCalls option

reyemxela:
This adds a second Repeat option for phone calls. I like having a single buzz for messages (sms/emails/etc.), but a continuing repeated buzzing for phone calls.

This seemed like a nice easy/small tweak, but let me know if there's anything that needs to change!


gfwilliams:
Thanks! This looks like a good addition - although I do have a concern that if this is installed as an upgrade to an older version, it might break for calls as `repeatCalls` won't be defined in the settings file until a setting is actually changed in the app?

I'll just put a note in the code

gfwilliams:
Nope - sorry, just checked - it's all fine since the code checks for `undefined` and sets it to 4 already :)

Thanks! Merging now :)

Issue #1040, opened 2021-12-09
after 2v14 release: Add settings menu item for left-handed wearers


ruswerner:
I've searched the wiki, the docs, GitHub and all the settings on the Bangle 2 and can't figure out how to flip the screen so I can wear the watch on my right arm (for left-handed users). The button should be on the left when looking at the watch.

Is this something that was considered? If not, can you please point me in the direction of where I would have to modify the firmware to accomplish this?

Cheers

gfwilliams:
There's nothing in there right now, but you can implement this basically by going to the `custom boot code` app and entering:

```
(function() {
  g.setRotation(2);
})();
```

However, quite a few other things (touchscreen, launcher, etc) don't work too well, so I think this would require some actual firmware changes to do nicely - inverting touch coordinates/directions and also fixing `g.setClipRect` when rotated

thyttan:
Too me this could make sense for us wearing the watch on left wrist as well. As it is now you can accidentally press the button when you angle your hand towards the watch. Having the button on the other side of the screen would eliminate that risk.

On a traditional watch it makes sense to have the wheel on right side for right-handed people since it's easier to turn it that way. But with a button there isn't really any benefits to it as i see it, other than familiarity. 

Another note, if implemented, would this mess up accelerometer readings?

gfwilliams:
Just saw another request for this at http://forum.espruino.com/conversations/375520/#comment16505901 so there's definitely interest.

I just made a change to the main Espruino firmware at https://github.com/espruino/Espruino/commit/f772bd51062b84626c595c395538182409bb77ff so that now the touchscreen obeys `g.setRotation(2);` - so actually if you stick that in, the watch is pretty much usable left-handed now.

However there's still an issue with scrollable lists - I'll look into that. But once done we can make it an option in `settings`

> Another note: If implemented, would this mess with accelerometer readings?

Right now they are left as-is. I guess it depends what you intended to use the readings for. Step counting/etc would work fine.

gfwilliams:
Ok, text rendering issue is now fine - so if you do as mentioned in this comment: https://github.com/espruino/BangleApps/issues/1040#issuecomment-989662278

The Bangle should now work pretty well left-handed.

thyttan:
Just testing what you suggest in https://github.com/espruino/BangleApps/issues/1040#issuecomment-989662278 on FW 2v15.31.

Touch screen-touches seems to work pretty good in apps, but behaves weirdly on the widget field.

[...]

PR #2999, opened 2023-09-07
Fix spirit level when screen rotated

matijatosic:
Take default rotation/reflection into account, as defined via the Settings app.

### DIFF
--- a/apps/slevel/ChangeLog
+++ b/apps/slevel/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Updated to work with both Bangle.js 1 and 2.
 0.03: Now also visible on Bangle.js 2
 0.04: Now work with different themes
+0.05: Take default rotation/reflection into account
--- a/apps/slevel/metadata.json
+++ b/apps/slevel/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "slevel",
   "name": "Spirit Level",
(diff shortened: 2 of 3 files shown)


thyttan:
Looking good!

I just have one hold up. Maybe the firmware should take care of rotating the accelerometer readings before passing it on instead, @gfwilliams? Adjacent previous discussion:

>> Another note: If implemented, would this mess with accelerometer readings?
>
>Right now they are left as-is. I guess it depends what you intended to use the readings for. Step counting/etc would work fine.

_Originally posted by @gfwilliams in https://github.com/espruino/BangleApps/issues/1040#issuecomment-1116034509_

Not to do with this PR: While testing I found some weirdness that's showing both before and after these changes. If the angle is very step the bubble will jump to the opposite side of the screen. But at that angle you're not wondering if the surface is level anymore so...

gfwilliams:
Yes, I think you're right - we should probably do that in firmware. The hit itself should be minimal, and we do it for touchscreen and I think even accelerometer 'tap' events already.

It's hard to think of any apps that wouldn't benefit from it being flipped - the only one I can think of that uses accelerometer right now is ball maze, but I guess that would be broken at the moment if done on a rotated screen

gfwilliams:
Just done :)

PR #3002, opened 2023-09-08
Prepare version 0.02 of Space Weaver

pavelmachek:
. Major changes are move towards binary format, which saves a lot of space, and including world map, so that there's something to play with without extensive setup.


gfwilliams:
Nice - thanks!

PR #3004, opened 2023-09-09
Add widhr (Last announced heartrate BPM Widget)

joyrider3774:
Hello,

I created a small widget that will show the last announced BPM using `Bangle.on("HRM", ...);` it does not explictly enable the heart sensor it only waits for such messages. For example i had setup the watch to take every 10 minute a heartrate reading and this will display to me in the widget the last reading it had gotten.

The widget also saves that last bpm value to a file in storage so that when the bangle reboots or after application quits it keeps displaying the previous value.

It's a very simple widget and initially based the code on lato peedometer but i changed a few things and did not use lato font eventually.

PS the branch is named wrong it is named `add_widhrm` but the folder i added is actually `widhr` because `widhrm` already existed

regards,
willems davy


gfwilliams:
Thanks! Looks good!

PR #3006, opened 2023-09-11
Trying to add date formatting options

dormio:
This is my first ever pull request. Like, on anything, ever. My coding experience is quite limited, so what follows may sound a little sad and discombobulated, but read on...

I am American and can't get used to DDMMYYYY. I was just trying to make the date on my watch into ISO format; you have it as a picture on the App Loader, but I can't actually figure out how to do it on my watch. I wrote something and tried it on the espruino emulator, but I can't figure out to how to make app.js talk to settings.js (on the emulator).

I have _attempted_ to add two new settings: "Date Format" and "Date Separator." Line 56 should probably start its own "sub-function" (I've never used js before; is this a thing with a name?) to return the correctly formatted date string, but I just left it how it appears.

Like I said, I didn't actually test this, and I cannot for the life of my figure out how to even do that. So I hope this works. And I hope someone actually reads this.

### DIFF
--- a/apps/vectorclock/app.js
+++ b/apps/vectorclock/app.js
@@ -5,6 +5,8 @@ var dowcol = settings.dowcol || g.theme.fg;
 var timecol = settings.timecol || g.theme.fg;
 var datecol = settings.datecol || g.theme.fg;
 var chimetype = settings.chimetype;
+var dateFormat = settings.dateFormat;
+var dateSeparator = settings.dateSeparator;
 
 function padNum(n, l) {
   return ("0".repeat(l)+n).substr(-l);
@@ -52,6 +54,16 @@ function draw() {
   let secondsText = padNum(d.getSeconds(), 2);
   let dowText = locale.dow(d);
   let dateText = locale.date(d, true);
+  let day = dateText.substr(0,2);
+  let month = dateText.substr(3,2);
+  let year = dateText.substr(6,4);
+  if (dateFormat=="DDMMYYYY") {
+    dateText = day + dateSeparator + month + dateSeparator + year;
+  } else if (dateFormat=="MMDDYYYY") {
+    dateText = month + dateSeparator + day + dateSeparator + year;
+  } else if (dateFormat=="YYYYMMDD") {
+    dateText = year + dateSeparator + month + dateSeparator + day;
+  }
   let width = g.getWidth() - 2;
 
   g.setFont("Vector", 256);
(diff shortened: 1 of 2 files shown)


pavelmachek:
To test: push it to your own branch on github, and tell github to prepare web pages from that. Then open that webpage with app loader, connect with your watch and update.

gfwilliams:
Hi - I think what you're missing is just installing the 'Languages' app at https://banglejs.com/apps/?id=locale with the USA locale.

You can see the code does `let dateText = locale.date(d, true);` normally - the locale module gives you the date in the correct format depending on what you'd set in the Languages app - see https://www.espruino.com/Bangle.js+Locale for a bit more information on how it works. The idea is we do it in one place, then every app can take advantage.

So I think it's best if I just close this PR - but I'll make sure there's something in the FAQ bit of the Wiki about this in case anyone else hits it

dormio:
@gfwilliams This makes sense, but I _still_ would like a YYYY-MM-DD format option; adding the US locale only gives me MM/DD/YY. Is there another workaround I'm missing? (I just got my watch on Wednesday, so my knowledge is still pretty shallow.)

gfwilliams:
Ahh, ok. I've just come up with a new `en_US 2` locale that does YYYY-MM-DD

... and now it shows you some examples of what it looks like too. I guess perhaps the next stage might be allowing a 'custom' locale in that app, but if you really need something specific you can just write your own file called `locale` to the Bangle

dormio:
Sorry to bother everyone again, and sorry if this is posted in an inappropriate place.

I was able to get my GitHub Pages up and running to upload my own version of Vector Clock to my watch, but I can't get the new settings I added to write. I am able to change them on my "Vector Clock settings" on my watch, but they don't stick if I leave the settings page and go back in. Does anyone have any advice or wisdom? (And where do I post stuff like this?)
[Link to my settings](https://github.com/dormio/BangleApps/blob/master/apps/vectorclock/settings.js)

thyttan:
Hi! Did you spot the available tutorials? 🙂

All espruino tutorials: https://www.espruino.com/Tutorials

General pointers developing for Bangle:
https://www.espruino.com/Bangle.js+Development

Getting started:
https://www.espruino.com/Bangle.js+Getting+Started

Adding settings: https://www.espruino.com/Bangle.js+App+Settings

bobrippling:
Yes I think the forum is generally where this kind of help/chat happens, but as a brief starter:

> I am able to change them on my "Vector Clock settings" on my watch

If you then use [the IDE](https://www.espruino.com/ide/) to read back the settings file, does it indicate that they're as you set them? (`require("Storage").readJSON("vectorclock.json", true)`)

> they don't stick if I leave the settings page and go back in

If the above is true, you could log out [this `settings` variable](https://github.com/dormio/BangleApps/blob/6c245a494acfaceeae87efffc2f7a0aca9fa00b5/apps/vectorclock/app.js#L3) in the IDE, after your app's loaded (type `settings` and hit enter)

PR #3008, opened 2023-09-11
ContourClock 0.31

xxDUxx:


bobrippling:
Thanks for the improvement - out of interest, was it the extras drawing that caused the freeze? How was it locking things up?

apps/contourclock/app.js
```diff
@@ -29,14 +29,17 @@
   }
   let showExtras = function() { //show extras for a limited time
     drawExtras();
-    if (extrasTimeout) clearTimeout(extrasTimeout);
+    //if (extrasTimeout) clearTimeout(extrasTimeout);
```
gfwilliams, line 32: What was the reasoning for doing this? It feels like this should really be in there...

gfwilliams:
Looking at the code change, I think it's likely there was an error from the code at some point, and it stopped execution before drawTimeout could be set again.

So these changes set drawTimeout right at the start, but they may not actually fix the occasional error that happens...

gfwilliams:
Thanks!

PR #3012, opened 2023-09-12
LCD Clock Plus

VitaliiTsilnyk:
A new clock based on the LCD Clock app.

Differences from the original LCD Clock:
* Two extra (4 total) ClockInfo areas at the bottom of the screen. ClockInfo areas are now smaller and only include the icon and the data.
* AM/PM labels moved to the right of the time and are only visible when 12-hour mode is enabled in the system locale settings.
* Month name label added.
* Background color changed to better mimic old-style LCD screens.
* Various small alignment ajustments.


gfwilliams:
Nice - thanks!

PR #3013, opened 2023-09-12
Casio Logo Widget

VitaliiTsilnyk:
A simple widget that displays Casio logo when clock screen is active.

Looks nice in combination with the LCD Clock / LCD Clock Plus.


bobrippling:
Nice! Out of interest, how does this look with the light theme?

VitaliiTsilnyk:
Good question. I only tested it with LCD Clock (which overrides the theme with its own) before submitting the PR. Just tested it with Anton Clock and it basically disappears since it's hardcoded to be white.

gfwilliams:
Thanks! As @bobrippling says, it wouldn't work - but it's the image itself is 1 bit so Espruino will automatically choose the right theme - all that's needed is to remove setColor (which I have just done)

VitaliiTsilnyk:
Just custom themes, with red foreground for example, etc. Making the logo the same color as the other text looks off.
image image

Looks much better when it's white (more like silver in real life) or black.
image image
Looks like an actual logo that's not part of the watch screen.

gfwilliams:
Ok, I just put a change in. I've just not seen people using themes like that normally...

PR #3016, opened 2023-09-13
Layout: Allow consistent rendering of buttons with custom BG color

ticalc-travis:
This attempts to solve a quirk with @nravanelli's useful contribution of allowing custom button colors. When using `bgCol`, the button background does change color, but this background color bleeds outside the button border into the corners, which does not happen with buttons that do not specify custom colors. The underlying problem seems to be that we really have two different concepts of “background color” here, one for the button itself, and the other for the rounded border around the button. Using `bgCol` confuses the rendering code into rendering both the button face and the space around it in the same color, when we often actually want two different colors.

I decided that the best solution was to revert `bgCol` to doing what it did before nravanelli's commits, which is to only affect the space outside the border, and instead add an additional field `btnFace` for specifying the button face BG color. This way both colors can be specified independently if needed. (The behavior of `col` and `btnBorder` seem fine, so I've left them alone and clarified their meaning WRT buttons in `Layout.md`.) I also think this would be potentially better for app compatibility, as it avoids changing the original behavior of `bgCol` for buttons. Since nravenelli's commits did not update the minified file `Layout.min.js`, it doesn't seem likely that anyone has released much code that depends on the behavior change they were making for `bgCol` (am I mistaken?).

Question: Should the `btnBorder` and `btnFace` fields be renamed `btnBorderCol` and `btnFaceCol` for consistency with the naming of `col` and `bgCol`? This might be a good time to decide for sure before people start writing code that uses them. :)


ticalc-travis:
`selected` is for the Bangle.js 1, right? That's a good call. It seems I missed a spot in the code and had broken that, so I've added a commit that appears to fix it in the emulator (which is my only way of testing Bangle-1-related stuff). Is it used anywhere else? (I was thinking it would be nice if the Bangle 2 could highlight the button in the selected color as it's tapped for visual feedback like the buttons in `E.showPrompt` and friends seem to do, but it doesn't look like that's implemented at the moment.)

I'm a bit skeptical of trying to do things like automatically combining theme colors with app-provided colors since the theme colors could be configured as anything by the user and it might not be easy to predict how certain combinations of colors might end up clashing when rendered. It seems like going ahead and providing direct control for each color involved would be more flexible and easier for the app author to ensure legible results. (While I would argue that custom colors should usually be used sparingly for usability reasons, one example of something I was considering doing with them in the app I'm working on is to sometimes set button background and border colors that are blended with the theme background to produce a “dimmed” appearance to represent currently unavailable actions.)

I'll also go ahead and rename the other color options to end in “Col”.

ticalc-travis:
> Yes, that would be really neat - potentially that could be done, yes. You'd just have to add a 'drag' handler and work out whether the button was pressed based on that (rather than using the touch handler)

Technically I think I know how to implement this, but there's a compatibility issue I'm not sure what to do about. Ideally I'd replace the touch handler with a drag handler as you said, but I noticed the Layout code sets a reference to its touch handler at `Bangle.touchHandler`, and from a quick search it does seem that a few apps rely on that to be able to unregister the handler and replace it with their own custom one. And in doing so they look for it as a `touch` event and not `drag`. So just changing the handler event would likely break something in those. It might have to be some sort of opt-in feature via an option, which means most apps that could benefit from it wouldn't automatically know to use it.

gfwilliams:
Do you have an example of those apps?

The use of touchHandler comes because of setUI: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_setUI_Q3.js#L13-L28

The idea is when `setUI` is used elsewhere (eg for showMenu) the touch handlers can get removed.

If I'm honest I'm not sure why touchHandler couldn't have just been added into the `mode:custom` setUI call like is shown on https://www.espruino.com/Reference#l_Bangle_setUI (because setUI is called there anyway)

ticalc-travis:
Ah, that's helpful to know. I didn't find `Bangle.touchHandler`, etc., in a quick doc search, so I wasn't sure if they had any use outside Layout itself.

A closer look at my grep for `Bangle.touchHandler` doesn't show as many results as I initially thought. Basically, I get:

* kbmorse: Seems to set `Bangle.touchHandler` for its UI, though it's not clear why. Doesn't appear to use buttons in Layout.js, so it doesn't look like a problem.
* menuwheel: Seems to explicitly want to extend setUI's touchHandler for whatever reason (does some check, then calls the original handler if it doesn't see what it's looking for). But does not appear to use Layout.js at all, so no chance of conflict there.
* messagelist: Has a comment explicitly indicating that it overrides Layout's button handler to do custom button processing, and it does so by replacing `Bangle.touchHandler` itself

So if we used a drag listener rather than touch, messagelist almost certainly would need to be adapted or else both handlers would end up active at the same time and likely conflict. But it seems like that could be the only one?

PR #3017, opened 2023-09-14
widhrt - improved image for widget to a full heart, color changes to red if HRM is on

hughbarney:
thank you Gordon for your support.


gfwilliams:
Thanks! just FYI you can turn on 'crop' in the image converter which will strip out excess passing around the edge of the image. You'd have to reposition it but the image base64 will be a bit smaller.

Not a big deal but worth bearing in mind for other stuff in the future

Issue #2993, opened 2023-09-01
⌚️🚀 Maintainers wanted! 


gfwilliams:
I've had to take a bit of time off the last few weeks with my kids and school holidays, and I haven't been as quick merging PRs as I'd like to be.

Would anyone like to get involved in checking and merging PRs? With the testing action on commits you can see pretty clearly if it's ok or not, so adding new apps is usually pretty straightforward. For changes to existing apps (especially default ones) it might sometimes need a bit of poking around and thought - not just whether the code looks ok, but thinking about whether we really want to change all the default alarms to display party parrot or whether it should be an option :)

@rigrig, @peerdavid, @nxdefiant, @bobrippling , @thyttan, @halemmerich off the top of my head I feel like you've all been contributing loads of great stuff recently, and fixing other people's code too - would you like to get involved?

And sorry if I missed anyone else from that list - please feel free to reply if you're interested!

bobrippling:
Hey, yeah I'd be interested in helping out!

gfwilliams:
That would be great! Thank you! I'll see about adding you as a collaborator

nxdefiant:
What I would like to see:
- PRs must be accepted my multiple (2? 3?) maintainers to get merged (I think github can do that)
- Maintainers can not merge/accept their own PR

thyttan:
I'd be happy to jump on board as well.

I think I align with @halemmerich's thinking. Especially with having a checklist to go through, that would help me a bunch :+1:

gfwilliams:
Thanks! I've just added @thyttan, @halemmerich and @bobrippling - I imagine there's more than enough of you for now - we may have to think of a way of stopping you from accidentally looking at the same PRs at the same time :)

@fanoush has contributed loads of quality stuff, but as you say I think it probably makes more sense for them to work on the main Espruino repository than the BangleApps one

I'm with @halemmerich on this - this is quite a small project in the scheme of things, and if I make someone admin it's because I trust their judgement. For most PRs I don't expect we'll need two people to sign off on it.

I added some initial thoughts (copy pasted from @halemmerich and added to) at https://github.com/espruino/BangleApps/wiki/App-Contribution - feel free to add to those/discuss them. I think it's probably good to be public about what kind of things we check - especially about things like when a fork is a good idea.

ProgramminCat:
I'd be interested in helping out!

gfwilliams:
[...]

PR #3018, opened 2023-09-16
HRM Widget+  adds bpm digits in lato font, plus some README, icon, screenshot fixes

hughbarney:


apps/widhrt2/widget.js
```diff
@@ -0,0 +1,49 @@
+(function(){
+  if (!Bangle.isHRMOn) return; // old firmware
+  var hp = Bangle.setHRMPower;
+  Bangle.setHRMPower = () => {
```
bobrippling, line 4: I think we can get away with this on Espruino JS, but `() => ...` functions don't have `arguments`, so we should perhaps call this a `function` to avoid breaking if Espruino changes to match the spec more

```suggestion
  Bangle.setHRMPower = function() {
```

apps/widhrt2/widget.js
```diff
@@ -0,0 +1,49 @@
...
+    hp.apply(Bangle, arguments);
+    WIDGETS.widhrt.draw();
```
bobrippling, line 6: Should we preserve the return value?

```suggestion
    var r = hp.apply(Bangle, arguments);
    WIDGETS.widhrt.draw();
    return r;
```

bobrippling:
Nice, LGTM! Just two little things on intercepting `setHRMPower`

apps/widhrt2/widget.js
thyttan: Intriguing,  will functions declared like so break:

`(arg1)=>{...}`

?

apps/widhrt2/widget.js
hughbarney: Happy to go with whatever you guys decide.
I only did it this way as that code became part of widhrt at somepoint and just copied it from there.


Just out of interest how are pull requests working these days ?
I am more than happy for my code to be reviewed and glad that we have this in place, it can only lead
to better apps and code.
I might have missed some announcements though.
Bob are you working for Gordon or were you volunteered etc ?
Is there a team of reviewers etc ?

apps/widhrt2/widget.js
bobrippling: So with explicit arguments, do you mean? For example:

```javascript
  Bangle.setHRMPower = (isOn, appID) => {
    hp.call(Bangle, isOn, appID);
    WIDGETS.widhrt.draw();
  };
```

Yes, that would be fine in this case - I suppose to make things behave as transparent as possible we should forward on the `this` value rather than hardcoding it as `Bangle` here, but you could say that's getting too academic

apps/widhrt2/widget.js
bobrippling: Ah gotya, in that case I'll resolve these comments and perhaps make these adjustments in a separate PR later - thanks for letting me know!

And yeah, Gordon asked if anyone would mind sharing the load (in #2993), what with holidays and so on. Just volunteering :)

bobrippling:
I'm happy to merge (and install on my watch!) if that's ok with you too @thyttan?

thyttan:
> I'm happy to merge (and install on my watch!) if that's ok with you too @thyttan?

I didn't look too closely - if you and @hughbarney are happy I trust you :+1:

I'd personally maybe have opted for the id to be `widhrtplus` to match the user facing name, like we did with `runplus`/Run+. But doesn't really matter :p

bobrippling:
> I'd personally maybe have opted for the id to be `widhrtplus` to match the user facing name, like we did with `runplus`/Run+

Yeah that might help it stand out more, what do you think @hughbarney?

gfwilliams:
> I'd personally maybe have opted for the id to be widhrtplus

Yes, I think this would be best. No other apps use `+` in the name, and I'd be a bit concerned that it could cause issues trying to store those files on some operating systems?

But otherwise this looks good

hughbarney:
I have changed id to widhrtplus.  I should point out that no files have a + in the name.

I'm rather confused by how to manage edits to a pull request once it has been submitted;  I  might have broken other changes that Rob did to my pull request above,   I have no idea how I would get such changes into my branch to merge them etc.

thyttan:
> I'm rather confused by how to manage edits to a pull request once it has been submitted; I might have broken other changes that Rob did to my pull request above, I have no idea how I would get such changes into my branch to merge them etc.

I wasn't really sure myself, so I found this: 
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

So the suggested changes you decide to commit in will be added to the branch you want to merge in with espruino/BangleApps:master, in this case hughbarney/BangleApps:master. Those will now be in your branch here on github, but not sync to your local copy if you cloned the repo to your computer. That should be resolved if you fetch or pull the changes from the remote copy on github (I think).

bobrippling:
There's no problem in this case, what you've done is spot on imo - a merge (of `upstream/master`) and extra changes committed followed by a push is a good way to check. If git sees that anything might be overwritten, it'll fail at the push, so that can be used to check nothing's been missed :)

(in this case I only had comments, no changes to your branch, so nothing is lost)

bobrippling:
Meant to say - I'm happy to merge if everyone else is?

hughbarney:
> I wasn't really sure myself, so I found this: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

Glad it is not just me.  I'e going to try and do my next app all through the fork on github - and try and use the 'update fork' button which I could not find, maybe I was not signed into github.

gfwilliams:
Thanks! Looks good to me! Sorry about the confusion on the naming - I saw `HRM Widget +` and assumed the app ID was `widhrm+` - thanks for making the change anyway, and I think that probably fits better with what we've done in other apps (run/runplus/etc)

PR #3019, opened 2023-09-16
pomoplus: new setting and minor changes

notEvil:
see https://notevil.github.io/BangleApps/?q=pomoplus

- adds setting "Show clock"

If set, returns to clock (`Bangle.showClock()`) after start/resume.

- minor changes
  - (changed*) buzz pattern of long break
  - added `g.clear()` and adjusted `.clearRect` in `drawTimerAndMessage`
  - replaced `g.getWidth()` and `g.getHeight()` with `w` and `h`

@bruceblore is this in conflict with your vision for this app?

*nothing changed since atm the last buzz immediately follows the second one

### DIFF
--- a/apps/pomoplus/ChangeLog
+++ b/apps/pomoplus/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02-0.04: Bug fixes
 0.05: Submitted to the app loader
+0.06: Added setting to show clock after start/resume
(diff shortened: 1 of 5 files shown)


bruceblore:
No conflict with my vision. I think that's a good feature to have, and people who do not like it can just turn it off. And I appreciate the code quality improvements.

gfwilliams:
Looks good to me! Thanks for the link to try it out.

I guess a nice addition at some point might be to try and use the `sched` library to schedule the timer (rather than boot code?). Having less boot code means app switching is a bit faster :)

PR #3023, opened 2023-09-22
Add Anton Clock + Remote for Spotify

fizitzfux:
I have merged the UI of antonclk with the Spotify controls of spotrem as I was annoyed at not being able to see the time and widgets and control Spotify at the same time.
The only new feature is the addition of a buzz on every spotrem action.
The UI of spotrem has been replaced with the antonclk UI, and all code for the submenus of spotrem has been removed as the clock has no submenus.


thyttan:
Hi! Thanks for contributing - especially fun for me that you wanted to use Remote for Spotify to build something ;)

I'll take a closer look at the PR in a minute, but address some things in your opening post first.

> I was annoyed at not being able to see the time and widgets and control Spotify at the same time.

It sounds like it would be worthwhile to add settings to show/hide the widget field! If you want to add that to spotrem directly I'd be happy to merge that!

>The only new feature is the addition of a buzz on every spotrem action.

I'd probably be happy to add this to spotrem as well if you'd like to add it there?

thyttan:
A readme describing the interactions could be good. Have a look at e.g. https://github.com/espruino/BangleApps/blob/master/apps/_example_app/README.md.

Maybe it would be better to drop the underscore from the id so it becomes `antonclkspotrem`, or maybe `antonspotclk`? It would jam a little bit more with whats said in the [BangleApps README](https://github.com/espruino/BangleApps#adding-your-app-to-the-menu) (at least according to my interpretation).

But also, did you see [Bluetooth Music Swipe Control Widget](https://thyttan.github.io/BangleApps/?q=widhid)? I think it does basically the same thing as `antonclk_spotrem` but on any clock/app? Maybe it would be a good idea to add to that instead if there's anything you miss there. What do you think @fizitzfux?

I've gone through [the checklist on app contributions](https://github.com/espruino/BangleApps/wiki/App-Contribution) and for the most part it looks good according to that :+1:

fizitzfux:
Hey! Thanks for your time and the quick and detailed replies.

> It sounds like it would be worthwhile to add settings to show/hide the widget field!

That seems like a feature that'd help some for sure, though not an exact solution to my particular case. I'd be fine giving a go at implementing it though as it does seem useful for others.

> I'd probably be happy to add this to spotrem as well if you'd like to add it there?

Sure!

> I've also been thinking of adding the possibility to have the spotrem time out to clock after say 10 minutes. Would that make sense to you?

To me that doesn't really make sense, though I could be misunderstanding what you mean.

> A readme describing the interactions could be good.

Oh yeah of course! Totally forgot about adding a readme.

> Maybe it would be better to drop the underscore from the id so it becomes antonclkspotrem, or possibly antonspotclk?

I chose to use an underscore in between the two names to show more clearly that it is just those two apps merged, instead of something totally new, and I've seen a few other apps with underscores in their name as well.

> did you see Bluetooth Music Swipe Control Widget?

Oh I actually did not! I must've overlooked that one when I was looking around. It seems to be exactly what I'm looking for, I didn't know widgets could actually be that powerful! If it does what it promises then that'd be perfect, though then this pr seems quite unnecessary.

I suggest dropping this pr, I'll try out `Bluetooth Music Swipe Control Widget` and I will be looking into creating a new pr for `spotrem` including the features mentioned above as I'd still love to help this community.

If you think this pr can still be worth looking into further I'm happy to do so, but otherwise you're free to close this one.

PS: sorry for my long reply, I wanted to make sure I commented on everything and I'm quite new to making public pr's.

thyttan:
> > It sounds like it would be worthwhile to add settings to show/hide the widget field!
> 
> That seems like a feature that'd help some for sure, though not an exact solution to my particular case. I'd be fine giving a go at implementing it though as it does seem useful for others.

Super if you do! If it's not for your own benefit though I could do it at some point in the future also. (As a sidenote, have you seen the digital or analog clock widgets?)

> > I'd probably be happy to add this to spotrem as well if you'd like to add it there?
> 
> Sure!

Looking forward to it! 🙂
 
> > I've also been thinking of adding the possibility to have the spotrem time out to clock after say 10 minutes. Would that make sense to you?
> 
> To me that doesn't really make sense, though I could be misunderstanding what you mean.

gfwilliams:
Thanks for sorting this and trying to roll changes back into the main app - the less duplicated code the better :)

> for future reference, underscores ok or not?

Yes, they're ok - a few apps seem to already use them. For cases like this i'd say it could be a good idea, but in general I'd avoid adding them to 'normal' apps

bobrippling:
Sorry about the silence - been on holiday for a little while. I think also `antonclk_spotrem` is approaching the app name length limit (I think it's around 32 chars?)

gfwilliams:
> I think it's around 32 chars?

It's actually 28 - `"antonclk_spotrem.app.js"` is 23, but yeah, ideally we want to try not to have huge filenames. Even with that, if a file like `"antonclk_spotrem.settings.js"` was added that's the absolute limit

PR #3029, opened 2023-09-23
Waternet v0.02 - Bigger Puzzle / Playfield tiles (more viewable on watch)

joyrider3774:
Hello,

People had issues viewing the playfield / puzzles on the watch, so i quickly updated waternet to version 0.02 which makes the playfield / puzzle tiles a lot bigger (16x16 instead of 10x10) so that things are much better viewable on the watch. 

Besides that the option to view widgets is removed, it now always runs fullscreen. 

And one final (small) fix was to the selector, when drawing it got cut off because i was using an int8array instead of int16array, but this only become visible in this version and did not really affect the initial version.

I also updated the screenshots to reflect the changes.

New version (pre)viewable on my github pages here 
https://joyrider3774.github.io/BangleApps/?id=waternet


thyttan:
I tried it - looks good! I'll keep this on my watch :)

PR #3030, opened 2023-09-24
[mtnclock] Add settings and the ability to show widgets

r-j-taylor:
This clock currently does not show the widgets bar. So, this PR adds a setting to enable the widgets bar but leaves it disabled by default to maintain the original behavior.


gfwilliams:
Thanks! However I'm afraid this is the wrong place to put the code for `Bangle.loadWidgets();` as it'll be trying to reload widgets each time the clock is drawn which could be slow.

Please could you keep drawWidgets where it is but then put the `if (...) { Bangle.loadWidgets(); Bangle.drawWidgets() }` right at the very bottom of the file?

apps/mtnclock/settings.js
```diff
@@ -0,0 +1,29 @@
...
+    STORAGE.writeJSON(FILE, SETTINGS);
+  }
+
+  // Show the menu
+  E.showMenu({
+    "" : { "title" : "Mountain Clock" },
+    "< Back" : () => back(),
+    'Show widgets': {
+      value: !!SETTINGS.showWidgets,  // !! converts undefined to false
+      format: value => value ? "On" : "Off",
```
bobrippling, line 22: For booleans we can leave this out and it'll default to a standard display :) (for example, [like in this commit](https://github.com/espruino/BangleApps/commit/42c2b5c8a2)'s settings file changes)

```suggestion
```

gfwilliams:
Thanks!

PR #3032, opened 2023-09-26
widminbate: monochrome version of `widminbat`

thyttan:
... using the system theme foreground color for the indicator bar.

Can be tried here: https://thyttan.github.io/BangleApps/?q=widminbate


gfwilliams:
Looks good - thanks!

PR #3034, opened 2023-09-28
Oxo Focus - play the computer as it learns to play naughts and crosses

hughbarney:


apps/oxofocus/app.js
```diff
@@ -0,0 +1,464 @@
...
+    }
+}
+
+Bangle.on('touch', function(zone,e) {
+  x = Object.values(e)[0];
+  y = Object.values(e)[1];
+
+  if (game_state == GAME_OVER) {
+    new_game();
+    return();
```
bobrippling, line 457: To fix the build we'd need to remove the `()` from this return

```suggestion
    return;
```

bobrippling (changes requested):
Nice app! Gave it a whirl (with the suggested fix) and it flows really nicely. Thanks for adding the screenshots too

gfwilliams:
Looks great - thank you!

PR #3040, opened 2023-10-08
tetris: Better controls and game over

pavelmachek:
Implement control by clicking, move down more by my click, and actually implement game over. Plus update metadata to display README and allow running in emulator.


gfwilliams:
Great - thanks!

PR #3041, opened 2023-10-10
backswipe: ignore `undefined` handlers

bobrippling:
`undefined` handlers are created when we [remove listeners] part-way through the array. This fixes `backswipe` for all firmwares, but leaves the door open for a potential firmware change.

See also https://github.com/espruino/BangleApps/issues/2953#issuecomment-1747428051

[remove listeners]: https://github.com/bobrippling/Espruino/blob/0f16231a4369bb2d154b9fab5a0d806122cddb39/src/jswrap_object.c#L1035-L1035


gfwilliams:
Great - thanks!

PR #3045, opened 2023-10-13
Astral clock update

jabituyaben:
had a needless while loop in here, which is now replaced by a one liner that makes it update much faster.


gfwilliams:
Great - thanks!

PR #3046, opened 2023-10-14
alarm auto expiration for events + fix repeat back

### LINKED ISSUE #2930: [alarm] Repeat setting changes from Once to 1*Month if Repeat menu canceled without changes
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

While editing an event, entering the Repeat menu while the setting is “Once”, then pressing Back to exit the menu without making a selection, results in an unexpected change of the repeat setting to once monthly. Under normal circumstances I would expect entering and immediately leaving a menu not to change any existing settings; this behavior is inconsistent with that of other menus in the interface. However, other repeat settings do not seem to be affected: Entering and backing out of the menu when a different repeat setting is active does leave the setting unchanged as expected. Only “Once” seems not to behave consistently.

Reproduction steps:
1. Open Alarms app
2. Select “New…”
3. Select “Event”
4. Scroll down to Repeat. Note default selection of Once
5. Tap Repeat menu, then immediately press Back.
6. Scroll down to Event and note the setting has changed to 1*Month. The expected result would have been for the setting to remain as Once.

I looked at the code to see if I could fix it, but I have a bit of trouble following how the repeat menu routine works exactly. I note that `showEditRepeatMenu()` gets `day` and `repeat` variables passed to it and that it seems to use a separate var `dow` to make some internal changes to `day` before the user makes a selection; but for some reason `repeat` gets manipulated directly straight away instead of a separate var as with `day`. This ultimately seems to be causing the “false” value of `repeat` to get unconditionally replaced with an object representing once monthly, and then this is getting saved back to the alarm when the user cancels the menu (despite the user not having selected any changes).

### Installed apps

launch (0.20), antonclk (0.11), about (0.14), widlock (0.08), widbat (0.11), welcome (0.14), sched (0.22), sched (0.22), widalarm (0.01), boot (0.58), notify (0.13), health (0.23), setting (0.61), alarm (0.41), mylocation (0.10), hrm (0.11), waypoints (0.01), gpsnav (0.08), sleeplog (0.15), widdevst (0.03), stlap (0.04), stlapview (0.03), batchart (0.12), dtlaunch (0.24), timerclk (0.05), widclk2 (0.00), multitimer (0.03)

Also reproducible in emulator

nxdefiant:
auto expiration for events:
- Set default value
- Add "delete after expiration" checkbox to events.

repeat menu: Do not change value on back (fixes #2930)
Instead of saving repeat argument as originalRepeat create a new changed repeatObj

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -46,3 +46,4 @@
 0.41: Fix a menu bug affecting alarms with empty messages.
 0.42: Fix date not getting saved in event edit menu when tapping Confirm
 0.43: New settings: Show confirm, Show Overflow, Show Type.
+0.44: Add "delete timer after expiration" setting to events.
--- a/apps/alarm/app.js
+++ b/apps/alarm/app.js
@@ -106,6 +106,9 @@ function showEditAlarmMenu(selectedAlarm, alarmIndex, withDate) {
   var isNew = alarmIndex === undefined;
 
   var alarm = require("sched").newDefaultAlarm();
+  if (withDate) {
+    alarm.del = require("sched").getSettings().defaultDeleteExpiredTimers;
+  }
   alarm.dow = handleFirstDayOfWeek(alarm.dow);
 
   if (selectedAlarm) {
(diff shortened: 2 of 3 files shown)


gfwilliams:
Looks good - thanks!

PR #3049, opened 2023-10-16
added support for hr raw exporter and astral clock to bangle.js 2

jabituyaben:
added support for hr raw exporter to bangle.js 2


apps/hrrawexp/metadata.json
```diff
@@ -2,13 +2,14 @@
   "id": "hrrawexp",
   "name": "HRM Data Exporter",
   "shortName": "HRM Data Exporter",
-  "version": "0.03",
+  "version": "0.04",
   "description": "export raw hrm signal data to a csv file",
   "icon": "app-icon.png",
   "tags": "",
   "supports": ["BANGLEJS"],
```
gfwilliams, line 9: Looks like you've got a duplicate line here - maybe just update this line rather than adding `"supports": ["BANGLEJS","BANGLEJS2"]`?

gfwilliams:
Hi - thanks for this - but looking at the astral clock changes, haven't you just broken it for Bangle.js 1?

Bangle.js 1 touchscreen only supports left/right swipes (no up/down) so as far as I can see this would no longer be usable on it? You could try with the emulator to check.

Maybe you should check if it's Bangle.js 1/2 and then then choose whether to do swipes or button presses?

OR you could do `Bangle.setUI({mode:"clockupdown"})` (https://www.espruino.com/Reference#l_Bangle_setUI) which automatically uses either taps or button presses depending on which device is used.

jabituyaben:
> Hi - thanks for this - but looking at the astral clock changes, haven't you just broken it for Bangle.js 1?
> 
> Bangle.js 1 touchscreen only supports left/right swipes (no up/down) so as far as I can see this would no longer be usable on it? You could try with the emulator to check.
> 
> Maybe you should check if it's Bangle.js 1/2 and then then choose whether to do swipes or button presses?
> 
> OR you could do `Bangle.setUI({mode:"clockupdown"})` (https://www.espruino.com/Reference#l_Bangle_setUI) which automatically uses either taps or button presses depending on which device is used.

Good spot - I've sorted this by checking which version of the Bangle it is and adding button function for Bangle 1. Also added note that it currently only supports the dark background mode for Bangle 2 at the moment.

gfwilliams:
Looks great - thanks for making that change!

PR #3051, opened 2023-10-16
Calendar: Add menu to jump to settings

nxdefiant:
- Add menu to fast open settings to edit holidays
- Drop "B2 Colors" setting, use theme dark indicator instead
- Display Widgets in menus


gfwilliams:
Looks good to me, but I'd be a bit concerned about removing the non-dithered colour stuff? That was put there at some point specifically because some folks complained about not being able to see the text properly against the dither pattern?

nxdefiant:
ok, I can re-add the color option. I just thought I spare the part to reload the color after settings change and folowing the theme to choose which color to use seemed for natural to me. Also text looks fine for one on the Bangle.js2 with dark theme.

gfwilliams:
Ok, thanks, I just tried this out though:

image

image

There's just no comparison - I can barely make anything out, and the numbers that aren't in the current month are pretty unreadable even on desktop. I think the B2 colours option needs to stay and be the default for Bangle.js 2.

gfwilliams:
Just so you know while testing I spotted a bug in the settings (which wouldn't open on a fresh install it seemed) so I've fixed that - but didn't bump the version of change the changelog so I think your changes should still merge in fine

nxdefiant:
ok, I've undone the change of the ndColors removal

gfwilliams:
Great - thanks!

PR #3053, opened 2023-10-17
grocery: Drop grocery.html

nxdefiant:
Drop app customiser as it is redundant with download interface and the latter has more features


gfwilliams:
Thanks!

PR #3054, opened 2023-10-18
scicalc: Add swipe up action

nxdefiant:
Swiping up has now the same effect as hitting the "<" button.

@berkenbu Hope this is ok with you. I find it easier to correct typos that quickly appear with a single swipe instead of navigating to the correct screen first.

### DIFF
--- a/apps/scicalc/ChangeLog
+++ b/apps/scicalc/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New App!
+0.02: Swiping up has now the same effect as hitting the "<" button.
--- a/apps/scicalc/README.md
+++ b/apps/scicalc/README.md
@@ -5,7 +5,7 @@ by the JS interpreter.
 
 ## Usage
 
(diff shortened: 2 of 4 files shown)


nxdefiant:
To be honest having left swipe for undo, right swipe to calculate result and accordingly up/down swipe to cycle between screens would feel more natural for me. But I did want to touch the original input actions.

berkenbu:
Original author of the calculator here: 
I have this (arguably arbitrary) mental image of the 3 input screens being arranged side by side, hence the lr swipe gesture to switch between them.
That being said, I am not married to that idea, if others find your arrangement more natural I am not against changing it to that.

gfwilliams:
This PR looks good to me - as it's just adding the swipe up functionality rather than changing things - so merging now.

Although I can definitely see the reasoning behind having swipe left as delete, and right as calculate result. While arranging the screens left/right makes sense, I think most users probably expect left swipe=delete/back, and given that arranging the screens up/down instead probably is preferable.

... so I'd be up for changing it. It might also be an opportunity to use `Bangle.setUI` to get user input as well?

nxdefiant:
ok lets do it

Issue #2188, opened 2022-10-21
[Settings] Calibration screen - widgets hidden but active


thyttan:
Fixed for me as well, thanks! Will come back re/on #1040. 

Just a small note: I had "Light Switch Widget" installed when trying calibrating this time and when I touched the crosshair when in the top right corner, Light switch widget would be displayed and toggle background illumination, instead of registering touch to the crosshair. So I had to try again by touching a little further down on the screen - which worked and the crosshair then moved to the next corner. Should I open a separate issue for this, @gfwilliams?

_Originally posted by @thyttan in https://github.com/espruino/BangleApps/issues/2181#issuecomment-1285306079_

thyttan:
Note to myself that this should be easily solvable with a prependlistener + stopeventpropagation combo now.

PR #3057, opened 2023-10-19
setting: Prepend 'LCD->Calibration' touch listener

thyttan:
...and stop event propagation.

Closes https://github.com/espruino/BangleApps/issues/2188

I thought about intercepting drag events as well. But didn't bother. Can add it if wanted.


gfwilliams:
Great! Thanks for sorting that out!

Issue #1690, opened 2022-04-11
[Multiclock] Widgets flash when switching to/from dk.face.js (iface index 4)


thyttan:
@jeffmer

### Affected hardware version

Bangle 2

### Your firmware version

2v13

### Describe the bug
The widget field flashes when switching to/from dk.face.js (iface index 4).
### Expected Behavior
The widget field should not flash, as it doesn't when switching between the other faces.
### Steps to reproduce
1. [Load Multi Clock v0.09 from main App Loader](https://banglejs.com/apps/?q=multi%20clock)
2. (from this step onward do on watch) Launcher->Settings->System->Select Clock->Multi Clock
3. Press physical button to load selected clock.
4. Cycle through clock faces by pressing repeatedly on upper or lower screen corner until widget field flashes.

thyttan:
I've looked at the it but couldn't figure out what in the code causes the bug.

jeffmer:
Worked out why this is happening. The other faces do not call g.flip() at the end of the drawclock routine and rely on the screen being updated by the Espruino idle routine. So the fix is simple - remove the g.flip() line in drawclock(). Will submit a PR when I get a chance as I also want to add a new face.

PR #3058, opened 2023-10-19
multiclock: fix widget field flashing

thyttan:
... when switching to the "dk "clock face.

Closes https://github.com/espruino/BangleApps/issues/1690


gfwilliams:
Thanks!

PR #3060, opened 2023-10-27
OpenHaystack: Keep advertising when connected

olivierbarriere:
A small modification to keep advertising as airtag when connected.

It still advertises both airtag and normal device name when connected.
Ideally it should be only airtag advertissment when connected (every 2000ms), and back to both (every 375ms) when disconnected.


gfwilliams:
Sounds like a good idea - thanks!

PR #3061, opened 2023-10-27
scicalc: Switch swipe handling

nxdefiant:
As discussed in #3054

- Use Swipe up/down to switch between screens, left to delete, right to calculate result
- Also use Bangle.setUI()


gfwilliams:
Looks excellent - thanks for doing that so speedily!

PR #3065, opened 2023-10-30
CalClock: Add buzzing before events

LeonMatthes:
The main reason why I developed CalClock is for me to not make sure I don't forget any upcoming events I have in my calendar.

So I added a subtle buzzing that notifies me of upcoming events.

TODO: In future I want to put this into the boot code, as then I could always be reminded of upcoming events, even if I have any other app open.


bobrippling:
Looks good, thanks!

PR #3066, opened 2023-10-30
sunrise: redraw the clock exactly when a minute ticks

bobrippling:
A minor tweak to sunrise - ensure we render the clock when a minute ticks over, so the time shown on screen lines up with other widgets (e.g. chimer).

Leaving as a draft until #3063 is merged, to avoid introducing conflicts.


bobrippling:
Looks like a temporary build issue, will retry after #3063 is merged

```
npm ERR! code E500
npm ERR! 500 Internal Server Error - GET https://registry.npmjs.org/@humanwhocodes/object-schema/-/object-schema-1.2.1.tgz - KV GET failed: 401 Unauthorized
```

gfwilliams:
Thanks! Odd about the build issue, but I'm happy with how that looks

PR #3068, opened 2023-10-31
Add cadence data to `recorder` app

Mineinjava:
Hi
This is my first contribution to this repository. If there is anything I missed, please let me know.
This PR adds cadence data to the GPX output file

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -38,4 +38,4 @@
 0.30: Add clock info for showing and toggling recording state
 0.31: Ensure that background-drawn tracks can get cancelled, and draw less at a time to make updates smoother
       plotTrack now draws the current track even if you're not actively recording
-0.32 Add cadence data to output files
+0.32: Add cadence data to output files
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -38,3 +38,4 @@
 0.30: Add clock info for showing and toggling recording state
 0.31: Ensure that background-drawn tracks can get cancelled, and draw less at a time to make updates smoother
       plotTrack now draws the current track even if you're not actively recording
+0.32 Add cadence data to output files
--- a/apps/recorder/interface.html
+++ b/apps/recorder/interface.html
@@ -99,11 +99,11 @@ function saveGPX(track, title) {
   track.forEach(pt=>{
     let cadence;
     if (pt.Steps && lastTime != 0){
-      cadence = pt.Steps * 60000 / (pt.Time.getUTCMilliseconds() - lastTime);
+      cadence = pt.Steps * 60000 / (pt.Time.getTime() - lastTime);
     }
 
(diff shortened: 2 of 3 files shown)


bobrippling:
Great first contribution :) Let us know when you're happy with the remaining items on your list and we can re-review

gfwilliams:
Thanks! I guess my only concern with this is the CSV file already stores steps and time, so it should be easy to work Cadence out from what's in the stored CSV data already - basically just looking at the difference between the last timestamp and this one and doing `steps * 60000 / timeDifference`?

So we could just have the changes you made to `apps/recorder/interface.html`, but with a bit of extra code in there to work out the Cadence at the time the file was read back?

Mineinjava:
> Thanks! I guess my only concern with this is the CSV file already stores steps and time, so it should be easy to work Cadence out from what's in the stored CSV data already - basically just looking at the difference between the last timestamp and this one and doing `steps * 60000 / timeDifference`?
> 
> So we could just have the changes you made to `apps/recorder/interface.html`, but with a bit of extra code in there to work out the Cadence at the time the file was read back?

That sounds like a better solution. I will revert and reimplement. Thanks

Mineinjava:
oops didn't mean to close it

Mineinjava:
reopening

Mineinjava:
It appears that `pt.Time.getUTCMilliseconds()` returns zero.

Mineinjava:
I got the cadence log in the gpx output to work. However, it appears that Strava doubles the cadence value. This is probably because gpx cadence for bikes is rotations/min whereas for running it is steps/min, and two steps is one cycle.

Source: [Garmin GPX schema](https://www8.garmin.com/xmlschemas/TrackPointExtensionv1.xsd):
```xml
<xsd:element name="cad" type="RevolutionsPerMinute_t" minOccurs="0"/>
```

I believe the standard is one revolution is two steps (one step per foot)

I will update the source code to divide by two as that appears to be the standard. I don't believe this will cause a problem on bicycles because step count while cycling is irrelevant.

thyttan:
Off topic:

Hi! Once you're done with this would you be interested in working on gpx-stuff in Gadgetbridge for implementing the 'Activity Tracks' feature for Bangle.js?

https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3153#issuecomment-1314419

gfwilliams:
This looks great - thanks for your work! 

Are you happy to merge this now? I know there are some items in the original PR note you haven't marked complete but even so the changes here seem useful

Mineinjava:
> Off topic:
> 
> Hi! Once you're done with this would you be interested in working on gpx-stuff in Gadgetbridge for implementing the 'Activity Tracks' feature for Bangle.js?

Sorry, I don't use gadgetbridge (I have an iphone) and I'm not sure I understand what that feature is.

thyttan:
Ah! Thanks anyways then 😉

gfwilliams:
Great - thanks!

Issue #3052, opened 2023-10-16
[cards] Widget bar obscuring QR codes


m-p-3:
### Affected hardware version

Bangle 2

### Your firmware version

2v19

### Describe the bug
When the watch goes into "lock" mode, the widget bar shows up shortly after, obscuring the top of the QR code and making it unscannable because the timing blocks at the top are cropped.

### Expected behavior
Make sure the widget bar doesn't show up when a QR code is displayed.

### Steps to reproduce
1. Create a test QR code in Catima
2. Upload card
3. Display card
4. Wait for BangleJS to lock
5. ![PXL_20231016_173015088](https://github.com/espruino/BangleApps/assets/3731432/e343155f-33ff-4ad2-8d86-0f62e1fa576a)

### Installed apps

`messageicons (0.05), widalarm (0.01), authentiwatch (0.07), files (0.08), widbt_notify (0.17), calclock (0.06), widclk (0.07), info (0.03), widlock (0.08), widslimbat (0.01), fileman (0.03), weather (0.25), sched (0.22), widmessages (0.05), powermanager (0.10), drained (0.03), owmweather (0.03), mylocation (0.10), agenda (0.14), quicklaunch (0.15), locale (0.17), qmsched (0.10), wid_edit (0.05), launch (0.20), boot (0.59), android (0.32), messagegui (0.74), messages (0.62), cards (0.01), alarmqm (0.01), setting (0.64), alarm (0.44)`

nxdefiant:
@glemco

Does the QR code usually paint on the full area? If so it looks like we shoud either
- remove load/drawWidgets() calls
-  Switch to widget_utils hide/swipeOn();

glemco:
@nxdefiant  yes it does. I could make it smaller but the resolution is already not so great so I didn't go down that path..
As soon as I have some time I'll look into it. I'm not sure how it works but since it should be possible to go back and forth from the app (with widgets) to the code (without), would removing the call to drawWidgets work?

nxdefiant:
If you really want to show widgets try http://www.espruino.com/Bangle.js+Hideable+Widgets

glemco:
It would be much nicer if I could unload and reload widgets, the code for QR is already having issues as it is, saving some memory might give ti some breath.
I'll give it a go, thanks!

gfwilliams:
Yes, I'm afraid widget unloading just isn't possible, so probably the drag-down is the way to go.

[...]

PR #3069, opened 2023-11-01
cards: hiding widgets while showing the code

glemco:
With this PR we hide widgets while showing the barcode/QR and show them back while going back to the card's info. In this way, widgets are not covering part of the displayed code

Fixes: https://github.com/espruino/BangleApps/issues/3052


gfwilliams:
Looks good to me! Just version + changelog needed?

glemco:
Sure, there you go!

bobrippling:
Thanks!

PR #3070, opened 2023-11-01
astral: setUI before loading widgets

bobrippling:
This fixes the sanity check warning:
```
Clock astral file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)
```

(there's no affected code between the old and new locations of `setUI()`)


gfwilliams:
Great - thanks for sorting that!

PR #3071, opened 2023-11-02
hardalarm and 28x6 for banglejs 2

jamespsteinberg:


jamespsteinberg:
Good idea fixed all. Still learning the banglejs library - will look at the sched library next

gfwilliams:
Thanks! That looks great! Those 28x6 screenshots look great on Bangle.js 1

On Bangle.js 2, it seems like maybe it'd be worth changing `backgroundColor` from `#2c2e3a` to `#000`? That would make it pure black so would avoid the dithering, which I reckon makes it a bit harder to read than is needed?

Merging now anyway...

jamespsteinberg:
Good to know for background color - wasnt sure what was causing that

PR #3075, opened 2023-11-03
contacts: Inital version of contacts app.

pavelmachek:
Simple app to view/edit contacts from phone and from web interface.


nxdefiant:
Cool idea, just one suggestion: How about using one of the [textinput libraries](https://github.com/espruino/BangleApps/blob/master/apps/kbswipe/README.md) to allow edit on device?

Like [noteify](https://github.com/espruino/BangleApps/blob/319d22fab675bd9ca0c81ab26c58b2aeb6d1299f/apps/noteify/app.js#L10) or (grocery)[https://github.com/espruino/BangleApps/blob/319d22fab675bd9ca0c81ab26c58b2aeb6d1299f/apps/grocery/settings.js#L77]

pavelmachek:
I attempted to do that -- text input seems to work for me. Let me know if I got it wrong (or if I need dependency in manifest.json or something like that).

pavelmachek:
Actually, what I'd really like to do would be gadgetbridge integration, so that call could be initiated from the contacts app on watch. But unfortunately I don't know enough about android to be able to code that. Help/hints would be welcome.

nxdefiant:
ah ok, I've missed that in the code. Yes you should also add
`"dependencies": {"textinput":"type"},`
to metadata

bobrippling:
I think this looks good, @nxdefiant happy for us to merge?

nxdefiant:
of course!

bobrippling:
Great, thanks both!

PR #3076, opened 2023-11-03
tetris: implement level/mode selection

pavelmachek:
Implement level selection, and allow user to select controls.


bobrippling:
Great stuff, love the tilt controls too!

PR #3079, opened 2023-11-03
waypoint_editor: allow text waypoint names, allow marking current GPS position

pavelmachek:
Plus some cleanups as a bonus.

I wanted to ask: would it be okay to merge this with "waypoints" app? app.js file is small enough that keeping them separate is just complicating things and taking more resources...

### DIFF
--- a/apps/waypoint_editor/ChangeLog
+++ b/apps/waypoint_editor/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New App!
 0.02: Display waypoint name instead of its index in remove menu and fix icon
+0.03: Use text input for waypoint names, allow marking waypoint with current GPS position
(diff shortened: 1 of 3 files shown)


pavelmachek:
(Plus, "waypoint_editor" name is descriptive but quite long. Merging this with waypoints would help with that, too).

bobrippling:
> would it be okay to merge this with "waypoints" app? app.js file is small enough that keeping them separate is just complicating things and taking more resources

> Plus, "waypoint_editor" name is descriptive but quite long. Merging this with waypoints would help with that, too

Sounds good to me - Gordon's done [something similar](https://github.com/espruino/BangleApps/commit/91728c147) in the past. Would you prefer me to merge this PR before your merging with the waypoints app?

pavelmachek:
I'd prefer to merge this and base new work on top of it (better for my workflow and not sure when I find time).

bobrippling:
Sounds good, in that case thanks for the PR!

PR #3080, opened 2023-11-05
Fix coloring on 28x6 app

jamespsteinberg:


bobrippling:
Thanks, looks great!

Adding a [link here](https://github.com/espruino/BangleApps/pull/3071#issuecomment-1792042377) in case anyone's interested in the rationale

PR #3083, opened 2023-11-07
settings: lcd calibration bug

d3nd3:
The code logic does not work because the function with no return, returns undefined. Ends up that both `prependListener` and `on` are called, creating 2 callbacks, yet only 1 is removed.

Noticed it from running lcd calibration, then touch anything else and it goes back into calibration.


thyttan:
I could trigger the bug just now and it was resolved when I switched to your suggested change.

Thanks!

PR #3084, opened 2023-11-07
setting: bump version and ChangeLog

thyttan:
After LCD calibration fix.


d3nd3:
I just realized this should probably edit settings app changelog, to keep it proper?

Issue #2887, opened 2023-07-17
[SleepLog] screen refresh issues on banglejs 2


m0n5t3r:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

Everything updated, the sleep log app mangles the output when switching between days (left / right swipe), see attached image

expected behaviour: display texts properly, not have junk displayed on the left of the screen, or the "calculating data, please wait" text visible at all times

### Installed apps

simple clock
simple pedometer
charging status
thermometer
launcher
message icons
weather
heart rate monitor
alarms widget
daylight saving
scheduler
health tracking
alarms & timers
astrocalc
tea timer
bluetooth widget
compass
notifications (default)
bootloader
batery level widget (vertical)
messages
sleep log
settings
my location
message ui
android integration

ticalc-travis:
I just had this happen to me (Bangle 2, 2v19) after having used Sleep Log for a few months. It occurred immediately after changing the time zone setting (actually, installing Daylight Saving to do it for me). Was fine the previous day, then the next day after the time change, the UI is unreadable just as in the screenshot, apparently permanently. Possible clue?

bobrippling:
I've had a skim through the code and noticed a [`width` being used in place of a `height`](https://github.com/espruino/BangleApps/pull/3085), I'm not sure it impacts the exact issue here but I've [deployed it](https://bobrippling.github.io/BangleApps/?id=sleeplog) if you wouldn't mind giving it a go?

@storm64 how does this look to you?

Out of interest, does the issue appear consistently now, or just temporarily after changing the timezone?

ticalc-travis:
It seems consistent: even after changing the time zone back to its original setting the problem persists. It also persists when viewing past logs, including dates from previous log files before the currently-written one. I'll watch and see if the situation changes once the current log where the time change happened gets rotated out in a few days.

It doesn't appear that #3085 patch improves anything, unfortunately.

bobrippling:
Thanks for the update and for trying out the patch even so - in that case the issue will be in the drawing code most likely, just an overlap of two draws. But yes, nice idea, let's confirm if it happens when the rotate-out occurs too

ticalc-travis:
I haven't had time to investigate further, but a quick confirmation that even after a log rotation the symptoms are still present so far. The data still seems intact and reasonable if I view it from the app launcher, but it's still a mystery why the app UI remains confused so persistently.

bobrippling:
[...]

PR #3085, opened 2023-11-08
Fix sleeplog drawing

bobrippling:
Draft for now - potential fix for #2887


storm64:
Hi @bobrippling,
the choice to use the width as height in this position might not be the best decision, but I did it on purpose. My thoughts might have been, that I didn't wanted to get the height and make it "zero based" (-1), as the app might only be possible to work on BJ2 and it has a square screen.

bobrippling:
Ah I get you, that's fair enough - thanks for the explanation!

PR #3089, opened 2023-11-11
Add weather and calendar data to iOS integration

Mineinjava:
Adds the ability for the bangle.js to receive weather data contained in custom notifications.

Calendar shortcut: https://www.icloud.com/shortcuts/baee5a95a55a4c5c8e6e582e09db020f
Weather shortcut: https://www.icloud.com/shortcuts/f21cd4c1753445a4aadec84b55c3e7e6
(requires OWM api key)

- [x] Calendar
    - [x] on Bangle.js
    - [x] ios shortcut
- [x] Weather
    - [x] on Bangle.js
    - [x] ios shortcut
    
**Caveats:**

While notifications are automatically dismissed, they take up the top of the screen for a few seconds
 - this can be avoided with a focus mode

Calendar does not send the `color` field
 - defaults to `0`
 - I couldn't figure out how to do this on the side of iOS. It may be possible to replicate this by assigning calendar names to colors.

Weather does not give a forecast--just the current weather conditions
 - this might be an issue with my understanding of the bangle.js weather library

relevant forums discussion: https://forum.espruino.com/conversations/387583/

### DIFF
--- a/apps/ios/ChangeLog
+++ b/apps/ios/ChangeLog
@@ -13,3 +13,4 @@
 0.13: Making ANCS message receive more resilient (#2402)
 0.14: Add settings page, allow time sync
       Allow negative/positive actions to pass through to message GUI
+0.15: Enable calendar and weather updates via custom notifications (via shortcuts app)
--- a/apps/ios/README.md
+++ b/apps/ios/README.md
@@ -29,6 +29,24 @@ for now. It's just a few bucks/pounds/euro's.
 If you like to try a free app first, you can always use NRF Toolbox or
 Bluetooth BLE Device Finder to find and connect your Bangle.
 
+### Weather and Calendar
+
+By using the `Shortcuts` app on your phone, you can send weather and calendar data to your watch. This works by sending a notification, which is read by the watch through ANCS. The watch then parses the notification for the data.
+
+While you may write your own shortcuts if you prefer (for example, to get weather from a different source), two are provided:
+
+- Calendar shortcut: https://www.icloud.com/shortcuts/4eac12548b4c424dbcdb1bd58cff338f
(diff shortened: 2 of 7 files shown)


Mineinjava:
As of the latest commit I am ready to merge this PR.

My only concern is the change I made to the weather lib. I exposed the `update()` function.

I also made a mistake in one of the shortcuts, returning the wrong field (`icon` instead of `id`)

nxdefiant:
Just wondering: by using android.calendar.json you bypass the alarm/scheduler events. So there won't be an alarm with the default infrastructure. Is that really intended?

Mineinjava:
>  by using android.calendar.json you bypass the alarm/scheduler events.

I'm not sure I understand what you mean. I thought that only calendar data was stored in `android.calendar.json`--the only two times I see the file written to are when calendar items are uploaded/deleted.

nxdefiant:
In the alarm app you can create/edit dated events, these are stored in sched.json. They do create an alert with the scheduler at the specified date and time.

I would have assumed that calendar entries from Android/iOS would land in the very same sched.json, so they would work with the existing scheduler infrastructure.

Obviously this is not the case with the android.calendar.json. That was just me wondering why.

Mineinjava:
> That was just me wondering why you bypass the alarm/scheduler infrastructure.

I did not implement the alarm/scheduler infrastructure because the original gadgetbridge implementation (apps/android/boot.js) didn't--my implementation mimics the gadgetbridge implementation.

bobrippling:
To check - are you happy with the changes? I'll look to merging if so

Mineinjava:
Yes. I am ready to merge

bobrippling:
Great, thanks for all your tinkering with this!

PR #3092, opened 2023-11-12
Add tag health

nxdefiant:
Please add the tag health to the navigation filter


nxdefiant:
My guess is that @gfwilliams wants to review this personally.

gfwilliams:
Thanks! That looks good! Please could you add it to the Android.html file too so Gadgetbridge users see it? https://github.com/espruino/BangleApps/blob/master/android.html

nxdefiant:
done, but untested from me (no Gadgetbridge).

gfwilliams:
Thanks! Yes, no need for something like that :)

PR #3094, opened 2023-11-14
skyspy: view sattelitte positions

pavelmachek:
Application to view satellite positions and whether watch can receive data from them.


pavelmachek:
Thank you, fixed. Yes, this is about GPS.

bobrippling:
Cool idea, I'd like to try this out once it stops raining here - happy to merge if you're ok with everything else @nxdefiant ?

apps/skyspy/skyspy.app.js
```diff
@@ -0,0 +1,261 @@
...
+
+function onSwipe(dir) {
+  display = display + 1;
+  if (display == 3)
+    display = 0;
+}
+
+Bangle.setUI({
+  mode : "custom",
+  back : function() {}, // optional - add a 'back' icon in top-left widget area and call this function when it is pressed , also call it when the hardware button is clicked (does not override btn if defined)
```
nxdefiant, line 255: Not sure about this part: `back : function() {}` doesn't this cause the fastload of the cloak if 'fastload' is installed?

The whole code doesn't seem to be compatible with fastload so I would remove this line

nxdefiant:
Please check the fastload issue (see my comment), for everything else: +1

Also consider yourself lucky if you live in a part of the world where it stops raining...seems I'm not...

bobrippling:
Nice spot! And thank you - I'll consider my two hours of non-rain that's predicated today lucky 😄

pavelmachek:
Thanks for the fastload hint, now should be fixed, this resolved other problems, too.

I wish you a non-rainy weather, but... you should be able to put watch by the window, and still see some satellites to be decoded. With AGPS and patience, you might even get a fix...

apps/skyspy/skyspy.app.js
thyttan: > Not sure about this part: back : function() {} doesn't this cause the fastload[1] of the cloak if 'fastload' is installed?

I'm pretty sure this is not the case (please enlighten me if I'm wrong!). It's the presence of the `remove` key that allows fastloading to trigger (*https://www.espruino.com/Bangle.js+Fast+Load#1-bangle-setui-remove-*).

The `back` key is used to :
```
back : function() {}, // optional - add a 'back' icon in top-left widget area and call this function when it is pressed , also call it when the hardware button is clicked (does not override btn if defined)
```
*https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setUI*

thyttan:
Added a comment regarding the back function on the conversation above :)

bobrippling:
Ah yeah, totally forgot it was `remove` not `back`, thanks! I reckon we're happy to merge still since there was no back behaviour anyway, @pavelmachek are you happy with skyspy as it is now?

pavelmachek:
Yes, happy with the merge as is.

nxdefiant:
opps sorry, mixed them up.

bobrippling:
np, me too!

and thanks @pavelmachek

PR #3095, opened 2023-11-14
waypoints: Merge waypoint_editor into waypoints application

pavelmachek:
Waypoint editor is reasonably small piece of javascript, so merge it with waypoints and make it always-available.


bobrippling:
Nice, thanks for persisting with this!

@gfwilliams is there anything special to do when an app's deleted?

thyttan:
It seems the precedent regarding deleting apps when the functionality has been reimplemented is:
- don't delete the app.
- similarly to [`gbridge`](https://banglejs.com/apps/?q=gbridge) change its description to "(NOT RECOMENDED) ..." and point to the new app.

I don't have experience with waypoints or waypoints_editor, so maybe this does not make sense here.

gfwilliams:
Looks good!

Yes, we'd keep the old app if we think it's got lots of installs, but I just checked the installs of waypoint_editor (click on the version in the app loader) and it's less than 1% so I think we could remove. Afterwards it'll appear in the app loader as 'unknown app' but can still be removed if needed.

bobrippling:
Ok great - I'll merge this, I suppose it's easy enough to reinstate if anyone needs the old app back. Thanks

nxdefiant:
Does that work for you?
Seems the launcher filters for app.type:
https://github.com/espruino/BangleApps/blob/master/apps/launch/app.js#L30

gfwilliams:
Good point - @pavelmachek did you actually test this?

I think it'll need type:app, then `"provides_modules" : ["waypoints"]` which is probably what should have happened before. Other apps that use it will need `"dependencies" : { "waypoints":"module" }`

pavelmachek:
Sorry about that, my testing was clearly not adequate :-(.

bobrippling:
I merged it, I should've tested - I've created #3105 which will catch this kind of thing in future

PR #3099, opened 2023-11-15
Fixes for alarm, settings & waypoints

nxdefiant:
alarm: defaultDeleteExpiredTimers detection
Fix new alarm when selectedAlarm is undefined

settings: Fix syntax error

waypoints:
- Do not register as type waypoint - show in launcher
- Fixes for Bangle.js 1
- hide some menu entries if textinput is not installed


nxdefiant:
tests ok

nxdefiant:
@gfwilliams Got a syntax error in settings from 681bd41945, should be ok now

nxdefiant:
Lets fix also the menu entry of #3095

gfwilliams:
Looks great, thanks! Sorry for introducing that syntax error - not sure how that happened. I'd made these changes a while back and only just pushed them so I must have forgotten to test :(

PR #3098, opened 2023-11-15
recorder: don't prompt the user when the columns change - force new file

bobrippling:
Resolves a small issue with #3081, see [commit comments](https://github.com/espruino/BangleApps/commit/dd96e33774bee9c0ce0aea9ffe336ce18a0d631c#r132592068) for more detail

Closes #3081

PR #3100, opened 2023-11-15
recorder: handle initial/non-settings load

bobrippling:
Resolves [an issue](https://github.com/espruino/BangleApps/pull/3098#issuecomment-1813310835) where the recorder widget might load without having a settings file, leading to:

```
>
Uncaught Error: Cannot read property 'forEach' of undefined
 at line 1 col 2805 in recorder.wid.js
...Recorders();settings.record.forEach(r=>{var recorder=recorders[r...
                                  ^
in function "getActiveRecorders" called from line 1 col 4959 in recorder.wid.js
...Headers(getActiveRecorders()).join(",")){options.force="new"...
                              ^
in function "setRecording" called from line 11 col 368 in recorder.app.js
...["recorder"].setRecording(v).then(function(){loadSettings();showMai...
                                     ^
```

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -41,3 +41,4 @@
 0.32: Add cadence data to output files
 0.33: Ensure that a new file is always created if the stuff that's being recorded has changed (fix #3081)
 0.34: Avoid prompting when creating a new file (#3081)
+0.35: Handle loading without a settings file (default record setting)
--- a/apps/recorder/metadata.json
+++ b/apps/recorder/metadata.json
@@ -2,7 +2,7 @@
   "id": "recorder",
   "name": "Recorder",
   "shortName": "Recorder",
-  "version": "0.34",
+  "version": "0.35",
   "description": "Record GPS position, heart rate and more in the background, then download to your PC.",
   "icon": "app.png",
   "tags": "tool,outdoors,gps,widget,clkinfo",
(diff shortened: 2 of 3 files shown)


thyttan:
>Thanks! And yeah I see, if you check your watch's filesystem, does `recorder.json` either not exist, or if it does, does the JSON within it omit a `record` property? If so - I have a fix: #3100

_Originally posted by @bobrippling in https://github.com/espruino/BangleApps/issues/3098#issuecomment-1813325293_

No and no seems to be the answer:
```
{"recording":false,"period":2,"record":["bat","baro","bthrm","hrmint","bthrv","hrm","gps"],"file":"recorder.log20231115g.csv"}
```

I'll try your fix now anyway.

bobrippling:
Thanks for checking - I've looked at the code more closely and I believe the issue is that `settings` isn't actually defined/in scope (but we don't get a `ReferenceError` because it's likely another widget/boot code might've left the variable lying around in the global scope).

To confirm - did you see similar console output after the initial test of this PR?

And I've pushed up a change, how does it go now?

thyttan:
That seems to do it, thanks for working it out! :)

>To confirm - did you see similar console output after the initial test of this PR?

Yes.

thyttan:
I'm happy to merge if you want? :)

bobrippling:
Ah great - yes merge away :) And thanks for testing again!

PR #3101, opened 2023-11-16
messages_light: Fix js issues

Rarder44:
this change fixes a problem related to js.

until recently this code worked:
var a=[];
//add something to "a"
a.length = 0; // to remove all elements

now it doesn't work anymore.
(I replaced with splice)


bobrippling:
Nice fix - I'll merge this but i expect it's something we might want to fix upstream in espruino, @gfwilliams? If so I don't mind having a go at it

PR #3102, opened 2023-11-16
Recorder: 1 second period with 1 decimal place

thyttan:
This will make logs with 1 second periods record time with 1 decimal place precision.

EDIT3 - my first solution seemed to behave weirdly to me. But after sitting on it I realized it was probably just floating point number math doing what it does. The discussion that followed here no longer applies.

### DIFF
--- a/apps/recorder/widget.js
+++ b/apps/recorder/widget.js
@@ -176,11 +176,11 @@
   };
   let getCSVHeaders = activeRecorders => ["Time"].concat(activeRecorders.map(r=>r.fields));
 
-  let writeLog = function() {
+  let writeLog = function(period) {
     entriesWritten++;
     WIDGETS["recorder"].draw();
     try {
-      var fields = [Math.round(getTime())];
+      var fields = [period===1?getTime().toFixed(1):Math.round(getTime())];
       activeRecorders.forEach(recorder => fields.push.apply(fields,recorder.getValues()));
       if (storageFile) storageFile.write(fields.join(",")+"\n");
     } catch(e) {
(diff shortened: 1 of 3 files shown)


thyttan:
So I guess it was just floating point math doing its thing, and not wrong per se. To get nicer formatting in the csv file, I've converted to string. 

Seems to give good results now.

bobrippling:
Yeah that's floating point - there's no way to represent some numbers with that level of precision, and the nearest representable number might not be displayable in decimal (like how we can't write 1/3 in decimal)

So:
```javascript
Math.round(n * 10) / 10
```

won't necessarily give you a number with a single decimal place. If you want that, you could try [`n.toFixed(1)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed), which will give you exactly 1 decimal place (in a string), for popping into the CSV if you prefer

apps/recorder/widget.js
```diff
@@ -176,11 +176,17 @@
...
 
-  let writeLog = function() {
+  let writeLog = function(period) {
     entriesWritten++;
     WIDGETS["recorder"].draw();
     try {
-      var fields = [Math.round(getTime())];
+      var time = getTime();
+      if (period===1) {
+        time = Math.round(time)+"."+(Math.round(time*10))%10;
```
thyttan, line 185: @bobrippling do you think it's better to use [n.toFixed(1)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) here like you hinted?

bobrippling: Yeah I could see that being more consistent for giving a single decimal place and I believe it'll produce the same answer we're after

thyttan:
I tested the current code by recoring just the time and battery status with periods 1 second and 2 seconds for ca 10 seconds each. Seems to work as it should.

bobrippling:
Nice - those changes LGTM, shall see if Gordon has anything to mention though  👀

bobrippling:
I think let's merge - we can pick up other parts in #2819 if needed

PR #3103, opened 2023-11-17
Sunrise 0.07: account for altitude, use suncalc module

bobrippling:
This continues the great job done by @g-rden in #3063, bringing in those changes but adjusting for the sinLUT changes mentioned in the previous PR.

PR #3104, opened 2023-11-18
Recorder: Period 1s + Log GPS => record on GPS event

thyttan:
This is a follow up to #3102 trying to implement what's suggested by @gfwilliams in https://github.com/espruino/BangleApps/pull/2819#issuecomment-1809812485: 
>Just a thought: ideally for 1 second we'd just output data as soon as we got the GPS reading (assuming we were supposed to log GPS - we might not be). That would help to stop us getting the blank GPS entries and would make sure we didn't miss any GPS readings too. It is possible to configure the GPS to give data at a higher rate, but I think that's fine - it's rare anyone would do that, and if they did they might even like being able to record all the data they get

Initial testing shows promise. Sitting indoors certainly not having a fix my watch is still getting GPS events prompting the recording of data.


thyttan:
oops, that's a lot of commits. Switched from wrong branch...

apps/recorder/widget.js
```diff
@@ -239,7 +247,7 @@
     reload();
     Bangle.drawWidgets(); // relayout all widgets
   },isRecording:function() {
-    return !!writeInterval;
+    return isRecording;
```
thyttan, line 250: I'm a bit unsure about my change to introducing and using the variable `isRecording`.

apps/recorder/widget.js
```diff
@@ -199,6 +200,9 @@
     var settings = loadSettings();
     if (writeInterval) clearInterval(writeInterval);
     writeInterval = undefined;
+    function writeOnGPS() {writeLog(settings.period)}
```
thyttan, line 203: unsure if this is the best place to define `writeOnGPS()`.

apps/recorder/widget.js
bobrippling: I suppose we could repurpose `writeInterval`, so it's either a number (the interval) or `true` in which case we know we're using gps as the driver for the writes. We could call it `writeSetup` and we'd just need to make sure we handle the three cases in each place it's used (`number | true | undefined`)

Possibly controversial, but typescript would help if we wanted to really go whole hog

apps/recorder/widget.js
bobrippling: Yes - the `removeListener` won't work because we'll have just created a new `writeOnGPS` when entering the `reload` function. What we'll want is to have it defined outside of `reload`, so it's captured once.

Something a bit like
```diff
+  function writeOnGPS() {writeLog(settings.period)}
+
   // Called by the GPS app to reload settings and decide what to do
   let reload = function() {
     var settings = loadSettings();
     if (writeInterval) clearInterval(writeInterval);
     writeInterval = undefined;
-    function writeOnGPS() {writeLog(settings.period)}
```

apps/recorder/widget.js
```diff
@@ -176,13 +176,16 @@
...
   let getCSVHeaders = activeRecorders => ["Time"].concat(activeRecorders.map(r=>r.fields));
 
-  let writeLog = function() {
+  let writeLog = function(period, fix) {
     entriesWritten++;
     WIDGETS["recorder"].draw();
     try {
-      var fields = [Math.round(getTime())];
+      console.log("\nfix lat:         " + fix.lat + "\nfix time (ms):   " + fix.time + "\ncpu time (s):    " + getTime());
+      var time = (fix.time&&fix.lat)?fix.time/1000:getTime();
```
thyttan, line 184: I'm deliberating on whether to use the GPS fix time for the log entries. 

Here the GPS time is used if the fix also contain coordinates. Otherwise the Bangle system time is used. The effect of this is:
- Before a fix with GPS coordinates is found the system time is used.
- If the GPS is no longer able to determine coordinates, the log will go back to using the system time.

If the difference between GPS time and Bangle system time is large enough this can give the result that a log entry gets a reported time that is older than previous log entries.

I believe an advantage of using GPS fix time is it's reporting becomes independent of the variable time before the fix is acted on by the recorder widget. This should make analytics (speed, etc.) more accurate. But it comes with some kinks as described above, and probably some other consequence I'm not thinking of.

apps/recorder/widget.js
thyttan: I did the change to `writeSetup`. Thanks :+1: 

As for typescript, I don't know it so don't have anything to say about it :P

apps/recorder/widget.js
thyttan: Thanks!

thyttan:
Sometimes getting a bug when starting a recording with current changes: 
```
>
Recorder for "bthrm"+not found
Recorder for "hrmint"+not found
Recorder for "bthrv"+not found
Uncaught undefined
in function "start" called from line 227 col 30 in recorder.wid.js
activeRecorder.start();
                     ^
in function called from line 228 col 8 in recorder.wid.js
      });
       ^
in function "reload" called from line 260 col 12 in recorder.wid.js
reload();
       ^
in function "reload" called from line 19 col 57 in recorder.wid.js
    if (WIDGETS["recorder"]) WIDGETS["recorder"].reload();
                                                        ^
in function "updateSettings" called from line 313 col 28 in recorder.wid.js
    updateSettings(settings);
                           ^
in function "setRecording" called from line 11 col 368 in recorder.app.js
...["recorder"].setRecording(v).then(function(){loadSettings();showMai...
                                     ^
>
```

Remember to think about this before possible merge of the PR.

apps/recorder/widget.js
nxdefiant: I would prefer a monotonic clock for recording, but since the bangle does not have one the system time is the next best thing. Also I if using e.g. [gpsautotime](https://github.com/espruino/BangleApps/tree/master/apps/gpsautotime) both time sources should have the same value.

>  believe an advantage of using GPS fix time is it's reporting becomes independent of the variable time before the fix is acted on by the recorder widget. This should make analytics (speed, etc.) more accurate,

I dcan't follow here. Yes the GPS time is very accurate but I don't think the system time is that inaccurate that it would make a difference.

Also: if in doubt, why not log both times?

thyttan: Yes, I'm not really sure. It's likely you have a better understanding of this.

My current assumption is GPS time is initially based on system time until it starts getting coordinates, at which time I'd thought the time would be given by the satellites. But I have not verified this.

nxdefiant: System time is not needed since their time is the main thing that the satellites are sending.

thyttan: But until the GPS receiver has contact with satellites it seemed to be using the Bangle system time rounded to nearest second, when I did some testing last night. But again, not really verified/falsified.

bobrippling:
> Sometimes getting a bug when starting a recording with current changes:
> ```
> Uncaught undefined
> in function "start" called from line 227 col 30 in recorder.wid.js
> activeRecorder.start();
>                      ^
> ```

Interesting - this is the code that's throwing:
https://github.com/espruino/BangleApps/blob/bf3809970ca18ed0085a80de41b7c865f9fed486/apps/recorder/widget.js#L215-L218

Specifically the `.start()`. Do you have any recorders that might not have a `start` function? (`require("Storage").list(/^.*\.recorder\.js$/)`)

Although I wouldn't expect `undefined` to be thrown from that. If you add some code before the start, we could see if we can narrow it down, maybe something like:
```diff
 activeRecorders.forEach(activeRecorder => {
+  try{
   activeRecorder.start();
+  }catch(e){
+    console.log("<" + activeRecorder.name + ">.start(), caught", e);
+  }
 });
```

apps/recorder/widget.js
bobrippling: That's fair! :D

apps/recorder/widget.js
```diff
@@ -194,11 +197,14 @@
...
   }
 
+  let writeOnGPS = function(fix) {writeLog(loadSettings().period, fix); }
+
   // Called by the GPS app to reload settings and decide what to do
   let reload = function() {
     var settings = loadSettings();
-    if (writeInterval) clearInterval(writeInterval);
-    writeInterval = undefined;
+    if (writeSetup && !writeSetup===true) clearInterval(writeSetup);
```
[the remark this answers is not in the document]
thyttan: Actually, now the second condition is not needed again like before, I think.

apps/recorder/widget.js
gfwilliams: > GPS time is initially based on system time until it starts getting coordinates

It should report the time the GPS receiver has, which is actually different to the Bangle's time. I think if the battery was totally flat and you restarted you'd find you didn't get any GPS time.

Personally, I think we should just use Bangle.js time - there's really not going to be any difference of more than a few seconds, and the downside of having the GPS time is it could just completely change when you get a fix and that'll throw off everything.

If you think it's something you really need then as @nxdefiant says you could just make a new app with an `app.recorder.js` file which added a new column for GPS time. But we definitely shouldn't add it by default, as it'll just use up even more storage space for recordings.

thyttan: Thanks!

apps/recorder/widget.js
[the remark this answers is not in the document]
bobrippling: The `writeSetup !== true` part? How so?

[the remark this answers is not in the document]
thyttan: Never mind! :) It now makes sense to me for it to be there.

apps/recorder/widget.js
thyttan: I dropped the commit/changes for using GPS time.

thyttan: I dropped the commit/changes for using the GPS time. Thanks for your feedback! :)

thyttan:
>If you add some code before the start, we could see if we can narrow it down, maybe something like:
> 
> ```diff
>  activeRecorders.forEach(activeRecorder => {
> +  try{
>    activeRecorder.start();
> +  }catch(e){
> +    console.log("<" + activeRecorder.name + ">.start(), caught", e);
> +  }
>  });
> ```
https://github.com/espruino/BangleApps/pull/3104#issuecomment-1817992368

I've tried to trigger this again with the current shape of the PR, and I don't get the error anymore it seems.

So I'll mark this as ready for review now.

bobrippling:
> I've tried to trigger this again with the current shape of the PR (including the above try catch statement), and I don't get the error anymore it seems.
> 
> So I'll mark this as ready for review now.
> 
> EDIT: But testing the code reset to not include the try catch statement just now, I got the `Uncaught undefined` almost immediately. Hmm... I'll keep this as draft.

Interesting, I wonder if this is unrelated to this specific PR/app, what with you seeing it on the [sunrise PR](https://github.com/espruino/BangleApps/pull/3103#discussion_r1408409559) too. Perhaps we leave in an empty `try { ... } catch (e) {}` block to stop the exception propagating, and look into it as part of another issue?

thyttan:
So now I've updated to fw 2v19.86 and run `Bangle.factoryReset()` from the Web IDE. Then updated to latest version of default apps and installed only ´recorder´ in addition.

And I hit it now again:
```

Uncaught undefined
in function "start" called from line 213 col 30 in recorder.wid.js
activeRecorder.start();
                     ^
in function called from line 214 col 8 in recorder.wid.js
      });
       ^
in function "reload" called from line 246 col 12 in recorder.wid.js
reload();
       ^
in function "reload" called from line 300 col 32 in recorder.wid.js
    WIDGETS["recorder"].reload();
                               ^
in function "setRecording" called from line 277 col 103 in recorder.wid.js
...rding(1,{force:"overwrite"});
                              ^
> 
```

It didn't happen the first few tries, but then it did.

I'll reintroduce your try catch statement to see if I'm more lucky with catching it again.

thyttan:
I also just caught it with v0.36 that's already on the development app loader:

```
Uncaught undefined
in function "start" called from line 211 col 30 in recorder.wid.js
activeRecorder.start();
                     ^
in function called from line 212 col 8 in recorder.wid.js
      });
       ^
in function "reload" called from line 239 col 12 in recorder.wid.js
reload();
       ^
in function "reload" called from line 19 col 57 in recorder.wid.js
    if (WIDGETS["recorder"]) WIDGETS["recorder"].reload();
                                                        ^
in function "updateSettings" called from line 292 col 28 in recorder.wid.js
    updateSettings(settings);
                           ^
in function "setRecording" called from line 270 col 103 in recorder.wid.js
...rding(1,{force:"overwrite"});
                              ^
> 
```

thyttan:
Got one! 

```
<Baro>.start(), caught undefined
```

apps/recorder/widget.js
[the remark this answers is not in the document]
thyttan: @bobrippling would it be better to do `if (typeof writeSetup === "number") clearInterval(writeSetup);`? Probably clearer right?

gfwilliams:
Looks good! Ready to merge now? I guess it's possible there is an exception regression on cutting edge Espruino firmware itself?

I did find an error and fix it on wednesday (4b61b98b81926688ddc0afc6f51dd2b6d9ac54ae) - as far as I can tell it looks ok now, but maybe there is something else?

thyttan:
> Ready to merge now?

Essentially yes! I'll just tag you on a new review comment in a minute and after that we can merge 🙂

bobrippling:
> Got one!
> 
> ```
> <Baro>.start(), caught undefined
> ```

Nice! 🎣 
Am I right in understanding that's after [the fix (`4b61b98b819`)](https://github.com/espruino/espruino/commit/4b61b98b81926688ddc0afc6f51dd2b6d9ac54ae)?

I wonder - [`Baro` / `getPressure`](https://github.com/bobrippling/BangleApps/blob/2a182375cb87e7577364411e2d921f3a2e74caba/apps/recorder/widget.js#L116-L143) can raise exceptions from time to time if it's in the wrong state when called, perhaps this is how it's triggered.

We could try to confirm this by upping the amount of times we call `Bangle.getPressure()`, perhaps if you kick off something like this from the IDE when your recorder's active, it might aggravate the bug and cause it to appear more:

```javascript
setInterval(() => Bangle.getPressure().catch(()=>{}), 1000) // or a shorter interval, even
```

thyttan:
@bobrippling I did the change to use typeof. Please merge ahead!

thyttan:
Not sure if it's what we're looking for, but I got this now:
```
>setInterval(() => Bangle.getPressure().catch(e=>console.log("aggravater caught:", e)), 1000) // or a shorter interval, even
=20
Uncaught Error: Conversion in progress
 at line 1 col 20
Bangle.getPressure().catch(e=>console.log("aggravater caught:",e)...
                        ^
> 
```

thyttan:
I modified the intervaled function and got this:

```
>setInterval(()=>{try {Bangle.getPressure()} catch (e) {console.log("aggrevator caught:", e)}}, 1000)
=3
aggrevator caught: undefined
Uncaught Error: Unhandled promise rejection: undefined
>

**Comment by thyttan** [comment]
If I shorten the interval:

```
>setInterval(()=>{try {Bangle.getPressure()} catch (e) {console.log("aggrevator caught:", e)}}, 100)
=3
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
aggrevator caught: Error: Error: Conversion in progress
```

**Comment by bobrippling** [comment]
### Pull Request

Awesome! Merged - thank you.

### Bug

It looks like we have two scenarios here:

The expected error from spamming the pressure API (the majority of the exceptions):
```
Uncaught Error: Conversion in progress
```

and the mysterious bug (caught twice above):
```
aggrevator caught: undefined
Uncaught Error: Unhandled promise rejection: undefined
```

From the first line, we know it's definitely a promise rejection from `getPressure()`.

I wonder if it's a coincidence that we see the exception a bit more (for example, in your first response). Either way, I believe the exception raised from `Bangle.getPressure()` may occasionally have an `undefined` message.

Our next step is to have a look at the firmware code - [this part seems relevant](https://github.com/espruino/Espruino/blob/67493a7d95c316e144eb4dbb9600fa014329503f/libs/banglejs/jswrap_bangle.c#L5094-L5105=), perhaps we have had an error but no exception? Shall dig into it more 🔍

**Comment by bobrippling** [comment]
Great - I've pushed up [a PR](https://github.com/espruino/Espruino/pull/2438) that handles the case where we don't have an exception but we have an error (and we presume it's I2C)

**Comment by bobrippling** [comment]
@thyttan the firmware should now give us a non-`undefined` exception. When you're on the next firmware and see the error, we should now be able to confirm that it's an I2C error communicating with the barometer.
I'm afraid that's as far as I know, I can't say where to go after that!

**Comment by thyttan** [comment]
> When you're on the next firmware and see the error, we should now be able to confirm that it's an I2C error communicating with the barometer.

```
>setInterval(() => Bangle.getPressure().catch(e=>console.log("aggravater caught:", e)), 500)
=6
Uncaught Error: Unhandled promise rejection: InternalError: I2C barometer error
> 
```
Confirmed!


Bonus error:
```
Uncaught undefined
in function "start" called from line 214 col 30 in recorder.wid.js
activeRecorder.start();
                     ^
in function called from line 215 col 8 in recorder.wid.js
      });
       ^
in function "reload" called from line 247 col 12 in recorder.wid.js
reload();
       ^
in function "reload" called from line 19 col 57 in recorder.wid.js
    if (WIDGETS["recorder"]) WIDGETS["recorder"].reload();
                                                        ^
in function "updateSettings" called from line 300 col 28 in recorder.wid.js
    updateSettings(settings);
                           ^
in function "setRecording" called from line 278 col 103 in recorder.wid.js
...rding(1,{force:"overwrite"});
                              ^
```

bobrippling:
Great to see! So it sounds like `getPressure()` needs a `try`/`catch` if we see this still, but otherwise I think we're sorted

PR #3105, opened 2023-11-18
sanitycheck.js: detect apps that may mistakenly have an app.js file

bobrippling:
This will help us catch more incorrect `type` settings in metadata


gfwilliams:
Looks good - thanks!

PR #3106, opened 2023-11-19
health/interace: Replace moving average for bpm with avg

nxdefiant:
Fixes also average bpm on 0 value

When hrm interval in health is off bpm data is only available when another app activates the heart rate monitor, for all other records bpm value is 0.

### DIFF
--- a/apps/health/interface.html
+++ b/apps/health/interface.html
@@ -139,6 +139,7 @@ function getDailyData(data) {
   var idx = DB_HEADER_LEN;
   for (var day = 0; day < 31; day++) {
     var dayData = {steps: 0, bpm: 0, movement: 0};
+    let bpmCnt = 0;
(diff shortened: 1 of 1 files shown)


bobrippling:
Nice spot!

@gfwilliams @thyttan thought I'd ask for a second review with this being a core app (for my part, I can't think of any potential problems and if anything @nxdefiant's fix is a very good-to-have bugfix)

nxdefiant:
After sleeping a night about this I will probably change the moving average to average. Just checked the app and [it seem to use an average.](https://github.com/espruino/BangleApps/blob/master/apps/health/app.js#L79)

nxdefiant:
replaced moving avg with avg

gfwilliams:
This looks great, thanks! Yes, that really would have knocked off any bpm readings.

I guess I probably never noticed since when testing I'd turned the heart rate sensor on in the settings

PR #3107, opened 2023-11-19
altimeter: use system variables to calibrate altimeter.

pavelmachek:
Use system variables to calibrate altimeter. That means that calibration works for other applications, too, and persists between altimeter starts.

Immediately show altitude after calibration changes, otherwise it is hard to set the calibration as the actual values lag behind reality.


bobrippling:
Cool, thanks!

PR #3110, opened 2023-11-23
Added Utils.readStorageJSON with relaxed JSON parser

gfwilliams:
Added Utils.readStorageJSON and relaxed JSON parser, and swap all interface.html over to using it.

This allows Bangle.js 2v20 (or cutting edge) and later to store a relaxed 'JSON' on internal storage which (while still normal JS) is smaller and faster (and preserves unicode better)

Fixes https://github.com/espruino/Espruino/issues/2429

It'd be great if you could take a look over this as I had to change a bunch of files.  I've tried some (eg openstmap) and they work ok, but it may well be I broke something :(

### DIFF
--- a/apps/openstmap/interface.html
+++ b/apps/openstmap/interface.html
@@ -151,7 +151,7 @@ TODO:
             if (filename=="") return;
             promise = promise.then(() => new Promise(resolve => {
               Util.readStorageJSON(filename, mapInfo => {
-                console.log(filename + " => " + mapInfo);
+                console.log(filename + " => " + JSON.stringify(mapInfo));
                 let mapNumber = filename.match(/\d+/)[0]; // figure out what map number we are
                 loadedMaps[mapNumber] = mapInfo;
                 if (mapInfo!==undefined) {
--- a/apps/openstmap/interface.html
+++ b/apps/openstmap/interface.html
@@ -150,15 +150,9 @@ TODO:
           files.trim().split(",").forEach(filename => {
             if (filename=="") return;
             promise = promise.then(() => new Promise(resolve => {
-              Util.readStorage(filename, fileContents => {
-                console.log(filename + " => " + fileContents);
+              Util.readStorageJSON(filename, mapInfo => {
+                console.log(filename + " => " + mapInfo);
                 let mapNumber = filename.match(/\d+/)[0]; // figure out what map number we are
-                let mapInfo;
-                try {
-                  mapInfo = JSON.parse(fileContents);
-                } catch (e) {
-                  console.error(e);
-                }
                 loadedMaps[mapNumber] = mapInfo;
                 if (mapInfo!==undefined) {
                   let latlon = L.latLng(mapInfo.lat, mapInfo.lon);
--- a/apps/owmweather/interface.html
+++ b/apps/owmweather/interface.html
@@ -2,35 +2,34 @@
   <head>
     <link rel="stylesheet" href="../../css/spectre.min.css">
   </head>
-  <body>   
+  <body>
     <h3>Set OpenWeatherMap (OWM) API key</h3>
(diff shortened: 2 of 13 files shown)


apps/openstmap/interface.html
```diff
@@ -150,15 +150,9 @@
           files.trim().split(",").forEach(filename => {
             if (filename=="") return;
             promise = promise.then(() => new Promise(resolve => {
-              Util.readStorage(filename, fileContents => {
-                console.log(filename + " => " + fileContents);
+              Util.readStorageJSON(filename, mapInfo => {
+                console.log(filename + " => " + mapInfo);
```
bobrippling, line 154: I appreciate this is just a console log, but to preserve the original behaviour, perhaps something like:
```suggestion
                console.log(filename + " => " + JSON.stringify(mapInfo));
```

nxdefiant:
tested read good on calendar, grocery, noteify, openstmap, sleepphasealarm, mylocation, also with special character "\xFC" working

apps/owmweather/interface.html
```diff
@@ -2,35 +2,34 @@
...
+
         var settings = {};
         function onInit(){
           console.log("Loading settings from BangleJs...");
           try {
-            Util.readStorage("owmweather.json", data=>{
-              if(data.length > 0){
-                settings = JSON.parse(data);
+            Util.readStorageJSON("owmweather.json", settings=>{
+              if("settings){
```
bobrippling, line 32: Minor typo
```suggestion
              if(settings){
```

apps/owmweather/interface.html
```diff
@@ -2,35 +2,34 @@
...
-    
+
         var settings = {};
         function onInit(){
           console.log("Loading settings from BangleJs...");
           try {
-            Util.readStorage("owmweather.json", data=>{
-              if(data.length > 0){
-                settings = JSON.parse(data);
+            Util.readStorageJSON("owmweather.json", settings=>{
```
bobrippling, line 31: I think we might have a behaviour change here to check - previously if the file didn't exist, we'd get called back with `""`, but now we'll end up trying `parseRJSON("")`.

Having said that, while `JSON.parse()` will throw on that, looking at [the code](https://github.com/espruino/EspruinoAppLoaderCore/blob/c97b7851f50cfff4e898c2264a337a17085ce463/js/utils.js#L405-L453) I think we might be ok, we only `match(EOF)` if we have a token, otherwise we just return `undefined`

Short on time atm but can give this a test later on

(same for `tinycmc`'s `interface.html` too)

apps/openstmap/interface.html
gfwilliams: Thanks!

apps/owmweather/interface.html
gfwilliams: Ouch - thanks! I wonder if we can lint the JS in HTML files somehow?

apps/owmweather/interface.html
gfwilliams: Thanks - I just tried and:

```
Utils.parseRJSON("")
=undefined
```

So I think we're fine there, *but* I really screwed this change up anyway as `settings` was a global var, which I should have updated with a `if (data) settings=data`. Hopefully I fixed that now!

apps/owmweather/interface.html
bobrippling: Good idea - I suppose we could add a check that does something like this? (except a little more refined)

```sh
awk '/<script>/, /<\/script>/' < html | sed '1d; $d' | node -c
```

apps/owmweather/interface.html
bobrippling: Great, on both counts there! Yeah I think we're good to go in that case

bobrippling (approved):
Looks grand, shall leave to you to merge @gfwilliams in case you're wanting to time it with any other changes

gfwilliams:
Thanks! I won't push this to banglejs.com/apps until next week just in case ;)

PR #3115, opened 2023-11-29
Presentor V0.08 -> V0.14

7kasper:
Finally I got back to using Bangle in school so that warrented me putting time into improvements in the presentor app.
It now works with windows 11 again also!

This PR fixes some issues and makes the experience a lot better. The app is still in beta I guess but it works well enough for me most of the time 😄

### DIFF
--- a/apps/presentor/ChangeLog
+++ b/apps/presentor/ChangeLog
@@ -11,3 +11,4 @@
 0.11: Fix mouse move getting stuck.
 0.12: Added support for mouse dragging action (click then drag).
 0.13: Removed mouse dragging (too buggy/unuseful).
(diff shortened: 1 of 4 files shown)


pavelmachek:
I guess it would be normal to advance by one version when the subversions were not in mainline?

gfwilliams:
This looks great!

> I guess it would be normal to advance by one version when the subversions were not in mainline?

I think if you're not making your changes public then increasing just by one is preferable but honestly it doesn't really matter.

If you were making the app available on your personal app loader then there's a really good reason for bumping the version, so other users who might want to use the absolute latest from your app loader can see when there are new versions.

PR #3117, opened 2023-11-30
gipy: new release

wagnerf42:
hi, this is a new release for gipy with a new renderer.
everything is now way more reactive including inside large cities.


bobrippling:
Looks great, thanks!

PR #3120, opened 2023-12-03
health: calculate distance from steps & introduce myprofile

nxdefiant:
- Adds myprofile to set various personal settings

Health:
- new submenu distance: multiplies step counts with stride length
- distance menu under step counting is only available if stride length is
    set in new app myprofile

### DIFF
--- a/apps/myprofile/README.md
+++ b/apps/myprofile/README.md
@@ -9,6 +9,7 @@ Configure your personal profile. All settings are optional and are only stored o
 | Birthday      | Used to calculate age           | year, month, day    | 'YYYY-MM-DD' | 01.01.1970    | -                                                                 |
 | HR max        | maximum heart rate              | BPM                 | BPM          | 60            | Use maximum value when exercising.<br/> If unsure set to 220-age. |
 | HR min        | minimum heart rate              | BPM                 | BPM          | 200           | Measure your heart rate after waking up                           |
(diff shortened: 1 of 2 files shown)


pavelmachek:
Should we make step length global setting? At least "pebble clock with distance" and "sixths" depend on that value already.

nxdefiant:
ok, I was not aware of that, this brings back [the "My Profile" idea](https://forum.espruino.com/conversations/391660/#comment17193275).

@hughbarney How far are you with this?

nxdefiant:
> [thyttan](https://github.com/thyttan) marked this pull request as ready for review [1 hour ago](https://github.com/espruino/BangleApps/pull/3120#event-11143151206)

uhm why?

thyttan:
Sorry, don't know how that happened!

nxdefiant:
I have added an initial version of "myprofile" to store personal data and use that instead.

nxdefiant:
I'm done for now.

nxdefiant:
Since "health" is installed by default I assume @gfwilliams  wants to review this personally.

gfwilliams:
This looks great! Thanks for your work on this!

I guess at some point in the future we could have an `interface.html` so it can be set from the app loader too, but this seems great for now.

... also moving `step goal` into this app would be great! It's currently in the health app (but also some pedometers) I think and would be really good to move to one place.

Merging now...

PR #3130, opened 2023-12-14
chess: Use Bangle.setBacklight()

nxdefiant:
instead of Bangle.setLCDTimeout() to turn off the LCD while the computer is thinking.


bobrippling:
Nice - thanks for the `finished` bugfix too

PR #3131, opened 2023-12-15
widhrzone: Read maximum HRM from myprofile

nxdefiant:
drop own settings


apps/widhrzone/widget.js
```diff
@@ -1,20 +1,18 @@
 (() => {
-  const config = Object.assign({
-    maxHrm: 200,
-  }, require("Storage").readJSON("widhrzone.json",1) || {});
+  const myprofile = require("Storage").readJSON("myprofile.json",1)||{};
```
bobrippling, line 2: Looks good - just this: do we need to cover the case where the user doesn't have `myprofile` installed or can we assume it's always present because it's now a dependency? (in which case, do we want to drop the `1` and the `||{}` ?)

nxdefiant: Because of the dependency it should be present. But dependencies can be uninstalled. In this case myprofile.maxHrm will be undefined and the comparisions in calczone() will never resolve to true which is ok for me.

bobrippling: Cool, that's fair - thanks

PR #3133, opened 2023-12-18
Update fr_FR.json

npalix:


bobrippling:
Thank you!

(for anyone following, see also #3134)

PR #3136, opened 2023-12-20
recorder: resume logs recorded within 4 hours

### LINKED ISSUE #3135: [Recorder] Day in filename is often wrong
### Affected hardware version

Bangle 2

### Your firmware version

2v19

### The bug

1. Open recorder
2. Start recording
3. Stop recording after a few data points have been recorded
4. view the file that was created

From my experience, about 50% of the recording files have a name that is a day later than the date it was recorded on.

My timezone on watch is set to GMT-8. This is likely the cause of the issue. 

image

### Installed apps

widlock (0.08), sched (0.22), sched (0.22), widalarm (0.01), messageicons (0.05), boot (0.59), notify (0.13), clock_info (0.07), gpsinfo (0.12), bikespeedo (0.06), gpsnav (0.08), dtlaunch (0.24), bwclk (0.33), widmsggrid (0.05), widminbt (0.01), widgps (0.09), slides (undefined), magnav (0.06), quicklaunch (0.15), runplus (0.22), kbswipe (0.09), clkinfostopw (0.03), mylocation (0.10), clkinfosunrise (0.03), configurable_clock (0.03), locale (0.17), sleeplog (0.15), widclk (0.08), gpsmagcourse (0.01), waypoint_editor (0.03), rebbleagenda (0.01), weather (0.26), clkinfocal (0.02), ios (0.15), waypoints (0.03), setting (0.68), messagegui (0.75), messages (0.62), fastload (0.06), fastreset (0.03), flashcount (0.01), widbatv (0.03), alarm (0.46), health (0.28), openstmap (0.27), recorder (0.37), myprofile (0.01)

bobrippling:
We now look at the latest log's last timestamp - if it was within four hours, we prompt the user to resume recording to that file.

This change also adds (manually runnable) tests for `setRecording()`

Fixes #3135


apps/recorder/ChangeLog
```diff
@@ -45,3 +45,4 @@
 0.36: When recording with 1 second periods, log time with one decimal.
 0.37: 1 second periods + gps log => log when gps event is received, not with
 	setInterval.
+0.38: Resume previous recorder logs, if within four hours - regardless of date
```
Mineinjava, line 48: Might want to clarify that the previous recorder log is only resumed on append

Mineinjava (commented):
Should it be clarified that the file name is in UTC not local time?
This unlikely to affect people not living in GMT+-1

Additionally, I feel like forcing the recorder to append should always append to the latest file, regardless of the time difference

bobrippling:
> Should it be clarified that the file name is in UTC not local time? This unlikely to affect people not living in GMT+-1

Very good point - I think that since we now look at the times within the file, the filename is less important (so long as it still sorts by date), so we could make it that the filename is in localtime (but keep the timestamps themselves in UTC), so when you look at the files they're in your timezone.

---

> Additionally, I feel like forcing the recorder to append should always append to the latest file, regardless of the time difference

That makes sense - how does this sound as a proposed flow? Any tweaks/additions?

```
on recording start:

if any previous recorder log (containing data):
  if within 4 hours:
    prompt the user: append, new, overwrite
  else:
    prompt the user: append, new

"new" will create a filename based on the current date (in localtime)

(the prompt will also have the existing cancel option)
```

(I'll leave this PR open here - feel free to add changes on if you like, I'm not sure when I'll next be free to pick it up)

apps/recorder/widget.js
```diff
@@ -256,10 +256,36 @@
...
-        settings.file = getTrackFilename();
+      function nextTrackFilename() { return "recorder.log" + date + trackNo.toString(36) + ".csv"; }
+      function mostRecentFile() {
+        const logs = require("Storage")
+          .list(/^recorder\.log/)
+          .map(x => x.replace(/^recorder\.log/, ""));
+
+        if(logs.length > 0){
+          logs.sort();
+          const latest = "recorder.log" + logs[logs.length-1].replace("\1", "");
```
nxdefiant, line 267: Not sure why: above you remove recorder.log and here you add it again

bobrippling: Ah duh - yes, I was thinking it'd impact the sort above, but it doesn't matter, shall fix

apps/recorder/widget.js
```diff
@@ -256,10 +256,34 @@
...
+        const logs = require("Storage").list(/^recorder\.log/);
+
+        if(logs.length > 0){
+          logs.sort();
+          const latest = logs[logs.length-1].replace("\1", "");
+          const f = require("Storage").open(latest, "r");
+          const timeIdx = f.readLine().replace("\n", "").split(",").indexOf("Time");
+          if(timeIdx >= 0){
+            let last, tmp;
+            while(tmp = f.readLine())
```
gfwilliams, line 270: there's a lint warning here - I think you can just use`while((tmp = f.readLine()))` to fix it

gfwilliams:
I like the idea, but have you actually tested this in practise? Because iterating over the entire file to find the last line could take quite a long time (10+ seconds depending on the track length) and I think that would provide quite a poor experience if you user just wants to start a run.

At minimum it should at least check the filename date and if it's different, skip the check.

... plus this feels like quite a lot of extra complexity? I'm not sure really sure if it's worth it?

bobrippling:
Good point - @Mineinjava is this an issue you still see?

Either way I'd like to keep the tests, I'll split out to another PR if necessary (and if you agree the tests are good to have)

Mineinjava:
> Good point - @Mineinjava is this an issue you still see?
> 
> Either way I'd like to keep the tests (and fix the [lint warnings](https://github.com/espruino/BangleApps/actions/runs/8307240476/job/22736224822?pr=3136)), I'll split out to another PR if necessary (and if you agree the tests are good to have)

Honestly I think the best solution is to 
- tell the user the name of the last log
- if they say "append," append regardless of timestamps

bobrippling:
Yeah that seems reasonable, I'll make some progress towards that as part of this PR

bobrippling:
This has been subsumed by #3431, closing

PR #3139, opened 2023-12-26
spotrem: further refactoring to shorten the code somewhat

thyttan:
Also fixes search and play feature that was broken on last version of spotrem.


bobrippling:
Nice, looking better every time!

PR #3141, opened 2023-12-30
added hrm sport mode setting

nxdefiant:


apps/sportmode/settings.js
```diff
@@ -0,0 +1,54 @@
...
+  E.showMenu({
+    "" : { "title" : /*LANG*/"HRM sport mode" },
+    "< Back" : () => back(),
+    /*LANG*/'Sport mode': {
+      value: settings.mode,
+      min: -1, max: SPORT_MODES.length-1,
+      format: v => v === -1 ? /*LANG*/"Auto" : SPORT_MODES[v],
+      onchange: v => {
+        settings.mode = v;
+        writeSettings();
```
bobrippling, line 50: If a user's got fastload enabled, this won't take effect until a full reload of the watch - perhaps we should set `hrmSportMode` here too?

nxdefiant: Actually leaving settings does a reload

bobrippling: So it does, nice!

nxdefiant: But I'm thinking of adding this line anyway. What if we want to open this setting from e.g. heart rate monitor.

bobrippling: Yeah that's a fair point, nice catch with the remove handler too

nxdefiant:
should be ok now

bobrippling:
Cool, thanks!

Issue #795, opened 2021-09-01
Recipe app?


gfwilliams:
Looks like there's a free recipe database at: https://www.themealdb.com/api.php

It'd be great to be able to store recipes on Bangle.js such that you could just step through each step one by one. Ideally with the quantities included for each step, but this doesn't seem to be handled by the API

bobrippling:
I reckon we can close this issue off - there's now a newly introduced recipe app (#795) with the potential to iterate further

gfwilliams:
Sounds great - thanks! Link to app at https://espruino.github.io/BangleApps/?id=followtherecipe

PR #3143, opened 2024-01-02
Create a recipe app mentioned in issue #795

Mel-Levesque:
As mentioned in the [issue #795](https://github.com/espruino/BangleApps/issues/795#issue-985237986), the app use [the meal db api](https://www.themealdb.com/api.php).
It displays the category and origin of the recipe, then the ingredients, and concludes with the instructions.
The app use Gadgetbridge to get a random recipe on application start.
But the more common usage is to choose a recipe from the App Loader and save it into the BangleJs. 

I hope you enjoy using it !


bobrippling:
Great stuff, just gave it a try through the app loader/`interface.html`, will be using this for my next meal adventure

PR #3144, opened 2024-01-02
lcars 0.31: Fix Battery vref due to nRF52840: jsvReadVRef now reads V…

d3nd3:
[E.getAnalogVRef](https://www.espruino.com/Reference#t_l_E_getAnalogVRef) no longer returns a mostly fixed 3.3. Thus breaking the calculation for battery volts.  Using 3.3 instead, works.

### DIFF
--- a/apps/lcars/ChangeLog
+++ b/apps/lcars/ChangeLog
@@ -28,3 +28,4 @@
 0.28: Battery Vref implemented correctly.
 0.29: Support fastload.
 0.30: Add many new colors to the settings and allows random colors on startup if enabled.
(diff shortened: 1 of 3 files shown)


bobrippling:
Thanks for investigating! @gfwilliams do you know if this is an intentional change for [`E.getAnalogVRef`](https://www.espruino.com/Reference#t_l_E_getAnalogVRef)?

gfwilliams:
I was intentional - there's a discussion on https://github.com/espruino/Espruino/commit/eabd13cfb8ce1c0a81708bb6e9f37f15b101a5fd#commitcomment-136080837

But on Bangle.js 2 VDDH isn't used so I have now reverted that change - so I think it's best not to merge this in.

PR #3147, opened 2024-01-04
Two app updates plus one new application

pglez82:
Hi,

First of all, sorry if commits from different applications should not be together. 

The two updates are very basic, the involve these two banglejs 1 apps (which I created):

- Intervals: changed the configuration to allow intervals with a resolution of 1s instead of 5s.
- Planeatarium: Added two new constellations.

The new application is for banglejs 2. It is called quoteclock. Basically, it's a clock that shows nice literary quotes every hour.

Best regards,
Pablo


bobrippling:
Thanks - I think in this case one PR is fine for the three apps.

Just one request, would you mind moving quoteclock's `setUI` call above its `Bangle.loadWidgets` call? This'll handle the warning we're getting around there ("_Clock quoteclock file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)_")

pglez82:
Sorry, I forgot about this completely. I have moved the call upwards and also formatted the code (sorry for posting unformatted code in my initial PR).

bobrippling:
No worries - thanks for the changes :)

PR #3148, opened 2024-01-06
New App: Line clock

deepDiverPaul:
Analog clock with improved readability.

app-screenshot


thyttan:
Looks nice, I will be using this for a while. Thanks :)

One thing first is the app declares stuff in global scope. Especially the const `storage` makes it so e.g. the watch hangs when trying to start the settings app since that variable name is taken. You have two possibilities here:
- delete the `remove` function of `setUI` so  that the watch will reset before loading the launcher/other apps. Or...
- follow this tutorial to implement fast loading to the launcher properly: https://www.espruino.com/Bangle.js+Fast+Load

I'll look some more and be back if i find anything else :+1:

deepDiverPaul:
Thanks @thyttan for pointing this out. I disabled the `remove` function for now. I will optimise for fast loading in a future release.

apps/line_clock/settings.js
```diff
@@ -0,0 +1,39 @@
...
+    function save() {
+      storage.write(SETTINGS_FILE, settings)
+    }
+
+    E.showMenu({
+      '': { 'title': 'Line Clock' },
+      '< Back': back,
+      'Show Lock': {
+        value: settings.showLock,
+        format: () => (settings.showLock ? 'Yes' : 'No'),
```
thyttan, line 24: There's a default boolean format with empty/filled checkbox. 

I'll take thsse Yes/No formattings out upon merging :+1:

See recent PR https://github.com/espruino/BangleApps/pull/3149 by @bobrippling for reference.

thyttan:
Thanks!

PR #3149, opened 2024-01-08
Remove unnecessary boolean formatters

bobrippling:
Remove manual boolean formatters - rely on pre-existing default formatter. This continues the work done by @alessandrococco in #1980 / 42c2b5c8a2 and introduces a change to `bin/sanitycheck.js` to catch future introductions.

### DIFF
--- a/apps/90sclk/settings.js
+++ b/apps/90sclk/settings.js
@@ -21,7 +21,6 @@
       '< Back': back,
       'Full Screen': {
         value: settings.fullscreen,
-        format: () => (settings.fullscreen ? 'Yes' : 'No'),
         onchange: () => {
           settings.fullscreen = !settings.fullscreen;
           save();
--- a/apps/banglexercise/settings.js
+++ b/apps/banglexercise/settings.js
@@ -11,7 +11,6 @@
     '< Back': back,
     'Buzz': {
       value: "buzz" in settings ? settings.buzz : false,
-      format: () => (settings.buzz ? 'Yes' : 'No'),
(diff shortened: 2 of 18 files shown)


bobrippling:
@thyttan what do you reckon to this?

thyttan:
Looks good I think! I tested it with just the sanitycheck update and it warned me about the apps, when I added the other commit in as well the warnings went away.

I'm not too familiar with either the sanitychecker or regex, so I don't know if this could give false positives somehow?

Would it make sense to look for "yes" (lower case initial) , "On" and "on" as well? I don't think there are instances of those yet, but maybe someone came to do it in the future.

Reading [the settings tutorial](https://www.espruino.com/Bangle.js+App+Settings#the-settings-page) it shows how to add boolean formatting (on/off in that case). Maybe it should not? But then there's no instruction on formatting. Maybe we add another example to the code block there, or format the "how many?" setting instead?

bobrippling:
Nice idea - I've popped in additions for "yes/on" (case insensitively) too.

For false positives, there is a small chance, but as this is just a script that produces warnings, we can always add the warnings to the [ignore list](https://github.com/espruino/BangleApps/blob/b5b65406c000881afdd56f88992d0578375bcc71/bin/sanitycheck.js#L93-L101), or tweak the regex if necessary.

I've opened a PR to the docs repo too - espruino/EspruinoDocs#715, nice spot!

bobrippling:
Shall merge this in tomorrow if all's ok :)

PR #3150, opened 2024-01-09
widget_utils: use correct variable to clear interval

flaparoo:
Hi
A clock I'm developing is using hideable widgets. I've noticed that every now-and-then the clock stops updating after showing the widgets. I'm using an interval (not timeout) to update the clock.
I think I found the culprit.

### DIFF
--- a/modules/widget_utils.js
+++ b/modules/widget_utils.js
@@ -109,7 +109,7 @@ exports.swipeOn = function(autohide) {
   };
 
   function anim(dir, callback) {
-    if (exports.animInterval) clearInterval(exports.interval);
+    if (exports.animInterval) clearInterval(exports.animInterval);
     exports.animInterval = setInterval(function() {
       offset += dir;
       let stop = false;


gfwilliams:
Thanks for spotting this! That's great to get sorted.

And no, unfortunately we can't bump the version - so we'd just have to wait until apps get updated/reinstalled.

thyttan:
Maybe the app loader could keep track of the latest date when an app was installed and when modules were last updated on github. With those data the user could be informed about apps that would benefit from being reinstalled.

flaparoo:
So, when/how is this module updated on the Bangle?
Can I check which version is installed?

thyttan:
It should already be updated on the Developmen App Loader, as that is built automatically by github from this esruino/BangleApps repo whenever it has new commits made to it.

So just going to the Development App Loader and reinstall the relevant app/-s. There's a toggle under 'More...'->'Advanced' to let you reinstall without first uninstalling.

Also, I sometimes find an outdated version of the App Loader is probably cached in my browser. At those times I usually open it in the browsers private mode to get the most recent version.

gfwilliams:
I updated the main side around an hour ago so if you reinstall the app from banglejs.com/apps now you'll get the latest version.

On banglejs.com/apps I disable cacheing, but I can't for GitHub so the development app loader often requires you to force a full reload.

You can't tell which version is installed apart from by looking at the app's source on your watch and trying to figure out what's in the module that's added at the start of the file.

Some versioning could be good but it's a lot of extra complexity, and modules don't change that often. If a module changes a lot we should probably switch it to an app so it has all the versioning support.

PR #3152, opened 2024-01-13
tetris: Reimplement original controls

nxdefiant:
Bring back old controls as "swipe" in menu

Also now exits with short button press

### DIFF
--- a/apps/tetris/tetris.app.js
+++ b/apps/tetris/tetris.app.js
@@ -220,7 +220,7 @@ function newGame() {
         drawTile(ct, ctn, ox+px*8, oy+py*8, false);
       }
     });
-  } else {
+  } else { // control != 4
     if (control == 2) { // Tilt
       Bangle.on("accel", (e) => {
         if (state != 1) return;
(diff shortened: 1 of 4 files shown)


apps/tetris/tetris.app.js
```diff
@@ -200,59 +200,80 @@ function linear(x) {
 
 function newGame() {
   E.showMenu();
-  Bangle.setUI();
-  if (control == 2) {
```
nxdefiant, line 204: All lines below setUI() were only moved into else clause of control == 4

apps/tetris/tetris.app.js
```diff
@@ -200,59 +200,80 @@ function linear(x) {
...
-        move(0, 1);
-        g.flip();
+
+    Bangle.on("swipe", (x,y) => {
+      if (y<0) y = 0;
+      if (moveOk(ct, x, y)) {
+        drawTile(ct, ctn, ox+px*8, oy+py*8, true);
+        px += x;
+        py += y;
+        drawTile(ct, ctn, ox+px*8, oy+py*8, false);
```
nxdefiant, line 220: Code copied from original commit

bobrippling:
Looks grand, thanks!

PR #3154, opened 2024-01-13
fwupdate: don't reset the watch after upload

### LINKED ISSUE #3153: Firmware Update fails if "Automatically reload watch after app App Loader actions" is on
If "Automatically reload watch after app App Loader actions" under "More..."->"Advanced Options" is turned on the firmware update is never executed. The watch just restarts without flashing the new firmware.

The firmware update works again after disabling above option and flashing another version (e.g. reflashing 2v19 before update to 2v20 succeeds).

bobrippling:
Closes #3153

### DIFF
--- a/apps/fwupdate/custom.html
+++ b/apps/fwupdate/custom.html
@@ -426,7 +426,7 @@ function handleUpload() {
     storage:[
       {name:"RAM", content:hexJS},
     ]
-  });
+  }, true);
 }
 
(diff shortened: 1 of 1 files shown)


bobrippling:
@gfwilliams can I request a review? I also notice @nxdefiant is as frequent on here as I am, perhaps another maintainer candidate?

bobrippling:
Sounds good - went for the argument object (and likewise in core)

gfwilliams:
Thank you!

PR #3156, opened 2024-01-14
Add Warpdrive watchface

felipemanga:
Submitting without a screenshot since I don't know how to take one on an actual watch.
This code really isn't going to run on Emscripten. 👀


nxdefiant:
[g.dump()](http://www.espruino.com/Reference#l_Graphics_dump) should create the screenshot

felipemanga:
Ah, nice! Thanks! Unfortunately the screenshot looks *terrible* compared to the real thing, but I've added it anyway.

thyttan:
Looks really rad!

Have you though about having the buttons start the launcher? https://www.espruino.com/Bangle.js+Clock#btn2-to-start-the-launcher It's common practice for clocks.

Also you might find further info on `Bangle.setUI()` intereseting: https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setUI

Having `Bangle.CLOCK=1` (see above about `Bangle.setUI()`) set makes it so other bootcode apps that want to do stuff on clock faces (e.g. `quicklaunch`) can do their thing.

felipemanga:
Thanks for the tip! I'm still looking through the docs, but I've added the launcher and the widgets (but these only show up when the screen is locked since they impact the framerate a lot).

d3nd3:
Hey, I recommend setting the font text color based on theme used, I had white theme and could not see the time at all! Just a headsup.

felipemanga:
I hadn't noticed that! For now I've added `g.setBgColor(0)`, but maybe it'd be nice to have a white background with black text later on.

d3nd3:
https://github.com/espruino/BangleApps/blob/243df4911a77217a504073e9935ebab56c520308/apps/widmessages/widget.js#L39C11-L39C55

This widget is bleeding into your app's time display via setting of BgColor somehow.

felipemanga:
Thanks for the heads-up and the detailed explanation! :)

bobrippling:
That will fix the background colour and I notice you set the font alignment, but you can reset other graphics state with `g.reset()` at the start of your drawing, which will catch other properties that might be left from other drawing code too

d3nd3:
I recommend you use this: 
https://www.espruino.com/Bangle.js+Hideable+Widgets
``` javascript
require("widget_utils").hide()
require("widget_utils").show()
```
Because I have noticed that when the widget re-draws its-self from its internal timeout, it sometimes also bleeds into the drawing, when you expect no widgets to display.
TLDR: You can't assume that widgets stop drawing just because you don't call Bangle.drawWidgets() (They are already loaded and have their own callbacks setup, where they can call Bangle.drawWidgets() or their own draw function).  That library handles this for you though, suppressing their draw callback.

felipemanga:
Oh, so that's what those functions are for. I had seen some code using them but I hadn't realized the point.

bobrippling:
Thanks - let us know when you're happy with the changes and we'll look at merging

felipemanga:
This should be ready for merging, unless I've accidentally missed any of your feedback.

apps/synthwave/app.js
```diff
@@ -0,0 +1,760 @@
...
+  int padstart = bitstart & 31;
+  int padend = bitend & 31;
+  int maskstart = -1 << padstart;
+  int maskend = unsigned(-1) >> (32 - padend);
+  if (wstart == wend) {
+    maskstart &= maskend;
+    maskend = 0;
+  }
+
+  int* row = (int*) &fb[y * stride];
```
bobrippling, line 141: Will this always be aligned to an `int` ? (can't remember if the chip we're on has strict alignment, so I'll just note this & close it, so if we do see bus errors etc, we can come back another time)

bobrippling (commented):
Looks good to me, shall merge likely tomorrow evening (GMT time) to give anyone a chance to test it out

apps/synthwave/app.js
felipemanga: It won't always be aligned, but the chip does support unaligned memory accesses so it's OK. :)

bobrippling: Sounds grand, thanks for the quick reply too!

bobrippling:
Thanks again!

PR #3157, opened 2024-01-14
calendar: Load holidays before events

nxdefiant:
Previously events were loaded and as result painted first and a following draw on a holiday would overwrite the event indicator

### DIFF
--- a/apps/calendar/calendar.js
+++ b/apps/calendar/calendar.js
@@ -280,18 +280,12 @@ const showMenu = function() {
       setUI();
     },
     /*LANG*/"Exit": () => load(),
-    /*LANG*/"Settings": () => {
-      const appSettings = () => eval(require('Storage').read('calendar.settings.js'))(() => {
+    /*LANG*/"Settings": () =>
+      eval(require('Storage').read('calendar.settings.js'))(() => {
         loadSettings();
         loadEvents();
         showMenu();
-      });
-      appSettings(() => {
-        loadSettings();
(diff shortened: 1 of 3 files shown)


apps/calendar/calendar.js
```diff
@@ -141,6 +141,7 @@ const calcDays = (month, monthMaxDayMap, dowNorm) => {
 };
 
 const drawCalendar = function(date) {
+  d1 = new Date();
```
bobrippling, line 144: Is this left over from testing?
```suggestion
```

apps/calendar/calendar.js
```diff
@@ -281,7 +282,11 @@ const showMenu = function() {
     },
     /*LANG*/"Exit": () => load(),
     /*LANG*/"Settings": () => {
-      const appSettings = eval(require('Storage').read('calendar.settings.js'));
+      const appSettings = () => eval(require('Storage').read('calendar.settings.js'))(() => {
+        loadSettings();
+        loadEvents();
+        showMenu();
+      });
       appSettings(() => {
```
bobrippling, line 290: The function passed here as an argument is unused - was there a different change intended here?

bobrippling (commented):
Looks good - two minor things

apps/calendar/calendar.js
nxdefiant: oops, yes.

bobrippling:
Nice, thanks!

PR #3158, opened 2024-01-15
Fix dragboard

flaparoo:
Hi
In the "dragboard", the touch region (upper bound) for selecting letters is not aligned with the rectangle for it - see line 38:
```
g.fillRect(0, (R.y+R.h)-26, (R.x+R.w), (R.y+R.h));
```
This fix will make it much easier to select letters.

### DIFF
--- a/apps/dragboard/ChangeLog
+++ b/apps/dragboard/ChangeLog
@@ -8,3 +8,4 @@
 0.08: Catch and discard swipe events on fw2v19 and up (as well as some cutting
 	edge 2v18 ones), allowing compatability with the Back Swipe app.
 0.09: Fix colors settings, where color was stored as string instead of the expected int.
+0.10: Fix touch region for letters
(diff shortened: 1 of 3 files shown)


thyttan:
Hi - thanks!

I haven't tried this change yet but I will 👍 Just a question before I do, is this supposed to be a better default for everyone or just you?

The current values are probably fitted to my particular Bangle and use pattern (I wrote the initial version), so I'm all for merging this if it should be better for everyone else.

flaparoo:
It's definitively better on mine (Bangle.js 2), but I would assume it's better for everyone. The touch coordinates should be (roughly) aligned with the screen coordinates, right? And that would mean if the rectangle for the letters basically starts at bottom edge - 26, then the touch region should start there too.
The only variable here, as far as I can tell, would be the touch-screen calibration.

thyttan:
I think I agree with this, it's better on my watch as well from the little testing I just did. Maybe I chose the old values before  touch screen calibration was a thing.

PR #3160, opened 2024-01-18
Bin Clock - add fullscreen option

shansou504:


bobrippling:
Nice addition, thanks!

PR #3161, opened 2024-01-18
forge: Compatibility with Fastload Utils

thyttan:
... by adding the string `"Bangle.loadWidget()";`. This way Fastload Utils thinks widgets would be present and so fastloads. But in the absence of Fastload Utils no widgets will be loaded, and so no overhead is added for the usual case.

Exiting forge will be done with a regular `load()` call in both cases, since no remove-handler is added to a `Bangle.setUI({})` call.

### DIFF
--- a/apps/forge/ChangeLog
+++ b/apps/forge/ChangeLog
@@ -1,1 +1,2 @@
 0.01: attempt to import
+0.02: Make it possible for Fastload Utils to fastload into this app.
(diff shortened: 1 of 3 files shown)


bobrippling:
To check, does forge render itself ok still if widgets have already been loaded?

thyttan:
Yup 👍 the widget field will persist from the previous app and the menu will render accordingly.

bobrippling:
Great, thanks for the image too!

PR #3170, opened 2024-01-28
drained: improve escape menu and remove clock type+tag

### LINKED ISSUE #3168: [Drained] Unable to exit clock with Drained as clock
### Affected hardware version

Bangle 2

### Your firmware version

2v19

### The bug

I was attempting to change the clock, but on the small screen I accidentally set Drained has my clock. I was expecting to be able to go back to the launcher to change this, but when you attempt to exit to the launcher, it will ask if you want to restore watch to full power. No matter which option you pick, it will return to the clock, which is Drained. You therefore can't exit Drained. To reproduce this issue, install drained and set it as your clock in settings.

### Installed apps

I cannot check my apps, because drained disables BLE when running.

bobrippling:
Fixes #3168

### DIFF
--- a/apps/drained/ChangeLog
+++ b/apps/drained/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Allow boot exceptions, e.g. to load DST
 0.03: Permit exceptions to load in low-power mode, e.g. daylight saving time.
       Also avoid polluting global scope.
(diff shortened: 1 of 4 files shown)


thyttan:
Nice!

PR #3171, opened 2024-01-29
New app: aviatorclk (and module "avwx")

flaparoo:
Hi

This is a clock designed for pilots. It's a port of the clock I've used for years on my Pebble.

It relies on the public AVWX API to retrieve the meteorological report. I've got a couple of [other](https://github.com/flaparoo/BangleApps/tree/avweather/apps/avweather) [apps](https://github.com/flaparoo/BangleApps/tree/flightdash/apps/flightdash) in the works which will use the same API, so I've decided to create the small "avwx" module/library (primarily so that the API key only has to be entered once).


apps/aviatorclk/aviatorclk.app.js
```diff
@@ -0,0 +1,283 @@
...
+Bangle.drawWidgets();
+
+// draw static separator line
+y = Bangle.appRect.y + mainTimeHeight + secondaryFontHeight;
+g.setColor(separatorColour);
+g.drawLine(0, y, g.getWidth(), y);
+
+// draw times and request METAR
+draw();
+if (settings.showSeconds)
```
nxdefiant, line 273: You might only want to draw settings if unlocked, else display updates can drain the battery faster

flaparoo: I assume you mean "draw seconds" (instead of "settings" - please correct me if I'm wrong).

If the seconds display is required or useful in flight (and therefore enabled through the settings), it's actually handy to have it constantly available (ie. without unlocking). I've also tried to minimise the code for the seconds updates (only the 2 seconds digits are updated every second).

I am considering implementing a way to "dynamically" enable the seconds display (double tap, or create 3 "tap zones" for scroll up / down and enable seconds display, etc.).

nxdefiant: right, its "seconds". Just wanted to make sure that you are aware of the consequences to battery life.

flaparoo: Thanks - yes, I'm aware of the extra power draw.
The readme points this out to the end user:
```
- **Show Seconds**: to conserve battery power, you can turn the seconds display off
```

bobrippling:
Looks great, thanks for the module extraction too!

PR #3173, opened 2024-01-30
Update lib.js

EnerGeh:
Added Molly (Signal fork) and Threema libre (official f-droid flavour of threema).


bobrippling:
Thanks!

EnerGeh:
Happy to help!

gfwilliams:
I just saw on https://github.com/espruino/BangleApps/commit/b705c4bc748778cdbb4007afbf6a6657df05b436#commitcomment-139431966 you're wondering if this went in?

The version number hasn't been incremented so even though it is live nobody knows.

But - I'm afraid this code does need some extra changes anyway.

I've just updated the `generate.js` file to make the comments clearer and right at the top of the file, but the file is auto-generated from the data in https://github.com/espruino/BangleApps/blob/master/apps/messageicons/icons/icon_names.json - so if the generation script is ever run again it'll undo your changes

I've modified `icon_names.json` as well so hopefully your changes will stay now too, and I bumped the version - but it might be a few days before that goes in

PR #3176, opened 2024-01-31
Improve boolean formatter check

bobrippling:
See also #3149


bobrippling:
@thyttan might I request a review? This is pretty much a follow-on from the [last one](https://github.com/espruino/BangleApps/pull/3149)

thyttan:
I'm not very good at regex at all, but I used an online parser/explainer and I think I understand how it matches. So looks good to me :+1:  Should I merge?

bobrippling:
That's alright, thank you and yes please!

PR #3181, opened 2024-02-06
[fwupdate] Add links to firmware changelogs; couple minor fixes

thinkpoop:
It'd be nice to have the firmware changelogs linked to from the firmware updater. The custom.html page seemed like the best spot for now, until a readme with more detail is made.

I wasn't sure if I should increase the version number or not, since this was more of a documentation fix?

I also fixed a couple small bugs (features?) in the page:

1) When viewed in gadgetbridge, some of the links try to open as a frame inside the popup. It didn't seem intentional, since other links opened in a new window.
 
2) When clicking the DFU or Advanced toggles, it jumps to the top of the page. Other js links on the page use preventDefault, so I assumed those should too.

Link to my updated app loader:
https://thinkpoop.github.io/BangleApps

### DIFF
--- a/apps/fwupdate/custom.html
+++ b/apps/fwupdate/custom.html
@@ -58,6 +58,15 @@
 
     <pre id="log"></pre>
 
+    <p><a href="#" id="changelog-btn">Firmware ChangeLog ▼</a></p>
+    <div id="changelog-div" style="display:none">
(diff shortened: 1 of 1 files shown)


bobrippling:
Looks good to me - no need for the version bump since it's not part of what's loaded onto the watch :)

Thanks for the bug fixes too!

PR #3182, opened 2024-02-06
aviatorclk: added tap event to scroll METAR and toggle seconds display

flaparoo:
This is an update for the "aviatorclk". Triggered by @nxdefiant 's comment when the app was originally submitted, I've added a way to easily toggle the seconds display (through double taps left or right).
And I found the taps useful to scroll the METAR display, too.

This PR also includes a minor bug fix (set the background colour prior to seconds updates).


bobrippling:
Thanks!

PR #3183, opened 2024-02-06
New app: flightdash

flaparoo:
This is a brand-new app, to display basic flight and navigation instruments.

It is battery-hungry (screen updates every second, with GPS and optionally also barometer on). It consumed ~10% battery running it for 2h straight. I think that's still acceptable, for what the app does.

I've used it already on several occasions, and I'm confident that the main app is working fine. Similar for the settings, at least for my neck of the woods (ie. with my list of airports, and using the optional avwx module integration). The interface page (to upload a list of airports to the Bangle) is functional, but is not as thoroughly tested.


pavelmachek (commented):
What goes o with eval() here? That should not really be required?

flaparoo:
Hi @pavelmachek 
The challenge was to create a list of menu entries, in most cases from a dynamic list of airports, which configure multiple settings values (the airport ID, latitude and longitude). The only showMenu option that works for this case is calling a function when the menu entry is selected (ie. I couldn't use the boolean, custom format or number range menu entry). However, as far as I know there's no way to pass any arguments to that function, so the best way I could come up with was to dynamically create a specific function per menu entry - which required eval.

If there is a better way to handle this, I'm all ears.

nxdefiant:
Where is the difference to e.g. alarm showing menu entries for a dynamic list of alarms?

flaparoo:
Thanks for the hint @nxdefiant - I didn't look at the alarms for a solution. I can see the workaround there is to set a timeout (of 10ms) to call a function, with arguments. That's clever, too, albeit also similarly hacky.

Before I change this app to use setTimeout, what is the concern with using eval?
I've added extra sanitisation to the variables used in the eval - in case the concern was about stability or even security implications.

nxdefiant:
Afaik the setTimeout is needed because showMenu() does some painting after calling onchange(). So a new showMenu() created in onchange() would be overpainted by the previous menu.

I've nothing against eval() btw as long as it does not burn the watch or make it evil to kill its carrier.

flaparoo:
Fair enough. I didn't notice any overpainting, but maybe that's just under certain conditions. And, given I'm not speaking to you from beyond the grave, the evals haven't killed me  - yet... ;-)

Anyway, I've pushed another commit with all evals replaced with setTimeouts.

apps/flightdash/flightdash.settings.js
```diff
@@ -0,0 +1,326 @@
...
+  var afterGPSfixMenu = 'destNearest';
+  function getLatLon(fix) {
+    if (!('fix' in fix) || fix.fix == 0 || fix.satellites < 4) return;
+    Bangle.setGPSPower(false, APP_NAME+'-settings');
+    Bangle.removeListener('GPS', getLatLon);
+    switch (afterGPSfixMenu) {
+      case 'destNearest':
+        loadNearest(fix.lat, fix.lon);
+        break;
+      case 'createUserWaypoint':
```
atjn, line 62:
```suggestion
      case 'createUserWaypoint': {
```

apps/flightdash/flightdash.settings.js
```diff
@@ -0,0 +1,326 @@
...
+          settings.userWaypoints = [];
+        let newIdx = settings.userWaypoints.length;
+        settings.userWaypoints[newIdx] = {
+          'ID': 'USER'+(newIdx + 1),
+          'lat': fix.lat,
+          'lon': fix.lon,
+        };
+        writeSettings();
+        showUserWaypoints();
+        break;
```
atjn, line 73:
```suggestion
        break;
      }
```

atjn (commented):
I suggest putting this switch case in a block, since it will not pass the eslint rule that I am trying to enable for all apps in #3188 :)

apps/flightdash/flightdash.settings.js
flaparoo: No worries - done.

bobrippling:
Looks good, happy to merge - will do so tomorrow most likely, to give any existing reviewers a last chance if they wanted to check anything

bobrippling:
Thanks!

PR #3185, opened 2024-02-09
android.html: allow zooming in Gadgetbridge

thinkpoop:
Update to android.html (Gadgetbridge) to allow zooming in the embedded AppLoader.

Most browsers have a "Force enable zoom" option, but the embedded webview isn't affected by it.

Link: https://thinkpoop.github.io/BangleApps/android.html


bobrippling:
Looks good - @gfwilliams was there a strong reason for the original attributes / any points before I merge?

gfwilliams:
If this is actually tested and works in Gadgetbridge, let's merge.

The main thing is to ensure that it forces readable font/button sizes and one column of apps. I think likely I just set the `meta` tag with the min/max because that's what I'd seen suggested for webviews and I didn't think people would really want to zoom anyway, but I don't really have strong opinions either way :)

PR #3187, opened 2024-02-10
Don't use octals

atjn:
I suggest disallowing 0-prefixed numbers in the official apps, as it is not allowed per the Ecmascript standard.

These numbers are disallowed because the original Ecmascript standard allowed some decimal numbers to be 0-prefixed, but also used the 0-prefix to encode octal numbers. Most of the time, octal numbers are simply converted to decimal numbers and everything is good, but sometimes these subtle differences lead to some very weird bugs. The newer strict Ecmascript standard therefore disallows 0-prefixed numbers completely to eliminate this class of bugs.


bobrippling:
Great find, sounds good to me!

PR #3188, opened 2024-02-10
Require switch blocks when declaring variables

atjn:
When declaring variables inside switch cases, the variables are also available in the other switch cases, with some special rules. This is basically never the intention, and can lead to weird bugs. I suggest requiring switch blocks when variables are declared, which limits the scope of the variable to the individual case.

### DIFF
--- a/apps/.eslintrc.json
+++ b/apps/.eslintrc.json
@@ -142,7 +142,6 @@
                 "SwitchCase": 1
             }
         ],
-        "no-case-declarations": "off",
         "no-constant-condition": "off",
         "no-delete-var": "off",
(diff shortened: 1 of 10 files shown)


bobrippling:
Sounds good, thanks!

gfwilliams:
I just want to check - have you tested all of these apps since you made the changes to them?

Because while I think this is a good thing, there may still be a possibility that wrapping the blocks has actually broken something if the code did actually use the defined variable later?

atjn:
I checked that the variables in the switch cases were only used in that case. I also tested a few apps on my watch just to make sure my approach was valid.

PR #3191, opened 2024-02-11
widbaroalarm: plot + fix interval on start

nxdefiant:
- Do not immediately measure on start, use previous value to always stay in interval
- Add plot pressure history to settings


bobrippling:
Nice ideas, thanks!

Issue #3048, opened 2023-10-15
[calendar] Swipe actions seem wrong in calendar app


jt-nti:
### Affected hardware version

Bangle 2

### Your firmware version

2v18

### The bug

The current calendar swipe actions seem like they're the wrong way round to me...

- Swipe left to go to the previous month
- Swipe right to go to the next month
- Swipe up (Bangle.js 2 only) to go to the previous year
- Swipe down (Bangle.js 2 only) to go to the next year

Swiping should be a drag gesture, for example how the banglejs menu works, dragging up to view items further down the menu.

Would it be ok to submit a PR to reverse the current swipe gestures in the calendar?

nxdefiant:
I'm the perpetrator that changed the calendar handling from tapping on the left/right side to swipe events.
Unfortunately I do not understand your problem, but I gather that you expect for example a swipe from left to right to go to the previous month instead of the next month?

I think the best solution is to add an option for that in the settings.

Sure PRs are welcome, but in this case I would like you to wait a few days because I'm also working on an update and I would hate to run into a merge conflict. (Or you do the change on my [branch](https://github.com/nxdefiant/BangleApps/tree/calendar)?)

jt-nti:
Swiping is definitely a nice improvement to the old tapping left and right, I just always swipe the wrong way! I wondered whether an option to switch the direction round might be ...well, an option :)

It would likely take me longer than a few days to getting round to working on it anyway, so there shouldn't be any danger of conflicts. It's probably worth keeping the issue open for a bit to get feedback on what the option might be called, and which direction would be most popular for the default.

gfwilliams:
Yes, I think this is probably just a personal preference thing? I think if someone really cares they can make a PR to add an option for it I guess? But if we change it around now we'll just get **another** issue later complaining it's the wrong way and swipe left should to go to the previous month.

... and I'm not even joking. It's happened before for other similar things.

jt-nti:
That related thread was definitely interesting. It would certainly be nice if things were consistent so rather than a calendar specific option, would a system-wide option be a possibility?

jt-nti:
No idea when I'll get a chance but if no one beats me to a PR I'll give it a go sometime.

nxdefiant:
You could try something similar to this as swipeinvert.boot.js:
```
{
  const setURIOrig = Bangle.setUI;
  Bangle.setUI = (mode, callback) => {
    if (typeof mode === "object" && mode.swipe) {
[...]

PR #3199, opened 2024-02-14
Add scroll and swipe direction system settings

jt-nti:
Contributes to #3048

### DIFF
--- a/apps/setting/ChangeLog
+++ b/apps/setting/ChangeLog
@@ -78,3 +78,4 @@ of 'Select Clock'
 0.67: Rename 'Wake on BTN1/Touch' to 'Wake on Button/Tap' on Bangle.js 2
 0.68: Fix syntax error
 0.69: Add option to wake on double tap
+0.70: Add scroll and swipe directions on Bangle.js 2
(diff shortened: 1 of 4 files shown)


jt-nti:
Hopefully the settings make sense and I got them the right way round! I split out horizontal and vertical scrolling so that it's possible to keep the current behaviour, not that any apps use the new settings yet.

nxdefiant:
What does the setting actually change? Am I'm missing something? I see the code for the setting, but I don't see it used somewhere.

nxdefiant:
Anyway iirc default apps must be reviewed by @gfwilliams

nxdefiant:
Wouldn't it be better to invert the direction globally instead of each app individually?

jt-nti:
Possibly but I'm not sure. There may be some apps that always want/need swipes to work one way or another, e.g. to move something around on the screem. Maybe if apps could opt out of the setting somehow? I also don't want to break anything related to entering numbers etc.

nxdefiant:
I would assume that those are the exception. I would still change it globally and handle those when a bug is filled.

gfwilliams:
I think having configurable swipe direction would be nice, but I don't think this way of doing it is a great idea I'm afraid. It'll cause a massive amount of confusion having this global setting which has effect in exactly 0 of the 600+ apps available.

We have the `Bangle.setUI` function which pretty much all apps use to work with swipes, so actually you could just override it like the https://banglejs.com/apps/?q=backswipe app does, and then change the sign of numbers coming out of the `drag` event.

I think at this point, that would be the best way of doing this - just a separate app with boot code that can run to change the default behaviour. Then for people who keep the default scroll direction there's no extra code installed/running - but if you want to change you just install the app, change the settings as you want them, and it 'just works'. It could also include (like `backswipe`) the ability to exclude certain apps if the user wanted

PR #3201, opened 2024-02-15
health: add hrm info to readme

thyttan:
reference:

> The health app on the Bangle only stores data every 10 minutes. The '3 minute' option is for the heart rate - it'll take 3 HRM samples every 10 minute slot and use the best/average value of HRM readings - but it only stores one reading per 10 minutes.

https://forum.espruino.com/conversations/386312/#comment16972306


bobrippling:
Nice, thanks for the tweak

PR #3202, opened 2024-02-19
Add new Scroll Settings app

### LINKED ISSUE #3048: [calendar] Swipe actions seem wrong in calendar app

jt-nti:
New app for configuring natural or reverse scroll and swipe directions

Also updates calendar app to default to natural scroll and swipe directions

Closes #3048


apps/swpcfg/README.md
```diff
@@ -0,0 +1,20 @@
+# Swipe Settings
+
+A service for configuring the Bangle.js scroll and swipe directions
+
+* **Scroll Direction** switch between natural (down motion scrolls down) and reverse (down motion scrolls up) vertical scrolling
```
thyttan, line 5: Isn't natural/reverse scrolling the other way around? Just so we get the terminology right.

One reference:
https://blog.logrocket.com/ux-design-natural-vs-reverse-scrolling/attachment/scroll-direction-natural-vs-reversed/

nxdefiant:
Uhm, why do you need the direction in calendar? I'm not happy with this..

apps/swpcfg/README.md
nxdefiant: I would name them "default" and "reverse"

thyttan: Good idea I think, since the default behavior is different across apps 👍

jt-nti:
> Isn't natural/reverse scrolling the other way around? Just so we get the terminology right.

I was definitely paranoid about getting the terminology right in the first PR attempt but I _think_ it's right. For example...

https://jessequinnlee.com/2015/07/25/natural-scrolling-vs-reverse-scrolling/

"Reverse: Swipe fingers up on trackpad, magic mouse, scroll-wheel, content goes down, scrollbar goes up."

This is what I think the calendar is currently doing which is why it keeps catching me out. 

"Natural: Swipe fingers up on trackpad, magic mouse, scroll-wheel, content goes up, scrollbar goes down."

This is what I'm expecting the calendar to do, a bit like the menus, i.e. I swipe up, the imaginary scroll bar moves down to reveal the next year below the current year.

If there are better names to make it clearer, that would definitely be good since it's very confusing, but the idea is that all the apps should work consistently and swipe the same way, so "default" and reverse should be global, rather than for individual apps.

nxdefiant:
After reading that article: Lets name them "default" and "apple"

jt-nti:
Another possibility would be to change the settings to just be boolean options to "reverse" the swipe direction.

thyttan:
> Another possibility would be to change the settings to just be boolean options to "reverse" the swipe direction.

I like the sound of that!

nxdefiant:
> This is what I'm expecting the calendar to do, a bit like the menus, i.e. I swipe up, the imaginary scroll bar moves down to reveal the next year below the current year.
showScroller is using drag, while calendar is using swipe. Just saying.

thyttan:
> ... , but the idea is that all the apps should work consistently and swipe the same way, so "default" and reverse should be global, rather than for individual apps.

I'm sympathetic to that sentiment. But I'll reference this thread again, https://forum.espruino.com/conversations/388188/#17045379 , as a counter argument. It is not obvious that all swipe/drag interactions are or are expected to be done in the same way.

I don't have strong feelings about this. But my suggestion would be to leave apps as they are. Then those that want to reverse swipe directions for an app can do it with `swpcfg`.

jt-nti:
That's a great thread but I'm not sure it's a counter argument. I completely agree that not all swipe/drag interactions should be done in the same way, which is why I was/am nervous about hijacking the drag event. That seemed to be the prefered direction in #3199 though. What I do think should be the same is that the same interactions should be consistent between apps, e.g. if I swipe up to increase the volume in one app, it would be confusing if I had to swipe down to increase the volume in another app.

Even hijacking the higher level 'updown' and 'leftright' events seem prone to causing disruption. I'm back to thinking a system setting which apps could use might still be the better option, e.g.

```
// load global settings
const STORAGE = require('Storage');
let bSettings = STORAGE.readJSON('setting.json',true)||{};

// ...

// Respond to user input
Bangle.setUI({mode: "updown"}, function(dir) {
  if (dir === settings.scrollUp) {
    n--;
    draw();
  } else if (dir === settings.scrollDown) {
    n++;
    draw();
  } else {
    n = 0;
    draw();
  }
});
```

Of course that is likely to result in even more discussions about which direction is which :)

In the case of the calendar app, which is what started this in #3048, it's the only thing I've noticed so far that seems inconsistent with other Bangle.js UIs. It's also the opposite way round to the calendar app on my phone which makes it doubly confusing.

nxdefiant:
> I'm back to thinking a system setting which apps could use might still be the better option, e.g.

Another idea: Instead of adding that to system settings you could keep it in swpcfg and remove its boot file.
Now in calendar you read the swpcfg settings and if it is present and value is set you invert the direction.

> it's the only thing I've noticed so far that seems inconsistent with other Bangle.js UIs

Not inconsistent with the apps I use, do you have an example?

thyttan:
From [#3199](https://github.com/espruino/BangleApps/pull/3199#issuecomment-1952256411):


> We have the `Bangle.setUI` function which pretty much all apps use to work with swipes, so actually you could just override it like the https://banglejs.com/apps/?q=backswipe app does, and then change the sign of numbers coming out of the `drag` event.

> 

> I think at this point, that would be the best way of doing this - just a separate app with boot code that can run to change the default behaviour. Then for people who keep the default scroll direction there's no extra code installed/running - but if you want to change you just install the app, change the settings as you want them, and it 'just works'. It could also include (like `backswipe`) the ability to exclude certain apps if the user wanted


I think that way of doing it makes sense, keep it entirely inside `swpcfg`. Through `swpcfg` settings page the user tells it on what apps it's boot code should reverse directions.

nxdefiant:
> I think that way of doing it makes sense, keep it entirely inside `swpcfg`. Through `swpcfg` settings page the user tells it on which apps it's boot code should reverse directions.

Agreed, should be simple to create a menu with the list of installed apps to select/unselect. `global.__FILE__` can be used to get current running app.

jt-nti:
I'm beginning to think that the general problem (consistent and configurable scroll directions) cannot easily be fixed in a seperate app, if at all. Adding hscroll and vscroll interface types to [Bangle.setUI](http://www.espruino.com/Reference#l_Bangle_setUI) that can be configured with a system setting might work, i.e. to distinguish scrolling from other direction specific swipe interactions in apps. Until something like that exists, the swipeinv app seems like a useful work around for issue #3048

thyttan:
I'll close this as we merged #3217. Open it again if you disagree :)

PR #3203, opened 2024-02-22
use a crisper svg logo

KTibow:
svgs are vector files. they can get small with optimization, smaller with manual tweaking, and even smaller with gzip compression, while having better resolution at the same time.

the logo on banglejs.com/apps/ looks blurry, and it gets worse on high density displays. this pr fixes it by moving it to an svg.

(i plan to make some more prs after this one that make the ui better in some other ways)


android.html
```diff
@@ -22,7 +22,7 @@
   <body>
     <header class="navbar-primary navbar">
       <section class="navbar-section" >
-        <a href="https://banglejs.com" target="_blank" class="navbar-brand mr-2" ><img src="img/banglejs-logo-sml.png" alt="Bangle.js">
+        <a href="https://banglejs.com" target="_blank" class="navbar-brand mr-2"><img src="img/banglejs-logo-small.svg" alt="Bangle.js">
```
bobrippling, line 25: Do we want to delete `banglejs-logo-sml.png`?

bobrippling (commented):
Nice idea - cc'ing @gfwilliams since it's a more user-facing change

android.html
gfwilliams: Maybe don't bother? It's not like it's a big file...

gfwilliams:
Looks good - thanks!

android.html
KTibow: I feel like if it were to be deleted it would be more a matter of semantics

bobrippling: Yeah my thought was that someone might change one without the other

PR #3204, opened 2024-02-22
flightdash: drop unnecessary boolean formatter

bobrippling:
This drops the warning we keep seeing (`Settings for flightdash has a boolean formatter - this is handled automatically, the line can be removed`). No change in behaviour, no need for a version bump.


bobrippling:
Thanks @thyttan!

thyttan:
Thanks!

PR #3208, opened 2024-02-25
Add http-server dependency

atjn:
Otherwise the `local` and `start` npm scripts do not work :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "eslint": "^8.14.0",
         "eslint-config-airbnb-base": "^15.0.0",
         "eslint-plugin-import": "^2.26.0",
+        "http-server": "^14.1.1",
         "npm-watch": "^0.11.0"
       }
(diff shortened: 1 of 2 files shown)


bobrippling:
Sounds good, cc'ing @gfwilliams with it being a central change

gfwilliams:
Thanks, yes, this is fine - although I do like to try and keep the dependencies as low as possible usually.

I'd be interested to know how much `local/start` commands are used though - it's something that was contributed but that I don't think was ever documented anywhere.

atjn:
I looked at [`testing`](https://github.com/espruino/BangleApps?tab=readme-ov-file#testing) and found both the online and offline options to be very complicated for me. I already have the code open in an editor, so it is every easy to run `npm install && npm run local`.

nxdefiant:
Fun, I did not knew the `npm run local`, I always did  `npm start` instead

PR #3209, opened 2024-02-26
Rename da_DA -> da_DK

atjn:
I don't think `da_DA` is a real language code, the common danish language code is `da_DK` for danish_denmark. I am guessing this was just an oversight by @copoer - but feel free to correct me if there was a reason to do it this way :)


bobrippling:
Looks good to me - shall give @copoer a day in case there was a reason, and I'll merge tomorrow if not

PR #3212, opened 2024-02-26
Warn about empty blocks

atjn:
The `no-empty` rule checks that there are no empty blocks defined in the code, similar to the check for unused variables.

I have disabled the check for empty catch blocks, because they are a useful shortcut for trying something and failing softly. Obviously it would be better to enforce explicit error handling for all try blocks, but I think this is one of the cases where the benefit doesn't outweigh the extra work for contributors.

Warnings about empty blocks can make the developer restructure their code to be more readable, have less overhead, and can help them find bugs. I actually think I found one in the `carcrazy` app. @Ronin0000 is it on purpose that level 7 is ignored? It seems like it should be included in the check at line 49 :)

I have tested these apps to the point where I'm convinced that I have introduced 0 new bugs.

### DIFF
--- a/apps/.eslintrc.json
+++ b/apps/.eslintrc.json
@@ -144,7 +144,7 @@
         ],
         "no-constant-condition": "off",
         "no-delete-var": "off",
-        "no-empty": "off",
+        "no-empty": ["warn", { "allowEmptyCatch": true }],
         "no-global-assign": "off",
         "no-inner-declarations": "off",
         "no-prototype-builtins": "off",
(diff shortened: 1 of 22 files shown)


atjn:
I don't think it makes sense to bump the version of all the apps, but if you disagree I will be happy to bump more/all of them:

- [ ] andark - not bumped because we only remove empty blocks, no change to executed code
- [x] aptsciclk - bumped
- [x] carcrazy - bumped
- [x] doztime - bumped
- [x] hwid_a_battery_widget - bumped
- [ ] novaclock - not bumped because we only remove empty blocks, no change to executed code
- [x] red7game - bumped
- [ ] spacew - not bumped because we only remove empty blocks, no change to executed code
- [x] tabanchi - bumped

PR #3213, opened 2024-02-26
Warn about unused variables

### LINKED ISSUE #3189: Work towards stricter linting for app code
If I enable the recommended Eslint checks that ensure variables are scoped and used correctly, I receive 6500 errors (on average about 10 errors per app). That seems problematic to me, especially because these errors are likely performance/power drain issues.

As I see it, you should not be able to merge apps here if they don't pass these checks. The checks are not hard to pass, it is things like deleting variables that are never used, and declaring variables so they aren't accidentally attached to the global scope.

Unfortunately there are so many errors that it is not trivial to go through them all and just fix the issues. Therefore I propose that the global Eslint rules are updated to be more strict, but then each app that doesn't pass the new rules will get its own local Eslint config that disables the new rules only for that app.

This solution would ensure that:
- New apps must pass the stricter rules, so the problem does not continue to grow
- App owners will become aware that their app has an issue, so some will probably go and fix that
- People like you and me can incrementally go through existing apps and fix them on a case-by-case basis

If you agree with this, I would be happy to write a script that automatically adds the necessary local Eslint configs.

As for which rules should be enabled, I think that all rules in the `eslint:recommended` set should be enforced for new apps.

I am eager to hear your opinion on this! Thanks for taking the time to read my wall of text :)

atjn:
I believe we have reached a consensus in #3189 to enable warnings for unused variables (except in function declarations). Because there are so many existing violations, I will not try to fix them here and now, but will expect future PR's to slowly cut down on the number of violations.

Closes #3189


atjn:
@gfwilliams @bobrippling This is ready for review now.

## Lint exemptions
I was able to add a file (`apps/lint_exemptions.js`) that contains a file path + a file hash which maps to a set of rules that should be ignored. Eslint will ignore the rules that are defined for each file path. A custom script will run before eslint runs, which makes sure that the files still match the hash. If the files do not match the hash, that entry is removed from the ignore file, which means Eslint will enforce the normal rules when it runs.

I have also create a script that can add a new entry in the ignore file for you. It works this way:
```sh
# template
node bin/exempt-lint.mjs LINTRULE FILEPATH

# example
node bin/exempt-lint.mjs no-unused-vars ./apps/_example_app/app.js
```
The example command will ensure that the `no-unused-vars` rule is not enforced for the main file in the example app.

## Fixing existing lint violations
I have tried to only fix lint where I am sure that the fixes do not change the code. It is however hard to be 100% sure because many of the apps use funky hacks, such as importing each other scripts as text and `eval`ing the code, then using variables that were declared in the `eval`ed script.

I have not deleted a lot of code, most of it is commented out. This is because it seems like the code was supposed to be there for something like debugging or for future unimplemented features. By commenting the code out we don't rob anyone of their hard work, but we reap the benefits of running less code. For users with pretokenisation turned on, this also makes their apps smaller.

I also found several apps that seem to have bugs which are causing the lint warnings. I have not fixed the bugs because I think this PR needs to be as safe as possible to merge, but I have added `TODO` comments to point out where the bugs are.

atjn:
I can downgrade the mjs files to cjs files if you want to. I was just guessing that you eventually would want to migrate to standard ES modules, so might as well start writing them like that now. I also prefer ES modules.

The update to nodejs v20 is required to run the relatively modern JS that I am using in the scripts. I can write some compat wrappers if you want, that would probably get the nodejs version down to at least v18.

atjn:
If I need to downgrade to cjs, I might need to write some ugly wrapper code because I need support for top-level await.

gfwilliams:
> I might need to write some ugly wrapper code because I need support for top-level await.

Ok, thanks - good to know. Let's leave as-is then.

I'm not so bothered about the other scripts but ideally I'd like `npm test` to run on the most possible node.js versions - but I think we're probably ok there anyway.

Not switching to ES modules was partly because as they're not supported by Espruino there's then a mix of coding styles which might be confusing to contributors - but I guess most users won't really even see this code,

nxdefiant:
https://packages.debian.org/search?keywords=nodejs I guess I have to refrain for some time..

nxdefiant:
Yes please, I'm usually not bypassing apt and I'm not keep to install node docker container either.

atjn:
@nxdefiant see #3240

bin/bulk-update-apps.mjs
```diff
@@ -0,0 +1,114 @@
...
+ * node bin/bulk-update-apps.mjs GITHASH --output
+ *
+ * This mode doesn't make any changes to your files, it outputs the ID's of all apps that would be bumped.
+ */
+
+import { exec } from "node:child_process";
+import fs from "node:fs/promises";
+
+const commitHash = process.argv[2];
+if (!commitHash || commitHash === "--output") {
```
bobrippling, line 27: If the user passes an extra argument, we won't catch it, e.g.
```bash
$ node bin/bulk-update-apps.mjs abc123 'message' --output --help
```
or even:
```bash
$ node bin/bulk-update-apps.mjs abc123 they accidentally forgot to quote the changelog message

# creates changes with "they" in the changelog
```

Then general way to parse arguments is either with a library or a loop, this is simple enough for a loop, e.g.
```javascript
for (let i = 2; i < process.argv.length; i++) {
    const arg = process.argv[i];

    if (arg === "--output") {
        outputFlag = true;
    } else if (!commitHash) {
        commitHash = arg;
    } else if (!changelogMessage) {
        changelogMessage = arg;
    } else {
        usage();
    }
}

if (!commitHash || !changelogMessage) usage();

function usage() {
    // print program usage to stderr and exit with an error
}
```

This also allows flexibility with where flag arguments (e.g. `--output`) can be placed (they traditionally come before positional arguments)

I think that command line programs shouldn't throw an error to exit - it leaves an ugly stack trace and exposes implementation details, they should log a message to the console with help on how the user can correct their error and exit with an exit code, but I don't mind if you want to make `usage()` throw an error

bin/bulk-update-apps.mjs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+let outputFlag = false;
+if (process.argv[3] === "--output" || process.argv[4] === "--output") {
+  outputFlag = true;
+}
+
+const gitOutput = await new Promise((resolve) => {
+  exec(
+    `git diff-tree --no-commit-id --name-only ${commitHash} -r`,
```
bobrippling, line 47: What if the user passes `commitHash` as `; rm -rf ~/` ?

Is there an execl / execv function we can use? (one that doesn't invoke `sh`)

bin/exempt-lint.mjs
```diff
@@ -0,0 +1,66 @@
...
+ * Run it like this:
+ * node bin/exempt-lint.mjs LINTRULE FILEPATH
+ *
+ * Example command:
+ * node bin/exempt-lint.mjs no-unused-vars ./apps/_example_app/app.js
+ */
+
+import fs from "node:fs/promises";
+
+const lintRule = process.argv[2];
```
bobrippling, line 19: Same here (and with `sync-lint-exemptions.mjs`) as above with dealing with all arguments :)

bobrippling (commented):
A bit late to checking this out but it looks really good - very useful to have. I've a few comments on catching invalid usage mostly

atjn:
@bobrippling The command line handling is very basic. But these scripts are only supposed to be used by you and me, not by users. The first comment in the scripts explain exactly how the script should be used, and if you accidentally use it incorrectly, you will catch the bug when you check the diff to see what you're merging.

I didn't see it as a major problem, so I decided that it would be a waste of time and complexity to handle those edge cases. But maybe you can explain the use case where you think this is a problem? Also, if you want to rewrite the CLI handling yourself, I would never be against that :)

bin/bulk-update-apps.mjs
atjn: > I think that command line programs shouldn't throw an error to exit

Maybe you're right, but throwing an error is just an easy way to make sure the program stops, and it gives us more information about what went wrong. Once again I will argue that the "users" are basically only you and me.

bin/bulk-update-apps.mjs
atjn: Again, since it is only used by us locally, we will probably be fine. Just don't do command injection on your own machine.

I don't think there is a safe way to run commands in Nodejs apart from using a library, but a quick solution to this could be to check that the input only contains a-z and 0-9 characters.

bobrippling:
Yes, I suppose it's end-user experience, sometimes i'll chuck `--help` or attempt to use a command to find it doesn't behave any differently, and it's nice to catch these things, a bit like how lints catch potential errors before they happen

PR #3214, opened 2024-02-26
Warn about undefined variables

atjn:
Undefined variables are almost never intentional. A variable is either undefined because there is a breaking bug in the code that needs to be fixed, or because the developer forgot to correctly initialize the variable. If the developer forgot to initialize the variable, it will be added to the global scope, which is basically a memory leak. In both cases a warning is good to make the developer aware of their mistake.

In the rare case that the developer wants to set something on the global scope, they can explicitly set it with `global.foo = ...`, which will not trigger this warning.

I think I can piggy-back on the decision in #3189 and say that this should be enabled as a warning. Maybe?

I have looked through some of the warnings and have already found a breaking bug thanks to them. The variable `time` that is defined in the `doztime` app at [line 177](https://github.com/espruino/BangleApps/blob/43d7fba0ca2799dc5ee793a2f9953624c9e177e3/apps/doztime/app-bangle2.js#L177) or [line 184](https://github.com/espruino/BangleApps/blob/43d7fba0ca2799dc5ee793a2f9953624c9e177e3/apps/doztime/app-bangle2.js#L184) was presumably a `var` declaration once, and has since then been updated to a `let` declaration. This restricts the scope it is defined in, and means that it can no longer be accessed at [line 194](https://github.com/espruino/BangleApps/blob/43d7fba0ca2799dc5ee793a2f9953624c9e177e3/apps/doztime/app-bangle2.js#L177). That mistake sometimes crashes the entire app. If this warning was enabled when that code was refactored, this breaking bug likely wouldn't have been merged.


gfwilliams:
Just to be clear here, we're not saying that `var interval = undefined` will cause a warning or anything like that?

It's basically just using a variable that hasn't previously been defined with `var/let/etc` or is listed in `.eslintrc`? 

If so this sounds great. Even if these don't cause a bug, where used in loop variables, using a global rather than a locally defined variable will slow down execution.

Only thing is, this causes almost 5000 warnings (https://github.com/espruino/BangleApps/actions/runs/8055575872/job/22002743195) so if committed it'll swamp any new warnings caused by PRs and make the CI tests effectively useless, so I don't think we can commit as-is.

Ideally, we'd fix all the existing warnings as part of this PR - do you think that's a possibility? Looks like many apps have multiple warnings so at least it's not 5000 complete files

atjn:
> It's basically just using a variable that hasn't previously been defined with var/let/etc or is listed in .eslintrc?

Yes. It is only catching code that you should basically never write. Valid use cases like `var foo = undefined` are still allowed.

> Ideally, we'd fix all the existing warnings as part of this PR - do you think that's a possibility? Looks like many apps have multiple warnings so at least it's not 5000 complete files

I can absolutely fix all the warnings, but there is no way that I can properly test so many apps. You should basically be expecting a certain percent of apps to be broken, then work to fix those later. I guess we can discuss this issue also in #3189

gfwilliams:
How is this looking after [the last merge](https://github.com/espruino/BangleApps/pull/3213)? I imagine some work needs to be done to:

* fix the 'easy' cases
* ignore other cases

atjn:
@gfwilliams yes. I will note that the previous merge only had 500 warnings, and that took many more hours than I expected to complete. Therefore my scan through these 5000 warnings will have to be very quick, and will probably mostly be to ensure that we are ignoring warnings for all legitimate use cases. I hope that is acceptable.

atjn:
Looking at the warnings, it might not be as bad as it looks. We'll see how much I am able to fix.

atjn:
@gfwilliams @bobrippling this PR is now ready to be merged :)

There isn't much to say; I fixed the simple cases and exempted everything else. I also found a few bugs that I added comments for.

nxdefiant:
Is it really neccesary to increment the version number? I just had to update 20 apps that basically work same as before.

Note: "App Manager" (files) allows to change the sort order of the apps. This is lost on update.

atjn:
The version bump was requested by @gfwilliams but I agree that while it is annoying, it is probably best to bump versions when making these changes. Thankfully this will be the last big bulk update.

The app manager issue is unfortunate, but it sounds more like a bug in app manager. Maybe you can sneak in a fix for this before this update is published?

bobrippling:
I'm afraid this is a huge amount to review, it's going to take a while.

We've already had a bug report from lint changes introducing bugs - can we tweak the linter config to increase our confidence in it? For example I'd like to see a config where #3257 and similar bugs can't occur.

atjn:
@bobrippling this PR is the config tweak that ensures #3257 and similar bugs can't occur :)

Unfortunately I think it is near impossible to make a change like this without introducing a few bugs. But on the flip side it has also found many breaking bugs in existing code, and will ensure that these errors are not introduced in the future. So I think it is a net-positive.

gfwilliams:
Ok, just been through - I think I spotted one potential breakage in apps/diract/diract.js (where the variable's value should have been kept between calls) but otherwise it looks good. I guess there may still be issues but at least I think having run through it should be reasonably unlikely.

Also spotted sometimes when a JS file that wasn't for Bangle.js changed the version was bumped when not needed, but that's minor.

Just merged - thanks for all your work on this!

atjn:
> Also spotted sometimes when a JS file that wasn't for Bangle.js changed the version was bumped when not needed, but that's minor.

Hmm yeah I didn't think of that when designing the bulk updater.

PR #3215, opened 2024-02-26
messagegui: swipe up/down on msg goes to newer/older msg

thyttan:
Swipe up/down on a shown message to show newer/older message.

I would personally rather map up/down to older/newer. But I think up/down mapped with newer/older makes more sense with how 'tick' and 'cross' swipes are done and how the messages are presented in the scroller list. 

Thoughts? @gfwilliams @bobrippling

The change can be tried by installing from my app loader: https://thyttan.github.io/BangleApps/?q=messagegui

### DIFF
--- a/apps/messagegui/ChangeLog
+++ b/apps/messagegui/ChangeLog
@@ -102,3 +102,5 @@
 0.73: Add sharp left+right nav icons
 0.74: Add option for driving on left (affects roundabout icons in navigation)
 0.75: Handle text with images in messages list by just displaying the first line
+0.76: Swipe up/down on a shown message to show the next newer/older message.
+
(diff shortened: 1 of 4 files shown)


gfwilliams:
I'm happy with this idea - but personally I'd prefer if the `udHandler` code was just in the `swipe` handler? I think it'd be a lot more readable.

thyttan:
@gfwilliams I did the change you suggested. Also refactored the function getting the index to a variable, see https://github.com/espruino/BangleApps/pull/3215/commits/66d422e3fd47c43945776b17693e364edf51c67b. Tried it on my watch and it works still.

gfwilliams:
Thanks! That looks good to me...

thyttan:
I'll run this for a day or two and mark as ready for review if everything seems good.

thyttan:
I believe this is fine.

bobrippling:
Yeah looks good to me too - I don't use `messagegui` but the change sounds fair.

PR #3217, opened 2024-02-27
swipeinv: new app

### LINKED ISSUE #3048: [calendar] Swipe actions seem wrong in calendar app

nxdefiant:
Attempts to fix #3048 as an alternative to #3202. Up for discussion.

Inverts swipe direction globally or per app, see settings. If global inversion is enabled, you can unselect the inversion per app and vice versa.

Ping @jt-nti @thyttan


jt-nti:
This looks like a more general version of the work around I've been using:— an app that specifically reverses the calendar swipes :)

I'm beginning to think that the general problem (consistent and configurable scroll directions) cannot easily be fixed in a seperate app, if at all. This app would definitely work around #3048, so that's good, but it has limitations:

- installing a separate app is not an obvious mechanism for configuring scroll directions
- it reverses both directions for any swipe interaction in an app (apps may use swipes differently for different interactions)

I still think it would be useful though, especially since it doesn't seem like a common problem,— the Desktop Launcher scrolls the way I expect for example.

To make consistent and configurable scrolling possible more generally, I think adding `hscroll` and `vscroll` interface types to [Bangle.setUI](http://www.espruino.com/Reference#l_Bangle_setUI) that can be configured with a system setting might work, i.e. to distinguish scrolling from other direction specific swipe interactions in apps.

thyttan:
>- installing a separate app is not an obvious mechanism for configuring scroll directions

I can see that that one might expect to find this kind of function as a setting in many cases. Proof being the different times the subject has come up. But the path with a separate app is probably a neat solution to the problem. To me it feels a little 'unix-like' which I'm sympathetic to - one problem one tool.

>- it reverses both directions for any swipe interaction in an app (apps may use swipes differently for different interactions)

The code that @nxdefiant has written here can be modified to give ability to set, per app, if both or just one axis should be inverted.

nxdefiant:
> The code that @nxdefiant has written here can be modified to give ability to set, per app, if both axes or just one should be inverted.

Sure it can be, but I would leave it as it is for now and wait for the feature request if it ever happens.

thyttan:
I tried it on my watch now. It works for apps using `Bangle.setUI` as expected.

Some additional thoughts:

Some apps use `Bangle.on("swipe", ()=>{...})` so for those it doesn't work - obviously. I don't know if those could be handled through `swipeinv` as well with additional logic. Otherwise many could be refactored to use `Bangle.setUI` which would probably be a good thing anyway (could be done over time, no need to do it now). But I think in some cases `Bangle.setUI` has already been called and the app adds on a swipe handler with a `Bangle.on` call. Maybe those can be refactored as well.

`quicklaunch`, `backswipe` and ´clock_info` uses `Bangle.on` and I don't think it will work refactoring them to `Bangle.setUI` since that would (I'm pretty sure - didn't test yet) remove the underlying `Bangle.setUI` called by the foreground app/clock face.

jt-nti:
It definitely seems like a useful work around as is, even with the limitations, so merging now makes sense to me.

nxdefiant:
ok, lets merge.

bobrippling:
Sounds good, thanks for the change

PR #3218, opened 2024-02-28
linuxclock - Bugfix Add 1 to month received from Date.getMonth (0..11)

signaleleven:
Date.getMonth() returns an integer from 0 to 11, so to display the date I think it's better to add 1.

I did not try to install the app by deploying the full website but I did test the simple change by modifying the app directly with the IDE.

For such a simple change I hope it's enough.


signaleleven:
Sorry, just saw the failure in the build. I bumped the version in metadata.js

bobrippling:
LGTM, nice spot!

PR #3219, opened 2024-02-28
rep: fix interface.html

bobrippling:
`reps` is global to `interface.html`, no longer assigned after fcebbf3ef4d4 / #3110. This resolves that issue


bobrippling:
Minor change to `interface.html`, no code changes, safe to merge

PR #3220, opened 2024-02-28
sched: permit buzz count to be infinite

bobrippling:
... so an alarm will buzz until it is stopped by the user.

Also updated `sched`'s readme to align it with the latest code.

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -23,3 +23,4 @@
 0.20: Alarm dismiss and snooze events
 0.21: Fix crash in clock_info
 0.22: Dated event repeat option
+0.23: Allow buzzing forever when an alarm fires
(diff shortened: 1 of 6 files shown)


bobrippling:
@thyttan does this look ok to merge?

thyttan:
Currently trying it on my watch.

- The main functionality of forever buzzing seems to work.
- Snoozing seems to work.
- I can't change the `Buzz count` settings from `null` - the sub-menu will not be displayed.
-  (also, instead of a user facing ´null´ I think it should say something else? could we achieve a `∞` infinity sign?)

image

(I have fastload utils installed)

bobrippling:
Great spot, thank you - I believe 6aa0b8d will fix that (and display the count as "Forever")

thyttan:
That fixed it for me - nice!

I think this is one of the apps/libs @gfwilliams should see the PR before we merge - so I'll not merge now.

bobrippling:
Sounds good, thanks for the checks!

gfwilliams:
Yes, this all looks good to me - thanks!

bobrippling:
Thanks both!

PR #3222, opened 2024-02-29
autoreset: bootloader app - universal timeout to clock face

thyttan:
Can be installed from my app loader: https://thyttan.github.io/BangleApps/?q=autoreset


bobrippling:
LGTM - I was thinking of something similar recently, but more of a watchdog timer. I woke up and my watch had hung (I have a nightly backup and I suspect the timing had interrupted it mid-draw, so it didn't get chance to schedule another draw).

I wonder whether something like this could help with that, some way of detecting if there's not been a draw for 10 minutes. I'll have a think anyway

PR #3224, opened 2024-02-29
multitimer: add setting for initial screen

bobrippling:
This also updates `multitimer.json` to hold settings as well as stopwatches.

@frigis1 is this change ok by you? No change in behaviour by default

### DIFF
--- a/apps/multitimer/app.js
+++ b/apps/multitimer/app.js
@@ -328,21 +328,22 @@ function editTimer(idx, a) {
   setUI();
 }
 
-function readAndConvJson() {
-  let json = require("Storage").readJSON("multitimer.json", true);
+function readJson() {
+  let json = require("Storage").readJSON("multitimer.json", true) || {};
 
   if (Array.isArray(json)) {
     // old format, convert
     json = { sw: json };
     require("Storage").writeJSON("multitimer.json", json);
   }
+  if (!json.sw) json.sw = [];
 
   return json;
 }
 
 function drawSw() {
   layer = 1;
-  const sw = readAndConvJson().sw;
+  const sw = readJson().sw;
 
   function updateTimers(idx) {
     if (!timerInt1[idx]) timerInt1[idx] = setTimeout(function() {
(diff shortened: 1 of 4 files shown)


thyttan:
My fw2v21.25 bangle starts loading the app, but then nothing seems to happen:

image

bobrippling:
Interesting - I found an old bug where the scroller wasn't saved, which I've fixed but I don't see how that would cause it to crash on startup. Do you still see it? And do you have anything in `multitimer.json` / would you mind sharing the [console output](https://www.espruino.com/ide/) when you start the app?

thyttan:
```
Uncaught Error: Can't read property 'initialScreen' of undefined
 at line 728 col 26 in multitimer.app.js
switch (readAndConvJson().initialScreen) {
```

I don't really understand the error yet, `readAndConvJson()` looks like it should be defined. What I don't find is that it should have a `initialScreen` property. But that's not what it complains about?

apps/multitimer/app.js
```diff
@@ -328,9 +328,21 @@ function editTimer(idx, a) {
   setUI();
 }
 
+function readAndConvJson() {
+  let json = require("Storage").readJSON("multitimer.json", true);
```
nxdefiant, line 332: Probably missing a default value e.g. || {}

nxdefiant:
Yes, the error means that readAndConvJson() returns undefined.

apps/multitimer/app.js
bobrippling: Thank you!

bobrippling:
Doh - yes I always tested with a file present, hadn't considered that if a user's never created a stopwatch they won't have the file. Pushed the fix (27cfb0e)

frigis1:
@bobrippling I'm afraid I haven't had time at all to update my stuff, so any fixes and updates are not only welcome but also greatly appreciated. Thank you!

thyttan:
Tried it now without running into bugs. Do you feel like it's merge ready?

bobrippling:
Thank you, yes I feel so - think it's taken a Saturday brain to fix it!

thyttan:
Ok - thanks then! :)

thyttan:
Oops - hadn't tried the settings page: 

```
 ____                 _
         |_| espruino.com
 2v21.25 (c) 2023 G.Williams
Multi Timer settings error: Error: Error: Can't read property 'initialScreen' of undefined
>g.dump()
```
image

bobrippling:
Doh! Great spot - pushed a fix in #3226 (quite pleased with that rhyme)

PR #3225, opened 2024-03-01
Fixes to Catalan locale: right icon for all regions, shorter long dat…

jordimas:
…a pattern and fix lang code


bobrippling:
Thanks!

PR #3227, opened 2024-03-02
sched: add day selection to interface.html

bobrippling:
This gives `sched`'s `interface.html` the ability to select which days an alarm will trigger on.

Deployed on [my gh-pages](https://bobrippling.github.io/BangleApps/?q=sched)

### DIFF
--- a/apps/sched/interface.html
+++ b/apps/sched/interface.html
@@ -163,6 +163,26 @@ function renderAlarm(alarm, exists) {
   tr.appendChild(tdTime);
   tdTime.appendChild(inputTime);
 
+  const tdDays = document.createElement('td');
+  tr.appendChild(tdDays);
+  const selectDays = document.createElement('select');
+  selectDays.multiple = true;
+  selectDays.classList.add('form-input');
+  selectDays.dataset.uid = alarm.id;
+  const days = ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"];
+  const options = days.map((day, i) => {
+    const option = document.createElement('option');
+    option.value = day;
+    option.text = day;
+    option.selected = alarm.dow & (1 << i);
+    selectDays.appendChild(option);
+    return option;
+  });
+  selectDays.onchange = (e => {
+    alarm.dow = options.reduce((bits, opt, i) => bits | (opt.selected ? 1 << i : 0), 0);
+  });
+  tdDays.appendChild(selectDays);
+
   const tdSummary = document.createElement('td');
   tr.appendChild(tdSummary);
   const inputSummary = document.createElement('input');
@@ -320,6 +340,7 @@ function onInit() {
         <tr>
           <th>Type</th>
           <th>Date/Time</th>
+          <th>Days</th>
           <th>Summary</th>
           <th>On?</th>
           <th></th>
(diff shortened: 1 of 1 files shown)


thyttan:
On my laptop it's hard to make anything out of the days section:
image

bobrippling:
That's fair - we could make the selection a little wider? How do you feel about its height?

thyttan:
It might be better if two days fit the display box. But currently I don't understand how to toggla a day? If I could see more of the boxes content I might have understood though.

nxdefiant:
I would say we give each entry two rows, that would give us additional horizontal space.

nxdefiant:
I meant the entire scheduler entry: type, date, day of week, message, on/off and deleted could probably be split into two lines so each entry can get wider.

thyttan:
Don't know if you want any feedback already, but here's a screenshot from my laptop after https://github.com/espruino/BangleApps/pull/3227/commits/1c620878cf3a1dced23678d8db38f36d4f6084d4 :

image

bobrippling:
I'm interested in what you think now in fact - we can rowspan a few of the cells to give this effect, although it looks a bit more noisy than the original 1-row approach:

image

Original:

image


Perhaps we stick with 1 row and just buff the height (and width) of the day selection, what do you both think?

thyttan:
Just looking at your screenshot with the changes, could the delimiters for the different alarms be as fat as the delimiter line under the header. Then if possible make that top delimiter even fatter? I think that could be nice.

I agree it currently looks a little more cluttered, but I still prefer it since I can actually see the info in days box. Extra space in the Summary field is also welcome!

nxdefiant:
How about: Row 1: type, date, summary
Row 2: options.

Also I think that [select2](https://select2.org/getting-started/basic-usage) would make the better select box then spectre.

bobrippling:
Pushed up some changes to declutter the table borders and align the type - I think `select2` is nice but let's see about merging the functionality change and we could discuss with @gfwilliams about bringing in a library or more thorough UI changes.

What do you think about the current layout? We could move options to the second row if this is a bit cluttered still.

image

nxdefiant:
Basically ok with me. But the on/off toggle could be below the summary input, which would mean the summary could take more space.

bobrippling:
Nice idea, let me see what I can rustle up

bobrippling:
Ok, what do you think of this edition?

image

nxdefiant:
Looks like the scrollbar is somehow missing

bobrippling:
Phew about the scrollbar! Thanks for the save for events & timers. I've also adjusted the layout similarly to your idea above - shifted the options around. Let me know what you both think:

image

nxdefiant:
I would put the date up, I believe it should be the first item since it is the most interesting one and probably the first to be edited. Any other opinions on this?

bobrippling:
Sorted, tada:

image

nxdefiant:
I hope I'm not pushing your patience too far, but maybe the type (event, alarm, timer) to the left?

bobrippling:
Not a problem, easy change to make

nxdefiant:
ok to merge :)

bobrippling:
Great! Here's the final version now it's built, shall merge shortly

image

bobrippling:
Thanks for helping hash out the design, both of you

gfwilliams:
Looks great - thanks everyone!

PR #3228, opened 2024-03-02
sleeplog: only write log if we have something to write

bobrippling:
@storm64 is this a change you'd be happy with? I found my storage would populate with files containing an empty array, which slowed down listing files (e.g. connecting from the app loader)


storm64 (approved):
This is a really nice and simple check, should work perfect. 👌
It should even prevent creation of possible unneeded files, if a entry at around datetime 0 has been triggered.

And thanks for letting me know. 😊

bobrippling:
My pleaure, thanks for the review!

PR #3229, opened 2024-03-02
a_dndtoggle: add settings page

thinkpoop:
Adds a settings page with "Quiet Mode": "Alarms", "Silent"
Adds a newline in the update message (e.g. "Noisy -> \nSilent")

Default setting is "Silent" (2) to match the old code.

I also considered changing "Noisy" to "Off", so it matches what Bangle's settings calls it and maybe language translations. But I left it as-is for now. 

App Link:
https://thinkpoop.github.io/BangleApps/?id=a_dndtoggle


apps/a_dndtoggle/a_dndtoggle.app.js
```diff
@@ -6,11 +6,16 @@ let current = 0|bSettings.quiet;
...
+		require('Storage').readJSON("a_dndtoggle.default.json", true) || {},
+		require('Storage').readJSON("a_dndtoggle.settings.json", true) || {}
+  );
+
 console.log("old: " + current);
 
 switch (current) {
 	case 0:
-		bSettings.quiet = 2;
+		bSettings.quiet = dndSettings.mode;
```
bobrippling, line 18: I'll leave it up to you, but we could save having an extra file (and the extra time to read it) and apply the default here and in settings, like so:

```suggestion
		bSettings.quiet = dndSettings.mode || 2;
```

apps/a_dndtoggle/a_dndtoggle.app.js
```diff
@@ -29,15 +34,15 @@ switch (current) {
...
 
 
 function exitApp(){
 
-require("Storage").writeJSON("setting.json", bSettings);
-// reload clocks with new theme, otherwise just wait for user to switch apps
-
+  require("Storage").writeJSON("setting.json", bSettings);
+	// reload clocks with new theme, otherwise just wait for user to switch apps
+
```
bobrippling, line 45: Looks like some whitespace change has accidentally made it it

```suggestion
require("Storage").writeJSON("setting.json", bSettings);
// reload clocks with new theme, otherwise just wait for user to switch apps

```

bobrippling:
Thanks for the change, just two little comments

apps/a_dndtoggle/settings.js
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    function buildMainMenu(){
+      const modes = [/*LANG*/"Noisy",/*LANG*/"Alarms",/*LANG*/"Silent"];
+      let mainmenu = {
+        '': { 'title': 'A_DND Toggle' },
+        '< Back': back,
+        /*LANG*/"Quiet Mode": {
+          value: settings.mode || 2,
+          min: 1, max: modes.length - 1,
```
bobrippling, line 17: To check I understand here, do we want mode to go between 1 (`Noisy`) and 3 (`Silent`)? Or do we want it to be 0-indexed, so 0 (`Noisy`) up to 3 (`Silent`) ?

bobrippling:
Thanks for the update, sounds good! I'd just spotted the settings page seems to just go between 1 and 2, so I'm not sure we get the first entry in the mode array, if you see what I mean

apps/a_dndtoggle/settings.js
thinkpoop: 0 Noisy is just there for being a placeholder.
1 - Alarms, 2 - Silent are the actual choices, so min = 1 and max = 2.

Mode could be an object with {1:"Alarms",2:"Silent"} but the array pattern is what I saw in other spots, so I went with that.

thinkpoop: _(No rush, I'm just adding to make sure I explain the best I can. I probably should've documented better.)_

The Bangle quiet mode options are: `0-Off`, `1-Alarms`, `2-Silent`

The current a_dndtoggle app is hard coded to toggle between `0-Off` and `2-Silent`. 
- If current mode is `1-Alarms` or `2-Silent`, it goes to `0-Off`. 
- If current mode is `0-Off`, it goes to `2-Silent`.

This new settings page lets you pick if it should go to `1-Alarms` or `2-Silent`.

I think my options were:
1. Have an array of `["Alarms", "Silent"]`, then add `1` to the result
2. Have an array of `[null,"Alarms","Silent"]`, and not allow picking `0`

I went with option 2, but instead of "null" I left it labelled as Noisy.

I've added a comment in settings.js and a note in the readme about the settings page options.

Let me know if I should add further comments, or if there's a better way.

bobrippling: No that's not a problem, it all makes sense! Thanks for adding the comments though, all very clear now - option 2 is easier for us to map across to the bangle options, or at least in my head :)

bobrippling:
Thanks!

PR #3230, opened 2024-03-03
Cards App added brightness option

phrogg:
Added an option to use full brightness when showing code

### DIFF
--- a/apps/cards/app.js
+++ b/apps/cards/app.js
@@ -18,16 +18,8 @@ Bangle.drawWidgets();
 let brightness;
 
 function loadBrightness() {
-  try {
-      const d = require('Storage').readJSON("setting.json", 1) || {};
-      brightness = Object.assign({
-          'brightness': 0.1
-      }, d || {});
-      return d;
-  } catch(e){
-    console.log(e.toString());
-    return;
-  }
+    const getBrightness = require('Storage').readJSON("setting.json", 1) || {};
+    brightness = getBrightness.brightness || 0.1;
 }
 
 //may make it configurable in the future
@@ -105,6 +97,7 @@ function printLinearCode(binary) {
 }
 
 function showCode(card) {
+  // set to full bright when the setting is true
   if(settings.fullBrightness) {
     Bangle.setLCDBrightness(1);
   }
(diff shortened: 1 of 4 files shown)


apps/cards/app.js
```diff
@@ -14,6 +14,22 @@
...
+let brightness;
+
+function loadBrightness() {
+  try {
+      const d = require('Storage').readJSON("setting.json", 1) || {};
+      brightness = Object.assign({
+          'brightness': 0.1
+      }, d || {});
+      return d;
+  } catch(e){
```
bobrippling, line 27: What exceptions were you seeing here? Was it from an older version of the code? `readJSON` won't through with `1` as the second argument, and nothing else throws in this code block

apps/cards/app.js
```diff
@@ -14,6 +14,22 @@
...
 
+// get brightness
+let brightness;
+
+function loadBrightness() {
+  try {
+      const d = require('Storage').readJSON("setting.json", 1) || {};
+      brightness = Object.assign({
+          'brightness': 0.1
+      }, d || {});
```
bobrippling, line 25: To save on holding all of settings in memory, we could do:

```javascript
const settings = require('Storage').readJSON("setting.json", 1) || {};
brightness = settings.brightness || 0.1;
```

And use just `brightness` below

bobrippling:
Thanks for the change! A few small suggestions

apps/cards/app.js
phrogg: Sure sounds good to me. Did not do much too developing until now. Tips are very welcome, thanks :) You cannot however use the variable name `settings` as it is already used for the settings of the cards app itself.

apps/cards/app.js
bobrippling: Not a problem, sounds like we can drop the try-catch :)

apps/cards/app.js
bobrippling: Cool, that's fair, we could call it whatever you like - give me a shout if you have any trouble making the changes

bobrippling:
Thanks for the adjustments, looks great!

PR #3232, opened 2024-03-03
widalarmeta: recheck latest alarm when alarms change

bobrippling:
This allows `widalarmeta` to update itself when alarms change, which happens if we modify them and use fastload utils to return to the clock face without a full widget reload.

@nxdefiant is this change alright by you?

### DIFF
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -24,3 +24,4 @@
 0.21: Fix crash in clock_info
 0.22: Dated event repeat option
 0.23: Allow buzzing forever when an alarm fires
+0.24: Emit alarmReload when alarms change (used by widalarm)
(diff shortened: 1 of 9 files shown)


nxdefiant:
Looks ok to me. I'm just wondering how you triggered this one? I have fastload installed and both the alarm app and the scheduler do a full restart on my watch.

The only use case where I have to manually restart is when editing the alarms using interface.html

bobrippling:
Ah sorry I didn't mention, I use [`multitimer`](https://espruino.github.io/BangleApps/?id=multitimer) which fastloads back to the clock, you're right - I don't see it with the alarm app.

Adding these here for completeness:
Steps to reproduce:
- Create an alarm (any method is fine)
- Observe `widalarmeta` correctly picks it up
- Disable the alarm in `multitimer`, fastload back to clock
- `widalarmeta` shows the time to the disabled alarm

PR #3233, opened 2024-03-03
widlockunlock: disable tap-to-lock if back button is present

bobrippling:
This avoids having to try to dodge the lock/unlock widget when going back

### DIFF
--- a/apps/widlockunlock/ChangeLog
+++ b/apps/widlockunlock/ChangeLog
@@ -1,2 +1,3 @@
 0.01: First commit
 0.02: Add tap-to-lock functionality
+0.03: Disable tap-to-lock if back button is present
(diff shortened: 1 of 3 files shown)


bobrippling:
@gfwilliams what do you think about stopping event propagation when the [back button's pressed](https://github.com/espruino/Espruino/blob/c1824f087b2618fb53711482c49323da4b6e4d67/libs/js/banglejs/Bangle_setUI_Q3.js#L117-L118)?

gfwilliams:
Looks good - maybe the description should mention it won't work if 'back' is showing though?

gfwilliams:
And sorry, yes - that's a good idea. We set the `handled=true` so that was obviously the intention, but at that point we didn't have stopPropagation. I'll add it now

bobrippling:
> Looks good - maybe the description should mention it won't work if 'back' is showing though?

Good idea - [done](https://github.com/espruino/BangleApps/pull/3233/commits/000543edf30abe5d5f3ab77cac7712bd37c4f431)

> at that point we didn't have stopPropagation. I'll add it now

Thanks!

gfwilliams:
Thanks!

PR #3234, opened 2024-03-03
smpltmr: reset clkinfo to "main" timer after no interaction

bobrippling:
@gfwilliams @sir-indy I'd like `smpltmr`'s `clkinfo` menu to reset to the "main" menu (showing the time remaining on the timer) after no interaction from the user.
To do this requires shuffling around some clkinfo code to create an API for selecting a given menu. If this seems reasonable to you both, I'll give it a test.

([git's `diff.colormoved`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---color-movedltmodegt) might be useful for seeing which lines have moved but are otherwise unchanged)

- [x] Test
- [x] Update [clock info docs](https://www.espruino.com/Bangle.js+Clock+Info)

### DIFF
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -294,9 +294,9 @@ exports.addInteractive = function(menu, options) {
     if (redraw) options.redraw();
   };
   const focus = (redraw) => {
+    Bangle.CLKINFO_FOCUS=true;
     if (!options.focus) {
       options.focus=true;
-      Bangle.CLKINFO_FOCUS=true;
       const itm = menu[options.menuA].items[options.menuB];
       if (itm.focus && itm.focus(options) === false)
         redraw = false;
@@ -362,6 +362,7 @@ exports.addInteractive = function(menu, options) {
 
     return true;
   };
+  if (options.focus) focus();
   delete settings; // don't keep settings in RAM - save space
   return options;
 };
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -292,13 +292,17 @@ exports.addInteractive = function(menu, options) {
     if (itm.blur) itm.blur(options);
   };
   const focus = (redraw) => {
+    let shown = false;
     if (!options.focus) {
       options.focus=true;
       Bangle.CLKINFO_FOCUS=true;
+      shown = true;
+    }
+    if (redraw) options.redraw();
+    if (shown) {
       const itm = menu[options.menuA].items[options.menuB];
       if (itm.focus) itm.focus(options);
     }
-    if (redraw) options.redraw();
   };
   let touchHandler, lockHandler;
   if (options.x!==undefined && options.y!==undefined && options.w && options.h) {
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -284,38 +284,39 @@ exports.addInteractive = function(menu, options) {
     E.stopEventPropagation&&E.stopEventPropagation();
   }
   Bangle.on("swipe",swipeHandler);
+  const unfocus = () => {
+    options.focus=false;
+    delete Bangle.CLKINFO_FOCUS;
+    options.redraw();
+  };
+  const focus = (redraw) => {
+    options.focus=true;
+    Bangle.CLKINFO_FOCUS=true;
+    if (redraw) options.redraw();
+  };
   let touchHandler, lockHandler;
   if (options.x!==undefined && options.y!==undefined && options.w && options.h) {
     touchHandler = function(_,e) {
       if (e.x<options.x || e.y<options.y ||
           e.x>(options.x+options.w) || e.y>(options.y+options.h)) {
-        if (options.focus) {
-          options.focus=false;
-          delete Bangle.CLKINFO_FOCUS;
-          options.redraw();
-        }
+        if (options.focus)
+          unfocus();
         return; // outside area
       }
       if (!options.focus) {
-        options.focus=true; // if not focussed, set focus
-        Bangle.CLKINFO_FOCUS=true;
-        options.redraw();
+        focus(true);
       } else if (menu[options.menuA].items[options.menuB].run) {
         Bangle.buzz(100, 0.7);
         menu[options.menuA].items[options.menuB].run(); // allow tap on an item to run it (eg home assistant)
       } else {
-        options.focus=true;
-        Bangle.CLKINFO_FOCUS=true;
+        focus();
       }
     };
     Bangle.on("touch",touchHandler);
     if (settings.defocusOnLock) {
       lockHandler = function() {
-        if(options.focus) {
-          options.focus=false;
-          delete Bangle.CLKINFO_FOCUS;
-          options.redraw();
-        }
+        if(options.focus)
+          unfocus();
       };
       Bangle.on("lock", lockHandler);
     }
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -284,12 +284,12 @@ exports.addInteractive = function(menu, options) {
     E.stopEventPropagation&&E.stopEventPropagation();
   }
   Bangle.on("swipe",swipeHandler);
-  const unfocus = () => {
+  const blur = () => {
     options.focus=false;
     delete Bangle.CLKINFO_FOCUS;
     options.redraw();
     const itm = menu[options.menuA].items[options.menuB];
-    if (itm.unfocus) itm.unfocus(options);
+    if (itm.blur) itm.blur(options);
   };
   const focus = (redraw) => {
     if (!options.focus) {
@@ -306,7 +306,7 @@ exports.addInteractive = function(menu, options) {
       if (e.x<options.x || e.y<options.y ||
           e.x>(options.x+options.w) || e.y>(options.y+options.h)) {
         if (options.focus)
-          unfocus();
+          blur();
         return; // outside area
       }
       if (!options.focus) {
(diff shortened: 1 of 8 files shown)


gfwilliams:
Am I right in thinking it's setting a global `Bangle.CLKINFO` for each call to `addInteractive`, so it wouldn't work in the case where you have two or more ClockInfos (LCDclock, pebble++, etc?)

addInteractive returns an object that already has `options.setItem` in it, so isn't there a way we could just pass that object in when we call `show/hide/get`, without having to make much in the way of changes?

But I think this need a bit more thought - what happens if you have multiple clockinfos, but I also think maybe this should only auto-reset to the timer if you're actually on the 'Timer' page? Like if you're showing something else completely, it shouldn't just always go back to the timer.

Another thought is maybe this should be an option in `clock_info` settings? Like whether you do want to auto-change, after how long, and what to?

I can imagine cases where someone might want to maybe default to heart rate or something, rather than just the timer?

bobrippling:
> Am I right in thinking it's setting a global `Bangle.CLKINFO` for each call to `addInteractive`, so it wouldn't work in the case where you have two or more ClockInfos (LCDclock, pebble++, etc?)

That's a good point, yes - multiple clockinfos wouldn't work very well!
 
> addInteractive returns an object that already has `options.setItem` in it, so isn't there a way we could just pass that object in when we call `show/hide/get`, without having to make much in the way of changes?

This would be ideal, but unfortunately that's only for the client of the clockinfo, the clockinfo itself doesn't have a way of getting hold of `setItem`. Perhaps exposing existing clockinfos (i.e. the return values from `addInteractive`) on `Bangle.CLKINFO` would work, removing them again when `remove()` is called on the clockinfo object.

> But I think this need a bit more thought - what happens if you have multiple clockinfos, but I also think maybe this should only auto-reset to the timer if you're actually on the 'Timer' page? Like if you're showing something else completely, it shouldn't just always go back to the timer.

Yes, I agree - I believe this to be a small tweak, we can cancel the timeout in `hide()`, meaning it only happens when that particular clkinfo's page & item are still visible:

https://github.com/espruino/BangleApps/blob/8961b8efdf189873def981ae0652d203de0d6181/apps/smpltmr/clkinfo.js#L82-L89

> Another thought is maybe this should be an option in `clock_info` settings? Like whether you do want to auto-change, after how long, and what to?

> I can imagine cases where someone might want to maybe default to heart rate or something, rather than just the timer?

bobrippling:
Yeah the lock event isn't a bad idea actually, ties it in with other possible events that all might change when the watch locks.

I like your idea of passing options to `show` (and also to `hide` and `run`) to permit that flexibility. I'd forgot the clockinfos are saved, perhaps that would work in fact, and not require a change to the module. Shall give it a whirl, thank you

gfwilliams:
Thanks! I just had another thought when you wrote that - the `clock_info` module could check for and call a `focus/unfocus` function on the active clock_info?

I guess it might be a useful addition so that maybe the icon/text can be changed when the thing is actually ready to accept being pressed? Or in this case, so it could change back to the default item when unfocussed?

bobrippling:
To check I've understood - you're suggesting we add extra (optional) `focus` and `unfocus` methods, and that's where (in this case) we could refocus the "main" item we'd like to show?

bobrippling:
I've changed the approach to what we discussed, shall be testing shortly - let me know what you think

apps/clock_info/lib.js
```diff
@@ -284,38 +284,49 @@ exports.addInteractive = function(menu, options) {
...
-        options.redraw();
+        focus(true);
       } else if (menu[options.menuA].items[options.menuB].run) {
         Bangle.buzz(100, 0.7);
-        menu[options.menuA].items[options.menuB].run(); // allow tap on an item to run it (eg home assistant)
+        menu[options.menuA].items[options.menuB].run(options); // allow tap on an item to run it (eg home assistant)
       } else {
-        options.focus=true;
-        Bangle.CLKINFO_FOCUS=true;
+        focus();
```
bobrippling, line 322: I'm not sure if this is needed here - `options.focus` is already `true`, was it just to set `CLKINFO_FOCUS` ?

gfwilliams: Good spot! It looks like this was added by @glemco to support quicklaunch/widhid - but it seems `options.focus=true` was in there already at that point, so this it does seem redundant.

We could take it out but we'd have to change the other apps, and I guess they are kind of independent of the clock_info module, so they'd have to do something like:

```
var clockinfo_focus = false;
try {
  clockinfo_focus = require("clock_info").clockInfos.some(i=>i.focus);
} catch { /* no module */ }
```

To catch the case that the module wasn't installed, which is not ideal I guess.

apps/smpltmr/clkinfo.js
```diff
@@ -71,13 +71,19 @@
     ]
   };
 
+  const restoreMainItem = function(clkinfo) {
+    clkinfo.menuB = 0;
+    // clock info redraws after this
```
gfwilliams, line 76: I just saw you have:

```
 const blur = () => {
    options.focus=false;
    delete Bangle.CLKINFO_FOCUS;
    options.redraw();
    const itm = menu[options.menuA].items[options.menuB];
    if (itm.blur) itm.blur(options);
  };
```

So does it actually redraw?

gfwilliams:
This looks great - it's nice having blur/focus and removing the duplication.

Apart from the redraw comment, this looks good - maybe you could just update `apps/clock_info/README.md` to mention the new functions and the fact that show/hide now have 'options' passed?

apps/clock_info/lib.js
bobrippling: Cool, and `options.focus = true` goes all the way back to the [introduction of focus](https://github.com/bobrippling/BangleApps/commit/11f63df3b1337c576701e8ee823d5745e1013749#diff-00d918fab679ddd859bdcd79e6c3e86474e1c74f8a25cc9c64d36784517f5146R220), so I reckon it's safe to remove both `focus` & `CLKINFO_FOCUS` being set here (or now, the call to `focus()`).

I don't understand the removal / other apps depending on this - `options.focus` is true at that point, so taking out the assignment won't have behaviour any apps can observe, right?

apps/smpltmr/clkinfo.js
bobrippling: Yes - I ran out of time to push this morning, [pushed now](https://github.com/espruino/BangleApps/pull/3234/commits/736b9f0433ad76d2867bd2d37401c1cc382d5b42)

bobrippling:
Thanks, quite pleased with the solution we've come up with. I've updated the docs, both in the README and on the docs repo too

apps/clock_info/lib.js
glemco: What @gfwilliams wanted to point out is that the global `CLKINFO_FOCUS` is independent on the clkinfo module (not the cleanest thing I have to admit). In this way apps like quicklaunch and widhid can check it without needing to know if the module is installed or not.
The main idea behind needing it in quicklaunch is that the app uses swipes to do actions, now if the clkinfo is focused, the swipe would both go to the next clkinfo and open the action from quicklaunch, which is not exactly what a user wants.
If `options.focus` carries the same information we can indeed use that, but there seems to be no better way than using that try/catch .
We could have something like a global variable (or handler/lib, you pick what is cleanest) where apps register themselves as the ones needing exclusive swipe control, other apps would check that and in this way we are not directly dependent on clkinfo and the functionality can be extended to other apps needing it.
Could be something as simple as:
```javascript
Bangle.EXCLUSIVE_SWIPE == null; // no app registered
Bangle.EXCLUSIVE_SWIPE == 'clkinfo'; // the clkinfo module is currently registered
function registerExclusiveSwipe (app) {
  Bangle.EXCLUSIVE_SWIPE = app;
}
function IsSwipeExclusive () {
  return Bangle.EXCLUSIVE_SWIPE == null;
}
```

bobrippling: Ah yes, I think there was misunderstanding on my part in that case, in this part of code, both `options.focus` and `Bangle.CLKINFO_FOCUS` will already be true, so there's no need to set them both to true again here (see 0943d202), but we need to preserve `CLKINFO_FOCUS`, which is done.

Slightly related (and I'd move this to another PR), but we could have [`E.stopPropagation`](https://github.com/espruino/Espruino/commit/35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b) in the clkinfo handlers now, which would avoid other apps needing to check `CLKINFO_FOCUS` at all, what do you both think?

gfwilliams:
Yes, this looks great! Thanks for all your work on this!

apps/clock_info/lib.js
glemco: That would be much neater actually, quite a nice function, I wasn't aware of it

thyttan: > , but we could have E.stopPropagation in the clkinfo handlers now, which would avoid other apps needing to check CLKINFO_FOCUS at all, what do you both think?

This works if the clock info handler is before other handlers in the related handler list. Otherwise other handlers will be called before we get to the clock info. IIrc. @gfwilliams

gfwilliams: > Slightly related (and I'd move this to another PR), but we could have [E.stopPropagation](https://github.com/espruino/Espruino/commit/35c8cb9be11ce76ea1829a4a8b40f2ab051bb60b) in the clkinfo handlers now, which would avoid other apps needing to check CLKINFO_FOCUS at all, what do you both think?

Yes, that sounds great. I think keeping `.focus` itself would be useful, but it'd be good to get rid of `CLKINFO_FOCUS`

As @thyttan says, we'd need to ensure we were first in the listener list, but I think we could handle that nicely using `prependListener` and then only adding `swipeHandler` when we're focussed (rather than always, like we do currently)

bobrippling:
Thank you too!

PR #3235, opened 2024-03-04
cscsensor: Automatically reconnect on error

nxdefiant:
Also lower connection interval for more stable connect.


bobrippling:
Nice, LGTM

PR #3237, opened 2024-03-05
Create a custom Eslint plugin

atjn:
@gfwilliams I have been working on a custom Eslint plugin for a little while. I am opening a draft PR here to hear your thoughts on it before committing too much time to it. The Eslint plugin should have its own repository and should be published to NPM, then installed here and in other repo's instead of the custom eslint configs we use right now.

I think there are many advantages to this approach. You will be able to have a single canonical description of the Espruino environment, whereas you right now have many duplicate eslint config files. Per the way I set it up, you will also be able to define more precisely what environment you want to support, as Espruino changes a lot depending on what hardware it uses.

You will be able to define different rulesets such as a BangleApps baseline. This is especially useful for devs that are developing apps outside of the official repo. They can set up Eslint with the same environment and rules as the official repo.

This also makes it possible to create custom rules. I have thought of several Espruino-specific rules that could be added, such as a check to make sure you are not relying on function hoisting, since that is not supported by Espruino. Some of the checks in the `sanitycheck` script might also be converted to Eslint plugins.

I have already defined most of the Espruino environment. The tree currently looks like:
test

The general use case is to import the `espruino` which gives you the base espruino environment. Then you also import the environment that specifies everything unique to your board, for example `bangle2`. If you have custom needs or an unsupported board, you can mix-and-match yourself. For example, `button-2` is for anything that has two buttons, and `stepper` is for anything that supports stepper functionality.

I hope that all made sense. What do you think?


gfwilliams:
I like the idea of detecting function hoisting, and maybe some other things that Espruino doesn't support or we know cause issues.

But then I just don't see there's the demand for this at the moment. The way people develop for Espruino is predominantly by running code on the device - so they're unlikely to ever really use this apart from via the BangleApps CI checks.

... but if someone's contributing a change to an app and relying purely on CI to check it's ok rather than bothering to test, that's a problem all on its own.

I did mention on your Lint thread that there are typescript decls already that could be leveraged, and even tools in the Espruino repo (like https://github.com/espruino/Espruino/blob/master/scripts/build_tern_json.js) that could serve as a way to automatically grab type info, automatically for specific Espruino build types. I feel like having a bunch of manually-written rules which are going to drift out of date over time will just do more harm than good.

If we could get the Web IDE to use eslint (currently we use jshint) with a way to automatically generate the rules, and eslint wasn't so heavyweight it made the IDE unusable on tablets/phones/chromebooks/etc then I guess this might be good, but I think that's quite a long way off.

At the end of the day, I think some linting is good, but it's definitely a case of diminishing returns.

I've merged your eslint changes, but over the last few weeks I've spent an awful lot of time looking at your PRs/issues - and don't have the time to continue at that level, when I think honestly most users of Espruino/Bangle.js would prefer that I spent my time on improvements/bugfixes on the interpreter.

atjn:
I understand if you want to focus on something else right now :)

Just to be clear, you're still interested in merging my last work on enforcing lint rules, right?

You have been extremely responsive, much more than I could ever expect. I fully understand if you wait a few days before looking over my issues and PRs because you deem something else more important. If you do that, you can naturally rate limit how much of your time I take up :)

gfwilliams:
> Just to be clear, you're still interested in merging my last work on enforcing lint rules, right?

Yes - I merged https://github.com/espruino/BangleApps/pull/3213 but I think https://github.com/espruino/BangleApps/pull/3214 now needs some work?

I just don't want to feel like there's this endless torrent of increasingly minor lint rule additions coming :)

PR #3238, opened 2024-03-05
android: posHandler-> wake + unlock + open associated app screen on android device

thyttan:
@gfwilliams @bobrippling 

This is only meant as a proof of concept. What it does is, when I do a right-swipe on a message on the Bangle, the related app screen to display the information on my android device is immediately opened. So I don't have to click the power button on the phone.

The wake + unlock part is the new functionality.

Only works if the Bangle is paired as companion device to Gadgetbridge/Android.

If we want this to go in I'll make the needed additions to the PR.

### DIFF
--- a/apps/android/boot.js
+++ b/apps/android/boot.js
@@ -315,7 +315,17 @@
   // Message response
   Bangle.messageResponse = (msg,response) => {
     if (msg.id=="call") return gbSend({ t: "call", n:response?"ACCEPT":"REJECT" });
-    if (isFinite(msg.id)) return gbSend({ t: "notify", n:response?"OPEN":"DISMISS", id: msg.id });
+    if (isFinite(msg.id)) {
+      return ()=>{
+        gbSend({ t: "notify", n:response?"OPEN":"DISMISS", id: msg.id });
+        // Wake the android device - if paired as Companion Device it will bypass keyguard and display the associated app screen directly.
+        gbSend({t:"intent", action:"android.intent.action.VIEW",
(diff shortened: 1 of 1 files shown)


gfwilliams:
That's cool - I'm all for putting this in, but I feel like it should be an option (default=off), **and** it shouldn't wake the phone for the DISMISS action? It'd also need documenting that you need to have enabled Intents in Gadgetbridge,

... although I wonder, is this an option in Gadgetbridge already? It always seemed odd to me that `OPEN` was kind of useless when your phone is off (which is most of the time)

I imagine for most people, their phone is in their pocket when they're using their watch, so something that turns the phone on automatically by default probably isn't appreciated.

thyttan:
>it shouldn't wake the phone for the DISMISS action?

Agreed!

>is this an option in Gadgetbridge already?

Not that I find when looking through at least.

>something that turns the phone on automatically by default probably isn't appreciated.

Also agreed.

thyttan:
Since the wake activity is a Gadgetbridge class it would be easy to implement this as a new special waking api-point instead of using the intents-api-point. I don't know what is preferable.

gfwilliams:
Maybe this should actually just be a Bangle.js option inside Gadgetbridge itself? It seems like it might be more sensible - since if we change Gadgetbridge to add the api and *then* change this we have to keep them in sync.

Then when we handle OPEN we can just check that setting and choose whether to wake or not?

thyttan:
That sounds better - yes

PR #3239, opened 2024-03-06
Add files via upload

naden:
fixed typos in readme, removed some unnecessary code


naden:
replaced icons with "Graphics.createImage"

bobrippling:
Looks good, would you be able to update the PR title with the changes? And it looks like there's a lint error introduced in `app.js`, with `hs` no longer being needed, if you could tweak that too?

naden:
removed no longer required import of "heatshrink"

naden:
> Looks good, would you be able to update the PR title with the changes? And it looks like there's a lint error introduced in `app.js`, with `hs` no longer being needed, if you could tweak that too?

good catch, fixed it

bobrippling:
Nice, thanks!

PR #3240, opened 2024-03-07
Support Nodejs 18

atjn:
It seems there is still a use case for nodejs 18, so this PR ensures everything runs on v18. I also downgraded the CI to v18 because as you say, it ensures the scripts work for as many people as possible :)


gfwilliams:
Thanks! As @nxdefiant noted, Debian/Ubuntu end up being way behind the times with Node.js releases

PR #3241, opened 2024-03-07
folderlauncher: add a way to move apps between folders from web interface

pavelmachek:
Web interface still can not do everything settings on phone can, still it is very useful to sort applications between folders.


bobrippling:
Thanks, LGTM - we don't need a version bump since there's no changes to JS files though, if you could roll it back?

pavelmachek:
No problem, rolled back, new version is in.

bobrippling:
Great, thanks!

PR #3242, opened 2024-03-08
regattatimer: fix settings lint

bobrippling:
No version bump: no functionality change.


bobrippling:
Fixed in 89e01a41dead7

PR #3245, opened 2024-03-08
Fixes typos and BTN1 to show launcher

jordimas:


bobrippling:
Looks good - seems there's [one warning](https://github.com/espruino/BangleApps/actions/runs/8209507908/job/22455177968?pr=3245) created in `app.js`, if you could tweak?

```
Clock rellotge file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)
```

Easiest is to move `loadWidgets()` to after `setUI` rather than in the callback

jordimas:
> Looks good - seems there's [one warning](https://github.com/espruino/BangleApps/actions/runs/8209507908/job/22455177968?pr=3245) created in `app.js`, if you could tweak?
> 
> ```
> Clock rellotge file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)
> ```
> 
> Easiest is to move `loadWidgets()` to after `setUI` rather than in the callback

Thanks!

It should be fixed now https://github.com/espruino/BangleApps/pull/3245/commits/11b0f7063b853876a127b80f7d9a2b5372a9b800

bobrippling:
Great, thanks! Nice idea with the swipe-on too

PR #3246, opened 2024-03-09
Home Assistant API Interface

ellabellla:
A Home Assistant App inspired by "Home Assistant Sensors". This app provides two features:

- Sending health, compass, accelerator, and battery information to Home Assistant
- Displaying Home Assistant state information

The app does this by connecting to Home Assistant via a rest api through the Banglejs GadgetBridge. 

The app makes requests to home assistant to render [templates](https://www.home-assistant.io/docs/configuration/templating/) and then displays the resulting information to the user. This allows the user to fetch and display any information they like from home assistant with minimal configuration and effort. On boot the app also sets up a background process that will set the state of a custom sensor defined by the user that can be tracked and viewed in home assistant.

The app also provides a interface.html that can be used to configure connection details, sensor metadata, and displayed templates through modifying the settings JSON. The settings are also validated before and after uploading to ensure the app functions correctly.

This app together with "Home Assistant" I feel provides a complete package for interfacing with home assistant. This app displays and tracks data, and the "Home Assistant" app provides controls for interacting with home assistant.


apps/hassio/interface.html
```diff
@@ -0,0 +1,106 @@
...
+          }
+        }
+
+        return true;
+      }
+
+      function getData() {
+        // show loading window
+        Util.showModal("Loading...");
+        Util.readStorage('hassio.json', data=>{
```
bobrippling, line 73: There's a little [`readStorageJSON` function](https://github.com/espruino/EspruinoAppLoaderCore/blob/bd301be3324775a8f464328ba9e34f750d503a2b/js/index.js#L266-L273) we can use here, it handles "espruino json" for you, which is a more lax format optimised for storage size :)

(see #3110 if you're interested in the details)

apps/hassio/metadata.json
```diff
@@ -0,0 +1,19 @@
...
+  "tags": "tool,sensors",
+  "supports": ["BANGLEJS2"],
+  "dependencies": {"android":"app"},
+  "interface": "interface.html",
+  "readme": "README.md",
+  "storage": [
+    {"name":"hassio.app.js","url":"hassio.app.js"},
+    {"name":"hassio.boot.js","url":"hassio.boot.js"},
+    {"name":"hassio.img","url":"hassio.img"},
+    {"name":"hassio.json","url":"hassio.json"}
```
bobrippling, line 16: For the settings, we can pop them in the `data` object, which means they won't get trashed by the app loader, for example:

https://github.com/espruino/BangleApps/blob/3f87305a2d773fd1249da95a18fabf8ca34180da/apps/worldclock/metadata.json#L17-L19

bobrippling (commented):
Looks great, given it a test and works nicely with my HA setup too. Just a couple of small suggestions :)

bobrippling:
Nice, thank you!

PR #3256, opened 2024-03-12
andark: support optional widgets loading and other settings

buzden:
I propose to add an alternative variant of an existing clock face `andark`. This face has different philosophy, because is loads widgets, but makes then be hidden by default, being shown when swiped. This is the main reason of creation an alternative clock face. Also, small change in the look is done, letters are not covered by clock hands.


bobrippling:
Looks good - what do you think about merging this into `andark` and putting the widget behind a setting? The general stance on this is that we duplicate apps for the BangleJS 1 (because of its constrained memory), but for a BangleJS 2, it's ok to make the apps a bit more flexible.

@HeiLaut would you be ok with this?

buzden:
> what do you think about merging this into `andark` and putting the widget behind a setting?

If there are no other objections and if the other change (the order of printing) is also acceptable to be added behind another setting -- yeah, I can do this.

bobrippling:
Great, we'll give @HeiLaut until the end of tomorrow to reply and then let's merge the apps together - we can always back this out if they change their mind.

HeiLaut:
Hi I have no problem with that. I'm not longer using the watch neither do I work on the click face. So I'm ok with merging them together.

bobrippling:
@buzden are you happy to merge the apps?

buzden:
> @buzden are you happy to merge the apps?

Yes, I am, but I was recently busy, so I didn't do this yet

bobrippling:
Cool, no worries!

buzden:
@bobrippling finally it's ready. I dared also to refactor the code a little bit during the addition of settings. Default settings are backwards compatible with the current state. The only non-backwards-compatible change is an addition of color indication of charging status, which I consider acceptable since you don't wear you watch during charging anyway.

bobrippling:
Great! I've given it a test and it works really nicely.

There's one small thing picked up - could you move the [`Bangle.setUI` call](https://github.com/espruino/BangleApps/blob/74d7e4d6dd69517e18f887bd6d2723f2d20b0338/apps/andark/app.js#L153) up above the [`Bangle.loadWidgets()`](https://github.com/espruino/BangleApps/blob/74d7e4d6dd69517e18f887bd6d2723f2d20b0338/apps/andark/app.js#L124-L128) / if loadWidgets statement? Then when the widgets load, they'll know if they're on a clock or not

(That's this warning: `Clock andark file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)`)

buzden:
> There's one small thing picked up - could you move the [`Bangle.setUI` call](https://github.com/espruino/BangleApps/blob/74d7e4d6dd69517e18f887bd6d2723f2d20b0338/apps/andark/app.js#L153) up above the [`Bangle.loadWidgets()`](https://github.com/espruino/BangleApps/blob/74d7e4d6dd69517e18f887bd6d2723f2d20b0338/apps/andark/app.js#L124-L128) / if loadWidgets statement? Then when the widgets load, they'll know if they're on a clock or not

Right! I've moved.

> (That's this warning: `Clock andark file calls loadWidgets before setUI (clock widget/etc won't be aware a clock app is running)`)

~~Where can I look at this warning?~~ Oh, warning must be as the result in CI, but CI does not start without approval for now, so I didn't manage to see the results

thyttan:
You can run the sanitychecker locally as well :)

>Note: It's a great idea to get a local copy of the repository on your PC, then run bin/sanitycheck.js - it'll run through a bunch of common issues that there might be. To get the project running locally, you have to initialize and update the git submodules first: git submodule update --init.

*https://github.com/espruino/BangleApps?tab=readme-ov-file#testing*

bobrippling:
> Right! I've moved.

Thanks!

> ~Where can I look at this warning?~ Oh, warning must be as the result in CI, but CI does not start without approval for now, so I didn't manage to see the results

Yeah that's unfortunate, and when the CI runs the warnings don't fail the build. I'm wondering if we should tweak it now we have things quite stable - a failing build won't block a merge / can be overridden. @gfwilliams @thyttan @halemmerich @atjn what do you all think? (and @buzden, with you having contributed :) )

gfwilliams:
Ahh - I think CI runs in your repo, but not ours for new committers.

@bobrippling what are you thinking? Just enable GitHub actions for all? I feel like there's not a great deal that could be done via GitHub actions (I guess someone can do a PR that totally changes the actions and then they get run, but even so...)

@buzden I think it's unlikely to be a swipeOn bug as it's used by so many clocks now.

I'm just wondering though if it's a fastLoad issue. Did you have `load widgets` turned off in settings? Does the issue happen when you go to the launcher and back?

Bangle.js can't 'fast load' into apps that don't have widgets in because the widgets are already loaded, but it figures out if widgets are loaded by checking if `Bangle.loadWidgets` is in the app's text. In this case it is in the text, but it may not always be called - so maybe it tries to 'fast load' back from the launcher (which does have widgets).

If that's the case you could probably fool it by using `(Bangle).loadWidgets()`, or you could do this, which will hide the widgets if they were already loaded:

```
if (settings.loadWidgets) {
  Bangle.loadWidgets();
  require("widget_utils").swipeOn();
} else if (global.WIDGETS) require("widget_utils").hide();
```

buzden:
> Did you have `load widgets` turned off in settings?

Indeed! The settings file was removed silently during update, and load widgets setting was reset to the default "false".

Unexpected for me thing is that this PR is merged, but in my list of apps with the default loader (https://banglejs.com/apps/) still shows old version 0.04, not the updated one 0.05. What are the rules of updating of this page after PR merging?

halemmerich:
It is done manually by @gfwilliams. The development apploader at [https://espruino.github.io/BangleApps/](https://espruino.github.io/BangleApps/) should be up to date as soon as the github deployment is done. There is some caching involved so it can take a few additional minutes to actually show up.

bobrippling:
> @bobrippling what are you thinking? Just enable GitHub actions for all? I feel like there's not a great deal that could be done via GitHub actions (I guess someone can do a PR that totally changes the actions and then they get run, but even so...)

I was thinking we change [`sanitycheck.js`'s exit](https://github.com/espruino/BangleApps/blob/964ba309facd13153d2159ac7b31493215c63812/bin/sanitycheck.js#L401-L401) to also exit with an error code for warnings too, just for CI, wdyt?

> If that's the case you could probably fool it by using `(Bangle).loadWidgets()`, or you could do this, which will hide the widgets if they were already loaded:
> 
> ```
> if (settings.loadWidgets) {
>   Bangle.loadWidgets();
>   require("widget_utils").swipeOn();
> } else if (global.WIDGETS) require("widget_utils").hide();
> ```

I like the sound of either of those fixes, the second one is probably a good choice with the clock face (mostly) using widgets

thyttan:
> Unexpected for me thing is that this PR is merged, but in my list of apps with the default loader (https://banglejs.com/apps/) still shows old version 0.04, not the updated one 0.05. What are the rules of updating of this page after PR merging?

This is a not too uncommon question. @buzden where could this info have been presented for you to happen on the answer "organically", do you think?

gfwilliams:
Ok, great - I just pushed that second fix.

@bobrippling we do set the exit code for errors, but I guess, yes - we could change it for 'unknown' warnings too.

Normally warnings are tagged with a line/file though so if you view the changed files they should pop up

bobrippling:
> @bobrippling we do set the exit code for errors, but I guess, yes - we could change it for 'unknown' warnings too.
> 
> Normally warnings are tagged with a line/file though so if you view the changed files they should pop up

Thanks - let me know what you think of #3408

gfwilliams:
Thanks - just added something there to mention it!

PR #3259, opened 2024-03-14
Improve linting of modules

atjn:
Apply the new linting rules from `apps` to the `modules` folder as well.

It is a little unfortunate that the eslint config for modules and apps need to be duplicated, I will try to make a PR later that puts most of the config in a shared file.

I am very confident in the fixes I have made, except for the fix in Slider, I am only 90% sure that I did that correctly. I have not tested any of the changes, but wanted to put this up in case you want to test and merge it yourself. Otherwise I plan on testing it myself at a later date :)


gfwilliams:
Thanks! I'm pretty sure the change to Slider will break it - I think `exFirst`/etc should persist outside the function. @thyttan any thoughts?

Otherwise it looks good, but `Layout.min.js` should be updated as that's the one that actually gets used

modules/Slider.js
```diff
@@ -112,6 +112,7 @@ exports.create = function(cb, conf) {
     // Function to pass to `Bangle.on('drag', )`
     o.f.dragSlider = e=>{
       "ram";
+      let exFirst, eyFirst;
       if (o.v.ebLast==0) {
         exFirst = o.c.horizontal?e.y:e.x;
```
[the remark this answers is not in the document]
gfwilliams: Thanks - it'd be really handy if you could test out the `o.v.exFirst` change as you'll be able to see more clearly if something's not going as expected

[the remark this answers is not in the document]
thyttan: I can look at it later tonight 👍

thyttan:
> There are no changes to Layout.min.js because the minifier removes unused variables :)

Now that you say that - does that mean with these lint additions we can expect app minification through the App Loader to be more stable? @atjn @gfwilliams

gfwilliams:
> we can expect app minification through the App Loader to be more stable?

Probably not I'm afraid. It looks like esprima has some real issues and unfortunately doesn't seem that well maintained any more. The big one was obviously it renames local vars like `a,b,c,d,e,f,g` but seems too dumb to realise if something within that scope is using the global `g`.

I guess it's possible we could just bodge something in to stop it using `g` though...

atjn:
@gfwilliams I have a lot of experience with minification systems and generally working with AST transforms. If you want, I can try to make a minification setup that works reliably.

modules/Slider.js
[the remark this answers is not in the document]
thyttan: I just made a PR here https://github.com/espruino/BangleApps/pull/3261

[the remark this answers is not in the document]
atjn: I have removed the incorrect fix from this PR, if you merge @thyttan's PR first, everything should work.

modules/suncalc.js
```diff
@@ -335,7 +335,7 @@ SunCalc.getMoonTimes = function (date, lat, lng, inUTC) {
 
 // export as Node module / AMD module / browser variable
 if (typeof exports === 'object' && typeof module !== 'undefined') module.exports = SunCalc;
-else if (typeof define === 'function' && define.amd) define(SunCalc);
-else window.SunCalc = SunCalc;
+//else if (typeof define === 'function' && define.amd) define(SunCalc);
+//else window.SunCalc = SunCalc;
```
bobrippling, line 339: For these two, banglejs won't be using them but I wonder if they were intended so this module can run elsewhere - I guess not since @gfwilliams it's your change?

bobrippling:
LGTM, I guess if we rebase/merge in `master` now #3261's merged, that'll sort the linting :)

modules/suncalc.js
nxdefiant: Thats from the source module:
https://github.com/mourner/suncalc/blob/master/suncalc.js#L313

gfwilliams: Yeah, I was wondering that. As it's an external module the less changes we can make the better really.

I'd be tempted to just leave it and add a comment to exclude it from the linter? I can definitely imagine we might end up with a case where an interface.html file wants to use it for something, so it'd still need to be able to be run in the browser at minimum.

atjn: > so it'd still need to be able to be run in the browser at minimum.

The browser would be supported with the `globalThis` solution. Then everything works except amd modules, but I don't think there is anything in the Banglejs ecosystem that would use AMD modules?

gfwilliams:
Now the sliders thing is out, it's just suncalc and then we're good.

> @gfwilliams If you want, I can try to make a minification setup that works reliably.

You could definitely look into it - the code that handles it is at https://github.com/espruino/EspruinoTools/blob/master/plugins/minify.js#L73-L112 and you can run that project in CLI for testing.

The main issue really is:

* It needs to run under Node.js *and* in the browser
* It needs to be lightweight enough that it can run in the browser on a budget Android phone
* It needs to run from the repo direct via GitHub pages (https://espruino.github.io/EspruinoWebIDE/) so ideally no bundling step needed
* We need to be able to bundle the relevant files up into https://github.com/espruino/EspruinoAppLoaderCore/blob/master/lib/espruinotools.js which is generated from https://github.com/espruino/EspruinoWebIDE/blob/master/extras/create_espruinotools_js.sh

I think your immediate reaction to this will be to pull in a NPM module for it and then use webpack/etc to put it into a bundle suitable for the web, but I'd prefer it if there were a way to have something we could just 'plug in' to what we have already (eg just a few JS files that can be included), rather than completely redoing what we have. I guess if we do that by bundling something up and storing the bundle in the repo, that could work...

modules/suncalc.js
bobrippling: Yeah I don't think so - I guess it's a question of whether we keep the module exactly the same and treat it as an external unlinted dependency, perhaps there are some others like this too?

atjn: I have merged the solution that I like best. Now suncalc will be exported via the CommonJS module if it exists, and otherwise it is set on `globalThis` which is generally speaking universally supported. No eslint comments are needed for the code to pass lint.

Let me know if you still prefer no changes and eslint comments.

nxdefiant: Staying as close as possible to upstream makes it easier to apply patches, but this one doesn't look to maintained.

bobrippling: Yeah I agree, I think we merge this - shall let @gfwilliams make the call though, in case he had any further thoughts

gfwilliams:
Have you tested any apps that use SunCalc after this change?

I think we're fine as we should use `exports` on Bangle.js, but in 2v21 (current) and earlier firmwares, `globalThis` isn't on the device, so if it were used it'd break

atjn:
> Have you tested any apps that use SunCalc after this change?

No not yet.

> I think we're fine as we should use exports on Bangle.js, but in 2v21 (current) and earlier firmwares, globalThis isn't on the device, so if it were used it'd break

There should be no change to behavior, as both `window` (used before my change) and `globalThis` are undefined in Espruino, so if it didn't break before, it still won't break now. (* assuming you don't use AMD modules)

gfwilliams:
Ok, thanks - hopefully that's fine then

PR #3261, opened 2024-03-14
Slider:fix lint warn var declared in global scope

thyttan:
exFirst -> o.v.exFirst
eyFirst -> o.v.eyFirst

Suggested here: https://github.com/espruino/BangleApps/pull/3259#discussion_r1524804030


thyttan:
I tried this change with my `slidertest` app - seems to work as expected!

atjn (approved):
Thanks for fixing this :)

PR #3262, opened 2024-03-16
chargent: allow a tap on the notification to silence it

bobrippling:
This allows a user to silence the buzz/notification once seen

### DIFF
--- a/apps/chargent/ChangeLog
+++ b/apps/chargent/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Added threshold
 0.04: Added notification
 0.05: Fixed boot
+0.06: Allow tap to silence notification/buzzing
--- a/apps/chargent/README.md
+++ b/apps/chargent/README.md
@@ -6,6 +6,8 @@ The first stage of charging Li-ion ends at ~80% capacity when the charge voltage
 
 This app has no UI and no configuration. To disable the app, you have to uninstall it.
 
+Tap the charged notification to prevent buzzing for this charging session.
+
 New in v0.03: before the very first buzz, the average value after the peak is written to chargent.json and used as threshold for future charges. This reduces the time spent in the second charge stage.
 
 Side notes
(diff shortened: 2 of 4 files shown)


bobrippling:
@thyttan would you be ok with me merging this? (or ok to merge it?)

thyttan:
Don't know if it's this PR or something else, but I just installed v0.06 on my watch and attached the charging cable. Started on ~60% went past both 80% and 90% without signaling. fw 2v21.51.

bobrippling:
Interesting, I'll take a look

thyttan:
I'll try again as well when the battery has come down, and connect with the web ide to look for errors.

thyttan:
I tried again but with v0.05 from standard dev app loader, also no reaction going from 75% to 85% (fw 2v21.53). So maybe something particular is up with my setup. Does it work for you?

thyttan:
And now I've tried v0.05 after a factory reset on fw 2v21.53 - still no reaction on my watch.

bobrippling:
That's strange, I see the notification when I charge mine ok, I wonder - do you see it if you allow your watch to sit at full charge for 5 mins?

thyttan:
Then I think go ahead with merging 👍 

Do you want me to do it?

As for my watch, I'll try you suggestion 🙂

bobrippling:
Thanks! Yeah let's see if we get more reports and I'll keep my eye on it, let me know if you still see issues and we can debug, even if it's an unrelated problem :)

thyttan:
Charged to 25% -> 100% today without noticing warning buzzes. I wasn't around it for the whole duration. When I got back it said 'Fully charged' over the widget field, some of the text went past the screen edge. The only way I got it to disappear was doing a regular `load()` with the hw button long press.

bobrippling:
Thanks, sounds like there's a bug - I'll see if I've any odd local config that might be interfering. I also see the notification part-way off the screen, which I'll sort a fix for

thyttan:
I think I didn't.test without fastload utils - maybe that's the culprit?

bobrippling:
Yes I suspect so - I've seen similar issues / possibly watch hangs when charging, need to investigate still though

bobrippling:
Cracked it - #3386

PR #3263, opened 2024-03-16
clkinfo: save clock info on `kill`/`remove`

bobrippling:
A follow-up to #3234, save clkinfo settings so on a fresh load we display what was last shown

### DIFF
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -224,12 +224,13 @@ exports.addInteractive = function(menu, options) {
       options.menuB = b;
     }
   }
-  options.save = () => {
+  const save = () => {
     // save the currently showing clock_info
     const settings = exports.loadSettings();
     settings.apps[appName] = {a:options.menuA, b:options.menuB};
     require("Storage").writeJSON("clock_info.json",settings);
   };
+  E.on("kill", save);
 
   if (options.menuA===undefined) options.menuA = 0;
   if (options.menuB===undefined) options.menuB = Math.min(exports.loadCount, menu[options.menuA].items.length)-1;
@@ -282,7 +283,6 @@ exports.addInteractive = function(menu, options) {
       oldMenuItem.removeAllListeners("draw");
       menuShowItem(menu[options.menuA].items[options.menuB]);
     }
-    options.save();
     // On 2v18+ firmware we can stop other event handlers from being executed since we handled this
     E.stopEventPropagation&&E.stopEventPropagation();
   }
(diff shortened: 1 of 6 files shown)


bobrippling:
@gfwilliams interested in your thoughts on this, or whether we detect this after a `focus`/`blur` (or even `show`/`hide`/`run`) and do it automatically. I think this is better since it's unlikely most clkinfos will want this functionality

gfwilliams:
Yeah... So right now we save whenever the user swipes?

I think this change is probably ok - we don't want to save whenever we change as I guess we may end up doing a bunch more writes than we need.

**Or...** we just save the state in `E.on('kill'` and `.remove`, and nowhere else? That is definitely preferable - no messing around writing all the time, and we're sure the state will be saved and then none of the clockinfos have to worry about it?

bobrippling:
Yeah exactly, we only save after a swipe currently. I prefer the idea of saving on `kill` and `remove` in fact, I'll alter this and that avoids apps/widgets having to manage it too

apps/clock_info/lib.js
```diff
@@ -333,6 +336,7 @@ exports.addInteractive = function(menu, options) {
   menuShowItem(menu[options.menuA].items[options.menuB]);
   // return an object with info that can be used to remove the info
   options.remove = function() {
+    save();
```
gfwilliams, line 339: we'd need to do `E.removeListener("kill",save)` here too I reckon or we will have a memory leak when doing fast load

gfwilliams:
Looks good apart from that possible memory leak? Might be worth running a fast load clock with it and just checking nothing gets leaked even after that change though

apps/clock_info/lib.js
bobrippling: Good catch, thanks!

bobrippling:
Thanks - shall run it on a fastload clock tomorrow and confirm the memory load

bobrippling:
It seems stable - this was with a clockface with two clock infos:

```javascript
>process.memory(true)
={ free: 6756, usage: 5244,  ... }

>load()
>process.memory(true)
={ free: 6756, usage: 5244,  ... }

>load()
>process.memory(true)
={ free: 6756, usage: 5244,  ... }

>load()
>process.memory(true)
={ free: 6756, usage: 5244,  ... }

>load()
>process.memory(true)
={ free: 6756, usage: 5244,  ... }

>load()
>process.memory(true)
={ free: 6756, usage: 5244,  ... }
```

gfwilliams:
Looks great, thanks!

PR #3264, opened 2024-03-16
Move eslint ignored files to the exemption system

atjn:
Now that you have an exemption system, I see no reason to have files listed in the eslint ignore file :)


bobrippling:
Nice!

PR #3266, opened 2024-03-16
Use Bangle's temperature instead of non-working service in advcasio watch

jordimas:
advcasio app uses an external service that has several problems:
* The app lives in an abandoned repository
* The app does not work for serveral reasons, including that the Openweathermap which was using is no longer free

This is a simple fix which uses Bangle's temperature instead (which we know that is not 100% accurate), which is better than the current situation where nothing is ever show.


apps/advcasio/metadata.json
```diff
@@ -18,8 +18,5 @@
   "storage": [
 			{"name":"advcasio.app.js","url":"app.js"},
 			{"name":"advcasio.img","url":"app-icon.js","evaluate":true}
-  ],
-	"data": [
-			{ "name": "advcasio.data.json", "url": "data.json", "storageFile": true }
```
bobrippling, line 23: @gfwilliams is there any way to clean up possible leftover data files on people's watches when they upgrade?

bobrippling (commented):
Nice fixes, thanks!

apps/advcasio/metadata.json
gfwilliams: Not as far as I'm aware, no. I guess we could add something but in the scheme of things I guess one or two files kicking around isn't the end of the world.

bobrippling: That's fair, thanks!

PR #3267, opened 2024-03-16
Remove app-specific eslint configs

atjn:
There are currently two apps that use their own configs. The config for `buffgym` did not really do anything. It required a few additional stylistic rules, but they are unfortunately deprecated in Eslint, so another solution needs to be thought of there.

The rule for `slpquiet` is an attempt to fix the warning that `sleeplog` is undefined. @bobrippling you should never need to add custom eslint configs to fix lint issues. The problem here was that sleeplog is being referenced as a local variable. You can access it on the `global` scope, then it is clear to both devs and tooling that you are using a global variable :)


gfwilliams:
Thanks!

bobrippling:
Thanks for the improved fix!

PR #3274, opened 2024-03-17
Revert changes to timerclk

### LINKED ISSUE #3273: [Timer clock / timerclk] countdown setup only increments in seconds after v0.06
### Affected hardware version

Bangle 2

### Your firmware version

2v21

### Describe the bug
After updating to v0.06 of Timer Clock, when using the watch faces countdown (swiping left on the watch face), it now only increments in the countdown in seconds, rather than based on where you swiped on the interface.

### Expected behaviour
Swiping in the appropriate place on the interface should increment the appropriate column, i.e. leftmost for hour, middle for minute, rightmost for seconds.

### Steps to reproduce
1. Update to v0.06 of Timer clock
2. Attempt to setup a countdown using minute or hour column.
3. Only seconds column changes regardless of swipe point.

### Installed apps

wid_a_battery_widget (0.04), cprassist (0.02), files (0.08), fileman (0.03), loadingscreen (0.01), chargeanim (0.02), dclock (0.10), widmessages (0.05), ptlaunch (0.15), score (0.01), widbt (0.09), qmsched (0.10), clock_info (0.07), wid_edit (0.05), messages (0.62), rebbleagenda (0.01), oxofocus (0.02), notifyfs (0.12), weather (0.26), alarm (0.46), widdst (0.05), spotrem (0.11), dragboard (0.10), drained (0.04), android (0.33), autoreset (0.01), setting (0.70), messagegui (0.76), widadjust (0.03), about (0.15), sched (0.24), widalarm (0.02), messageicons (0.07), themesetter (0.05), 2047pp (0.04), acmaze (0.04), diceroll (0.02), dtlaunch (0.25), timerclk (0.06), boot (0.60), poweroff (0.04), a_dndtoggle (0.02), lightswitch (0.10), icons (0.02), popconlaunch (0.05)

atjn:
Closes #3273

The `timerclk` app uses indirect variable access and it fooled me into believing that this change was safe.

### DIFF
--- a/apps/lint_exemptions.js
+++ b/apps/lint_exemptions.js
@@ -183,7 +183,7 @@ module.exports = {
     ]
   },
   "timerclk/timer.js": {
-    "hash": "f44702e61ad833b53fb28d222ff0fbd77d0307c6eb69c5ecebb8626a1cb20ed4",
+    "hash": "3b70606f1742b34bf4769f6e8703144ff0c3ba401a3a21f97ddcab5432c94db5",
     "rules": [
       "no-unused-vars",
       "no-undef"
@@ -197,7 +197,7 @@ module.exports = {
     ]
   },
   "timerclk/alarm.js": {
-    "hash": "6f521742cc6a6d6f215e499084a596ffc3494d1e1f950b204368c051a18773de",
+    "hash": "3c280b0858d43f10b12947e5ae939c6825a620e6cb3687c6a7549c1a000b6241",
     "rules": [
       "no-unused-vars",
       "no-undef"
--- a/apps/timerclk/ChangeLog
+++ b/apps/timerclk/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Use 'modules/suncalc.js' to avoid it being copied 8 times for different apps
 0.05: Improve responsiveness and detection of swipes on main clock screen
 0.06: Minor code improvements
+0.07: Fix regression in 0.06
--- a/apps/timerclk/alarm.js
+++ b/apps/timerclk/alarm.js
@@ -18,7 +18,7 @@ var editIndex = 0;
 var drawInterval;
 var drawIntervalTimeout;
 var buttons;
-//var dragBorderHrsMins=0, dragBorderMinsSecs=0;
+var dragBorderHrsMins=0, dragBorderMinsSecs=0;
 
 function update() {
   if (drawInterval) clearInterval(drawInterval);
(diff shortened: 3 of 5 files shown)


thyttan:
I don't have time to verify this myself right now. But if @Aracnoss or @atjn has tried this change on the watch and it works now - I'll merge 🫡

nxdefiant:
Code looks good, but have not tested either.

atjn:
I have tested the changes now and I can reproduce the issue on v0.06 but not on this v0.07, should be good to merge 👍

thyttan:
Nice - thanks!

bobrippling:
Thanks for the quick fix! I wonder whether we could have a program go through and pick out possible shared global variables like this, which we could add (temporarily) to our list of globals?

This would help spot similar problems elsewhere and also hide lint warnings if anyone does change `timerclk` / remove it from lint exemptions in the meanwhile

bobrippling:
> I wonder whether we could have a program go through and pick out possible shared global variables like this

A quick hacky scan:

```bash
$ grep -Eo '^(var )?[a-zA-Z_][a-zA-Z0-9_]+' apps/*/*.js | perl -e {script}
```

where `{script}` is:

```perl
while(<>){
  chomp;
  /(.*):(.*)/;
  $f = $1;
  $v = $2;
  next if $v =~ /^(require|Bangle|exports|if|layout|set(Timeout|Interval)|let|function|WIDGETS|Layout|const|atob|setWatch|Graphics|for|class|try|NRF)$/;
  $v =~ s/^var\s+//;
  $m{$v}->{$f}++;
}

for $v (keys %m) {
  $r = $m{$v};
  if(length keys %$r > 1) {
    $t = 0;
    $t += $r->{$_} for keys %$r;
    print "$v: $t hits across multiple files\n";
  }
}
```

atjn:
Very cool!

atjn:
I am not sure I understand how this would helpful though. Let's say you add these to the globals as a temporary measure, now: 
- Most files will still need to be in the exemption system because they commit other lint crimes
- If you decide to fix one of the files, you will forget to fix the common global because you don't see a warning for it
- It is now OK to add new apps that misuse these common globals, so the problem will grow bigger
- If you try to remove the globals from the scope, you will be overwhelmed with issues and probably decide to put them back in

But maybe you see something that I don't?

bobrippling:
Yes, all very good points - my main thought was to spot anything that would be a concern, and my second thought was how could we prevent this kind of regression in future.

For example, without the knowledge of this in our heads, someone could come along tomorrow and take out `dragBorderHrsMins`, based on the linter's advice, and we'd be back to having the bug.

PR #3277, opened 2024-03-19
sleepphasealarm: Fix json dates

nxdefiant:
Fix handle dates saved as object with milliseconds

see https://forum.espruino.com/comments/17322019/


thyttan:
I'm unsure how this will play with https://github.com/espruino/Espruino/commit/805c317d67ac9d73bd1bcbf017ed521b941a3586 ?

nxdefiant:
Date.toISOString() is the old behaviour restored. The ternary should handle both variants `new Date(typeof o === 'string' ? o : o.ms)`

nxdefiant:
Yes, sleep phases recorded with these firmware versions have the other data format saved in storage.

thyttan:
Ok - thanks for the fix and clarifications! 🙂

PR #3278, opened 2024-03-20
`locale`: Deprecate currency information

### LINKED ISSUE #3269: Deprecate currency information in `locale`
The `locale` app gives the name of the currency that is most commonly used in that locale. I suggest removing that information because I don't think it is useful. If you need to show a price in a local currency, you need to know what the local currency is, but also what the conversion is between your currency and the local currency. If you don't have both, none of them is useful. To get the conversion rate, you need to connect to some online service, and if you can do that, it should be simple to also get the local currency for at given locale.

My argument that it is not useful seems to be confirmed by the fact that no apps in this repository uses the currency information.

The reason I care about this, is because in #3268 I propose that you move to use the CLDR dataset for locales, but CLDR does not provide information about which currency each locale uses. Therefore to get that information, you need to import a different dataset and I could not find any industry standard datasets like CLDR, so it is a [somewhat questionable source](https://www.npmjs.com/package/country-locale-map). If this information isn't useful anyways, I think it is best to remove it.

If you agree, I would propose that we redesign the module such that if you query the currency information, it will always return `GBP` and will log a warning that currency information is no longer supported.

atjn:
Closes #3269

The currency information now always returns British pounds no matter the locale.

Using the currency method logs a deprecation warning. The TS types are also updated to mark the API as deprecated.

The raw currency data is removed because we don't use it.

I have tested these changes and could not find any issues.


apps/locale/ChangeLog
```diff
@@ -20,3 +20,4 @@
       Add second 'dp' argument for decimal places in distance/speed/temp (fix #1523)
 0.17: Fix regression where long month names were 'undefined' (fix #1641)
 0.18: Fix lint warnings, change anv->janv for fr_BE and fr_CH
+    : Deprecate currency information
```
bobrippling, line 23: Let's pop a version in here

```suggestion
0.19: Deprecate currency information
```

apps/locale/locale.html
```diff
@@ -212,7 +209,7 @@
     }
     return s.substr(0, i + 3) + r + (d ? '${locale.decimal_point}' + d: '');
   },
-  currency: n => ${currency},
+  currency: n => {console.log("Warning: Currency information is deprecated");return ${js("£")}+exports.number(n)},
```
bobrippling, line 212: What do you think of:

```suggestion
  currency: n => {console.warn("Warning: Currency information is deprecated");return ${js("£")}+exports.number(n)},
```

typescript/types/locale.d.ts
```diff
@@ -3,10 +3,12 @@ declare module Locale {
 
   type Locale = {
     name: string,
+    /** @deprecated */
```
bobrippling, line 6: Let's add a message and link to this MR and #3269 so people can see the rationale too (and perhaps for the console warning above too)

bobrippling (commented):
LGTM, a few little things

apps/locale/locale.html
atjn: I agree, but doing that would make the package incompatible with most versions of Espruino,  `console.warn` was added this year: https://github.com/espruino/Espruino/issues/1302#issuecomment-1873917322

Given that `console.warn` is just an alias of `console.log`, I don't think it is worth it.

typescript/types/locale.d.ts
atjn: Sure, will do later

apps/locale/ChangeLog
atjn: The rationale here was that we didn't technically have to release this yet. But I forgot that anything added to this repo automatically becomes the released version. Will fix it properly later (need to change the version in `metadata.json` too).

gfwilliams:
Looks good - thanks!

@bobrippling is that locale.ts file autogenerated, so Espruino itself should be updated to add the deprecated label?

apps/locale/locale.html
bobrippling: Yeah fair point 👍

bobrippling:
`locale.d.ts` is manually generated in fact, we could move the `.d.ts` to sit next to [its source js](https://github.com/espruino/Espruino/blob/c1824f087b2618fb53711482c49323da4b6e4d67/libs/js/banglejs/locale.js), but I think that would complicate trying to then import them into BangleApps, assuming they change not very frequently

gfwilliams:
Yes, I think it's best as-is... Thanks!

PR #3280, opened 2024-03-20
sleeplog: Add wear detection based on measured temperature

Ishidres:
SleepLog currently detects if the watch is being worn based on the HRM. This has never worked for me and when I don't wear my Bangle.JS 2 it usually considers that time span as deep sleep, making it unusable.

This PR contributes new settings as well as some functions that allow to change the wear detection based on the temperature measured by the Bangle.JS 2.

I tested this for around a week and found out that a temperature `>= 29°C` seems to be a good value to consider it being worn.


apps/sleeplog/settings.js
```diff
@@ -345,6 +347,19 @@
...
+        min: 20,
+        max: 40,
+        wrap: true,
+        noList: true,
+        format: v => v + "°C",
+        onchange: v => {
+          settings.wearTemp = v;
+          writeSetting();
+        }
+      },
```
bobrippling, line 362: What do you think about merging this with the `Wear detection using temperature` setting? If the user sets it to the minimum value (which we could have as 19.5), it turns it off, then we just present a single menu item to them?

bobrippling (commented):
Looks good to me, @storm64 are you ok if I merge this? (after any comment discussion)

apps/sleeplog/settings.js
Ishidres: Do you mean similar to the recent changes done to sched where if the lowest number gets selected it changes to "Forever" (in our case it would be "disabled") ?
```
// apps/sched/settings.js

    /*LANG*/"Buzz Count": {
      value: settings.buzzCount == null ? 4 : settings.buzzCount,
      min: 4,
      max: 15,
      step: 1,
      format: v => v === 4 ? "Forever" : v,
      onchange: v => {
        settings.buzzCount = v === 4 ? null : v;
        require("sched").setSettings(settings);
      }
    },
```
Sure, why not! :+1: I originally also wanted to implement it like this but thought this might make the user overlook the "disabled" option, but we can do it that way of course. I checked "Allow edits and access to secrets by maintainers" in this PR so if you want to, you can just edit it the way you want to.

Ishidres:
While wearing the BangleJS 2 today for some reason the app did not record any sleep data. I don't know why but maybe it's worth delaying the merge again until I figure out what the problem was.

apps/sleeplog/settings.js
bobrippling: Yeah exactly that, well spotted! And thanks - I'll let you make the change for now but nice to know we can help out if needed :)

storm64:
Back then on creation I decided to not use the temperature as it might be a very personal threshold. But I don't have a problem to add this as a additional feature.

For the implementation I would suggest to move the code changes into the `checkIsWearing`-function inside the library to let the `boot.js` be as small as possible.
You might just pass the temperature as additional argument behind `data` to the function and copy your code there.

bobrippling:
Those changes look good to me. One question:

> While wearing the BangleJS 2 today for some reason the app did not record any sleep data. I don't know why but maybe it's worth delaying the merge again until I figure out what the problem was.

Did you get to the bottom of this? Or does it seem to be stable now?

Ishidres:
Yes, it was because I accidentally removed some lines which I readded in https://github.com/espruino/BangleApps/commit/7cc67ceb278786834212b12e1f9fd34656e91634 so this should be fixed and can be merged now! 👍

bobrippling:
Great, thanks for the explanation!

PR #3282, opened 2024-03-21
recorder:fix ref error missing `settings` obj

thyttan:
would get:
```
Uncaught ReferenceError: "settings" is not defined
 at line 198 col 49 in recorder.wid.js
writeLog(settings.period);
                 ^
```

Not sure this is the best way to fix this. Maybe call `loadSettings()` inside `writeOnGps` function instead. But that would lead to many storage interactions.

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -47,3 +47,5 @@
 	setInterval.
 0.38: Tweaks to speed up track rendering
 0.39: Minor code improvements
+0.40: Fix reference error from missing `settings` object in widget on 1 sec
+	recordings with gps.
--- a/apps/recorder/metadata.json
+++ b/apps/recorder/metadata.json
@@ -2,7 +2,7 @@
   "id": "recorder",
   "name": "Recorder",
   "shortName": "Recorder",
-  "version": "0.39",
+  "version": "0.40",
   "description": "Record GPS position, heart rate and more in the background, then download to your PC.",
   "icon": "app.png",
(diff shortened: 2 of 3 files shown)


thyttan:
Also, not sure how I could have missed this when ´writeOnGps()` was added. And I have used the functionality to record on gps event before, I'm almost certain.

gfwilliams:
Thanks! Was this something someone reported? Or you just noticed there was an issue?

I'm not a great fan of having `settings` having to stay loaded (especially as it'll be out of date and won't be reloaded by `reload` - so if the user changes the period in the app and starts recording again I think it would break)

I noticed that the lint check threw up some issues on this PR (that existed before but weren't reported) so I've just made some changes in cb54fb59a03253cd61a9dd8e251568929e3ef2b9 which I think will fix this and fix those errors too.

Please could you take a look and make sure I haven't broken anything?

thyttan:
@gfwilliams 

> Was this something someone reported? Or you just noticed there was an issue?

I was looking at why the "Charge Gently" app didn't work for me, connecting to the Web IDE to see if there were any error messages when going past the battery threshold. In the process I used the recorder as a means to drain the battery - e.g. logging GPS every 1 sec. Doing that the error showed up in the console field. 

But like I said, I didn't see the error when developing the feature. I guess I must have missed it on some last tweaks or something before it was merged..

> Please could you take a look and make sure I haven't broken anything?

I'll load the changes to my watch and try it.

PR #3283, opened 2024-03-21
alarm, sched: Fix #3281

### LINKED ISSUE #3281: [sched] Corrupted alarm when it is snoozed through midnight
### Affected hardware version

Bangle 2

### Your firmware version

2v21

### Bug description

When an alarm in the evening is repeatedly snoozed until past midnight, its scheduled time becomes invalid and the alarm app is unusable.
- The next alarms do not run, including the snoozed one.
- The alarm app can't be opened. It freezes on "loading…" until I reboot the watch.
- In the app loader, the scheduler app doesn't shows a time for this alarm. Putting the time back and uploading it to the watch solves the problem.
- In sched.json, the alarm changes from `{t:72000000,del:false,on:true,rp:true,as:true,dow:127,last:13,vibrate:"::",msg:"Bedtime",ot:72000000}` to `{t:86400000,del:false,on:true,rp:true,as:false,dow:127,last:13,vibrate:"::",msg:"Bedtime",ot:72000000}`. If I understand correctly, `t:86400000` indicates an invalid time.

### Expected behavior

- An alarm snoozed late in the evening continues to ring after midnight (use case example: a bedtime alarm).
- An invalid alarm doesn't silently prevent the other alarms to ring.
- And ideally, an invalid alarm can be repaired on the watch itself. 

### Steps to reproduce

1. Change the time of the watch to 23:25.
2. Schedule an alarm for 23:30.
3. Snooze the alarm (or let the it auto-snooze) when it rings at 23:30, 23:45 and 00:00.
4. The alarm app won't open anymore.

### Installed apps

widbthide (0.01), chargent (0.05), swiperclocklaunch (0.06), alarm (0.46), iconlaunch (0.19), swp2clk (0.04), notify (0.13), pomoplus (0.06), boot (0.60), sched (0.24), kbmulti (0.08), mylocation (0.11), daisy (0.11), setting (0.71), health (0.29), widalarm (0.02)

nxdefiant:
Fix wrap around when snoozed through midnight.

### DIFF
--- a/apps/sched/lib.js
+++ b/apps/sched/lib.js
@@ -4,6 +4,7 @@ exports.getAlarms = function() {
 };
 // Write a list of alarms back to storage
 exports.setAlarms = function(alarms) {
+  alarms.forEach(e => e.t %= 86400000); // Also fix #3281 from other apps, e.g. multitimer
   return require("Storage").writeJSON("sched.json",alarms);
 };
 // Return an alarm object based on ID
--- a/apps/sched/lib.js
+++ b/apps/sched/lib.js
@@ -35,7 +35,7 @@ exports.setAlarm = function(id, alarm) {
     if (alarm.timer) { // if it's a timer, set the start time as a time from *now*
       var time = new Date();
       var currentTime = (time.getHours()*3600000)+(time.getMinutes()*60000)+(time.getSeconds()*1000);
-      alarm.t = currentTime + alarm.timer;
+      alarm.t = (currentTime + alarm.timer) % 86400000;
     }
     alarms.push(alarm);
   }
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -49,3 +49,4 @@
 0.44: Add "delete timer after expiration" setting to events.
 0.45: Fix new alarm when selectedAlarm is undefined
 0.46: Show alarm groups if the Show Group setting is ON.  Scroll alarms menu back to previous position when getting back to it.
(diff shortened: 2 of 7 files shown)


nxdefiant:
Default app, requesting review from @gfwilliams

gfwilliams:
Thanks! Yes, this looks good to me!

I guess you could argue the temporary fix in alarms.app.js isn't really needed but it's not like it's hurting anything.

**But:** I was just looking at this and thinking about whether it caused issues in timers:

https://github.com/espruino/BangleApps/blob/db32e9df6ee021062274cf3da2100f4b99a4612d/apps/sched/lib.js#L36-L38

And I could be wrong, but it looks like if you set a 10 minute timer at 5 mins to midnight, that is broken too?

Before I merge this, do you think it's worth actually changing your fix so you do the `e.t %= 86400000` in `exports.setAlarm` where it deals with timers as well?

nxdefiant:
Just tested, you are right about the timer. I'll put the modulo in.

apps/sched/lib.js
```diff
@@ -4,6 +4,7 @@ exports.getAlarms = function() {
 };
 // Write a list of alarms back to storage
 exports.setAlarms = function(alarms) {
+  alarms.forEach(e => e.t %= 86400000); // Also fix #3281 from other apps, e.g. multitimer
```
nxdefiant, line 7: I tested multitimer and found the same issue, so I added this catch all.

nxdefiant:
done.

PR #3285, opened 2024-03-22
New App: Counter 2 (dual counter), minor improvement for clockcal

Stuff-etc:
I am still getting used to git, so I'm sorry that there are two different things in one PR.

- I made a new app with TWO counters (only for Bangle2)
- Fixed a minor inconvenience in clockcal (you can now exit the full-size-calendar with the button)


foostuff:
This is my old account which I will no longer be using. Just wanted to confirm that Stuff-etc is now the maintaining account for ClockCal.

thyttan:
I believe you could make the ui a little snappier by changing from using `.on('drag' ...)` to `.on('swipe' ...)`. I haven't tried it to compare though. 

Swipes only went left/right earlier, and I guess that's why you opted for your drag solution in the first place?

Stuff-etc:
Thanks for the tips.
I mainly used drag because that's what was used in a tutorial or some other app or something when I first created clockcal.
While I can't use it for the counter, because I need to know on _which side_ of the screen the dragging happened, I think I could put it into the next version of clockcal. Even if it might not be snappier, the code will be cleaner :)

While I do really like the concept of doing one thing, but doing that thing really well (quicklaunch), having the Calendar expand on swipe-up is a core feature for me.
But I think I could implement some kind of toggleable quicklaunch-compatibility mode, where I will only check for a tap on the calendar zone and leave the rest to quicklaunch.

thyttan:
I saw you already have 'ignore' in the settings for the different directions. So that already gives compatibility with quicklaunch I think? :)

Stuff-etc:
I guess that's true.

thyttan:
Two notes for now,

Could you make it so the background of `counter2` uses the system background colour? Now it looks a little weird on light theme with the widget field being white and the app rectangle being black.

There are new warnings for apps when they don't use a variable keyword (var, let, cons). It would be neat if those were sorted for `clockcal`. But I will not hold up merging the PR because of these warnings if you feel you don't have time.

Stuff-etc:
Thanks for being so thorough. I'll try to fix both of them today.

apps/clockcal/app.js
```diff
@@ -211,8 +212,11 @@ function action(a) {
     g.reset();
     if (typeof secondInterval !== "undefined") clearTimeout(secondInterval);
     if (DEBUG) console.log("action:" + a);
+    state = "unknown";
+    console.log("state -> unknown");
     switch (a) {
         case "[ignore]":
+            drawWatch();
```
thyttan, line 219: Is a `drawWatch()` needed here? I'm not against it, but seems unnecessary too me (I might overlook something or have a misconception).

Stuff-etc: kinda but not really: I was playing with different states (so the hardware button would do different things: when I am showing the calendar it should redraw the watch, if I am in the watch, it should open the menu.
So I change the state to "unknown" when the launcher is called and whenever I draw the watch, I'm back in the watch-state. Somehow the app hangs when I try to go to drawWatch() after an showAlert(), but I wanted to investigate that further in the future. That's why I left the state = unknown in and call the drawWatch() in the [ignore].
Still things to do, but for now I think it doesn't hurt too much.

thyttan:
@Stuff-etc Do you still want to work on this PR or would you rather I look at merging it as is?

foostuff:
I am on vacation right now and while I am not done working on the apps, I AM done with this step :)
I guess from now on I will work on my fixes in branches, so I don't involuntarily update my PRs with every minor commit.

PR #3291, opened 2024-03-23
Two new apps: Elapsed Time Clock and Vaporwave Sunset Clock

paul-arg:
New app I created that allows you to calculate the time difference between now and any target date.
New clock I created with a Vaporwave Sunset Theme


paul-arg:
Sorry I pushed new commits on my branch and I see that it modified this MR as well.
I'm not very experienced with Github.

But that's just as well, because the second app I made is ready too.

thyttan:
If you want to set up different branches to work in parallel see e.g: https://git-scm.com/docs/git-switch

paul-arg:
Got it, thanks, I will use branches in the future!

paul-arg:
It should be all good now 🙂

thyttan:
The check fails because the tags in the metadata are expected to be all inside one string. So `"tag1,tag2,tag3"` and not a list of strings - so change that and well see it clears the tests then :+1: 

@bobrippling @gfwilliams maybe the sanity checker could use a fix for catching this? Now its own logic failed since it didn't anticipate the contributor could add a list of strings (which is an understandable misinterpretation of `"tags": "",                 // comma separated tag list for searching`) :
```
TypeError: app.tags.split is not a function
    at /home/runner/work/BangleApps/BangleApps/bin/sanitycheck.js:130:37
    at Array.forEach (<anonymous>)
    at Object.<anonymous> (/home/runner/work/BangleApps/BangleApps/bin/sanitycheck.js:118:6)
    at Module._compile (node:internal/modules/cjs/loader:1356:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1414:10)
    at Module.load (node:internal/modules/cjs/loader:1197:32)
    at Module._load (node:internal/modules/cjs/loader:1013:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:128:12)
    at node:internal/main/run_main_module:28:49

Node.js v18.19.1
Error: Process completed with exit code 7.
```

I'll maybe do a PR to sanitycheck.js .

paul-arg:
Just updated the tags

gfwilliams:
@thyttan yes, a PR to make sanity checker more robust would be great!

Also it looks like there's a  `src` field that shouldn't be in metadata too? But I'll put a fix in for that now

gfwilliams:
Just done - so I guess we're good to merge?

thyttan:
@gfwilliams I've waited with going through the rest of the contributions checklist or testing the app until the checks passed. So I have not checked everything yet.

thyttan:
#### Before I merge:
- The black countdown characters of Elapsed Time Clock are hardcoded black, making them invisible when using dark system theme. So that needs a tweak. I suggest using the theme foreground color.

- There are warnings regarding some of the colour constants not being used. Could you either remove those or comment them out? See the "Files changed" tab on this PR.

#### Misc. feedback you don't need to act on (I will merge anyway):
- I really dig the font of the vaporwave sunset clock! :)

- Since the contrast is low and the font small I have trouble reading the day and date strings.

gfwilliams:
>  I suggest using the theme foreground color.

If it helps, this is `g.theme.fg`

paul-arg:
@gfwilliams @thyttan just aded support for dark theme and commented out the unused colours :)

thyttan:
Great - thanks for the tweaks and the cool apps!

PR #3296, opened 2024-03-25
kineticscroll - Better scrolling behaviour

halemmerich:
This makes the behaviour more predictable while scrolling slowly.
* Prevents the kinetic scrolling that sometimes happened when trying to stop scrolling
* Directly passes the drag events y-diff through while the finger is still touching


thyttan:
Very nice improvements - thanks!

PR #3300, opened 2024-03-26
widmsggrid: fix drawing outside of widget field

thyttan:
... when the circle containing single digit numbers was drawn.


halemmerich:
Thanks!

PR #3302, opened 2024-03-27
New clock face: Pokemon Album

paul-arg:


thyttan:
Looks really cool and clean. The nostalgia is flowing!

But I think copyright/licensing doesn't allow us to keep this in the repo - please tell me if I'm wrong. Maybe this would be fair use or similar?

On this note though - I'm not sure how icons8 keep some of these icons without problems: https://icons8.com/icons/set/pokemon 

Maybe the icons are different enough or they have asked permission from the copyright owners (The Pokémon Company/Nintendo I suppose).

gfwilliams:
Thanks! I think this looks really cool.

The copyright is a tricky one - normally I'd be tempted to keep it and just be very quick to remove it if anyone complained, but given Nintendo is suing everyone at the moment it's probably best not to have it in the official repo.

I guess you could still host on https://paul-arg.github.io/BangleApps/?id=poke_album ? worst case seems you get a very unlikely DMCA notice and have to delete it.

I know we had something with a really cool Nyan Cat clock before as well. It's hard to know what to do in these cases - if we do a 'clock background' library then at least you can easily pull in your own backgrounds, but it'd still be nice to have a way to share the ones you made

thyttan:
> It's hard to know what to do in these cases - if we do a 'clock background' library then at least you can easily pull in your own backgrounds, but it'd still be nice to have a way to share the ones you made

@paul-arg That background library exists now if you want to use it: https://espruino.github.io/BangleApps/?id=clockbg

PR #3304, opened 2024-03-28
BT States Widget fix: handle colors themes better

paul-arg:
I might be wrong but I think there is a mistake here, because cyan is much more readeable on a black background, and same for blue on a white background

### DIFF
--- a/apps/widbtstates/widget.js
+++ b/apps/widbtstates/widget.js
@@ -17,7 +17,7 @@
     };
     var colours = (_a = {},
         _a[1] = {
-            false: "#fff",
+            false: "#000",
             true: "#fff",
         },
         _a[2] = {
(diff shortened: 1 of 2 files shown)


paul-arg:
same when BT is on but disconnected, white on white is invisible, so setting black on white

bobrippling:
Looks good to me - let's apply this change to [the typescript source](https://github.com/espruino/BangleApps/blob/8223946db1a0d1bfe088cf75708ca973a8a293a9/apps/widbtstates/widget.ts#L33) though, and we can keep the JS changes as-is or regenerate using (`cd typescript && npm i && npm run build`)

bobrippling:
Not a problem, thanks for the change :)

PR #3305, opened 2024-03-28
Battery Widget Themed fix: center widget vertically

paul-arg:
this just brings the battery icon up by 2 pixels so that is is centered vercially on the 24px widget band


halemmerich:
Great, thanks!

PR #3307, opened 2024-03-28
messagesoverlay - Rendering and UI improvements

halemmerich:
* Renders short messages in big font and centered up to as many lines as fit
* Longer messages are all rendered in smaller font on all their pages to keep them consistent
* Adds Settings
* Messages can be cleared automatically when bangle is locked
* Adds visual indicator for multiple messages in queue

### DIFF
--- a/apps/messagesoverlay/ChangeLog
+++ b/apps/messagesoverlay/ChangeLog
@@ -5,3 +5,7 @@
 0.05: Fix the overlay keeping the LCD on
 0.06: Better low memory handling
       Fix first message beeing displayed again on unlock
+0.07: Adds settings
+      Automatic discard of oldest messages
+      Indicator for multiple messages in queue
+      Some optimization in the rendering code
--- a/apps/messagesoverlay/ChangeLog
+++ b/apps/messagesoverlay/ChangeLog
@@ -9,3 +9,4 @@
       Automatic discard of oldest messages
       Indicator for multiple messages in queue
       Some optimization in the rendering code
+      Track handler changes done by background code
--- a/apps/messagesoverlay/README.md
+++ b/apps/messagesoverlay/README.md
@@ -2,13 +2,17 @@
 
 This app handles the display of messages and message notifications as an overlay pop up. 
 
-It is a GUI replacement for the `messages` apps.
+It is a GUI replacement for the `messagesgui` app.
 
 Messages are ephemeral and not stored on the Bangle.
 
 ## Usage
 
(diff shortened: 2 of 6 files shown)


thyttan:
How is the indicators for the messages queue supposed to work?

For me it works like this:
*# of messages* : *# of indication angles*
1:0
2:1
3:1
4:2
5:2
6:2
7:2

halemmerich:
Yes, thats how it is implemented. The values are chosen pretty randomly, I just wanted to see at a glance roughly how many messages I can expect to be stacked. The messages are redrawn over the whole area so we could easily do one additional line per message up to a value.

thyttan:
I notice that the watch screen will light up when I clear notifications on the phone. This is regardless of if I already cleared the message on the watch or not. I don't think this is intended?

I think ideally dismissing on the phone would never light up the watch screen, only clear the message if still shown.

halemmerich:
Hm, I thought I fixed that behaviour. A quick test on my watch seems to work fine in that regard. It checks the message for the new property. Maybe that is not enough for all types of messages? I tested with Conversations.

thyttan:
Maybe I'm not on your latest commit. Will try again

thyttan:
Yes, sorry - you seem to have fixed it! :)

I'll merge now :+1: Thanks!

thyttan:
@halemmerich 

Maybe we want to do something about these now that the lint exemption was removed. Sorry - didn't notice it before. 

```

/home/runner/work/BangleApps/BangleApps/apps/messagesoverlay/lib.js
Warning:   520:5   warning  'toRemove' is not defined  no-undef
Warning:   530:9   warning  'toRemove' is not defined  no-undef
Warning:   535:27  warning  'toRemove' is not defined  no-undef
Warning:   536:18  warning  'toRemove' is not defined  no-undef
Warning:   608:3   warning  'quiet' is not defined     no-undef
Warning:   674:3   warning  'bpp' is not defined       no-undef
Warning:   676:25  warning  'bpp' is not defined       no-undef
Warning:   676:45  warning  'bpp' is not defined       no-undef
Warning:   683:5   warning  'bpp' is not defined       no-undef
Warning:   701:63  warning  'bpp' is not defined       no-undef
Warning:   708:31  warning  'bpp' is not defined       no-undef
Warning:   709:50  warning  'bpp' is not defined       no-undef
Warning:   718:7   warning  'bpp' is not defined       no-undef
```

halemmerich:
Yes, I overlook those regularly as well... Will do a patch for those as a new PR.

PR #3309, opened 2024-03-29
Handle fast loading for Vaporwave Sunset Clock

paul-arg:
Here is the memory leak test I did:

memory_leak_test

everything seems ok.

let me know if some adjustments must be made


halemmerich:
Thanks, looks good!

halemmerich:
Found two small issues, fixed them in https://github.com/espruino/BangleApps/pull/3311.

PR #3313, opened 2024-04-01
[messagegui] Fix: When user taps on a new message, clear unread timeout

paul-arg:
So that the user can read the long message uninterrupted :)


bobrippling:
Looks good but I don't use messagesgui, @thyttan or @halemmerich are you able to review?

thyttan (approved):
I tried it and it works well - thanks!

Will leave the merging to @gfwilliams for this.

gfwilliams:
Sounds like a great idea - thanks for the PR!

PR #3315, opened 2024-04-01
`recorder` Remove unused variable

atjn:
The exemption for this file was broken in #3297 but the warning was not fixed.

The code includes an unused variable that seems safe to remove. I have tested the app and it does not seem to break in any way.

The code seems to be written by @gfwilliams, maybe you can check it to be sure?


gfwilliams:
Thanks - yes, this looks good! Should marginally speed up the rendering of graphs too

PR #3316, opened 2024-04-01
Better wrapping of the text base (dynamic instead of hardcoded)

jordimas:
Rellotge app: Better wrapping of the text base (dynamic instead of hardcoded)


bobrippling:
Great, thanks!

PR #3318, opened 2024-04-01
Temporary fix for minified file lint

atjn:
The minified file `kineticscroll/boot.min.js` includes an optimization that is perfectly fine in minified files, but terrible in source files.

We need to figure out how to deal with minified files while also enforcing good lint rules. I am currently working on a better minifier for the app loader. If that is succesful, we don't need to include minified files here, so the problem would solve itself.

For now, the temporary fix is to exempt the file from the warning.

### DIFF
--- a/apps/lint_exemptions.js
+++ b/apps/lint_exemptions.js
@@ -1481,5 +1481,11 @@ module.exports = {
       "no-unused-vars",
       "no-cond-assign"
     ]
+  },
+  "kineticscroll/boot.min.js": {
+    "hash": "a5106fa601cdbd2179e32265a8b77e2df62ba44ca0f4b94ce1e594c95a47e20d",
+    "rules": [
+      "no-cond-assign"
+    ]
   }
 };


halemmerich:
Can't we just exclude all `*.min.js` files? Other apps have minified files and small changes in code could create minified files the linter then finds objectionable. That would essentially mean that we need to add all minified files to the exemptions at some point.

atjn:
We could do that, but some apps only have the minified version of some code, so that would give them a free pass to include really bad code.

There are currently only 9 apps that have minified versions of their source code, so I think this solution will work fine for much longer than it needs to. If everything goes smoothly, I can have a functional minifier ready in a week.

atjn:
If you prefer it, you can also just let the warning stay until we decide on a better solution.

gfwilliams:
Adding this exemption looks good for now, but in general I agree - probably best to not bother checking minified files.

Generally we minify with the Closure compiler and that tends to do a certain amount of checking as it runs anyway

PR #3319, opened 2024-04-01
`launch` Fix broken reset

atjn:
Commit 9cbcff08ac20efe55ceb6b5e45addc7f29397ebe removed the `drawMenu` function, but did not remove the call to `drawMenu` on line 62. This means that the `App source not found` message would never go away by itself.

This fix restores the old behavior of removing the `App source not found` message after 2 seconds.

### DIFF
--- a/apps/launch/ChangeLog
+++ b/apps/launch/ChangeLog
@@ -20,3 +20,4 @@
 0.18: Add 'back' icon in top-left to go back to clock
 0.19: Fix regression after back button added (returnToClock was called twice!)
 0.20: Use Bangle.showClock for changing to clock
+0.21: Fix a bug where a warning would stay on screen forever
--- a/apps/launch/ChangeLog
+++ b/apps/launch/ChangeLog
@@ -20,4 +20,4 @@
 0.18: Add 'back' icon in top-left to go back to clock
 0.19: Fix regression after back button added (returnToClock was called twice!)
 0.20: Use Bangle.showClock for changing to clock
-0.21: Fix a bug where a warning would stay on screen forever
+0.21: Make the "App source not found" warning less buggy
--- a/apps/launch/app.js
+++ b/apps/launch/app.js
@@ -42,47 +42,51 @@ let apps = launchCache.apps;
 if (!settings.fullscreen)
   Bangle.loadWidgets();
 
-E.showScroller({
-  h : 64*scaleval, c : apps.length,
-  draw : (i, r) => {
-    var app = apps[i];
-    if (!app) return;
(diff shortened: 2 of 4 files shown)


atjn:
I am sorry that the git diff looks so crazy. I have not changed anything in the function, it is just indented.

nxdefiant:
> I am sorry that the git diff looks so crazy. I have not changed anything in the function, it is just indented.

https://github.com/espruino/BangleApps/pull/3319/files?w=1 should make it clearer. Default apps are reviewed by @gfwilliams

atjn:
> https://github.com/espruino/BangleApps/pull/3319/files?w=1 should make it clearer.

That is a great trick, thanks!

> Personally I do not understand the change?!?

To rephrase it, on line 62 there is a call to the function `drawMenu`, but that function does not exist. This PR solves that.

gfwilliams:
Thanks - this is a great spot, but not *quite* right I think, because the `remove` handler for `showScroller` will remove the `lockHandler`, and ideally we ought to re-add it inside `drawMenu` so that when the screen locks we go back to the clock.

... so `drawMenu` wants to include most of the lines after too:

```
g.flip(); // force a render before widgets have finished drawing

// 10s of inactivity goes back to clock
Bangle.setLocked(false); // unlock initially
let lockTimeout;
let lockHandler = function(locked) {
  if (lockTimeout) clearTimeout(lockTimeout);
  lockTimeout = undefined;
  if (locked)
    lockTimeout = setTimeout(Bangle.showClock, 10000);
}
Bangle.on("lock", lockHandler);
```

But then this looks great!

atjn:
Good catch, I misunderstood how that code was used. I have included the code in the function, and when testing, it now goes back to the watch face after the 10 secs.

While I was at it, I also added a call to disable the scroller before the warning is shown. Before, you could scroll the warning away which looked very buggy. You could probably also click on a scroller item even though the item was not displayed.

gfwilliams:
Great, thanks!

Just in future, if you need to remove some user interface item, best to just call `Bangle.setUI()` - it's what `E.showScroller()` would do internally anyway.

There's quite a strong argument that actually `E.showMessage` should call setUI internally, but it's possible it might break something in apps that use it.

atjn:
@gfwilliams I see thanks for letting me know, maybe that should be noted on the API page, which is what I looked at to determine that `E.showSroller()` was the correct solution: https://www.espruino.com/Reference#l_E_showScroller

> To remove the scroller, just call [E.showScroller()](https://www.espruino.com/Reference#l_E_showScroller)

PR #3320, opened 2024-04-02
New bug in bigdclock.app.js

deirdreobyrne:
The value of "width" was being reset somewhere, which meant that the battery usage feature wasn't working - it was resetting to "battery empty" whenever the clock updated.

So "width" has been renamed to "batteryUsedWidth", which is better anyway, as it is more descriptive of what the variable does.

### DIFF
--- a/apps/bigdclock/ChangeLog
+++ b/apps/bigdclock/ChangeLog
@@ -6,3 +6,4 @@
 0.06: Tell clock widgets to hide.
 0.07: Better battery graphic - now has green, yellow and red sections; battery status reflected in the bar across the middle of the screen; current battery state checked only once every 15 minutes, leading to longer-lasting battery charge
 0.08: Minor code improvements
+0.09: Something was changing the value of the "width" variable, which caused the battery usage feature to malfunction. The "width" variable has been renamed - the cause remains a mystery.
--- a/apps/bigdclock/bigdclock.app.js
+++ b/apps/bigdclock/bigdclock.app.js
@@ -12,7 +12,7 @@ Graphics.prototype.setFontOpenSans = function(scale) {
 
 var drawTimeout;
 var lastBattCheck = 0;
-var width = 0;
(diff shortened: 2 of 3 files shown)


gfwilliams:
Thanks! That's interesting - it might be interesting to see what's overwriting `width` as whatever it is likely shouldn't be changing global - maybe it's some widget/similar.

Please can you add a changelog entry and bump the version and we can change this?

deirdreobyrne:
Done (I think - git is still confusing to me). No idea as to how to check what might be changing `width`. Problem started after I updated my apps, and installed a BLE advertising app.

gfwilliams:
Just to add, for debugging, maybe since now you're not using `width` you could look at the value it gets set to, and that might give you some hint as to which app is causing the problem

deirdreobyrne:
It gets set to zero. So no help there :/

gfwilliams:
Actually that is some help :) Maybe check in `WIDGETS` and see if any of those have zero width, and then see if when you can get those widgets to show up, `width` changes?

bobrippling:
One other way to check would be to do this in [the IDE](https://www.espruino.com/ide/):
```javascript
Object.defineProperty(global, "width", { set: () => { throw new Error("tried to set width"); } })
```

then watch the console for a log appearing that looks like this:
```
Uncaught Error: tried to set width
 at line 1 col 13
throw new Error("tried to set width");
                    ^
in function called from line 1 col 5
width=3
    ^
```

gfwilliams:
@bobrippling that's really cool. It made me realise we should try and put some of these tips into a page. I've started a Wiki at https://github.com/espruino/BangleApps/wiki/Debugging-Tips - if you have any more magic tricks like this, please could you add them?

bobrippling:
Time served debugging minified webapps! Good idea, I'll add to that if I think of any more.

That tip about `global["\xff"]` is very useful btw

PR #2719, opened 2023-05-03
Ratchet launch: add caching

bobrippling:
@jsphpl hope this is ok - I've added app caching (taken from the launcher app) to the ratcher launcher

PR #3321, opened 2024-04-02
Fix bugs in `ratchet_launcher`

atjn:
In #2719, the app was updated to access storage via the variable `s` which is not defined. This was flagged by the new linting rules, which is how it caught my attention. Unless you are lucky and something else has defined `s` for you, this means the launcher cannot even open.

It turns out there is also a bug with the default `blankImage`, it is not supported by the `Layout` library. I assume that this worked when the app was designed, but that later revisions of Espruino disallowed images with no height. I have added newlines to give the image a bit of height, which makes it work again.

### DIFF
--- a/apps/lint_exemptions.js
+++ b/apps/lint_exemptions.js
@@ -798,12 +798,6 @@ module.exports = {
       "no-undef"
     ]
   },
-  "ratchet_launch/app.js": {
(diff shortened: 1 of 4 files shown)


thyttan:
Thanks!

PR #3323, opened 2024-04-02
`health` Remove unused variable

atjn:
`health` sets a single variable in the global scope that seems like it can be removed safely. I have tested this change and could not find any issues with it.


atjn:
Requesting @gfwilliams since this is a default app.

gfwilliams:
Thanks! Yes, that looks like some old code got left in there...

PR #3324, opened 2024-04-02
`gpsinfo` Fix memory leaks

atjn:
`gpsinfo` currently adds a lot of random numbers to the global scope, this should keep the variables local.

I am unfortunately not able to test these changes because my watch refuses to aquire a GPS fix. Maybe some of you can give it a quick check? I am pretty confident that the changes work :)

### DIFF
--- a/apps/gpsinfo/ChangeLog
+++ b/apps/gpsinfo/ChangeLog
@@ -9,3 +9,4 @@
 0.10: Show satellites "in view" separated by GNS-system
 0.11: Show number of packets received
 0.12: Fix number of packets received
+0.13: Minor code improvements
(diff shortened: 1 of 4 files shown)


bobrippling:
For the GPS fix, does the [AGPS `interface.html`](https://banglejs.com/apps/?id=assistedgps) help your watch acquire a fix at all?

atjn:
I tried uploading the AGPS data for GPS only, then ran the app again. This is the app after appr. 15 minutes of being left untouched outside:
image

gfwilliams:
Does it work if you run the standard unmodified app?

I should add that since this is an app, there's not really a big issue with it creating global variables. There's not going to be some kind of memory leak or anything - although obviously it's nice to tidy the code.

bobrippling:
Ok - let's merge this. About the GPS connection, give it a whirl where you suspect there may be better signal, if you want to check it out

Issue #1759, opened 2022-04-27
App Loader needs to fully reset Bangle.js before uploads


gfwilliams:
See: https://github.com/espruino/BangleApps/pull/1758

The app loader calls `Comms.reset()` which in turn sends `reset()` to Bangle.js before file uploads - with the intention that there's nothing running in the background that'll print to the console, hog the CPU, or otherwise screw everything up.

However, over time, the addition of boot code (`app.boot.js`) which gets called even after `reset()` has really screwed this up. Now you can call `reset()` and all kinds of stuff is still loaded

It seems that we could maybe hack around this by adding extra commands like `Bangle.removeAllListeners();E.removeAllListeners();NRF.removeAllListeners();clearInterval();clearWatch();`after `reset()`

But the ideal solution would be to just not load all this stuff in the first place, which I think needs a modification to the actual Espruino firmware.

gfwilliams:
There's now an option to not reset the bangle

PR #3326, opened 2024-04-03
fastload - Add warning about automatic reload apploader option

halemmerich:
This adds a warning to the fastload README regarding bad behaviour in combination with the new automatic reload of the apploader.

### DIFF
--- a/apps/fastload/README.md
+++ b/apps/fastload/README.md
@@ -1,10 +1,16 @@
+#### ⚠️EXPERIMENTAL⚠️
+
 # Fastload Utils
 
-*EXPERIMENTAL* Use this with caution. When you find something misbehaving please check if the problem actually persists when removing this app.
+Use this with caution. When you find something misbehaving please check if the problem actually persists when removing this app.
 
 This allows fast loading of all apps with two conditions:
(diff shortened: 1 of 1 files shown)


thyttan:
Thanks!

bobrippling:
This is interesting - I've been using auto-reload in the app loader and not seen any major issues, but would be keen on discussing a fix even so. Do you have any plans so far to resolve this?

Perhaps we could have the apploader tweak a setting (or touch a file) or change how it performs the upload (e.g. [`reset()`](https://www.espruino.com/ReferenceBANGLEJS2#l__global_reset)) so that fastload isn't active during that time?

nxdefiant:
I'm using fastload and auto-reload and for single app updates this works fine. When updating multiple apps at once it sometimes fails.

halemmerich:
I can usually reproduce it with single app upgrades in under five tries. Did not yet look deeper into this but would very much like a fix. I think we might be hitting #1759 again with the boot code running and doing stuff.
I suspect the "cleanup" done to workaround this is not enough for fastload, since that replaces the global `load()`.

bobrippling:
Cool, if you can get a console dump next time it happens, let's investigate more (and move the conversation to #1759 - nice find)

Issue #3287, opened 2024-03-22
Create a Library for custom clock background images?


gfwilliams:
Right now we have a bunch of different clocks which have background images:

* Patriotic/flag clock https://banglejs.com/apps/?id=patriotclk
* Image background clock https://banglejs.com/apps/?id=imgclock
* Wave clock/floral clock and way more besides

Especially the first two have 'customisers' where the user can choose/upload what background they want.

... but plenty of existing clocks would look really cool if they could have a custom background too.

It'd be really handy if we could have just one super library/app that handled backgrounds, and that other clocks could use:

* single color by default
* different color each time it loads
* choose an image from a list
* choose a flag from a list
* upload an image
* allow other apps to implement the `background` library type to allow them to do something else with existing clock faces - like generative backgrounds 
* possibly even animation at some point?

So I guess ideally a library would provide:

* draw a rectangle of background
* draw the whole screen
* get as a layer for g.drawImages

On Bangle.js 1 we had some special stuff to avoid flicker (since there was no offscreen buffer) but maybe we just ignore that for now as pretty much all new apps are Bangle.js 2 only.

Maybe at some point it could also be able to add a border around an image/graphics instance to allow existing clocks to easily add themselves to a background while still being easily visible.

PR #3328, opened 2024-04-04
Add 'clock backgrounds' app, allow configurable backgrounds in Pebble++

gfwilliams:
The issue related to this is #3287

Not sure what you all think about this?  I added it to Pebble++ clock face but it should be easy to add to others - I guess ideally when we merge we'd do so with a bunch of other apps converted too

Not sure if the API should change - for instance I wonder if we should be able to get info about the uploaded image:

* At minimum, whether it's dark or light (so the clock knows what to use to stand out against it)
* bounding box for 'clear areas' in the image

Also right now we can only upload one image, but it's possible to upload more. It's just how to make a nice UI since we don't really want to have to download the uploaded images each time for preview purposes.

...also not sure if we need a `.remove` method. Right now it doesn't use much memory at all (just a few vars) so it's not a big deal for fast load, but I guess it's possible it could become an issue.


thyttan:
Good ideas! Tagging @paul-arg , you may have some as well?

gfwilliams:
Any thoughts on this? @bobrippling?

I guess it doesn't matter too much adding a new app/library - but do we think this'll annoy pebblepp users? Hopefully not though.

apps/clockbg/lib.js
```diff
@@ -0,0 +1,25 @@
...
+// eg require("clockbg").fillRect({x:10,y:10,w:50,h:50})
+//    require("clockbg").fillRect(Bangle.appRect)
+exports.fillRect = function(rect,y,x2,y2) {
+  if ("object"!=typeof rect) rect = {x:rect,y:y,w:1+x2-rect,h:1+y2-y};
+  if (settings.img) {
+    g.setClipRect(rect.x, rect.y, rect.x+rect.w-1, rect.y+rect.h-1).drawImage(settings.img).setClipRect(0,0,g.getWidth()-1,g.getHeight()-1);
+  } else if (settings.style == "color") {
+    g.setBgColor(settings.color).clearRect(rect);
+  } else
+    console.log("No background set");
```
bobrippling, line 23: Perhaps we throw here? Or a sensible default like fill with `g.style.bg` ?

bobrippling (commented):
Looks like a good feature to me. I can think of a few clocks that might use something like this, possibly the [newly added Meridian one](https://github.com/espruino/BangleApps/pull/3347) - @Spioune thought I'd tag you here if you were interested :)

apps/clockbg/lib.js
gfwilliams: Good point, thanks - I think the log is fine but adding a default would be helpful

gfwilliams:
Ok, well I think I'll merge anyway then and see what happens. Hopefully the available backgrounds can be extended with some better ones and we can add some new clocks. I quite like the analog one!

PR #3330, opened 2024-04-04
counter2: fix lint warnings

thyttan:
I have not tested these changes on the watch. But they seem innocent enough to me.

Edit: Works on my watch.

### DIFF
--- a/apps/counter2/app.js
+++ b/apps/counter2/app.js
@@ -9,23 +9,22 @@ var s = Object.assign({
   colortext: true,
 }, require('Storage').readJSON("counter2.json", true) || {});
 
-f1 = (s.colortext) ? "#f00" : "#fff";
-f2 = (s.colortext) ? "#00f" : "#fff";
-b1 = (s.colortext) ? g.theme.bg : "#f00";
-b2 = (s.colortext) ? g.theme.bg : "#00f";
+const f1 = (s.colortext) ? "#f00" : "#fff";
+const f2 = (s.colortext) ? "#00f" : "#fff";
+const b1 = (s.colortext) ? g.theme.bg : "#f00";
+const b2 = (s.colortext) ? g.theme.bg : "#00f";
 
-var counter = 0;
 var drag;
 
-screenwidth = g.getWidth();
-screenheight = g.getHeight();
-halfwidth = screenwidth / 2;
-halfheight = screenheight / 2;
+const screenwidth = g.getWidth();
+const screenheight = g.getHeight();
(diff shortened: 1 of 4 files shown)


apps/counter2/app.js
```diff
@@ -9,23 +9,23 @@ var s = Object.assign({
...
-screenheight = g.getHeight();
-halfwidth = screenwidth / 2;
-halfheight = screenheight / 2;
+const screenwidth = g.getWidth();
+const screenheight = g.getHeight();
+const halfwidth = screenwidth / 2;
+const halfheight = screenheight / 2;
 
-counter = [];
+counter = []; // FIXME: How does this relate to `var counter = 0;` 8 lines above this one.
```
bobrippling, line 25: I suppose the `= 0` is a dead store and we can just delete the above line and change this like so:
```suggestion
var counter = [];
```

apps/counter2/settings.js
```diff
@@ -1,20 +1,20 @@
...
+    const defaults={
         counter0:12,
         counter1:0,
         max0:12,
         max1:0,
         buzz: true,
         colortext: true,
     };
-    settings = Object.assign(defaults, require('Storage').readJSON(FILE, true) || {});
+    let settings = Object.assign(defaults, require('Storage').readJSON(FILE, true) || {});
```
bobrippling, line 11: This can be `const` too, in fact - we don't change what `settings` references :)

apps/counter2/app.js
```diff
@@ -9,23 +9,23 @@ var s = Object.assign({
...
 counter[0] = s.counter0;
 counter[1] = s.counter1;
-defaults = [];
+let defaults = [];
```
bobrippling, line 28: This can also be `const`, since we keep the same array around

apps/counter2/app.js
```diff
@@ -56,7 +56,7 @@ function updateScreen() {
 }
 
 Bangle.on("drag", e => {
-  c = (e.x < halfwidth) ? 0 : 1;
+  let c = (e.x < halfwidth) ? 0 : 1;
```
bobrippling, line 59: One more! This can be `const` too

bobrippling (commented):
Nice! Just minor suggestions

atjn:
Pro tip: Eslint can automatically fix `let` to `const` for you, just run:

```sh
npx eslint ./apps/counter2/*.js --rule 'prefer-const: warn' --fix
```

halemmerich:
Thanks!

PR #3331, opened 2024-04-04
clockcal: fix lint warnings

thyttan:
no-undef

I think these changes should be fine. Will test it on my watch as well.

Edit: works on my watch.


bobrippling:
Looks good, thanks for fixes!

PR #3334, opened 2024-04-05
messages/messagegui: setting to temporarily wake on face up or twist

thyttan:
Add setting to wake on face up or twist when message is displayed automatically.

Can be tried on https://thyttan.github.io/BangleApps

### DIFF
--- a/apps/messagegui/app-newmessage.js
+++ b/apps/messagegui/app-newmessage.js
@@ -3,3 +3,18 @@ BUZZ_ON_NEW_MESSAGE is set so when messagegui.app.js loads it knows
 that it should buzz */
 global.BUZZ_ON_NEW_MESSAGE = true;
 eval(require("Storage").read("messagegui.app.js"));
+
+// Add temporary listeners for unlocking watch. `settings` is init:ed in the eval:ed file "messagegui.app.js".
+if (settings.tempWakeOnFaceUp || settings.tempWakeOnTwist) {
+  const options = Bangle.getOptions();
+  if (!options.wakeOnFaceUp && !options.wakeOnTwist) { // If the user already uses a one arm method of unlocking it's unnecessary to do anything more.
+    if (settings.tempWakeOnFaceUp) {
+      Bangle.setOptions({wakeOnFaceUp:true});
+      E.on("kill", ()=>{Bangle.setOptions({wakeOnFaceUp:false});});
+    }
+    if (settings.tempWakeOnTwist) {
+      Bangle.setOptions({wakeOnTwist:true});
+      E.on("kill", ()=>{Bangle.setOptions({wakeOnTwist:false});});
+    }
+  }
+}
--- a/apps/messagegui/app-newmessage.js
+++ b/apps/messagegui/app-newmessage.js
@@ -11,11 +11,9 @@ if (settings.tempWakeOnFaceUp || settings.tempWakeOnTwist) {
   if (!options.wakeOnFaceUp && !options.wakeOnTwist) { // If the user already uses a one arm method of unlocking it's unnecessary to do anything more.
     if (settings.tempWakeOnFaceUp) {
       Bangle.setOptions({wakeOnFaceUp:true});
-      E.on("kill", ()=>{Bangle.setOptions({wakeOnFaceUp:false});});
     }
     if (settings.tempWakeOnTwist) {
       Bangle.setOptions({wakeOnTwist:true});
-      E.on("kill", ()=>{Bangle.setOptions({wakeOnTwist:false});});
     }
   }
 }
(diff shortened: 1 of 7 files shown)


thyttan:
Running this currently to see if anything misbehaves.

apps/messagegui/app-newmessage.js
```diff
@@ -3,3 +3,18 @@ BUZZ_ON_NEW_MESSAGE is set so when messagegui.app.js loads it knows
 that it should buzz */
 global.BUZZ_ON_NEW_MESSAGE = true;
 eval(require("Storage").read("messagegui.app.js"));
+
+// Add temporary listeners for unlocking watch. `settings` is init:ed in the eval:ed file "messagegui.app.js".
+if (settings.tempWakeOnFaceUp || settings.tempWakeOnTwist) {
```
thyttan, line 8: @atjn This gives the lint warning:
` Warning:    8:5   warning  'settings' is not defined  no-undef`

It's defined correctly in the evaluated file.

Should we add an exemption for this maybe?

apps/messagegui/app-newmessage.js
```diff
@@ -3,3 +3,18 @@ BUZZ_ON_NEW_MESSAGE is set so when messagegui.app.js loads it knows
 that it should buzz */
 global.BUZZ_ON_NEW_MESSAGE = true;
 eval(require("Storage").read("messagegui.app.js"));
+
+// Add temporary listeners for unlocking watch. `settings` is init:ed in the eval:ed file "messagegui.app.js".
```
atjn, line 7:
```suggestion
/* global settings -- `settings` is init:ed in the eval:ed file "messagegui.app.js" */
// Add temporary listeners for unlocking watch.
```

apps/messagegui/app-newmessage.js
atjn: The only real solution to this, is to not use `eval`. `eval` has many downsides, one of them being that it is impossible to analyze it with tools like Eslint. Preferably, Espruino should support `require("messagegui.app.js").settings` or something like it, because then it is clear where the variable is coming from. Unfortunately I cannot find a simple solution in Espruino that would support this pattern without using `eval`. Maybe you have some ideas?

The quick and dirty solution is to define `settings` as a global variable. You already have a comment about this to help human readers understand the code, you just need to format it so that Eslint understands it as well. See my suggestion for how to format it correctly :)

thyttan: I realized that it can be initiated as undefined first in this filed and then be defined in the evaluation. Like with the latest commit. That avoids the lint warning. I think we can go with this - what do you think?

atjn: Yes that seems like a fine solution as well, but beware that if the declaration in the source file is changed to a `let` or `const`, it will break the `eval`, and unless you are testing the app thoroughly, you might not be aware that you are breaking something.

atjn (approved):

thyttan:
@gfwilliams Current implementation here toggles global `Bangle` options to wake on faceup/twist. My thought with this is the user configuration in the settings app applies - would they do that if just adding local handlers as well? 

If the listeners were local there would not be a need to reset the `Bangle` options on `kill`.

gfwilliams:
Thanks - it's tricky - did you test this? My feeling is that as long as we exit the messages app with `load` rather than fast loading (which is what we do it seems) then the boot code in `.boot0` will call `Bangle.setOptions` and should overwrite any custom settings? Assuming the wakeOnXX are in the options passed anyway...

So it should actually be fine without having to restore it?

thyttan:
Ah - I probably misunderstood what setOptions do. I thought the options change would persist. I'll refactor.

gfwilliams:
> I thought the options change would persist.

It does, but because `.boot0` calls `setOptions` again each time it should overwrite them (if they're specified in the object passed to setOptions - but they should be).

thyttan:
@gfwilliams Do you think it makes sense to add this to `messages` and `messagegui`? Or would it be better to do it as a bootloader app that triggers on notification messages?

gfwilliams:
I suppose it brings up a question about having settings in `messages` - potentially the settings app  in `messages` should try and call into `messagegui` to find out what settings `messagegui` wants to add (or just have a separate settings menu) - as apps that override `messagegui` won't implement all the functionality that `messagegui` adds (or may even have more settings).

But that's a very good point - yes, from my point of view if there's a way to implement something as an app that has zero impact for users that don't install it, that'd be really neat and I'd be all for that. It's also potentially easier for users to find (just by searching in the app loader) than if it's buried in settings.

So in the boot code you'd have to check if the `messagegui` app was running before enabling wake on twist? If your code is basically `if (__FILE__=="messagegui") Bangle.setOptions({wakeOnTwist:true});` the impact is going to be tiny (probably 1ms at most?)

thyttan:
closing in favour of https://github.com/espruino/BangleApps/pull/3351

Issue #3332, opened 2024-04-04
sleeplog: Suddenly almost always logged as awake on stable and development branch


Ishidres:
### Affected hardware version

Bangle 2

### Your firmware version

2v21

### Describe the bug
Since around March 28 I notice that the sleep reported by sleeplog is almost always reported as "awake" but never asleep for a long consecutive time (see screenshot).

![screenshot](https://github.com/espruino/BangleApps/assets/17363426/ab2d80e2-6c5f-4bff-8866-ea8289b98a10)
- In the screenshot above I was not wearing the Bangle.JS from 19:40 to 21:40, however it only marked the beginning and end of that period as "not worn".
- The time from around 01:00 to 09:00 should be marked as sleep or consecutive sleep, however it's marked as almost only awake, though I did wear the band during that time.

### Steps to reproduce

- I should add that for testing I installed the stable branch sleeplog version for the duration from 01:00 to 09:00 and the development version (that also contains the latest temperature threshold wear detection features) after I woke up at around 09:00.
- I reset the thresholds to the default values and set the temperature wear threshold to 29°C.
- I did a full reinstall of sleeplog several times where I also removed all files generated by sleeplog (including source files as well as recorded data)

I first suspected it might be due to the latest temperature checks that were added but it also happens on the stable branch. I also deleted some apps that I suspected might potentially cause issues like `activityreminder` but this changed nothing.

I'm currently using my own fork but as you can see in the changes at https://github.com/espruino/BangleApps/compare/master...Ishidres:BangleApps:master there are no significant changes that should cause any problems with sleeplog.

### Installed apps

widlock (0.08)
notify (0.13)
quicklaunch (0.15)
authentiwatch (0.07)
slevel (0.04)
weather (0.26)
thermom (0.07)
bee (0.03)
intervalTimer (0.01)
pomoplus (0.06)
noteify (0.03)
widbatv (0.03)
info (0.03)
barometer (0.04)
torch (0.11)
iconlaunch (0.19)
scicalc (0.03)
qmsched (0.10)
messagesmusic (0.05)
messages (0.62)
boot (0.60)
widalarm (0.02)
messageicons (0.07)
circlesclock (0.26)
agenda (0.15)
run (0.18)
kbmulti (0.08)
2047pp (0.04)
android (0.34)
icons (0.02)
inspire (0.03)
stopwatch (0.06)
smpltmr (0.09)
widmessages (0.06)
mylocation (0.11)
setting (0.71)
[...]

PR #3335, opened 2024-04-06
Sleeplog: Revert wear detection from temperature

thyttan:
Revert the wear detection from temperature as suggested in https://github.com/espruino/BangleApps/issues/3332#issuecomment-2041062180 while keeping the lint fixes that was merged after.

### DIFF
--- a/apps/lint_exemptions.js
+++ b/apps/lint_exemptions.js
@@ -59,13 +59,6 @@ module.exports = {
       "no-unused-vars"
     ]
   },
-  "sleeplog/settings.js": {
-    "hash": "bd5e3e1382321df6682ef1cb718b0e15ab355422bef77278eb086f213f643021",
-    "rules": [
-      "no-unused-vars",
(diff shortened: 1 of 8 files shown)


halemmerich:
I do not use sleeplog but do we need to remove the wear detection by temperature completely? Form a glance at the code it seems we could just disable it by default and collect some data on how well it is actually working for those who would like to try it. It at least seems to be working for some people.

Ishidres:
I'm currently testing the new bleeding edge firmware and so far it seems to work again, in other words, upgrading the firmware to bleeding edge *might* fix https://github.com/espruino/BangleApps/issues/3332. I'll still test it for another night and if it still works then I suggest to not revert the wear detection changes and keep them, especially because I seem to be the only one with this issue (e.g. as reported by @thinkpoop https://github.com/espruino/BangleApps/issues/3332#issuecomment-2041325286, he doesn't seem to have any issues with it).

PR #3337, opened 2024-04-07
Add new thunder app

jt-nti:


halemmerich:
Thanks! As an idea: if you do not need the explicit setting for the current unit for the distance you could maybe just use the locale module. Calculate in meters and let the module care about conversion for display. See [https://www.espruino.com/Bangle.js+Locale](https://www.espruino.com/Bangle.js+Locale).

jt-nti:
Thanks for the tip, the locale module definitely looks useful. I'll give it a try but I'll probably at least keep an option to override it for reasons... https://www.thepoke.com/2021/09/21/how-to-measure-like-a-brit-went-viral-and-it-absolutely-nails-it/ :)

PR #3339, opened 2024-04-07
TeaTimer 0.6

Koell:
- add newline to header so time is not split
- save counter before leaving the help screen, so it does not get reset


thyttan:
Thanks!

PR #3347, opened 2024-04-09
New Clock : Meridian Clock

Spioune:
This is my first contribution to Espruino.
An attempt to create a simple and elegant clock.

screenshot


thyttan:
Looks very nice!

One thing I spotted now is you probably want to redraw a bigger area of the background around the clock info's. E.g. the calendar one can end up drawing outside of the boxes and then some of it will not be cleared when cycling to the next clock info.

Spioune:
Thanks @thyttan ! I used apple watch face as an inspiration.

The last commit should fix the rendering issue.

thyttan:
Nice! I'll take a look.

Did you mean to add the bluetooth mouse app to the same branch an PR?

Spioune:
Oops... no I was just playing on my local fork and pushed to the wrong branch

thyttan:
The redraws work on my watch now - thanks!

I don't know where the problems lies and maybe it's outside of this clock face. But doing:
1. Select topright clock info
2. Select topleft clock info (without unselecting the other one first)
3. Swipe with intention to switch topleft clock info. But `quicklaunch` is triggered instead.

Spioune:
I tried to reproduce the problem but cannot. What is quicklaunch ?

Spioune:
I installed quicklaunch and I was able to reproduce the bug.
Also, I tried with another clock (Pebble++ Clock) and the bug also exists.
The weird thing is that for both my clock and Pebble++ Clock only the top-left clock_info is buggy...

thyttan:
Ok - thanks for Meridian Clock! I'll merge now since the problems we found seem to not be the fault of this app. 

But we should at least open an issue describing the problem and maybe linking to this discussion. Do you think you could do that @Spioune ? Otherwise I'll get to it.

thyttan:
Also, here's a tutorial if you'd want to add fast loading to the clock face: https://www.espruino.com/Bangle.js+Fast+Load

Spioune:
Thanks @thyttan, I will look into fast loading.

Any idea why I cannot find Meridian Clock on the app loader ?
[https://banglejs.com/apps/?q=meridian](https://banglejs.com/apps/?q=meridian)

thyttan:
Yes - it's already available on the development app loader (accessible via `More...` tab of the regular appLoader).

The regular app loader is manually updated by Gordon maybe once a week.

PR #3349, opened 2024-04-10
bthrm - Move caching of characteristics into settings

halemmerich:
Moves the caching of device characteristics out of the library code and into the settings. Reduces the lib by about 1/6 and potentially saves on flash writes since the connect is now strictly read only. Caching now only happens once during initial pairing with the sensor.

### DIFF
--- a/apps/bthrm/lib.js
+++ b/apps/bthrm/lib.js
@@ -318,20 +318,6 @@ exports.enable = () => {
       return result.then(()=>log("Handled characteristics"));
     };
 
-    let createServicePromise = function(service) {
-      log("Create service promise", service);
-      let result = Promise.resolve();
-      result = result.then(()=>{
-        log("Handling service" + service.uuid);
-        return service.getCharacteristics().then((c)=>createCharacteristicsPromise(c));
-      });
-      return result.then(()=>log("Handled service" + service.uuid));
-    };
-
-    let attachServicePromise = function(promise, service) {
-      return promise.then(()=>createServicePromise(service));
-    };
-
     let initBt = function () {
       log("initBt with blockInit: " + blockInit);
       if (blockInit && !powerdownRequested){
--- a/apps/bthrm/lib.js
+++ b/apps/bthrm/lib.js
@@ -302,22 +302,6 @@ exports.enable = () => {
       });
     };
 
-    let createCharacteristicsPromise = function(newCharacteristics) {
-      log("Create characteristics promis ", newCharacteristics);
-      let result = Promise.resolve();
-      for (let c of newCharacteristics){
-        if (!supportedCharacteristics[c.uuid]) continue;
-        log("Supporting characteristic", c);
-        characteristics.push(c);
-        if (c.properties.notify){
-          addNotificationHandler(c);
-        }
-
-        result = attachCharacteristicPromise(result, c);
-      }
-      return result.then(()=>log("Handled characteristics"));
-    };
-
     let initBt = function () {
       log("initBt with blockInit: " + blockInit);
       if (blockInit && !powerdownRequested){
--- a/apps/bthrm/lib.js
+++ b/apps/bthrm/lib.js
@@ -239,6 +239,7 @@ exports.enable = () => {
       }
     };
 
+    let initialDisconnects = true;
     let buzzing = false;
     let onDisconnect = function(reason) {
       log("Disconnect: " + reason);
@@ -252,7 +253,7 @@ exports.enable = () => {
       supportedCharacteristics["0x2a37"].active = false;
       if (!powerdownRequested) startFallback();
       blockInit = false;
-      if (settings.warnDisconnect && !buzzing){
+      if (settings.warnDisconnect && !buzzing && !initialDisconnects){
         buzzing = true;
         Bangle.buzz(500,0.3).then(()=>waitingPromise(4500)).then(()=>{buzzing = false;});
       }
@@ -415,6 +416,7 @@ exports.enable = () => {
 
       return promise.then(()=>{
         log("Connection established, waiting for notifications");
+        initialDisconnects = false;
         clearRetryTimeout(true);
       }).catch((e) => {
         characteristics = [];
@@ -435,6 +437,7 @@ exports.enable = () => {
       isOn = Bangle._PWR.BTHRM.length;
       // so now we know if we're really on
       if (isOn) {
+        initialDisconnects = true;
         powerdownRequested = false;
         switchFallback();
         if (!Bangle.isBTHRMConnected()) initBt();
--- a/apps/bthrm/lib.js
+++ b/apps/bthrm/lib.js
@@ -16,6 +16,14 @@ exports.enable = () => {
 
   log("Settings: ", settings);
 
+  //this is for compatibility with 0.18 and older
+  let oldCache = require('Storage').readJSON("bthrm.cache.json", true);
+  if(oldCache){
+    settings.cache = oldCache;
+    require('Storage').writeJSON("bthrm.json", settings);
+    require('Storage').erase("bthrm.cache.json");
+  }
+
   if (settings.enabled && settings.cache){
 
     log("Start");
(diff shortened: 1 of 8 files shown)


thyttan:
Under the files tab there are new lint warnings: `Unchanged files with check annotations `

```
/home/runner/work/BangleApps/BangleApps/apps/bthrm/lib.js
Warning:   331:9  warning  'attachServicePromise' is assigned a value but never used  no-unused-vars

/home/runner/work/BangleApps/BangleApps/apps/bthrm/settings.js
Warning:   24:9  warning  'bthrm' is assigned a value but never used  no-unused-vars
```

Do you think it makes sense to add a commit fixing those on this PR?

halemmerich:
Yes I think both of those are just forgotten. `attachServicePromise` probably makes other stuff obsolete on removal as well. Will do that.

apps/bthrm/lib.js
```diff
@@ -316,31 +280,32 @@ exports.enable = () => {
...
-    var attachCharacteristicPromise = function(promise, characteristic) {
+    let attachCharacteristicPromise = function(promise, characteristic) {
       return promise.then(()=>{
         log("Handling characteristic:", characteristic);
         return createCharacteristicPromise(characteristic);
       });
     };
 
-    var createCharacteristicsPromise = function(newCharacteristics) {
+    let createCharacteristicsPromise = function(newCharacteristics) {
```
thyttan, line 305: Like you suspected - this function is no longer used as well now.

thyttan:
- I feel like I see very big improvements in time to connection and getting readings when in the BT HRM app. It's even quicker than the internal hrm! (v0.18 seems to not get a connection for me at all on fw2v21.82)

- The `recorder` works with default mode of `bthrm`. But I don't get readings for the `recorder` "Log BT HR" entry when using "both" mode of `bthrm`.

- If I run the `bthrm` app on the watch and then connect with the Web IDE I get errors and the app hangs:

```
        var e=null==d.bgCol?a:g.toColor(d.bgCol);if(e!=a||"txt"==d.t...
        ^
debug>
In debug mode: Expected a simple ID, type 'help' for more info.
Found BANGLEJS2, 2v21.82
ERROR: Ctrl-C while processing interval - removing it.
Interpreter error: [
  "CALLBACK"
 ]
New interpreter error: CALLBACK
>
Connected to Web Bluetooth, Bangle.js 92f9
>
{"t":"status","bat":16,"chg":0}
{"t":"ver","fw":"2v21.82","hw":2}
{"t":"force_calendar_sync","ids":[]}
Disconnected from Web Bluetooth, Bangle.js 92f9
        var e=null==d.bgCol?a:g.toColor(d.bgCol);if(e!=a||"txt"==d.t...
        ^
debug>
In debug mode: Expected a simple ID, type 'help' for more info.
ERROR: Ctrl-C while processing interval - removing it.
Interpreter error: [
  "CALLBACK"
 ]
New interpreter error: CALLBACK
```

halemmerich:
> * I feel like I see very big improvements in time to connection and getting readings when in the BT HRM app. It's even quicker than the internal hrm! (v0.18 seems to not get a connection for me at all on fw2v21.82)

Great, that was kind of the reason for refactoring :)

> 
>     * The `recorder` app works with "default" mode of `bthrm`. But I don't get readings for the `recorder` "Log BT HR" entry when using "both" or "default" mode of `bthrm`.

There should be no changes in how recording and events are handled. I have a look at this, did not explicitly test.

>     * If I run the `bthrm` app on the watch and then connect with the Web IDE I get errors and the reading seems to hang:
> 
> 
Hm, how do you get into debug mode? The code seems minified, maybe from message/messagegui judging by the `"txt"==` in there?

thyttan:
> Hm, how do you get into debug mode?

I don't know. This is all after factory resetting the device and not doing anything other than installing and setting up the relevant apps - bthrm, recorder and run.

The debug thing seemed to happen only when bthrm were already running before connecting the web ide.

halemmerich:
I can reproduce on 2v21, but only with fully updated apps. Some change outside of bthrm between the stock app versions and the current dev apploader behaves badly in combination with the new bthrm changes.

halemmerich:
Wrapping the Bangle.emit in timeouts to decouple from the bthrm module scope helps with the debugger thing. The UI still stops working but `Bangle.on("BTHRM", print)` on the console works just fine on the console while the UI is frozen.

halemmerich:
The last two commits seem to workaround the problem for me. It seems to happen during `prepareLazyRender` in the layout module, maybe there is some bug hiding there.

thyttan:
Nice - I'll try it shortly!

We get this lint warn: 
```
 /home/runner/work/BangleApps/BangleApps/apps/bthrm/bthrm.js
Warning:   126:10  warning  'showStatusInfo' is defined but never used  no-unused-vars
```

But in the code:
```
// This can get called for the boot code to show what's happening
function showStatusInfo(txt) {
```
So it might be used somewhere else.

Maybe this should be added to the known warnings. Or if there's another way to permanently exempt just that warning, even if the app is updated?

halemmerich:
Yes, that is used by the library code to write the tiny status info in the bottom of the BTHRM app. This will need an exemption. I'll try if the linter is happy if this method is put onto the global object explicitly.

halemmerich:
> * The `recorder` app works with "default" mode of `bthrm`. But I don't get readings for the `recorder` "Log BT HR" entry when using "both" or "default" mode of `bthrm`.

This should be fixed now, there was an (old) typo preventing the recorder to start BTHRM on it's own.

thyttan:
> I'll try if the linter is happy if this method is put onto the global object explicitly.

Seems to work! Would exporting the function have worked as well I wonder?

My watch does not reliably react to the hardware button when running the app now. I would suspect it's the changes to how `bthrm` uses the layout module that had made it so. Also the lock widget acts weird. And I'm a little unsure, but after a while the watch seems to get completely stuck until I hold down the hw button and kick the watchdog.

Now when connecting with the Web IDE while `bthrm` is running I can get this in the console:

```
Unable to retrieve board information. Connection Error?
Connected to Web Bluetooth, Bangle.js 92f9 (No response from board)
```

halemmerich:
Exporting would probably have worked as well for the linter, but the boot code would then need to require the app which would not be great for RAM use if it is not in use.

The slowdown/hanging is very weird, I can see that sometimes too. That might come from the change to drawing every time an event arrives instead of once per second.

halemmerich:
Layouting takes about 1 second, the events arrive at similar pace (but draw for every event) so the CPU is a 100% all the time. Lazy layout is even slower, that can't be right, can it?

thyttan:
> Lazy layout is even slower, that can't be right, can it?

Certainly beats the purpose at least 🙃

halemmerich:
I think I got it now. Updating the layout on every redraw is not needed, so the biggest chunk of CPU time was wasted anyway. This should behave a lot better. Funny how it was just on the border of working before with drawing once a second with ~1s draw times. It is now closer to 200ms per draw, much more reasonable.

thyttan:
Everything seems to work now - and much better than before the refactor. Very good work! I think I should merge or do you want to do anything else before I do, @halemmerich ?

Just a small note (does not stop me merging the PR):
The `recorder` app works for getting HR readings from my Polar H10 now. It also add to headings `BT Battery,Energy expended,Contact,` - but I don't get readings for those. I don't know what this means, maybe other hr bt straps supplies them but not the H10.

halemmerich:
I get Battery and Contact from mine. Did not yet see one that does Energy, but it is part of the same few bits delivered in the heart rate BLE characteristic so I just did implement it, it might well be incorrectly implemented but I think it is as the spec defines it. Since the headers and one additional comma per column are the only few bytes stored by the recorder I think it is not really a problem to store the empty columns. If it gets problematic they could be put behind a setting. Ideally including the code that handles the relevant characteristics in the library to save on processing as well.

This should be good for now, you can merge.

thyttan:
Thank you for the great improvements!

halemmerich:
Thanks for the detailed testing!

PR #3350, opened 2024-04-10
Rest - Workout Timer App

devsnd:
An app to keep track of time when not lifting things and keep track of your sets when lifting things.


apps/rest/rest.app.js
```diff
@@ -0,0 +1,322 @@
...
+
+const MAIN_MENU = 'MAIN_MENU';
+const SET_REST = 'SET_REST';
+const SET_SETS = 'SET_SETS';
+const EXERCISE = 'EXERCISE';
+const REST = 'REST';
+const DOIT = 'DOIT';
+
+let mode = MAIN_MENU;
+let lastRender = MAIN_MENU;
```
thyttan, line 227: We get a lint warning here. `'lastRender' is assigned a value but never used` . Please either remove the line or comment it out.

apps/rest/rest.app.js
```diff
@@ -0,0 +1,322 @@
...
+    return;
+  }
+  if (mode === REST) {
+    restUntil = Date.now();  // skipping rest!
+    redrawApp();
+    return;
+  }
+}
+
+setWatch(buttonPress, BTN, { repeat: true, debounce: 25});
```
thyttan, line 305: I think you want to add `edge:"falling"` to the options passed here. Without it the app seemed to try and load my clock face when trying to skip the rest phase.

I'm not sure what debounce does here? removing it seems to have no effect on my watch.

thyttan:
Really fun UI and "Just DO IT!" screen :)

thyttan:
Not handling dark theme:
image

apps/rest/rest.app.js
devsnd: I set the debounce just so that the button would react to shorter presses. I have to long-press to go to my watch-face, do you maybe have another app that interferes here? I'll add the `edge: falling`, but I don't know how that will make a difference either TBQH... If that works better for you, that great :)

thyttan: Reacting on the falling edge makes sure there is no edge left for the next screen to react to.

thyttan: > do you maybe have another app that interferes here?

Yes it's the `fastreset` app that's the difference 🙂

devsnd:
Thanks for your comments @thyttan, I also fixed the dark theme compatibility!

thyttan:
Thanks for the fixes!

I just noticed you have changes to `fileman` here also. Could you do an interactive rebase and drop those commits for this branch?

devsnd:
Whoops, sorry about that, it's fixed!

thyttan:
Thanks! For the app and for the fixes - really fun and useful!

PR #3351, opened 2024-04-10
msgwakefup: new app to activate wake on face up on msg auto display

thyttan:
Temporarily activate wake on face up function when a new message is auto displayed.


apps/msgwakefup/boot.js
```diff
@@ -0,0 +1,5 @@
+// If doing regular loads, not Bangle.load, this is used:
+if (global.__FILE__=="messagegui.new.js") Bangle.setOptions({wakeOnFaceUp:true});
+
+// If Fastload Utils is installed this is used:
+Bangle.on("message", (_, msg)=>{if (Bangle.CLOCK && msg.new) {Bangle.setOptions({wakeOnFaceUp:true});}});
```
bobrippling, line 5: Out of interest - do we get the intended functionality if we have just this `.on("message", ...)` code?

bobrippling:
Looks good! Happy to merge once we've decided on that comment ⬆️

apps/msgwakefup/boot.js
thyttan: I don't think so. In that case the the option would be set, then the bangle reset (thus setting the option to default value again) before loading messagesgui (and here the `__FILE__` logic sets the face up option again).

(reference https://github.com/espruino/BangleApps/pull/3334#issuecomment-2044596680)

thyttan:
I seem to be missing something making this set the option or leave it on when it shouldn't. So wait with the merge. Will mark as draft again.

apps/msgwakefup/boot.js
thyttan: My understanding is that on fastload we don't run through the boot code, so the `__FILE__` check will not run since it only runs one time on boot.

bobrippling: Gotya - out of interest, if you use `prependListener` here, does that help? Then the events will be:
- This handler, which sees there's a `.new` `"message"` and activates `wakeOnFaceUp`
- The message handler, which does whatever (vibrate the watch, alert the user)
- The user then twists their wrist and the screen wakes?

Whereas previously perhaps the message handler ran first and stopped this code from executing?

thyttan: You may be right - I hadn't thought of that. I think generally the option gets set when I expect it to. But maybe I've noticed some time that it didn't.

bobrippling: Strange - let me know if you'd like a hand testing/investigating

thyttan:
Ready for review and/or merge

bobrippling:
Looks good, thanks!

PR #3352, opened 2024-04-10
fileman: allow managing files in the app loader

devsnd:
This adds a small interface to the app loader to manage files in the app loader and not just on the watch. You can list, delete and show files in the loader and it shows the storage stats.

### DIFF
--- a/apps/fileman/ChangeLog
+++ b/apps/fileman/ChangeLog
@@ -1,4 +1,4 @@
 0.01: New app!
 0.02: Improve handling of large amounts of files (fix #579)
 0.03: Update RegExp use (Was using backreference instead of character code)
-0.04: Allow managing files in the app loader
+
--- a/apps/fileman/ChangeLog
+++ b/apps/fileman/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02: Improve handling of large amounts of files (fix #579)
 0.03: Update RegExp use (Was using backreference instead of character code)
+0.04: Allow managing files in the app loader
--- a/apps/fileman/manage_files.html
+++ b/apps/fileman/manage_files.html
@@ -60,9 +60,10 @@
       },
       deleteFile: function (file) {
         if(confirm(`Really delete ${file}?`)) {
-          Puck.eval(`require("Storage").erase("${file}")`);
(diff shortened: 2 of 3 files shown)


apps/fileman/ChangeLog
```diff
@@ -1,3 +1,4 @@
 0.01: New app!
 0.02: Improve handling of large amounts of files (fix #579)
 0.03: Update RegExp use (Was using backreference instead of character code)
+0.04: Allow managing files in the app loader
```
nxdefiant, line 4: There is no need to increment the version number

apps/fileman/manage_files.html
```diff
@@ -0,0 +1,77 @@
...
+      getStats: function () {
+        Puck.eval('require("Storage").getStats()', (stats) => {
+          this.stats = Object.entries(stats);
+        });
+      },
+      deleteFile: function (file) {
+        if(confirm(`Really delete ${file}?`)) {
+          Puck.eval(`require("Storage").erase("${file}")`);
+          this.loadFiles();
+          this.getStats();
```
bobrippling, line 65: Do we want to do this after `Puck.eval`'s done? (in the callback)

bobrippling (commented):
Looks good - do you have this deployed to your gh pages for testing? And as @nxdefiant said, we don't need the version bump

apps/fileman/ChangeLog
devsnd: Ok, I removed it. It's true that there's nothing to update on the watch, but how are people going to know about this new feature in the app loader?

apps/fileman/manage_files.html
devsnd: That's a bug! Thanks for the review, I fixed it.

apps/fileman/ChangeLog
bobrippling: Same kind of way - they'd find new versions of the app in the app loader and they'll come across the interface changes there, although you are right in a way, in that they'd have to find the app in the app loader to start with.

Perhaps people would find it useful if the app loader summarised available `interface.html`s for their installed apps?

bobrippling:
Thanks for the fix and the html!

PR #3353, opened 2024-04-12
boot: fix `throw` stopping clock w/o wdgts loading

thyttan:
This surrounds the `throw` statement with a timeout and is then called only after the clock has loaded. This seems a little weird maybe? But it works. I also tried changing to a `conole.log` instead of the `throw` inside a timeout. That also works, as well as just removing the throw alltogether.

I am not sure this is the best way to solve this. @halemmerich @bobrippling @nxdefiant @gfwilliams do you have other suggestions? Maybe a fix should be somewhere else completely.

E.g. the new [Meridian Clock](https://espruino.github.io/BangleApps/?q=meridian) or [two of them clock](https://espruino.github.io/BangleApps/?q=2ofthemclk) have no widgets.

### DIFF
--- a/apps/boot/ChangeLog
+++ b/apps/boot/ChangeLog
@@ -69,3 +69,6 @@
 0.58: "Make Connectable" temporarily bypasses the whitelist
 0.59: Whitelist: Try to resolve peer addresses using NRF.resolveAddress() - for 2v19 or 2v18 cutting edge builds
 0.60: Minor code improvements
+0.61: Fix `throw` would stop clocks without widgets from being loaded if
+	called by `load(".bootcde")` which is the case when using
+	`Bangle.showClock()`.
--- a/apps/boot/bootloader.js
+++ b/apps/boot/bootloader.js
@@ -3,7 +3,7 @@ var s = require("Storage").readJSON("setting.json",1)||{};
 /* If were being called from JS code in order to load the clock quickly (eg from a launcher)
 and the clock in question doesn't have widgets, force a normal 'load' as this will then
 reset everything and remove the widgets. */
-if (global.__FILE__ && !s.clockHasWidgets) {load();throw "Clock has no widgets, can't fast load";}
+if (global.__FILE__ && !s.clockHasWidgets) {load();setTimeout(()=>{throw "Clock has no widgets, can't fast load";},0);}
 // Otherwise continue to try and load the clock
 var _clkApp = require("Storage").read(s.clock);
 if (!_clkApp) {
--- a/apps/boot/metadata.json
+++ b/apps/boot/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "boot",
   "name": "Bootloader",
-  "version": "0.60",
+  "version": "0.61",
   "description": "This is needed by Bangle.js to automatically load the clock, menu, widgets and settings",
   "icon": "bootloader.png",
   "type": "bootloader",


halemmerich:
Wrapping the throw in a timeout effectively is the same as just removing it, I think that would not be good. On a clean install "two of them" works just fine for me. Even with fastload there are no problems. Meridian gives me a problem with regard to clock_info, but that is probably not related.

thyttan:
@halemmerich I investigated some more:

### To reproduce:

1. Call Bangle.factoryReset() in Web IDE console field.
2. Uninstall antonclk via the App Loader
3. Install 2ofthemclk
4. Call `load(".bootcde")` from the Web IDE console field. (Works this time)
5. Call `load(".bootcde")` again from the Web IDE console field. (This time stuck on "Reloading..." message)
5. Call `load(".bootcde")` once more from the Web IDE console field. (This time stuck on "Loading..." message)
6. Remove the `throw` statement from `.bootcde` and upload the change via the Web IDE.
7. Call `load(".bootcde")` from the Web IDE console field. (Now works repeatedly)


### Web IDE console interaction:

```
>Bangle.factoryReset()
Erasing Storage Area...
Erase complete.
Writing initial storage contents.../
Write complete.
=undefined
>
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
Uncaught Storage Updated!
>[{"id":"boot","version":"0.60","files":"boot.info,.boot0,.bootcde,bootupdate.js","type":"bootloader"},{"id":"launch","version":"0.20","files":"launch.info,launch.app.js,launch.settings.js","data":"launch.json,launch.cache.json","type":"launch"},{"id":"antonclk","version":"0.11","files":"antonclk.info,antonclk.app.js,antonclk.img","type":"clock"},{"id":"about","version":"0.15","files":"about.info,about.app.js,about.img"},{"id":"widlock","version":"0.08","files":"widlock.info,widlock.wid.js","type":"widget"},{"id":"widbat","version":"0.12","files":"widbat.info,widbat.wid.js","type":"widget"},{"id":"widbt","version":"0.09","files":"widbt.info,widbt.wid.js","type":"widget"},{"id":"widid","version":"0.03","files":"widid.info,widid.wid.js","type":"widget"},{"id":"welcome","version":"0.14","files":"welcome.info,welcome.boot.js,welcome.app.js,welcome.settings.js,welcome.img","data":"welcome.json"},{"id":"notify","version":"0.13","files":"notify.info,notify","type":"notify"},{"id":"health","version":"0.29","files":"health.info,health.app.js,health.img,health.boot.js,health,health.settings.js","data":"health.json"},{"id":"setting","version":"0.71","files":"setting.info,setting.app.js,setting.img","data":"setting.json"},{"id":"sched","version":"0.25","files":"sched.info,sched.boot.js,sched.js,sched.img,sched,sched.settings.js,sched.clkinfo.js","data":"sched.json,sched.settings.json","type":"scheduler"},{"id":"widalarm","version":"0.02","files":"widalarm.info,widalarm.wid.js","type":"widget"},{"id":"alarm","version":"0.47","files":"alarm.info,alarm.app.js,alarm.img,alarm.settings.js","data":"alarm.settings.json"},"BANGLEJS2","2v21.82",725356,1713003104,70418672]
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
>{"id":"antonclk","name":"Anton Clock","type":"clock","src":"antonclk.app.js","icon":"antonclk.img","version":"0.11","tags":"clock","files":"antonclk.info,antonclk.app.js,antonclk.img"}
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
>OK
OK
OK
OK
OK
OK
OK
OK
OK
OK
load(".bootcde")
=undefined
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
>load(".bootcde")
=undefined
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
Uncaught Clock has no widgets, can't fast load
>load(".bootcde")
=undefined
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
Uncaught Clock has no widgets, can't fast load
>
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
>
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
>
>load(".bootcde")
=undefined
 ____                 _
         |_| espruino.com
 2v21.82 (c) 2024 G.Williams
> 
```

bobrippling:
To check I've understood, is the issue that if we have the clock set to one without widgets, we see this error? (with fastload?)

Looking at the code, the intention behind the `throw` seems to be to return early (we can't `return` since we're not in a function), so one approach might be to swap to an iife?

```diff
+(() => {
   ...
-  if (global.__FILE__ && !s.clockHasWidgets) {load();throw "Clock has no widgets, can't fast load";}
+  if (global.__FILE__ && !s.clockHasWidgets) {load();return;}
   // Otherwise continue to try and load the clock
   ...
+})()
```

Although I wonder why it stops clocks without widgets from loading. I'll look into it but at a guess, is the throw interrupting something the widgetless clock has registered?

thyttan:
> To check I've understood, is the issue that if we have the clock set to one without widgets, we see this error? (with fastload?)

Yes, the `load('.bootcde')` is what's not working.

gfwilliams:
Ok, so I think the issue here is that we can't call `load` from within a `load(...)` call, because it's then ignored? The whole point of the `throw` is to stop execution while allowing `load()` to run and reload the clock cleanly, so we don't actually want to change that.

`load(".bootcde")` triggers this, because __FILE__ is then set to `".bootcde"` instead of undefined (although the same code is run in either case).

And you can trigger it with `Bangle.showClock();` **if** the current clock doesn't support fast load, because via `Bangle.load` it ends up calling `load(".bootcde")`: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_load.js

However in that case we shouldn't be throwing the error anyway, because we're not actually trying to fast load - we were being called during `load` which by means we're not fast loading anyway, and it all would have been fine.

We could check `global.__FILE__!=".bootcde"` but then the code in `Bangle.load` would bypass this in the fastload case, which defeats the point of the check: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_load.js#L5-L6

So I think the change needs to happen in `Bangle.load` so it calls `load(undefined)` in the .bootcde case. Please can you try adding this as custom boot code, but without the changes in this PR and see if that fixes it?

```
Bangle.load = (function(name) {
  if (Bangle.uiRemove) {
    Bangle.setUI(); // remove all existing UI (and call Bangle.uiRemove)
    __FILE__=name;
    if (!name) name = ".bootcde";
    setTimeout(eval,0,require("Storage").read(name)); // Load app without a reboot
  } else load(name!=".bootcde"?name:undefined); //<---- changed here
});
```

If it does I'll make a change to the firmware.

thyttan:
@gfwilliams Yes, `Bangle.load()` and `Bangle.showClock()` both work with that as custom boot code (`Bangle.load()` seems to work even without the your fix).

gfwilliams:
> Bangle.load() seems to work even without your fix 

Yes, but `Bangle.load(".bootcde")` probably breaks?

Ok, cool - I'll put that fix in then

thyttan:
> Yes, but Bangle.load(".bootcde") probably breaks?

Yes that's right!

PR #3354, opened 2024-04-13
[messagegui] fix: handle the unreadTimeout better

paul-arg:
This MR does two things:

- move the code that sets the `unreadTimeout` to a new function called `resetReloadTimeout`
- each time a new message is received, call `resetReloadTimeout`

This way, if you receive a lot of consecutive messages, you are guaranteed to be able to read the last one, because your timeout starts from its reception and not from the reception of the first one like before.


bobrippling:
Makes sense to me although I don't use this app - @gfwilliams I believe you're the main contributer to it, ok to merge?

gfwilliams:
Looks good to me - but just to check, you're **sure** that it does still time out in normal usage?

Because the timeout is no longer set when the app starts - only from `showMessage`?

PR #3361, opened 2024-04-17
recorder: fix accessing WIDGETS

bobrippling:
On clocks without `WIDGETS`, the previous code would hit a `TypeError`

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -52,3 +52,4 @@
 0.41: Fix exit from plots and graphs would easily react twice, going back two
 	levels instead of one.
 0.42: Minor code improvements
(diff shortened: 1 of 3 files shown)


gfwilliams:
I wasn't even aware there was a recorder clockinfo! :)

Looks good - I guess in an ideal world we wouldn't even add the recorder clockinfo if there are no widgets loaded - but it seems like loadWidgets is called after the clockinfos are loaded, so we won't ever know.

thyttan:
> but it seems like loadWidgets is called after the clockinfos are loaded, so we won't ever know.

@gfwilliams We could emit an event when widgets are loaded that clock-info could react to and add/remove the recorder entry. May be over engineering 🙃 ?

PR #3362, opened 2024-04-17
[thunder] Use the locale module to display distance

jt-nti:


jt-nti:
This PR changes the default units setting to use the locale module, based on a suggestion by @halemmerich. This will break the units if anyone is already using the app _and_ has changed the settings but it's still a new app, which is unlikely to be widely used, and it's even more unlikely that anyone has updated the settings, so hopefully that's ok!

bobrippling:
LGTM, I don't think the default change is a worry :)

Issue #3355, opened 2024-04-13
[quicklaunch] Wrongly starts app when swiping clock info


Spioune:
### Affected hardware version

Bangle 2

### Your firmware version

2v21

### Bug

On a clock with 2 (or more?) clock info (like Pebble++ Clock):
1. Select the top right clock info
2. Select the top left clock info
3. Swipe left/right/up/down with the intention to switch the top left clock info will trigger quicklaunch instead.

Based on my tests this only happens on the top left clock info and this only happens if you select the top right clock info prior.

### Steps to reproduce

1. Install a clock with 2 clock info like [Pebble++ Clock](https://banglejs.com/apps/?id=pebblepp)
2. Install [quicklaunch](https://banglejs.com/apps/?id=quicklaunch)
3. In Settings > Apps > Quick Launch, setup a quick action on swipe left, right, up or down
4. Start Pebble++ clock from the main menu
5. Select the top right clock info
6. Select the top left clock info
7. Swipe left, right, up or down
8. The quicklaunch is triggered

This bug seems to happen on every clock with more than 2 clock info (#3347)

### Installed apps

alarm (0.46), widalarm (0.01), sched (0.22), setting (0.68), health (0.28), about (0.14), launch (0.20), boot (0.59), antonclk (0.11), widlock (0.08), widbat (0.12), widbt (0.09), widid (0.03), welcome (0.14), notify (0.13), quicklaunch (0.15), clock_info (0.09), pebblepp (0.05)

bobrippling:
I'd expect this code to cover it:

https://github.com/espruino/BangleApps/blob/c2ea454a3bc7bfaefbc39a5b376510bb9d7d4ec8/apps/quicklaunch/boot.js#L36-L36

Perhaps the clkinfo is defocussed and then this handler runs? But this should be covered by this code:

https://github.com/espruino/BangleApps/blob/c2ea454a3bc7bfaefbc39a5b376510bb9d7d4ec8/apps/clock_info/lib.js#L286-L287

thyttan:
Is it maybe a case of using `.on(...) and not `.prependListener(...)` somewhere? I wouldn't expect that to be it - but maybe?

thyttan:
Ok - I think what happens is (these are guesses at the moment)
1. `quicklauch`'s swipe handler is registered before the clock_info's on the clock face so the `stopEventPropagation` hasn't been called yet so `quicklaunch` runs. 
[...]

PR #3364, opened 2024-04-17
clock_info: prepend swipe listener if possible

thyttan:
attempt to fix https://github.com/espruino/BangleApps/issues/3355


bobrippling:
Thinking about this more, I suppose it's a good change to have since we want to be the first listener always, as when we don't have focus we're totally "transparent", if you will:

https://github.com/espruino/BangleApps/blob/b934653ff0c91fe649f9add2068de2fbb7e65d2d/apps/clock_info/lib.js#L256

thyttan:
Yes I think something like this should go in. I was thinking maybe wait a week as to not conflate some bug that might arise. But the risk feels small so we could also just go ahead.

bobrippling:
Yeah I think we merge it, it's low risk

thyttan:
I'll merge this now then.

PR #3366, opened 2024-04-18
clkinfo: make `CLKINFO_FOCUS` a counter

### LINKED ISSUE #3355: [quicklaunch] Wrongly starts app when swiping clock info

bobrippling:
This avoids issues when we have multiple clkinfos visible. The first clkinfo may want to focus itself, but a second clkinfo may then unfocus, deleting the first's effects (i.e. setting `CLKINFO_FOCUS`).

In detail:

`CLKINFO_FOCUS` will be set (to `true`) by the first handler (left hand clkinfo) and then unset (/ `delete`d) by the second handler (right hand clkinfo).

This is done in [the `touchHandler`](https://github.com/espruino/BangleApps/blob/c2ea454a3bc7bfaefbc39a5b376510bb9d7d4ec8/apps/clock_info/lib.js#L313-L324).

This commit makes `CLKINFO_FOCUS` a count.

- When we have a tap on a previous clkinfo (previous as in, earlier in the event handler list), it'll focus itself, `CLKINFO_FOCUS++`
- The later handler comes along, unfocuses itself, `CLKINFO_FOCUS--`
- `CLKINFO_FOCUS` is now `1`, so clkinfo is registered as being focussed still, meaning quicklaunch will work because of:

> https://github.com/espruino/BangleApps/blob/c2ea454a3bc7bfaefbc39a5b376510bb9d7d4ec8/apps/quicklaunch/boot.js#L36-L36

Fixes #3355

### DIFF
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -289,7 +289,7 @@ exports.addInteractive = function(menu, options) {
   Bangle.on("swipe",swipeHandler);
   const blur = () => {
     options.focus=false;
-    delete Bangle.CLKINFO_FOCUS;
+    Bangle.CLKINFO_FOCUS--;
     const itm = menu[options.menuA].items[options.menuB];
     let redraw = true;
     if (itm.blur && itm.blur(options) === false)
@@ -298,7 +298,7 @@ exports.addInteractive = function(menu, options) {
   };
   const focus = () => {
     let redraw = true;
-    Bangle.CLKINFO_FOCUS=true;
+    Bangle.CLKINFO_FOCUS = (Bangle.CLKINFO_FOCUS || 0) + 1;
     if (!options.focus) {
       options.focus=true;
       const itm = menu[options.menuA].items[options.menuB];
@@ -341,7 +341,7 @@ exports.addInteractive = function(menu, options) {
     Bangle.removeListener("swipe",swipeHandler);
     if (touchHandler) Bangle.removeListener("touch",touchHandler);
     if (lockHandler) Bangle.removeListener("lock", lockHandler);
-    delete Bangle.CLKINFO_FOCUS;
+    Bangle.CLKINFO_FOCUS--;
     menuHideItem(menu[options.menuA].items[options.menuB]);
     exports.loadCount--;
     delete exports.clockInfos[options.index];
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -298,7 +298,7 @@ exports.addInteractive = function(menu, options) {
   };
   const focus = () => {
     let redraw = true;
-    Bangle.CLKINFO_FOCUS = (Bangle.CLKINFO_FOCUS || 0) + 1;
+    Bangle.CLKINFO_FOCUS = (0 | Bangle.CLKINFO_FOCUS) + 1;
     if (!options.focus) {
       options.focus=true;
       const itm = menu[options.menuA].items[options.menuB];
--- a/apps/clock_info/ChangeLog
+++ b/apps/clock_info/ChangeLog
@@ -8,3 +8,4 @@
 0.07: Developer tweak: clkinfo load errors are emitted
 0.08: Pass options to show(), hide() and run(), and add focus() and blur() item methods
 0.09: Save clkinfo settings on kill and remove
+0.10: Fix focus bug when changing focus between two clock infos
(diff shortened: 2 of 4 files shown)


thyttan:
> I think the counter solution seems like a good and relatively simple fix!
> 
> I seem to remember some discussion about removing `Bangle.CLKINFO_FOCUS` altogether in favour of just using `.prependListener` and `.stopEventPropagation`. But since it wouldn't be compatible with older firmwares it was not acted on I think. But I don't remember where.

*https://github.com/espruino/BangleApps/issues/3355#issuecomment-2064710458*

Like I said I think this seems like a good thing to merge. But want to let @gfwilliams chime in as well. I also haven't tried it yet.

bobrippling:
Thanks, I'll tag @glemco too, having done the original `CLKINFO_FOCUS` code :)

glemco (commented):
Looks good to me. I probably lost a little bit of touch with this code. What is a use case for having _nested_ focus?
I tested this code initially with something like circlesclock and there, tapping on one clkinfo sets the focus, tapping on another should unfocus (blur) the first and focus the new one (in this case just re-setting the variable to true).
Is blur no longer called before the next focus in this scenario? Can we have two clkinfo concurrently selected?

thyttan:
> Is blur no longer called before the next focus in this scenario?

It's not guaranteed to. It depends on the ordering of the touch handlers. If one is selected and you immediately click the other one it might happen that you `focus` the new and then the previous is `blur`:ed, leaving `Bangle.CLKINFO_FOCUS` `undefined` where it ought to be true.

> Can we have two clkinfo concurrently selected?

No, that doesn't happen afaik.

bobrippling:
Yeah, @thyttan's pretty much summed it up, to add one extra bit - it's less of a nested focus and more that the count prevents one clkinfo from clobbering what another one might set, if one gets focused before the other one's blur handler has run.

I've tested this and it looks to have resolved the bug in `quicklaunch`

apps/clock_info/lib.js
```diff
@@ -298,7 +298,7 @@ exports.addInteractive = function(menu, options) {
   };
   const focus = () => {
     let redraw = true;
-    Bangle.CLKINFO_FOCUS=true;
+    Bangle.CLKINFO_FOCUS = (Bangle.CLKINFO_FOCUS || 0) + 1;
```
gfwilliams, line 301: Please could we make this just `(0|Bangle.CLKINFO_FOCUS) + 1;`?

It's just a bit more robust (eg if `Bangle.CLKINFO_FOCUS=="1"` the result of running this code is the string `"11"`)

bobrippling: Sure, sorted 👍

gfwilliams:
Thanks! This looks good, however I honestly thought this worked properly at some point!

As @glemco mentioned the idea was that the previous item would be blurred before the next was focussed - but maybe that never actually happened, and I only tested the case where I tapped outside the current clkinfo and then onto the next one, rather than tapping directly on the next :)

glemco (approved):

thyttan:
OK - thanks everybody!

bobrippling:
Thanks for the reviews and discussion :)

PR #3369, opened 2024-04-20
drained: change enable/disable BLE depending on state

bobrippling:
This makes it more clear when the user enables bluetooth, and allows them to also turn it back off

### DIFF
--- a/apps/drained/ChangeLog
+++ b/apps/drained/ChangeLog
@@ -3,3 +3,4 @@
 0.03: Permit exceptions to load in low-power mode, e.g. daylight saving time.
       Also avoid polluting global scope.
 0.04: Enhance menu: enable bluetooth, visit settings & visit recovery
+0.05: Enhance menu: permit toggling bluetooth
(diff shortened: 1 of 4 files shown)


thyttan:
Looks good! There's a lint warning for older parts of the code:

```
/home/runner/work/BangleApps/BangleApps/apps/drained/app.js
Warning:   3:5   warning  'drainedInterval' is not defined  no-undef
Warning:   3:37  warning  'drainedInterval' is not defined  no-undef
```

Do you want to fix that while you're at it?

bobrippling:
Ah yes - this is in generated code, typescript doesn't like us indexing into `global` but I have made it clear we're accessing a global in a [different way in typescript](https://github.com/espruino/BangleApps/blob/154a9075d5fb775f2e5eeabf980b1ba4ab705810/apps/drained/app.ts#L4).

I've added [a commit](https://github.com/espruino/BangleApps/pull/3383/commits/a759eb5cba07552189ff5a5942d12901fe5e9bbb) to stop the linter looking at generated files too, to keep the noise down :)

thyttan:
Thanks 🙂

PR #3370, opened 2024-04-20
kbmulti: reset caps when empty and add vibrate setting

bobrippling:
This turns caps back on if we clear the input (to match the initial state) and adds a setting to vibrate on interaction

### DIFF
--- a/apps/kbmulti/ChangeLog
+++ b/apps/kbmulti/ChangeLog
@@ -8,3 +8,5 @@
 	edge versions of 2v18. Falls back on `Bangle.on()` for backwards
 	compatability.
 0.08: Minor code improvements
+0.09: Recapitalise when returning to the start, and add setting to vibrate on
+	interaction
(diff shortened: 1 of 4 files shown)


thyttan:
Nice additions!

What do you think of the situation when
1. no character written
2. user toggles to lower case
3. user does a backspace swipe

?

Should it reset to upper case? Now the lower case is preserved.

bobrippling:
Thank you :)

So that shouldn't happen, because we only perform the upper casing if we've been able to delete a character (line 79), I can't get it to happen on my watch, do you see it?

https://github.com/espruino/BangleApps/blob/d6483983061bc998632d9e690ff1ba110e3c3f52/apps/kbmulti/lib.js#L77-L83

thyttan:
No, sorry for being unclear. It doesn't reset for me either. I was thinking that it could be a little confusing maybe that it didn't restore upper case in that situation. But I don't know really 🤷‍♂️

thyttan:
Happy to merge if you are!

thyttan:
OK thanks!

bobrippling:
Thanks for the review + merge!

PR #3372, opened 2024-04-20
clock_info.d.ts: improve `img` type

bobrippling:
... by reusing the `Image` type


bobrippling:
Merging: low risk, typescript-only

PR #3373, opened 2024-04-21
runplus - Add lock, gps, pulse indicators to karvonen screen

halemmerich:
To do this cleanly I did refactor the karvonen code into a library. @thyttan Do you have an opinion on changing to a library and requiring it?


thyttan:
Don't believe I have any objections against switching to require. IIrc you guided me to the current solution🙂

thyttan:
I took it for a run the other day also - worked well!

PR #3374, opened 2024-04-21
PhysTrax: v0.01

ekwawu:
Allows user to record heart rate data and download to csv file.


apps/phystrax/metadata.json
```diff
@@ -0,0 +1,16 @@
...
+  "description": "Tracking physiological measurements to support active learning in classrooms",
+  "tags": "health",
+  "interface": "interface.html",
+  "readme": "README.md",
+  "supports": ["BANGLEJS2"],
+  "storage": [
+    {"name":"phystrax.app.js","url":"app.js"},
+    {"name":"phystrax.img","url":"app-icon.js","evaluate":true}
+  ],
+  "data": [{"wildcard":"heart_rate_data.?.csv"}]
```
bobrippling, line 15: Looks like a small issue here

> `App phystrax data file wildcard heart_rate_data.?.csv does not include app ID`

The fix being something like this (we'd also need to update `interface.html` too):

```suggestion
  "data": [{"wildcard":"phystrax_hrm.?.csv"}]
```
(or whatever you'd like to call the file)

bobrippling (commented):
Great stuff! Just the lint issue in `metadata.json` if you could resolve that

apps/phystrax/metadata.json
ekwawu: Okay, I can do that. Thanks!

bobrippling: Cool, I've matched the ones in other sources and merged - thanks for the app!

PR #3375, opened 2024-04-21
ha: add value triggers

bobrippling:
... for example, to adjust light brightness:

image

Currently this works through a non-GadgetBridge shim I have, which pretends to be GB in order to respond to these messages. Interested in what people think.

@peerdavid would you find anything like this useful to include in `ha`? I can move this to another app (although the changes shouldn't interfere with the original behaviour)

@thyttan I've made a few tweaks to the slider library too, mainly:
- Pass `e` through to the callback (the drag event)
- Reset `dragActive` at the end of a drag
- Always send through the final event (so users can detect `e.b == 0` i.e. finger lift up)

### DIFF
--- a/apps/ha/ChangeLog
+++ b/apps/ha/ChangeLog
@@ -9,3 +9,4 @@
 0.09: Improve web interface, arrows in UI
 0.10: Issue newline before GB commands (solves issue with console.log and ignored commands)
 0.11: Minor code improvements
+0.12: Add slider functionality
--- a/apps/ha/README.md
+++ b/apps/ha/README.md
@@ -3,10 +3,12 @@ This app integrates your Bangle.js into the Home Assistant.
 
 
 # How to use
-Click on the left or right side of the screen to select the triggers that you configured. 
+Click on the left or right side of the screen to select the triggers that you configured.
 Swiping left or right works as well.
 
 Click in the middle of the screen to send the trigger to Home Assistant via Gadgetbridge.
+
+If the trigger is a value one (has `.value = true` in the config), a slider will be displayed for you to alter the value that's sent along with the trigger.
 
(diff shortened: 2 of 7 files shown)


peerdavid:
Looks fantastic!

thyttan:
> I've made a few tweaks to the slider library too, mainly:
> 
>     * Pass `e` through to the callback (the drag event)
> 
>     * Reset `dragActive` at the end of a drag
> 
>     * Always send through the final event (so users can detect `e.b == 0` i.e. finger lift up)

Ok - those seem like good changes to me! :+1: 

Can you update the Slider.md documentation as well?

With this `ha` will be the first app on the official app loader that uses the slider library - I'm glad to see it! ;)

bobrippling:
> Looks fantastic!

> Ok - those seem like good changes to me! 👍
> 
> Can you update the Slider.md documentation as well?

Thanks! Good call, done - let me know if there's anything you'd like that's not in there :)

> With this `ha` will be the first app on the app loader to use the slider library - I'm glad to see it! ;)

🚀 🎉 

This will double next - I'm thinking of using two sliders for lights which have temperature and brightness - but I'll leave that to another PR.

thyttan:
> Thanks! Good call, done - let me know if there's anything you'd like that's not in there :)

No that's good from what I can see. If you think other changes to Slider.md would be nice please go ahead.

bobrippling:
Thanks - I've finished testing and made a few small usability tweaks for when sliders are shown too. Happy to leave this for a little while to test more though. And I'll see what other docs to add as I work on version 2!

bobrippling:
Cool, happy with how it works for now and have given it further testing - thanks for the library!

PR #3377, opened 2024-04-22
Remapping of minute intervals to add accuracy

enricpineda:
I've changed the minute intervals, so it reflects more accuretly the hour in Catalan.


apps/rellotge/rellotge.js
```diff
@@ -129,51 +129,51 @@
     var m = d.getMinutes();
 
     let t;
-    if (m >= 0 && m < 2) {
+    if (m = 0) {
```
bobrippling, line 132: for example here:
```suggestion
    if (m == 0) {
```

bobrippling (commented):
Thanks - looks like the linter's found a few small issues that will stop the app from working how I imagine you intend :)

apps/rellotge/rellotge.js
enricpineda: Surely my sight is not as good as it used to be. Didn't see it coming, but now it's corrected.

enricpineda:
The esperanto file is still in progress, I'm translating it step by step.

apps/rellotge/rellotge.js
bobrippling: Thanks - easy missed!

bobrippling:
> The esperanto file is still in progress, I'm translating it step by step.

No rush - we could always make this PR a draft in the meanwhile

enricpineda:
Now the esperanto translation is finished! Hope everything is OK!

bobrippling:
Thanks!

PR #3378, opened 2024-04-22
analogquadclk: fix some fastloading logic

thyttan:
Fixed some hiccups. 

I ran some ram tests and the usage seemed to settle at 3784 when loaded and 2216 after a subsequent `Bangle.setUI()` call on a factory reset watch, fw 2v21.107.

There are still some things I didn't fix/change:
- When the hands are redrawn part of a clock info graphics can be cleared when it shouldn't 
- The tip of the minutes hand are overdrawn at times. Not sure if it's the hidden widget field, clock info or other that's the culprit.
- Just noticed the draw timeout can still survive the unloading also after my attempted (maybe very naive) fix a9c04d3bdcc5da0561be47ffc9dd1f99d359ebe9.

### DIFF
--- a/apps/analogquadclk/app.js
+++ b/apps/analogquadclk/app.js
@@ -32,7 +32,6 @@ const hand_hour_bg = get_hand(HOUR_LEN, HOUR_W, 8, HOUR_BACK);
 const hand_minute = get_hand(MIN_LEN, 4, 3, MIN_BACK);
 const hand_minute_bg = get_hand(MIN_LEN, MIN_W, 6, MIN_BACK);
 
-
 // schedule a draw for the next minute
 function queueDraw() {
   if (drawTimeout) clearTimeout(drawTimeout);
@@ -212,7 +211,6 @@ if (clockInfoItemsBangle) {
       });
   }
 }
-
 
 // Add the 4 clockinfos
 const CLOCKINFOSIZE = 50;
--- a/apps/analogquadclk/app.js
+++ b/apps/analogquadclk/app.js
@@ -236,6 +236,10 @@ Bangle.setUI({
   remove: function() {
     if (drawTimeout) clearTimeout(drawTimeout);
     clockInfoMenuA.remove();
+    clockInfoMenuB.remove();
+    clockInfoMenuC.remove();
+    clockInfoMenuD.remove();
+    require("widget_utils").show(); // re-show widgets
   }
 });
 // Load widgets
--- a/apps/analogquadclk/app.js
+++ b/apps/analogquadclk/app.js
@@ -87,6 +87,21 @@ function draw() {
   lastModified = getHandBounds();
   //g.drawRect(lastModified); // debug
 }
+// Show launcher when middle button pressed
+Bangle.setUI({
+  mode: "clock",
+  remove: function() {
+    if (drawTimeout) clearTimeout(drawTimeout);
+    clockInfoMenuA.remove();
+    clockInfoMenuB.remove();
+    clockInfoMenuC.remove();
+    clockInfoMenuD.remove();
+    require("widget_utils").show(); // re-show widgets
+  }
+});
+// Load widgets
+Bangle.loadWidgets();
+require("widget_utils").hide();
 
 // Clear the screen once, at startup
 background.fillRect(0, 0, W - 1, H - 1);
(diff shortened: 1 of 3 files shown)


apps/analogquadclk/app.js
```diff
@@ -86,6 +86,21 @@ function draw() {
...
+// Show launcher when middle button pressed
+Bangle.setUI({
+  mode: "clock",
+  remove: function() {
+    if (drawTimeout) clearTimeout(drawTimeout);
+    clockInfoMenuA.remove();
+    clockInfoMenuB.remove();
+    clockInfoMenuC.remove();
+    clockInfoMenuD.remove();
+    require("widget_utils").show(); // re-show widgets
```
bobrippling, line 98: You might also want a:
```javascript
require("widget_utils").cleanup();
```

thyttan: Not aware of that one - will look!

thyttan: Should I just use the cleanup or both?

> Using Fast Loading? You'll have to re-display widgets with require("widget_utils").show() when your app unloads.

*https://www.espruino.com/Bangle.js+Hideable+Widgets*

bobrippling: ~~I think both, `show()` to bring the widgets back in, then cleanup to detach the `widget_utils` specific listeners, etc~~

Ignore me, I've just seen that `show()` calls `cleanup()`, doh

https://github.com/espruino/BangleApps/blob/f3b639f3e920fa74210b1aee2f9c62b16f21b61e/modules/widget_utils.js#L16-L18

But I did spot - `cleanup` deletes `Bangle.appRect`, which I think is a bug - the app we're fastloading into might rely on this for drawing, it should restore it instead. If that sounds right to you, I'll open a PR to fix it

gfwilliams: Thanks - but I think `delete Bangle.appRect; ` is fine. If you delete a built-in field then it doesn't really delete it - what it's doing is deleting the 'overwritten' appRect (because when you hide widgets the appRect has to be overwritten with one that includes the full screen)

apps/analogquadclk/app.js
```diff
@@ -86,6 +86,21 @@ function draw() {
   lastModified = getHandBounds();
   //g.drawRect(lastModified); // debug
 }
+// Show launcher when middle button pressed
```
gfwilliams, line 89: Was there a reason for moving this upwards?

It's just that if (it's unlikely I know) `setUI` gets called again after this, clockInfoMenuA/etc aren't defined at this point.

apps/analogquadclk/app.js
```diff
@@ -31,12 +32,11 @@ const hand_hour_bg = get_hand(HOUR_LEN, HOUR_W, 8, HOUR_BACK);
 const hand_minute = get_hand(MIN_LEN, 4, 3, MIN_BACK);
 const hand_minute_bg = get_hand(MIN_LEN, MIN_W, 6, MIN_BACK);
 
-
 // schedule a draw for the next minute
 function queueDraw() {
   if (drawTimeout) clearTimeout(drawTimeout);
   drawTimeout = setTimeout(function() {
-    drawTimeout = undefined;
+    //drawTimeout = undefined;
```
gfwilliams, line 39: what's the reason for this one? It feels like a good thing to do in general, as at this point the timeout isn't valid any more.

gfwilliams: Oh, sorry - just spotted your reasoning for this. Got to say I'm not convinced about it - I'm not really sure I see how it can help

gfwilliams:
Thanks! What a pain - that was stupid of me not to check that the fast load stuff was unloading properly!

I totally agree on making the hour move with minutes too, and the rest of the stuff. I've had it running for a whole here and it's odd - the TL, TR and BL clockinfos seem to not be overwritten by the clock hands, but BR is?

gfwilliams:
Ok, I just pulled in these changes including the move of setUI (albeit with a check). Also I had to move all function declarations to `let x = function`.

Tested fastload and it seems to be able to reload itself again without losing memory now.

Also made the hour hand move with the minutes, and fixed the clockinfo overlap issue

I doubt that's fixed the issue with hands getting overwritten though - if you find a way to reproduce it, let me know!

thyttan:
> Also I had to move all function declarations to let x = function.

Ah, right. Forgot about that!

apps/analogquadclk/app.js
bobrippling: Ah nifty! Every day's a school day

PR #3379, opened 2024-04-22
multitimer: handle undefined alarm.data

bobrippling:
A minor bugfix to prevent exceptions in the multitimer event handler


bobrippling:
No issues from a few days of use, merging

PR #3380, opened 2024-04-22
sched: fix snoozing an alarm after the snooze period

bobrippling:
We now set the snooze period from when the user hits snooze, not when the alarm sounded.

For example:
- Snooze period is 10 mins
- Alarm sounds for 11 mins
- User hits snooze
- Alarm is scheduled/snoozed until 23 hours 59 minutes in the future


bobrippling:
Central app change, tagging @gfwilliams

gfwilliams:
Looks great - thanks!

PR #3382, opened 2024-04-24
[thunder] Show lightning image in dark themes

jt-nti:


jt-nti:
I'm trying to get the lightning image to show up properly in a dark theme but no luck so far. Any tips gratefully received!

stweedo:
I think the easiest way to get the image to show correctly is to use two versions, one for each theme, and draw the appropriate one based on the `setting.json` value of `theme.dark`. 

```
const isDarkTheme = require("Storage").readJSON("setting.json", 1).theme.dark;

// Define the image source based on the theme
var imageSource;
if (isDarkTheme) {
  imageSource = require("heatshrink").decompress(atob("h0awkBiMQ/4ABAoIGOAoX/AoIGKCYQGGFBQTCAyIAPgIGGIAoFCAxITDAxIFDAxATEJwIMEAw4TEAwo="));
} else {
  imageSource = require("heatshrink").decompress(atob("h0awkBiMQ/4ABAoIGOAoX/AoIGKCYQGGFBQTCAyIAPgIGGIAoFCAxITDAxIFDAxATEJwIMEAw4TEAwo="));
}

var layout = new Layout({
  type: "v", c: [
    { type: "txt", font: "6x8:2", label: "", id: "time", fillx: 1 },
    { type: "txt", pad: 8, font: "20%", label: "", id: "distance", fillx: 1 },
    { type: "h", c: [
      { type: "btn", font: "6x8:2", label: "Flash", cb: l => flash() },
      { type: "img", pad: 4, col: g.theme.fg, bgCol: g.theme.bg, src: imageSource },
      { type: "btn", font: "6x8:2", label: "Boom", cb: l => boom() }
    ]},
  ]
}, {
  lazy: true,
  back: load,
});
```

I also noticed that repeat presses of the "Flash" button causes the timer to slow down more and more with each press. Resetting this timer within the `flash` function like you do in the `boom` function would prevent this from happening. You could make a separate function that clears the timer and then call that when either button is pressed. 

Also, thank you for the cool app!

jt-nti:
Thanks for having a look @stweedo! Using two versions is definitely a good backup plan but the icons library seems to work with a single image. Maybe [the settings](https://github.com/espruino/BangleApps/blob/master/apps/icons/gen/generate.js#L12) are the trick? (I also wondered about drawing the lightning but that might be a step too far for my artistic abilities, even for something so simple!
I'll definitely have a look at fixing the flash button as well. Thanks again!

bobrippling:
I believe there's a trick todo with having the image be a single bit per pixel and then you can choose the colour of this bit, but I can't rememeber. @gfwilliams will know what I'm on about

gfwilliams:
Yes - if you make an image with the [image converter](https://www.espruino.com/Image+Converter) and it's 1bpp (I think even 2bpp works) *without a palette* then Espruino will use the current foreground and background colours when rendering it. In the image converter preview, what's white will be the fg color, what's black is the bg color.

Just don't use `Optimal 1 bit` when converting as that adds a colour palette to the image, which Espruino will then honour over the theme.

gfwilliams:
I just converted it - I think what you want is:

```
atob("DhoBH+B/gf4P8D/A/gP4H+B/AfwH8B//////7/+A/APwD4A+APADwA4AcAHABgAYAA==");
```

The old image was 4 bit for some reason even though it was only black/white, so was using 4x the space that was needed.

And @stweedo if you can at all, please avoid having two images in apps as it also means they won't work if someone chooses colors that aren't  black/white for the theme. Are there any apps you're aware of that are using two images right now?

jt-nti:
Fantastic, thanks @bobrippling @gfwilliams!

jt-nti:
Yay, it works, thanks for all the help!

bobrippling:
LGTM, thanks!

PR #3383, opened 2024-04-26
New app: ctrlpad

bobrippling:
A new app, which lets you drag down from the top of your watch to show a little control drawer.

This also upgrades typescript to `5.4.5` and brings in the latest espruino definitions (`main.d.ts`).


thyttan:
Tried it a little -  I don't think it's my style but it's pretty cool!

I have some suggestions, but don't feel you need to act on them. And if you do want to add them it can be done in a future PR.

- It's kind of nice that it uses drags. But I wonder if swipes would make for a better experience - because then the animation could be tweaked to look 'optimally' nice whereas now you can't know how fast the user drags down/up and if the watch can update fast enough.
- since I use `quicklaunch` it would make sense to me if this app could be disabled on the clock face via e.g. a setting or some clever logic.

bobrippling:
Thanks for checking it out!

- I currently prefer the drag, as I like to be able to peek at it (I'm adding statuses to the drawer) rather than a swipe to pull it down. Could be a good setting to have / try out though, although I find my watch handles the drag pretty quickly (or perhaps I'm slow!)
- Yes that's a fair point - out of interest, do you use `quicklaunch` by swiping from the top of your watch? Does `quicklaunch` correctly fire (and `ctrlpad` remain hidden) if you swipe about 40 pixels down from the top?

thyttan:
> do you use quicklaunch by swiping from the top of your watch? Does quicklaunch correctly fire (and ctrlpad remain hidden) if you swipe about 40 pixels down from the top?

Yes it works like you say - the clever logic was already there then! 😉 Neat 🙂

bobrippling:
Ah great to hear! I'll see if I can find a way to make that more discoverable, I've added to the docs for now too, thanks again for giving it a try

apps/ctrlpad/main.js
```diff
@@ -0,0 +1,314 @@
...
+                            NRF.wake();
+                    }
+                    return on !== tap;
+                }
+            },
+            {
+                text: "DnD",
+                cb: function (tap) {
+                    var on;
+                    if (on = !!origBuzz) {
```
thyttan, line 130: Typo `=` instead of `==`?

bobrippling: This is intentional in fact:
```typescript
let on;
if(on = !!origBuzz){
  ...
  // We can only assign a function to Bangle.buzz
  // but origBuzz has type `undefined | (() => Promise<void>)`
  //
  // So how does this typecheck?
  // Within the `if` condition, typescript can see that we're testing `origBuzz`,
  // so it knows it's not `undefined`.
  //
  // If we moved the assignment out of the `if`,
  // we'd lose this type-narrowing
  Bangle.buzz = origBuzz;
  ...

}
```

thyttan: OK thanks!

thyttan:
I'll go ahead and merge then? 🙂

bobrippling:
Sounds good, thanks for the review too!

thyttan:
Thanks!

thyttan:
Just noticed two bugs - sorry should have noticed before :)

- When toggling a button the graphics are not updated to show this until the next touch screen interaction. So I will press e.g. the dnd button and the icon will not get filled, but then I drag or touch the screen and it fills with blue.
- The BLE button never gets unfilled after trying to switch it off.

bobrippling:
> Just noticed two (potentially) bugs - sorry should have noticed before :)
> 
> * When toggling a button the graphics are not updated to show this until the next touch screen interaction. So I will press e.g. the dnd button and the icon will not get filled, but then I drag or touch the screen and it fills with blue.

Interesting! I don't see this, I wonder if it's a firmware thing. I'm on `2v21`

> * The BLE button never gets unfilled after trying to switch it off.

I suspect this might be related to the above, unless you see any errors in [the IDE](https://www.espruino.com/ide/) when tapping the BLE button?

And no worries, thanks for the report :)

PR #3384, opened 2024-04-27
exstats: fix steps being negative

bobrippling:
This is definitely not the correct fix - in my case:
```
>state.startSteps
=336651
>Bangle.getStepCount()
=47487
```

What I expect has happened is that `startSteps` is from a previous run and hasn't been reset overnight, whereas the step count has. I'm not familiar with exstats, @gfwilliams / @GrandVizierOlaf what do you think?

### DIFF
--- a/apps/run/ChangeLog
+++ b/apps/run/ChangeLog
@@ -17,3 +17,4 @@
 0.16: Added ability to resume a run that was stopped previously (fix #1907)
 0.17: Ensure screen redraws after "Resume run?" menu (#3044)
 0.18: Minor code improvements
+0.19: Fix step count bug when runs are resumed after a long time
--- a/apps/run/metadata.json
+++ b/apps/run/metadata.json
@@ -1,6 +1,6 @@
 { "id": "run",
   "name": "Run",
-  "version": "0.18",
+  "version": "0.19",
   "description": "Displays distance, time, steps, cadence, pace and more for runners.",
   "icon": "app.png",
   "tags": "run,running,fitness,outdoors,gps",
--- a/apps/runplus/ChangeLog
+++ b/apps/runplus/ChangeLog
@@ -25,3 +25,4 @@ Write to correct settings file, fixing settings not working.
 0.22: Ensure screen redraws after "Resume run?" menu (#3044)
 0.23: Minor code improvements
 0.24: Add indicators for lock,gps and pulse to karvonen screen
+0.25: Fix step count bug when runs are resumed after a long time
--- a/apps/runplus/metadata.json
+++ b/apps/runplus/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "runplus",
   "name": "Run+",
-  "version": "0.24",
+  "version": "0.25",
   "description": "Displays distance, time, steps, cadence, pace and more for runners. Based on the Run app, but extended with additional screen for heart rate interval training.",
   "icon": "app.png",
   "tags": "run,running,fitness,outdoors,gps,karvonen,karvonnen",
--- a/modules/exstats.js
+++ b/modules/exstats.js
@@ -102,6 +102,7 @@ var stats = {};
 const DATA_FILE = "exstats.json";
 // Load the state from a saved file if there was one
 state = Object.assign(state, require("Storage").readJSON(DATA_FILE,1)||{});
+state.startSteps = Bangle.getStepCount()  - (state.lastSteps - state.startSteps);
 // force step history to a uint8array
 state.stepHistory = new Uint8Array(state.stepHistory);
 // when we exit, write the current state


gfwilliams:
> hasn't been reset overnight, whereas the step count has

336,651 is a lot of steps for one day :) I can't remember 100% but it's possible that `getSteps` step count just increases constantly from boot time. Do you think it's possible that the whole Bangle got reset between runs?

I think what we need is right after loading the stats from a file: https://github.com/espruino/BangleApps/blob/8e6ff3bbe3c468ac6db4e98d2787dc6cf2413b25/modules/exstats.js#L100

We do:

```
state.startSteps = Bangle.getStepCount()  - (state.lastSteps - state.startSteps);
```

So we take `state.lastSteps - state.startSteps` which is the amount of steps we've actually recorded, and then just subtract that from the current step count so even if the amount of steps went up between times we still return the same number?

bobrippling:
> 336,651 is a lot of steps for one day :) I can't remember 100% but it's possible that `getSteps` step count just increases constantly from boot time. Do you think it's possible that the whole Bangle got reset between runs?

I must've racked that up sleep walking! But yes, I expect my watch had been reset between runs, or the run was from a long time ago (same thing as the former I suppose).


> So we take `state.lastSteps - state.startSteps` which is the amount of steps we've actually recorded, and then just subtract that from the current step count so even if the amount of steps went up between times we still return the same number?

That makes sense, I've updated the PR - thanks for your help

gfwilliams:
Sorry for the delay, I missed this.

Did you test it worked? I have to say I didn't when I made the suggestion :)

thyttan:
> Did you test it worked? I have to say I didn't when I made the suggestion :)

Neither have I I'm afraid! I'm not really sure on what to test for actually 🙃

bobrippling:
Giving it a whirl, think I'll have to fake the step count in order to confirm it works though - shall report back

gfwilliams:
Hopefully @bobrippling can see if it fixes his original issue?

I was just trying to test here and I realised that you can't just use exstats.js like it says in the header any more:

```
var ExStats = require("exstats");
var exs = ExStats.getStats(["step"], {});
```

It fails with an exception about `options.notify.dist.increment`...

Looks like it's https://github.com/espruino/BangleApps/blob/master/modules/exstats.js#L230-L232 from one of your commits 2 years ago @thyttan - I think it was intended to work without it defined, but it wouldn't have worked...

There's some weirdness going on with how/where stuff is used. Looks like run/runplus set:

```
  notify: {
    dist: {
      value: 0,
      notifications: [],
    },
    step: {
      value: 0,
      notifications: [],
    },
    time: {
      value: 0,
      notifications: [],
    },
  },
```

But `value` isn't used anywhere. Should that be `increment`?

I made some changes to exstats to try and make it work without defining the extra notify stuff in 89ea6d00450307e26b5adacde809f42176d998af - might be worth a check to see I haven't done anything dumb?

gfwilliams:
@bobrippling thanks! I think you can just put a large count in extstats.json. 

Or this is the one from my watch which is on ~98k: 
[exstats.json](https://github.com/espruino/BangleApps/files/15261462/exstats.json)

thyttan:
@gfwilliams I don't remember from top of my head... But that may explain why the time 'stood still' when I was using `runplus` during a roller ski session the other day!

I can try out your code change today 🙂

bobrippling:
> @bobrippling thanks! I think you can just put a large count in extstats.json.
> 
> Or this is the one from my watch which is on ~98k: [exstats.json](https://github.com/espruino/BangleApps/files/15261462/exstats.json)

Thanks - seems to have fixed my original issue, works for your `exstats.json` and works when I modify `exstats.json` to have the large step count mentioned in the PR post. I'm happy for this to be merged :)

image

thyttan:
I reinstalled `runplus` and also installed `run` from the development app loader just now. 

Sadly I can now see steps being negative if I leave the `run` app and then come back to it. And the same is true for `runplus` too. 

In addition on `runplus` the time still doesn't start counting up when starting recording. However if I leave the app and open it again then the time seems to be displayed correctly and is counting up.

@gfwilliams re:
> Looks like it's https://github.com/espruino/BangleApps/blob/master/modules/exstats.js#L230-L232 from one of your commits 2 years ago @thyttan - I think it was intended to work without it defined, but it wouldn't have worked...

I believe you got me mixed up with @GrandVizierOlaf since our profile pixel graphics are a bit similar. @GrandVizierOlaf do you have some familiarity with this still to add some context maybe? :)

In conclusion, maybe you should wait with updating banglejs.com/apps till we work this out?

bobrippling:
Ah dang, maybe we revert this change if we want to get the app loader updated? I'm a bit short on time today/over the weekend but shall see if I can track it down

gfwilliams:
@thyttan Yes, looks like I just saw the green pixel icon and just assumed it was you. Sorry about that!

Ok, I'll revert this change for now. It's an odd one as it definitely feels like it'd work ok, and I thought I had success when I tried it here yesterday,

thyttan:
@gfwilliams 
I now tried this fix (`runplus` v0.25) on a factory reset watch fw2v21.166 and didn't seem to get the problems I had yesterday. 

The procedure was factory reset -> update default apps -> install `runplus` and `recorder`. 

Will test some more.

gfwilliams:
Ok, thanks - so we think this fix probably was ok and I should merge again? It definitely feels like a reasonably simple change.

bobrippling:
Thanks for re-checking and reintroducing!

PR #3385, opened 2024-04-27
chargingstatus: also buzz on disconnect from charging

bobrippling:
This buzzes after a slight delay, to allow other "charging" handlers to run and not interrupt the buzz.


pavelmachek:
Yes, please. It is way too easy to connect charger badly, where it disconnects immediately after connecting.

bobrippling:
Thanks - yes that's my thinking exactly, nice to hear if the connection becomes unplugged

thyttan:
Thanks!

PR #3386, opened 2024-04-27
chargent: fix manual/tap hiding of notification

bobrippling:
Introduced by and documented in #3262.

This also shortens the message length to fit on the screen

### DIFF
--- a/apps/chargent/ChangeLog
+++ b/apps/chargent/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Added notification
 0.05: Fixed boot
 0.06: Allow tap to silence notification/buzzing
+0.07: Fix notification-tap silencing and notification length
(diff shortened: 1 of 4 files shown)


bobrippling:
@thyttan does this fix that [notification problem](https://github.com/espruino/BangleApps/pull/3262)?

thyttan:
> does this fix that [notification problem](https://github.com/espruino/BangleApps/pull/3262)?

I tried this the other day and no, not for my setup as it seemed... 🤷‍♂️

thyttan:
I'm sorry - much of this has probably been user error on my part :)

After reading the readme more carefully I see it's supposed to warn when battery indication in UI said 100 % for a few minutes. I mistakenly believed it should have buzzed at UI indication 80 %.

thyttan:
Thanks!

bobrippling:
Ah no worries! Thanks for helping test it out - wouldn't have found the fix without you :)

PR #3387, opened 2024-04-27
owmweather: upgrade to One Call API 3.0

gsuarezc:
[One Call API 2.5 will be deprecated in June 2024](https://openweathermap.org/api/one-call-api).
Upgrade to One Call API 3.0.
Add pressure to weather.json

### DIFF
--- a/apps/owmweather/lib.js
+++ b/apps/owmweather/lib.js
@@ -1,7 +1,10 @@
 function parseWeather(response) {
   let owmData = JSON.parse(response);
 
-  let isOwmData = (owmData.lat && owmData.lon) && owmData.current.weather && owmData.current;
+  let isOwmData = false;
+  try {
+    isOwmData = (owmData.lat && owmData.lon) && owmData.current.weather && owmData.current;
+  } catch (_e) {}
 
   if (isOwmData) {
     let json = require("Storage").readJSON('weather.json') || {};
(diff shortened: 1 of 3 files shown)


apps/owmweather/lib.js
```diff
@@ -1,19 +1,20 @@
 function parseWeather(response) {
   let owmData = JSON.parse(response);
 
-  let isOwmData = owmData.coord && owmData.weather && owmData.main;
+  let isOwmData = (owmData.lat && owmData.lon) && owmData.current.weather && owmData.current;
```
bobrippling, line 4: Now we're going deeper, do we want to guard against `.current` missing here? Perhaps:
```suggestion
  let isOwmData = false;
  try {
    isOwmData = (owmData.lat && owmData.lon) && owmData.current.weather && owmData.current;
  } catch (_e) {}
```

apps/owmweather/lib.js
```diff
@@ -1,19 +1,20 @@
...
-    weather.wdir = owmData.wind.deg;
-    weather.wind = owmData.wind.speed;
-    weather.loc = owmData.name;
-    weather.txt = owmData.weather[0].main;
+    weather.hum = owmData.current.humidity;
+    weather.temp = owmData.current.temp;
+    weather.code = owmData.current.weather[0].id;
+    weather.wdir = owmData.current.wind_deg;
+    weather.wind = owmData.current.wind_speed;
+    weather.loc = owmData.name ? owmData.name : "";
```
bobrippling, line 15: We can write this a little shorter if you prefer:
```suggestion
    weather.loc = owmData.name || "";
```

Similarly below with `.pressure`

bobrippling (commented):
Thanks for the fixes! Some small suggestions

apps/owmweather/lib.js
gsuarezc: Implemented.

apps/owmweather/lib.js
gsuarezc: Implemented.

bobrippling:
Thanks!

PR #3390, opened 2024-04-29
Enforce and fix typescript lints

atjn:
@bobrippling disabled linting for typescript projects in #3383. I don't understand why typescript projects should be exempt from the linting rules, and I could not find any explanation in the PR either.

I propose to reenable linting rules for typescript files and fix the lint issues, as done in this PR. I have also checked that none of the project files are temporarily exempt, so that means these few changes are enough to fix all lint issues in all typescript projects.


thyttan:
I merged the change. I believe the reasoning is that since the js files were generated from the ts source the warnings should not (never?) correspond to a real problem. 

What prompted the change: https://github.com/espruino/BangleApps/pull/3383#discussion_r1582617557

But I'll let you @atjn and @bobrippling talk this out more, I'm hardly familiar with these things.

atjn:
> I merged the change. I believe the reasoning is that since the js files were generated from the ts source the warnings should not (never?) correspond to a real problem.

That is a fair assumption, but the linter is testing for many things that typescript does not. They complement each other well :)

It would be possible to add plugins to the linter so that it checks the typescript files directly, but it really doesn't matter whether the js files or ts files are linted, as long as one of them is.

> What prompted the change...

Thanks for the pointer. Given that I have fixed that issue, I am guessing there is no problem now?

.eslintignore
```diff
@@ -4,15 +4,3 @@ apps/gipy/pkg/gps.js
...
-apps/btadv/
-apps/clkinfostopw/
-apps/ctrlpad/
-apps/drained/
-apps/folderlaunch/
-apps/popconlaunch/
-apps/rep/
-apps/widChargingStatus/
-apps/widbtstates/
-apps/widhid/
```
bobrippling, line 18: Can we run the linter on ts files and tweak this to ignore the generated js files?
```suggestion
apps/btadv/app.js
apps/clkinfostopw/clkinfo.js
apps/ctrlpad/main.js
apps/drained/app.js
apps/drained/boot.js
apps/drained/settings.js
apps/folderlaunch/app.js
apps/folderlaunch/configLoad.js
apps/folderlaunch/settings.js
apps/folderlaunch/types.d.js
apps/popconlaunch/boot.js
apps/popconlaunch/settings.js
apps/rep/app.js
apps/rep/settings.js
apps/widChargingStatus/widget.js
apps/widbtstates/widget.js
apps/widhid/wid.js
```

apps/ctrlpad/main.ts
```diff
@@ -193,8 +193,9 @@
 			{
 				text: "DnD",
 				cb: tap => {
-					let on;
-					if(on = !!origBuzz){
+					let on = false;
+					if(origBuzz){
```
bobrippling, line 197: Here my intention was to indicate via code that `on` is derived from `origBuzz`, we can keep this and satisfy the linter:
```suggestion
					if(((on = !!origBuzz)){
```

apps/drained/app.ts
```diff
@@ -1,9 +1,9 @@
 const app = "drained";
 
 // from boot.js
-declare var drainedInterval: IntervalId | undefined;
-if(typeof drainedInterval !== "undefined")
-  drainedInterval = clearInterval(drainedInterval) as undefined;
+const globals: { drainedInterval: IntervalId | undefined } = global as any;
+if(typeof globals.drainedInterval !== "undefined")
+  globals.drainedInterval = clearInterval(globals.drainedInterval) as undefined;
```
bobrippling, line 6: The `declare var` indicates that this is a global access, so I would prefer that we don't modify this

apps/drained/boot.ts
```diff
@@ -2,7 +2,7 @@
 const { battery: threshold = 5, interval = 10, keepStartup = true }: DrainedSettings
   = require("Storage").readJSON(`drained.setting.json`, true) || {};
 
-drainedInterval = setInterval(() => {
+(global as any).drainedInterval = setInterval(() => {
```
bobrippling, line 5: This removes cross-file type safety and makes it easier to introduce a bug if `drainedInterval` is changed in one file and not another, let's back it out.

Happy to have the `declare var` here too if you'd like to also indicate that we have a global variable

bobrippling (commented):
Thanks for this - I was short on time and didn't want to introduce lint warnings so am glad we've been able to take a look at it.

What I'd like is to eventually have is the typescript files linted and the generated files ignored, to avoid anyone making manual changes to the generated files which will the be lost - this will avoid PRs like #3304 where there's then a round-trip needed to change the source

.eslintignore
atjn: This seems like a big maintenance burden but if you don't mind it, that is fine with me :)
An alternative could be to make a plugin that checks to see if there is a pair of ts and js files with the same name, and if true, it instructs the parser to skip the js file.

apps/ctrlpad/main.ts
atjn: I don't like this solution since you're not honoring the idea with the lint rules, which is that it is hard to read and understand that this is both an assignment and a boolean check. But I don't want to argue any more about it, so this is fine.

apps/drained/boot.ts
atjn: That is a very good argument that I did not think about, agreed.

atjn:
Thanks for the reply @bobrippling. I did not think about the case where you declare a global  variable in a type-safe manner. To support that case, the linting needs to happen in the `ts` file, and the output `js` file must be ignored because it lacks the type information that makes it okay to use that variable.

Therefore I have almost completely remade the PR so that it now includes a typescript linter, just as you requested :)

bobrippling:
Nice! Yeah that all looks great, thank you.

One thing I'm interested in your opinion on - I think unlike the original lint changes where we wanted to bump app versions when things like variable `let`s were introduced, here we can confirm that some apps have no effective changes, for instance:
- the corresponding `.js` file is unchanged
- the changes (in the output js) are trivially bug-free, e.g.:
  - addition of scope-brackets (and no block-scoped variables affected)
  - `o.a` -> `o["a"]`

So I would say we can avoid the version bump in these cases. There's just the `Array` -> `[...]` change where we perhaps keep the version bump out of caution, or perhaps that me being too cautious, what do you think?

.eslintignore
bobrippling: Yes I like the sound of that plugin - leave it with me

atjn:
Very nice!

For apps where there is no change to the output, I have not bumped the version.

As for the other cases you mention, I bumped them out of caution just because there can be minor differences in their behavior and that is especially true in Espruino which does not always conform 100% to JS standards. If you request it, I don't mind unbumping them (realistically the changes are probably fine), but I would personally just bump them to be safe. Also be aware that I have not tested them yet, I didn't want to spend time on that before I knew if you approved the changes. I'll probably have time on the next few days :)

bobrippling:
Ah of course - great stuff. I think that's perfectly reasonable, let's stick with your approach in that case. And no rush :)

atjn:
I have tested the apps as best I could and could not find any breakage. I believe this is ready to be merged :)

bobrippling:
Cool - thanks again!

PR #3393, opened 2024-04-29
messagesoverlay - Handle prependListener and setWatch callback variable name change

halemmerich:
* Listeners added with `prependListener` were not tracked,
* 2v21.104 renamed the internal variable for callbacks stored with setWatch from `callback` to `cb`. This should handle both variants


halemmerich:
This has a bug where watches are not backed up correctly because no index can be found.

halemmerich:
Works fine for me.

thyttan:
Thanks!

PR #3391, opened 2024-04-29
backswipe:fix could not blcklist messagegui.new.js

thyttan:
fix for https://github.com/espruino/BangleApps/issues/3388

PR #3395, opened 2024-05-01
autoreset: add settings

thyttan:
- black-/whitelist
- timeout length


bobrippling:
Nice, thanks! Do you think we'll start using the include/exclude logic in other apps, now we have it in `autoreset` and `backswipe` (#3391) ? Wondering whether if we have a third case, it would benefit being in a library

thyttan:
Yes maybe we're getting to the point that it makes sense

PR #3397, opened 2024-05-02
widdevst: add ability to only redraw when a peripheral changes

bobrippling:
This change tweaks the drawing of `widdevst` to be fired by a change of a peripheral (such as a HRM event), so that:
- if peripherals are on, it'll redraw as regularly as it did before this change
- if periperhals are off, it'll redraw at the same (lower) frequency, or possibly not at all if the next option is enabled (saving draw time & battery)

The setting change allows us to configure the widget to only draw when a peripheral changes, at the cost of not having directly up-to-date bars for storage & memory.

@notEvil what do you think? I'm going to try this out for a little while and see how it compares time-wise to the current implementation

### DIFF
--- a/apps/widdevst/ChangeLog
+++ b/apps/widdevst/ChangeLog
@@ -1,3 +1,4 @@
 0.01: First version
 0.02: Support for Bangle.js 2
 0.03: Update storage usage and perform GC every minute
+0.04: Add ability to only redraw when a peripheral state changes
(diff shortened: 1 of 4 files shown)


notEvil:
Looks good to me, except `"redrawBar"` which should probably be `"redrawBars"` :)

bobrippling:
Doh, thanks!

bobrippling:
Tested ok, thanks for the look over :)

notEvil:
@bobrippling Just found that the Readme suggests redrawBars to be opt-out (`true` by default) while `false` is actually the default, if I'm not mistaken.

bobrippling:
Good spot, thank you - I've pushed up #3400 to sort it, if you could give it a glance over

PR #3399, opened 2024-05-03
Implements more asserts and reading testcases from json files

halemmerich:
This is mostly a base for discussion in [https://forum.espruino.com/conversations/395324/](https://forum.espruino.com/conversations/395324/) and extends the `runapptests.js` by @gfwilliams .


halemmerich:
I think this is now ok to merge. It is only missing a merged https://github.com/espruino/EspruinoAppLoaderCore/pull/61 and the update of `core` to pull the change into this repo.

bobrippling:
Thanks - I've merged #3410 which brings in those app loader core changes. Will leave to @gfwilliams to approve with this being a reasonably central change.

Unrelated / for another time, but I wonder whether we attempt to run the app tests as part of [CI](https://github.com/espruino/BangleApps/blob/master/.github/workflows/nodejs.yml), wdyt?

bin/runapptests.js
```diff
@@ -71,97 +89,408 @@ var emu = require(BASE_DIR+"/core/lib/emulator.js");
...
   });
-  // Running tests
-  runTest(TEST);
+  p.finally(()=>{
+    console.log("\n\n");
+    console.log("Overall results:");
+    console.table(testState);
+
+    process.exit(testState.reduce((a,c)=>{
+      return a + ((c.result == "SUCCESS") ? 0 : 1);
```
bobrippling, line 488: This could potentially reach 256, in which case it's equivalent to `exit(0)`, perhaps we should:
```suggestion
      return a || ((c.result == "SUCCESS") ? 0 : 1);
```

gfwilliams:
This looks great - thanks for your work on it!

halemmerich:
> Unrelated / for another time, but I wonder whether we attempt to run the app tests as part of [CI](https://github.com/espruino/BangleApps/blob/master/.github/workflows/nodejs.yml), wdyt?
I think that would be great. Don't know if we would hit some github limits when we get to a substantial amount of tests, but generally that should be a good idea.

gfwilliams:
Yes, I think at some point it'd be great to enable these. At least when it's just a few apps it's not going to be a huge amount of extra CPU use.

PR #3400, opened 2024-05-05
widdevst: fix default, keep original behaviour

bobrippling:
See #3397 / discussion at the end


bobrippling:
Confirmed [here](https://github.com/espruino/BangleApps/pull/3397#issuecomment-2094849741), merging

PR #3404, opened 2024-05-07
fuzzyw: add toggle for animation + minor bug fixes

aGoodUsername:
I didn't like the animation on the text, so I made it possible to turn it off.

In addition I made settings actually work, so now widgets work as well. I also made hours work properly with pretokenisation (probably – sometimes it worked, sometimes it didn't, maybe the fix is the same and I've just been lucky), and finally added a lot of small things like proper indentation, definition of all variables and adding semicolons.

### DIFF
--- a/apps/fuzzyw/fuzzyw.app.js
+++ b/apps/fuzzyw/fuzzyw.app.js
@@ -58,20 +58,14 @@
   let getTimeString = function(date) {
     let segment = Math.round((date.getMinutes()*60 + date.getSeconds() + 1)/300);
     let hour = date.getHours() + Math.floor(segment/12);
-    let f_string = fuzzy_string.minutes[segment % 12];
-    for (let i = 0; i < f_string.length; i++) {
-      if (f_string.charAt(i) == '$') {
-        if (f_string.charAt(i+1) == '1') return f_string.slice(0, i) + fuzzy_string.hours[hour % 12] + f_string.slice(i+2);
-        return f_string.slice(0, i) + fuzzy_string.hours[(hour+1) % 12] + f_string.slice(i+2);
-      }
-    }
-    // the more elegant solution that unfortunately gets optimized to smithereens
-    /*
+    // add "" to load into RAM due to 2v21 firmware .replace on flashstring issue
+    f_string = ""+fuzzy_string.minutes[segment % 12]; 
     if (f_string.includes('$1')) {
-      //return f_string.replace('$1', fuzzy_string.hours[hour % 12]);
+      f_string = f_string.replace('$1', fuzzy_string.hours[(hour) % 12]);
     } else {
-      //return f_string.replace('$2', fuzzy_string.hours[(hour + 1) % 12]);
-    }*/
+      f_string = f_string.replace('$2', fuzzy_string.hours[(hour + 1) % 12]);
+    }
+    return f_string;
   };
 
   let draw = function() {
--- a/apps/fuzzyw/fuzzyw.app.js
+++ b/apps/fuzzyw/fuzzyw.app.js
@@ -59,7 +59,7 @@
     let segment = Math.round((date.getMinutes()*60 + date.getSeconds() + 1)/300);
     let hour = date.getHours() + Math.floor(segment/12);
     // add "" to load into RAM due to 2v21 firmware .replace on flashstring issue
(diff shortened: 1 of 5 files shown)


apps/fuzzyw/fuzzyw.app.js
```diff
@@ -31,113 +31,135 @@
...
+    let segment = Math.round((date.getMinutes()*60 + date.getSeconds() + 1)/300);
+    let hour = date.getHours() + Math.floor(segment/12);
+    let f_string = fuzzy_string.minutes[segment % 12];
+    for (let i = 0; i < f_string.length; i++) {
+      if (f_string.charAt(i) == '$') {
+        if (f_string.charAt(i+1) == '1') return f_string.slice(0, i) + fuzzy_string.hours[hour % 12] + f_string.slice(i+2);
+        return f_string.slice(0, i) + fuzzy_string.hours[(hour+1) % 12] + f_string.slice(i+2);
+      }
+    }
+    // the more elegant solution that unfortunately gets optimized to smithereens
```
thyttan, line 68: Is this a bug with the pretokenizer then?

aGoodUsername: If I print before the .replace in the original code, I get e.g. "half past *$1" and after I get "half past *", but only with pretokeniser, and not always. Sometimes I get "half past *six". Maybe 50% of the time?

bobrippling: Sounds like a possible memory issue in the firmware, perhaps todo with the code that handles template strings? @gfwilliams what do you think?

gfwilliams: Thanks - yes, this seems to be some firmware issue. I just tried:

```
let f_string = "half past *$1";
print(f_string.replace('$1', "nine"));
```

And when uploaded pretokenised I see the issue, but:

```
let f_string = "half *$1";
print(f_string.replace('$1', "nine"));
```

works (the string is small enough it's loaded to RAM) - I think it's some issue with how `.replace` works on flash-based strings.

Please could you change the code to just do:

```
// add "" to load into RAM due to 2v21 firmware .replace on flashstring issue
f_string = ""+fuzzy_string.minutes[segment % 12]; 
if (f_string.includes('$1')) {
  f_string = f_string.replace('$1', fuzzy_string.hours[(hour) % 12]);
} else {
  f_string = f_string.replace('$2', fuzzy_string.hours[(hour + 1) % 12]);
}
```

That should fix it without you having to add the code to iterate, because adding "" to the string will load it into RAM and then it'll work, and then I'll see what I can do to fix this properly in the firmware

aGoodUsername: This does work for me at least.

gfwilliams: Ok, just fixed the firmware. So cutting edge or 2v22 (when released) should now have this fixed, but I guess we want to make sure this still works with older firmwares so the `""+` is a good fix

bobrippling: Thanks! @thyttan I appreciate you started the review, happy to let you finish it :)

apps/fuzzyw/fuzzyw.app.js
```diff
@@ -31,113 +31,129 @@
...
-  if (f_string.includes('$1')) {
-    f_string = f_string.replace('$1', fuzzy_string.hours[(hour) % 12]);
-  } else {
-    f_string = f_string.replace('$2', fuzzy_string.hours[(hour + 1) % 12]);
-  }
+  let getTimeString = function(date) {
+    let segment = Math.round((date.getMinutes()*60 + date.getSeconds() + 1)/300);
+    let hour = date.getHours() + Math.floor(segment/12);
+    // add "" to load into RAM due to 2v21 firmware .replace on flashstring issue
+    f_string = ""+fuzzy_string.minutes[segment % 12];
```
thyttan, line 62:
```suggestion
    let f_string = ""+fuzzy_string.minutes[segment % 12]; 
```

to avoid the lint warning that came with the latest commit.

apps/fuzzyw/fuzzyw.app.js
thyttan: Happy it was worked out - thanks!

thyttan:
Works well on my watch - thanks!

PR #3413, opened 2024-05-11
Port additional tests and fix GB events in runapptests.js

halemmerich:
* Fixes GadgetBridge events being empty when sent with `"t":"gb"`
* Ports remaining tests from `android/test.js`
* Fixes logging for single value assertiions
* Adds a new test for messagesoverlay
* Replaces sleep command with advanceTime which manipulates intervals,timers and Date.now() and returns instantly
* Add resetCall command to reset the recorded calls


gfwilliams:
Ok, thanks - the 12 sec delay isn't great if we're trying to run this in a CI test...

I checked and the emulator just uses `Date.now()` to get the current time in milliseconds... I wonder if the test harness could just override that and allow an additional value to be added to what it returns?

halemmerich:
Hm, overriding `Date.now()` on its own does not seem to influence timers.

```
setTimeout(()=>{print("waited");}, 20000);
setInterval(()=>{print("waited");}, 20000);

const delorean = (jump)=>{
  Date.now = (o)=>{return ()=>{
    return o() + jump;
  };}(Date.now);
  for(let c of global["\xff"].timers){
    if(c) c.time -= jump * 1000;
  }
};

print("Before", Date.now());
print("Before", global["\xff"].timers);

delorean(22000);

print("After", Date.now());
print("After", global["\xff"].timers);
```
This does seem to work and should be enough for my usecase but it feels massively hacky. No idea what other stuff gets broken by this. Might be enough for tests anyway.

gfwilliams:
Ahh, I mean override `Date.now` in Node.js, not in Espruino!

But yes, messing with the timers like that is probably good enough for a test at the moment? Maybe it can be another command type and if it causes issues later we can extend it

halemmerich:
I think just renaming sleep would be ok. Having sleep that actually pauses execution will eventually be used so maybe it would be better to not even allow that as long as there are ways around having to do this? I implemented it specifically to deal with waiting for timers.
Maybe we just name it "speedup" or something like that?

halemmerich:
> could that be why you were getting negative values?

I had removed the override inside before but it was just a quick and dirty test. There might've been something else wrong.

I have just removed the override for `Date.now` as my current tests don't need it. It is now named `advanceTimers` to keep `sleep` free for a more complete implementation when it is actually needed.

gfwilliams:
Thanks! Looks great!

PR #3418, opened 2024-05-14
multitimer: preserve original `uiRemove`

bobrippling:
e.g. a `showMenu()`'s `setUI` call.

Without this, a `showMenu`'s call doesn't get its `remove` callback called.


bobrippling:
Tested for a few days, works ok

PR #3419, opened 2024-05-15
runplus: add ability to focus an individual stat

bobrippling:
This allows users to focus on a particular stat - I found when I was running that I couldn't glance at my pace very easily, whereas with this I can leave the pace zoomed/focused as I run.

 @halemmerich @thyttan I think you're the main contributors to `runplus`, what do you think?

### DIFF
--- a/apps/runplus/ChangeLog
+++ b/apps/runplus/ChangeLog
@@ -26,3 +26,4 @@ Write to correct settings file, fixing settings not working.
 0.23: Minor code improvements
 0.24: Add indicators for lock,gps and pulse to karvonen screen
 0.25: Fix step count bug when runs are resumed after a long time
+0.26: Add ability to zoom in on a single stat by tapping it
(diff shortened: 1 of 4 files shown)


thyttan:
I think this can be a good idea (although personally have not felt the need).

I noticed:
- If I zoom in on the "TIME" stat directly after starting the app, and without pressing HW button to "START", the zoomed in time will already have started counting. On the overview this is not the case, there it shows "00:00".
- Some titles, e.g. "CADENCE", doesn't fit on screen when zoomed in.

bobrippling:
Thanks for checking it out - I've improved the drawing to match how its done in the layout code - we only update the zoomed version now when the stat changes (768437219e7d664f5).

This stops the time counting up, but it does mean the initial time shown may differ if the user zooms when paused, since we'll get the value of the time rather than trying to remember what it was when the user paused. To deal with this we now only show a stat if we're active (5eb8f56), which feels like the least hacky approach.

Also I've dropped the title size for long titles such as "cadence" (9b252e9)

bobrippling:
Tested this out, works well - I'll merge it later on but give me a shout if you spot anything :)

PR #3423, opened 2024-05-18
openhaystack: work with other advertisements

bobrippling:
This means if boot code runs before openhaystack's (which uses [`ble_advert`](https://github.com/espruino/BangleApps/blob/master/modules/ble_advert.ts) or stores its advertising data on `Bangle.bleAdvert`), the openhaystack code will add its own advert to the array, rather than overwrite.

@olivierbarriere is this ok by you? I'm currently testing it / waiting for reports to come through

### DIFF
--- a/apps/openhaystack/ChangeLog
+++ b/apps/openhaystack/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New App!
 0.02: Keep advertising when connected
+0.03: Work with other advertisements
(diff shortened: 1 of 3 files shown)


olivierbarriere:
I don't have other apps with ble advertissements to test with, but it looks good to me, thanks!

bobrippling:
Thanks - I've decided to merge this in with the `ble_advert` module change, in #3438.

PR #3425, opened 2024-05-18
Contacts 0.04: Added call function to contacts app, Refactoring

devsnd:
- Refactored contacts app
 - Added function to trigger calling a contact on the phone (Requires GadgetBridge with "Allow Intent" permission)

### DIFF
--- a/apps/contacts/contacts.app.js
+++ b/apps/contacts/contacts.app.js
@@ -3,11 +3,8 @@
 var Layout = require("Layout");
 
 var wp = require('Storage').readJSON("contacts.json", true) || [];
-// var wp = [];
 
-var key; /* Shared between functions, typically wp name */
-
-function writeContact() {
+function writeContacts() {
   require('Storage').writeJSON("contacts.json", wp);
 }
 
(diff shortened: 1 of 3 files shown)


apps/contacts/contacts.app.js
```diff
@@ -2,108 +2,102 @@
...
+function callNumber (number) {
+  Bluetooth.println(JSON.stringify({
+    t:"intent",
+    target:"activity",
+    action:"android.intent.action.CALL",
+    flags:["FLAG_ACTIVITY_NEW_TASK"],
+    categories:["android.intent.category.DEFAULT"],
+    data: 'tel:' + number,
+  }))
+
```
nxdefiant, line 20: did you check what happens when not connected (Gadgetbridge not installed or similar)?
+1 for refactoring

devsnd: I only tested in the emulator and it does nothing then. You can always send some data over the wire if the watch is connected via Bluetooth, but I don't know how I could detect if the connected device is an Android and if so, if GadgetBridge is used :man_shrugging:

pavelmachek:
Thanks for doing this. Patches look ok on a quick check. And yes, if someone created a nicer icon, that would be great, too :-).

apps/contacts/contacts.app.js
bobrippling: Yeah, if there's no connected device it'll print and the print gets ignored - I believe printing to some devices/consoles hangs but `Bluetooth` is ok

bobrippling:
Nice! LGTM - tested and looks good too

PR #3427, opened 2024-05-19
Allow using either the free OWM API 2.5 or the paid OWM API 3.0

devsnd:
The previous updates of owmweather broke the app for me, as the open weather map API 3.0 requires a new API key that can only be acquired using a paid subscription plan.

In this patch I make this change optional and allow to either use the free API v2.5 or to use the paid v3.0 API.

### DIFF
--- a/apps/owmweather/lib.js
+++ b/apps/owmweather/lib.js
@@ -3,21 +3,40 @@ function parseWeather(response) {
 
   let isOwmData = false;
   try {
-    isOwmData = (owmData.lat && owmData.lon) && owmData.current.weather && owmData.current;
+    isOwmData = (
+        // Paid API 3.0: onecall
+        (owmData.lat && owmData.lon && owmData.current.weather && owmData.current)
+        ||
+        // Free API 2.5
+        (owmData.coord && owmData.weather && owmData.main)
+  );
   } catch (_e) {}
 
   if (isOwmData) {
     let json = require("Storage").readJSON('weather.json') || {};
     let weather = {};
-    weather.time = Date.now();
-    weather.hum = owmData.current.humidity;
-    weather.temp = owmData.current.temp;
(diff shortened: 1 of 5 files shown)


apps/owmweather/lib.js
```diff
@@ -43,7 +62,13 @@ exports.pull = function(completionCallback) {
     "location": "London"
   };
   let settings = require("Storage").readJSON("owmweather.json", 1);
-  let uri = "https://api.openweathermap.org/data/3.0/onecall?lat=" + location.lat.toFixed(2) + "&lon=" + location.lon.toFixed(2) + "&exclude=minutely,hourly,daily,alerts&appid=" + settings.apikey;
+  let uri;
+  if (!!settings.useOneCall) {
```
halemmerich, line 66: The linter is not happy about the !! here. I think you can just remove this since the `if` can handle all truthy/falsy values.

halemmerich (commented):
Thanks for catching this! I had wondered why my weather was not updated but did not yet investigate myself. It works fine for me with the free API.
Maybe the 3.0 API should be the default since the 2.5 is announced to be switched off next month. It needs a subscription plan but there are still 1000 free calls a day and it seems they can be limited to prevent actual costs happening.

devsnd:
Since the API 2.5 is about to be discontinued (or already is), I'll close this PR. I have myself signed up for a paid API key, but I will probably never hit any of the free API limits, so this is a non-issue. I just found it sad that it requires all banglejs users you to hand out their payment information, but it is how it is.

devsnd:
And thanks a lot for your review @halemmerich !

zwiglm:
I also got a paid subscription. From eg Postman works fine. Within the app, I still get the error message "com.android.volley..." when trying to force-update the weather.
Any suggestions?

zwiglm:
Hmm. Sorry. Now working

PR #3416, opened 2024-05-14
promenu: handle boolean menu entries, back functionality and scroll and use typescript

bobrippling:
This PR:
- Fixes promenu's display/overwriting of boolean menu items
- Handle back functionality, both via `.back` and `"< Back"` items - an entry is displayed to go back and the `setUI` back widget is shown
- Scroll support is also added

@Ronin0000 is this ok by you?

PR #3430, opened 2024-05-21
promenu: fix removal of final menu item

bobrippling:
A small bug that crept in as part of #3416


bobrippling:
Tested for a day or two, no issues

PR #3431, opened 2024-05-21
runplus: add option to always resume

bobrippling:
For example, when doing reps with 1 minute rests, a user can configure runplus so it automatically resumes if started within the minute

### DIFF
--- a/apps/runplus/app.js
+++ b/apps/runplus/app.js
@@ -26,6 +26,10 @@ let settings = Object.assign({
   B5: "step",
   B6: "caden",
   paceLength: 1000,
+  resume: {
+    promptAfter: 10000,
+    default: false,
+  },
   notify: {
     dist: {
       value: 0,
@@ -64,10 +68,10 @@ function onStartStop() {
   }
 
   var running = !exs.state.active;
-  var shouldResume = false;
+  var shouldResume = settings.resume.default;
   var promise = Promise.resolve();
 
-  if (running && exs.state.duration > 10000) { // if more than 10 seconds of duration, ask if we should resume?
+  if (running && exs.state.duration > settings.resume.promptAfter) { // if more than N seconds of duration, ask if we should resume?
     promise = promise.
       then(() => {
         screen = "menu";
--- a/apps/runplus/app.js
+++ b/apps/runplus/app.js
@@ -1,5 +1,4 @@
 let runInterval;
-let statusTimeout;
 let screen = "main"; // main | karvonen | menu | zoom
 // Run interface wrapped in a function
 const ExStats = require("exstats");
@@ -27,10 +26,7 @@ let settings = Object.assign({
   B5: "step",
   B6: "caden",
   paceLength: 1000,
-  resume: {
-    promptAfter: 10000,
-    default: false,
-  },
+  alwaysResume: false,
   notify: {
     dist: {
       value: 0,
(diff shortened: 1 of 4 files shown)


bobrippling:
@halemmerich @thyttan does this seem ok to you both? I think you're the main users of `runplus`

halemmerich:
I'm not quite sure what you mean by this. Is it auto resume after manually stopping for a slow/rest period? The target would be to have only the high intensity periods tracked/shown?

bobrippling:
Yeah exactly, like if you're doing reps with a lot of standing about in between - I found I'd have to press start then wait and tap the resume button, sometimes forgetting during a session

thyttan:
By automatically, it's not that it sets a timer and starts by itself after i runs out? Just that the menu is not displayed within the window of the timeout and the records will be appended to the last log?

bobrippling:
Yeah that's poor choice of phrasing on my part - it's exactly the existing behaviour, just with the ability to customise the time period (so like you say, currently the user is always prompted to resume the run or start anew after 10s)

halemmerich:
How about letting a short button press pause and resume (including showing paused as state in the bottom right) and longer button press stop the recording? That way there is no additional setting and it is always possible to actually end the run. It would allow more flexible intervals.
Maybe even write a line into the recorder log when the pause starts and ends instead of actually pausing the recorder to make later analysis easier. That would allow to check for pulse recovery etc. during the rest intervals.
Thinking about it, maybe short press could just signify a new interval/round start/stop and otherwise everything would just run like normal? The UI could do whatever it wants with the intervals, i.e. reset exstats, show averages/values only for the last interval or whatever.

halemmerich:
How does actually stopping work? Say I configure a 5 minute auto resume time, do I have to kill `runplus` and go to the recorder to stop the run?

bobrippling:
For a 5 minute resume time, yes, you'd have to restart `runplus` - but that'd be it, it would then kick off a new recording (in fact you'd get a prompt - see below snippets), you wouldn't need to tangle with the recorder app manually:

https://github.com/espruino/BangleApps/blob/e856e381fad97cea0c88acd5cbb78c448c1abe0b/apps/runplus/app.js#L84-L91

https://github.com/espruino/BangleApps/blob/e856e381fad97cea0c88acd5cbb78c448c1abe0b/apps/recorder/widget.js#L267-L273

halemmerich:
So if a later rework to include laps or other handling of the start/stop/pause/resume mechanics is still on the table I'm fine with this change. Maybe we should signify if the timer has run out in the UI by replacing the STOP by PAUSE or something like that as long as it would autoresume on button press.

bobrippling:
Sounds good, sorted the timer UI (dbc9833), let me know what you think

apps/runplus/app.js
```diff
@@ -49,9 +54,15 @@ let statIDs = [settings.B1,settings.B2,settings.B3,settings.B4,settings.B5,setti
...
 // ---------------------------
 
+function inPauseWindow() {
+  return exs.state.duration <= settings.resume.promptAfter;
+}
+
 function setStatus(running) {
-  layout.button.label = running ? "STOP" : "START";
-  layout.status.label = running ? "RUN" : "STOP";
+  const paused = isInPauseWindow();
```
thyttan, line 62: ... this doesn't match :)

apps/runplus/app.js
```diff
@@ -49,9 +54,15 @@ let statIDs = [settings.B1,settings.B2,settings.B3,settings.B4,settings.B5,setti
 let exs = ExStats.getStats(statIDs, settings);
 // ---------------------------
 
+function inPauseWindow() {
```
thyttan, line 57: This and ...

thyttan, line 57: This and ...

apps/runplus/app.js
thyttan, line 62: ... this don't match :)

bobrippling: That's what I get for last minute changes! Thank you

bobrippling:
I've played around with this more and I think the timeout isn't worth the complexity, my use case would be to have the timeout be ridiculously high anyway - I've changed it to a simple toggle.

Given this a test and it looks to work well, let me know what you think :)

thyttan:
Thanks!

bobrippling:
Thanks for the review/merge!

PR #3436, opened 2024-05-28
sixths v0.04: make height auto-calibration useful and slow ticks to save power

pavelmachek:


pavelmachek:
No, sorry, the lint warnings can not be fixed at this point. There's too many of them, and that would invalidate testing. They would be actually useful if one got them during the development, and if system presented all of them at the time, but I just spent 30 minutes trying to make the robot happy.

Can I ask to merge 0.04 as-is, as that was what I was testing? I'll get them fixed for 0.05, but coming with warnings post-testing and one at a time is annoying and stupid.

bobrippling:
Yeah of course, no problem - I've tweaked a couple but no need to if you've other priorities :)

bobrippling:
And thanks for the PR!

pavelmachek:
Thanks for pull. I took a look at your changes, and I'll likely translate most variables from "var" to "let", as I believe that has better properties.

PR #3440, opened 2024-05-31
fileman: improve showing files in browser, was truncated before

devsnd:


thyttan:
Thanks for the improvement!

PR #3443, opened 2024-06-06
[datetime_picker] create app and use it in [alarms] app

lauzonhomeschool:
Creates a datetime_picker app that allows to select both the date and time with swipes in the same screen.
When the alarms apps detect a datetimeinput (datetime_picker) app, it replaces all the individual fields that constitute the datetime fields (for Events) with one single dateime field that is editable using the app.  If not found, everything displays as before.  

Also implemented the setting menuDateFormat for value "mmdd", where we'll show the datetime as "mm-dd" in the menu when the Event is in the same year as the current year (to let more space to display the Event's Message), yyyy-mm-dd otherwise.  The default value of this setting will just display the datetime as before (i.e. no change to end user experience unless they change this setting).

Also fixed the position in the alarms menu when getting out of an alarm Group, to scroll back to the same position as before entering the group.

I tested the new app for more than a week and everything seems fine.


apps/alarm/app.js
```diff
@@ -119,6 +123,7 @@ function showNewMenu(group) {
 }
 
 function showEditAlarmMenu(selectedAlarm, alarmIndex, withDate, scroll, group) {
+  console.log(scroll);
```
gfwilliams, line 126: Please could you remove this print statement?

gfwilliams:
Thanks! I like the idea of having an optional date picker app.

Please could you add `"provides_modules" : ["datetimeinput"]` to the metadata of the datetimepicker, and a short description of how to use it in code to the README? Maybe in future others will make different picker apps, and if so `provides_modules` prompts the user if they try and install 2 at the same time (and allows apps to request a picker app to be installed).

Also, I haven't tried it yet, but looking at the screenshot and the README I can't figure out where I'd swipe to change which thing, so something in the readme about where you have to swipe and in what axis to change what item might be handy.

Also, just wondering - do we need the app? I know it's a good demo, but I wonder whether most users would find having a not super useful app always in the launcher would be a bit annoying? You can always look at the keyboard apps for an example of what they do in the metadata

lauzonhomeschool:
Good ideas, I'll work on it and submit the corrections.

lauzonhomeschool:
I deleted app.js, the build now gives the error "Error: App datetime_picker has no entrypoint".  If I try to add `type: "datetimeinput"`, the build says it's not an allowed type.

gfwilliams:
Add - try `"type":"module"` - that should do it

lauzonhomeschool:
Indeed, that worked, tx.

gfwilliams:
So are you happy with this now? From my point of view I think we're ok to merge

bobrippling:
Going off the thumbs up, looks good - merging

PR #3445, opened 2024-06-07
Burn - Simple Calorie Counter

slynchDev:
Calorie count persists to kcal.txt and clears at midnight. Increment (step) value can be changed as needed


apps/burn/burn.js
```diff
@@ -0,0 +1,254 @@
...
+let drag, y, x;
+
+// If the hardware version is Bangle.js 2, set the coordinates accordingly
+if (BANGLEJS2) {
+  drag = undefined;
+  y = 45;
+  x = 5;
+} else {
+  // If the hardware version is Bangle.js 1, set the coordinates accordingly
+  y = 100;
```
thyttan, line 151: Lint warning: `'y' is assigned a value but never used`

So seems like you could just remove it?

apps/burn/burn.js
```diff
@@ -0,0 +1,254 @@
...
+  x = 25;
```
thyttan, line 152: And likewise here: `'x' is assigned a value but never used`

thyttan (changes requested):

apps/burn/.prettierignore
thyttan, line 1: I think this is just for formatting locally on your machine? We'll probably want to not have that file on the repo. What do you think @bobrippling @halemmerich ?

thyttan:
Thanks! I left some comments. 

If you want to let the calculation run in the background you could implement it in a `burn.boot.js` file or as a widget. But no need to do that in this PR - just a feature suggestion for the future.

Also I'd prefer if filenames were renamed from 
`burn.js`, `burn_icon.png` and `burn-icon.js`
 to simply 
`app.js`, `app.png` and `app-icon.js` as is common in the repo.

slynchDev:
thank you, I'll make your suggested changes

thyttan:
OK - looking forward to see it when your ready again! 🙂

apps/burn/.prettierignore
bobrippling: Yeah I agree, thanks for taking this :)

PR #3447, opened 2024-06-08
fix(astrocalc): handle when moon rise/set don't occur on current day

paul-arg:
I noticed a bug in the Moon > Times menu.

My (very basic) understanding is that the `rise` and `set` members of the `times` object returned by the `getMoonTimes` function car actually be `undefined`, depending on if the moon rise or set actually occur on the current day.

This MR handles these cases.

### DIFF
--- a/apps/astrocalc/ChangeLog
+++ b/apps/astrocalc/ChangeLog
@@ -6,3 +6,4 @@
 0.06: Fix azimuth (bug #2651), only show degrees
 0.07: Minor code improvements
 0.08: Minor code improvements
+0.09: Fix: Handle when the moon rise/set do not occur on the current day
--- a/apps/astrocalc/astrocalc-app.js
+++ b/apps/astrocalc/astrocalc-app.js
@@ -232,8 +232,6 @@ function drawSunShowPage(gps, key, date) {
   return null;
 }
 
-let m;
-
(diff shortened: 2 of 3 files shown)


thyttan:
Looks good to me. Just bump version in metadata and changelog and I'll merge 👍

thyttan:
Thanks! 🙂

thyttan:
```

! Removed lint exemptions for 'apps/astrocalc/astrocalc-app.js' because it has been edited

✔️ Synchronized all lint exemptions



/home/runner/work/BangleApps/BangleApps/apps/astrocalc/astrocalc-app.js

  303:10  warning  'getCenterStringX' is defined but never used  no-unused-vars

  314:5   warning  'm' is assigned a value but never used        no-unused-vars


✖ 2 problems (0 errors, 2 warnings)

```


*https://github.com/espruino/BangleApps/actions/runs/9431360748/job/25979927607#step:5:21*


Sorry - just looked at the lint warnings and as said above there was an exemption removed since the file was changed. 


Would you mind removing those unused variables so we get rid of the warnings? They have nothing to do with your changes but it would be neat fixing them as well.

thyttan:
I fixed the warnings from the linter. Will merge now and test on my watch as soon as the development app loader shows the new version. 

Thanks again!

Issue #3275, opened 2024-03-17
[Sleep Log] Gadgetbridge sync support


joserebelo:
I would like to implement sync support for the sleep logs to Gadgetbridge.

I see that we already send real-time events to Gadgetbridge, but they are currently not being handled. Aditionally, any events that happen while disconnected will be lost.

I was envisioning a mechanism similar to #2889, where we request all the sleep logs since a timestamp and update existing records in Gadgetbridge, along with a button in the preferences to trigger a full fetch of all data present on the Bangle, as that might take some time to execute.

@storm64 do you have any thoughts on this?

gfwilliams:
I could be wrong here, but as far as I can see from https://github.com/espruino/BangleApps/blob/master/apps/sleeplog/boot.js#L27 sleep log mainly uses the 10 minute 'movement' value from the built-in health event.

This is great because it's not really using any extra power, but also the `health` app is already storing this information too, all the time. And it's synced already via the `{t:'act', mov:...}` - we may just not be storing it in Gadgetbridge?

... so in a way, we may just be able to run [this algorithm](https://github.com/espruino/BangleApps/blob/master/apps/sleeplog/boot.js#L160-L177) on data that's already in gadgetbridge? Looks like sleeplog does check wear state and whether it's charging though, which we don't store (but charge state at the very least could be sent/stored trivially, and I guess if it's not worn we won't have a BPM reading for that 10 min health event). 

That's one thought, but also sleeplog saves everything to a storage file. Perhaps we should consider having some kind of generic file sync built into Gadgetbridge rather than setting out to do it just for sleeplog?

* metadata.json already has a `data` field listing data files the apps make. Maybe if `sync:true` is set there, we could make sure that's pushed through to `appname.info` in the bangle
* when syncing Gadgetbridge could scan those files and figure out a list of files to sync, and could then automatically sync them

Espruino has two different storage types (one fixed length and binary, one appendable and text). There's already the code to handle syncing the appendable one for tracks thanks to @thyttan, but I guess we might need to add something to deal with the binary one.

thyttan:
@gfwilliams 

> This is great because it's not really using any extra power

I also like this aspect of `sleeplog`. But every time I use it for a while I end up suspecting the app of being a big drain on battery somehow. Do you have this experince as well? Could it be writes to storage? I haven't looked at this deeply yet.

gfwilliams:
[...]

PR #3448, opened 2024-06-09
Burn - Simple Calorie Counter

slynchDev:
Fixes applied per your suggestion (prior PR)


slynchDev:
My apologies, I should have kept the prior PR and updated that branch rather than create a new PR (still learning the process here). I applied your suggestions: remove a few unused variables, removed .prettierignore, renamed all app files per the standard in the repo. Thanks

thyttan:
Thank you for the app and the quick changes! Always fun to see new contributors! :)

slynchDev:
Thank you! I did have a couple questions if you don’t mind?

1) I see there a centralized health service to store calories information
like this?

2) Any chance you could elaborate on your boot.js feature request
(calculate on startup)?

Sean

thyttan:
>Thank you! I did have a couple questions if you don’t mind?

Sure - I'll answer to the best of my abilities. :)

>1) I see there a centralized health service to store calories information
like this?

There's the `health` app that log some stats, https://thyttan.github.io/BangleApps/?q=health, is that what you're referencing? @gfwilliams has proposed to add more stats to `health` here: https://github.com/espruino/BangleApps/issues/3275#issuecomment-2006364430

>2) Any chance you could elaborate on your boot.js feature request
(calculate on startup)?

The suggestion was based of my misunderstanding of the app. I thought the idea was that it would automatically estimate calories burned based on activity. If that had been the case it would have made sense to have logic running at all times, also when other apps/the clock was running, to keep track on the activity/calorie burn. But as I understand it the idea is you use the app to count caloric intake?

slynchDev:
Thanks so much. Yes, it’s simply used to count caloric intake.

Ok I will take a closer look at the health app to see if / when there’s an
integration opportunity.

Thanks!

PR #3450, opened 2024-06-10
VPW clock: do not remove VGA8 font for fast loading

paul-arg:
As per this discussion: https://forum.espruino.com/conversations/395233

Do not remove the VGA8 font as part of the remove `handler`


thyttan:
Thanks!

Issue #3435, opened 2024-05-27
Discussion: HW buttons should act on `'rising'` edge


thyttan:
I recently watched "Theo - t3.gg" discuss a twitter post from John Carmack talking about using edge `'rising'` vs `'falling'` and user experience. The take home message being to act on the `'rising'` edge when it's possible. Here's the video: https://www.youtube.com/watch?v=yaMGtiPckAQ

In the name of responsivness I wonder if it would make sense to adhere to this way of thinking when programming the hardware button on the Bangle.js.

I have personally made changes from `'rising'` to `'falling'` edge responses as to not leave a `'falling'` edge for a subsequent screen to pick up - I was seeing some bug related to this ([one reference](https://github.com/espruino/BangleApps/pull/3297#discussion_r1539005248)). Now I find myself wanting to go the other way.

To avoid unexpected/unwanted interactions on a left over `'falling'` edge, could some default behaviour of disregarding just the `'falling'` edge event if it occurred within a short time after the `setWatch` was called be implemented? If not in `setWatch` then maybe in `setUI`? Or could this be solved by changing the default of reacting on `'both'` to instead use `'rising'`?

Really I just wanted to discuss if others also thinks this is worthwhile. Maybe it's a lot of work, complicating the logic and not backwards compatible. But I want as much snappiness as possible out of the watch! :P

thyttan:
Probably relevant comment:

>Yes, responding on the falling edge looks good - and was certainly the reason for going back 2 levels. 
>
>I forget the original reasoning for using falling in Bangle.js (I think someone wanted to be able to detect difference between long/short presses?) but now everything uses falling, so really most apps should try and use that where possible to avoid this issue.
>
>I'd prefer not to change the `setWatch` default just because it's so low level and shared with other devices, but I think ideally we should just try and use `setUI` wherever we can and avoid this and the whole `isBTN3?BTN3:BTN1` issue (I should really update the tutorials)

_Originally posted by @gfwilliams in https://github.com/espruino/BangleApps/pull/3297#discussion_r1540748465_

thyttan:
Maybe I can try this out by [overwriting] `setWatch` and/or `setUI` to see how it would behave. If it works I could make it into an app as per the linked tutorial.

gfwilliams:
Yes, you could try overwriting those functions and see what happens - overwriting setWatch might be enough, and maybe it's something you could even provide as an app.

[...]

PR #3452, opened 2024-06-11
Discussion: Proposals for new ways to interact with setUI

thyttan:
This is not meant to be merged - only for providing an example/proposal for changes to how we can interact with Bangle.setUI. This is an addition to the discussion at https://github.com/espruino/BangleApps/issues/3435.

The diff has comments explaining my thoughts.

### DIFF
--- a/apps/spotrem/app.js
+++ b/apps/spotrem/app.js
@@ -98,25 +98,39 @@ let swipeHandler = function(LR, _) {
   }
 };
 
+let buttonHandler = (n)=>{load();};
+
 // Navigation input on the main layout
 let setUI = function() {
 // Bangle.setUI code from rigrig's smessages app for volume control: https://git.tubul.net/rigrig/BangleApps/src/branch/personal/apps/smessages/app.js
-  Bangle.setUI(
-    {mode : "updown",
+  Bangle.setUI({
+      mode : "updown",
       remove : ()=>{
-        Bangle.removeListener("touch", touchHandler);
-        Bangle.removeListener("swipe", swipeHandler);
-        clearWatch(buttonHandler);
-        widgetUtils.show();
-      }
+       widgetUtils.show();
+      },
+      /**
+       * TODO: Proposal: Handlers added at the same level as the `mode : "updown"` entry
+       * will replace those that come with the mode - if there were any of that type.
+       * The handlers in the `extra` object below will still be added in addition to these ones.
+       **/
+      swipe : swipeHandler,
+      btn : buttonHandler,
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - extending to add the handlers looks good, but the whole `extra` thing seems like a bit of a mess. I think if you want to do something like that, it should really just be using `custom` and implementing its own behaviour

apps/spotrem/app.js
```diff
@@ -98,25 +98,39 @@ let swipeHandler = function(LR, _) {
...
+       widgetUtils.show();
+      },
+      /**
+       * TODO: Proposal: Handlers added at the same level as the `mode : "updown"` entry
+       * will replace those that come with the mode - if there were any of that type.
+       * The handlers in the `extra` object below will still be added in addition to these ones.
+       **/
+      swipe : swipeHandler,
+      btn : buttonHandler,
+      btnEdge : "rising", // TODO: Proposal: a new, optional, way to choose what edge the button(-s) should act on.
```
thyttan, line 118: @gfwilliams what do you think of the idea to be able to specify the desired edge like this or some other way?

gfwilliams: Well, I was going to say this didn't feel great but I couldn't think of a better way...

But then I was thinking: "What if you want to do different things on Rising *and* falling edges?"

So what if we had `btn` (as before) but then `btnPress`/`btnRelease` callbacks could be specified? That feels like maybe a tidier solution? What do you think?

gfwilliams: ... the only downside of this is I guess it's not easily backwards compatible - but if we want this stuff to work outside of the `mode:custom` then maybe that doesn't matter as it wouldn't work now anyway

thyttan: I think I agree with your callbacks solution being tidier.

But just regarding this: 
> But then I was thinking: "What if you want to do different things on Rising and falling edges?"

I'm not saying it's the better solution, just another one. If "both" was specified in my example above that could be passed on to the `setWatch` call inside setUI and then the `buttonHandler` function would be responsible for handling both edges.

If you feel callbacks are better that carries much weight for me :)

thyttan: @gfwilliams 

I think I agree with your callbacks solution being tidier.

But just regarding this: 
> But then I was thinking: "What if you want to do different things on Rising and falling edges?"

I'm not saying it's the better solution, just another one. If "both" was specified in my example above that could be passed on to the `setWatch` call inside setUI and then the `buttonHandler` function would be responsible for handling both edges.

And if you feel callbacks are better that carries much weight for me :)

PR #3455, opened 2024-06-15
New library: add a double picker and a triple picker

paul-arg:
Following the discussion here: https://forum.espruino.com/conversations/397052

I created a module that provides a double picker and a triple picker.

screenshot (1)
screenshot

If you want to test quickly you can add `|https://raw.githubusercontent.com/paul-arg/BangleApps/more_pickers/modules` to the Modules URL in the Communication section of the IDE Settings, and use the code snippet provided in the Markdown file.

### DIFF
--- a/modules/more_pickers.js
+++ b/modules/more_pickers.js
@@ -1,0 +1,198 @@
+/* see more_pickers.md for more information */
+
+exports.doublePicker = function (options) {
+  var menuIcon = "\0\f\f\x81\0\xFF\xFF\xFF\0\0\0\0\x0F\xFF\xFF\xF0\0\0\0\0\xFF\xFF\xFF";
+
+  var R = Bangle.appRect;
+  g.reset().clearRect(R);
+  g.setFont("12x20").setFontAlign(0, 0).drawString(menuIcon + " " + options.title, R.x + R.w / 2, R.y + 12);
+
+  var v_1 = options.value_1;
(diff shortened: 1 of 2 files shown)


modules/more_pickers.js
```diff
@@ -0,0 +1,198 @@
...
+    var txt_1 = options.format_1 ? options.format_1(v_1) : v_1;
+    var txt_2 = options.format_2 ? options.format_2(v_2) : v_2;
+
+    g.setFontAlign(0, 0)
+      .setFontVector(Math.min(30, (R.w - 110) * 100 / g.setFontVector(100).stringWidth(txt_1)))
+      .drawString(txt_1, 47.5, 113)
+      .setFontVector(Math.min(30, (R.w - 110) * 100 / g.setFontVector(100).stringWidth(txt_2)))
+      .drawString(txt_2, 128.5, 113)
+      .setFontVector(30)
+      .drawString(options.separator ?? "", 88, 110);
```
bobrippling, line 33: @gfwilliams looks like the IDE is out of sync with espruino:

image

`??` is parsed and handled fine

bobrippling:
This is very cool! Thanks

modules/more_pickers.js
gfwilliams: Yes - it uses Acorn and Tern for parsing and type inference, but I guess they are older versions that don't support ??

PR #3457, opened 2024-06-16
Add BLE privacy setting

bobrippling:
This follows on from espruino/Espruino#2506, providing an easy way for a user to configure BLE privacy.

Currently the simple modes are supported:
- Off
- On
- On with the device's name hidden

---

This also updates typescript types for the settings object, and brings in typings from espruino (espruino/Espruino#2516)

### DIFF
--- a/apps/boot/ChangeLog
+++ b/apps/boot/ChangeLog
@@ -70,3 +70,4 @@
 0.59: Whitelist: Try to resolve peer addresses using NRF.resolveAddress() - for 2v19 or 2v18 cutting edge builds
 0.60: Minor code improvements
 0.61: Instead of breaking execution with an Exception when updating boot, just use if..else (fix 'Uncaught undefined')
+0.62: Handle setting for configuring BLE privacy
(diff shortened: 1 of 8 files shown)


gfwilliams:
This looks pretty good to me.

Am I right in thinking this needs the absolute latest firmware though? I'm thinking if you run this on 2v22 then `NRF.setSecurity` might actually throw an exception because it doesn't know about the `privacy` field? Worth trying I guess.

If it silently ignores it then we're good, if not maybe `bootupdate.js` needs to check the firmware version is 2v23 or later

bobrippling:
> This looks pretty good to me.
> 
> Am I right in thinking this needs the absolute latest firmware though? I'm thinking if you run this on 2v22 then `NRF.setSecurity` might actually throw an exception because it doesn't know about the `privacy` field? Worth trying I guess.
> 
> If it silently ignores it then we're good, if not maybe `bootupdate.js` needs to check the firmware version is 2v23 or later

Thanks - yes this needs 2v23, but running it on an older firmware is ok, the [`privacy` field is ignored](https://github.com/espruino/Espruino/blob/49b8ac8e97efaac9f562ad879e7cf3f927a57dd8/targets/nrf5x/bluetooth.c#L2191-L2196) (same for if `ESPR_BLE_PRIVATE_ADDRESS_SUPPORT` isn't defined).

Also confirmed on my watch (still on 2v22)

gfwilliams (approved):
Thanks - this looks great then! Feel free to merge if you want to :)

bobrippling:
Cool, thank you!

PR #3458, opened 2024-06-16
promenu: fix values going below minimum, improve callbacks

bobrippling:
- Fixed values going below minimum, if minimum was 0 (i.e. falsy)
- Only call callbacks if something's changed


bobrippling:
Had no issues, merging

PR #3459, opened 2024-06-16
widdst: ensure DST is applied when leaving settings

bobrippling:
... for example, when fastload is installed, so we don't jump back to the clock with the old DST settings.


bobrippling:
Tested: full load is performed when changing DST setting, fastload when changing non-DST setting.
When fastload isn't installed, a full load is performed when changing DST, no errors occur

PR #3460, opened 2024-06-17
Correct interface.js path in interface.html example

ticalc-travis:
I used the example code for interface.html and couldn't get it to work until I made this change (which I had seen in all the other public source code). So I'm assuming this was an error or an outdated example that should be updated.


bobrippling:
Nice, thanks for the fix/diligence!

Issue #3128, opened 2023-12-13
[lcars] Temperature is just a black bar


Zombiefleischer:
### Affected hardware version

Bangle 2

### Your firmware version

2v19

### The bug

I only get a black bar that spans all the way to the left side instead of a temperature reading from the watch face. Other weather related infos are working correctly. It was working for a short while after I reseted the watch to factory settings and installed only the needed apps, but an hour or so later it stopped working so I installed a bunch of other apps. Also to note, it works fine on something like the Weather Clock or the Weather App.
![20231213_123331](https://github.com/espruino/BangleApps/assets/44850408/31055337-d353-42c3-aa27-d45bce0418a7)

I use Tiny Weather Forecast as my source and the banglejs gadgetbridge as bridge.

### Installed apps

Anton Clock
About
Lock Widget
Bluetooth ID Widget
Welcome
Scheduler
Alarms Widget
Bootloader
Bluetooth Widget
Notifications (default)
Health Tracking
Settings
Message Icons
Message Widget
Messages
Android Integration
Message List
A Battery Widget (with percentage)
App Manager
Weather
Desktop Launcher
GPS Info
Barometer
Gadgetbridge Music Controls
OpenStreetMap
Thermometer
GPS Time
Languages
Dice-n-Roll
Weather Clock
LCARS Clock
Alarms & Timers
Clockinfo Clock
Clock Info Module
Clock Info Widget

bobrippling:
Looks like it might be a colour problem - @peerdavid anything like this ring a bell?

CountMurphy:
Its an issue with the 2.19 firmware. I have the same problem. Downgrading to 2.18 fixes it.

bobrippling:
Thanks for checking - sounds like it's related to [this espruino commit](https://github.com/espruino/Espruino/commit/eabd13cfb8ce1c0a81708bb6e9f37f15b101a5fd#commitcomment-135446714) and #3144, so should be fixed when [this commit](https://github.com/espruino/Espruino/commit/91e45d60f1693ed61c93c0163b7c30d6168cc7d9) makes it to firmware (v2.20 at a guess)

CountMurphy:
I'll report back whenever 2.20 comes out

CountMurphy:
2.20 still has the issue.

bobrippling:
[...]

Issue #3312, opened 2024-03-30
Improve how we manage font ability to display charsets in translation


jordimas:
### Affected hardware version

Bangle 2

### Your firmware version

2v12

### The bug

**Background**

Currently when uploading the Bangle translations to the app, we remap any chars that we don't think we can display in Espruino's built in fonts. See:  

https://github.com/espruino/EspruinoAppLoaderCore/blob/master/js/appinfo.js#L77

In the mappings:

https://github.com/espruino/EspruinoAppLoaderCore/blob/master/js/utils.js#L75

things like "á" are mapped to "a".

**Problem**

This has an impact in many languages. For example, in Spanish or Catalan "Sí" means "Yes" but "Si" (without accent)" means "If". Removing the accent has an impact on meaning.  Additionally it is a spelling mistake.

Some [Espurino fonts](https://www.espruino.com/Fonts) are able to display iso-8859-1 characters and others do NOT. By doing this we remove the possibility for ALL the fonts to show these characters. This provides a suboptimal experience since we remove chars that can be displayed in some fonts.

**Proposed solution**

The goal is to display the iso-8859-1 when the font allows so for a better internationalised experience.

An alternative approach that will allow to display characters when it is possible:

1. When [processing the translation](https://github.com/espruino/EspruinoAppLoaderCore/blob/master/js/appinfo.js#L77
) do not do any mapping, just make sure that the chars are that are valid iso-8859-1.

2. When we draw the string (e.g DrawString) do the mapping ONLY if the font does not support all iso-8859-1 chars.

bobrippling:
Thanks for the detailed issue - a good offshoot of #3109, @gfwilliams tagging for visibility

gfwilliams:
Thanks - this was the idea actually. While perhaps not clear in naming, `convertStringToISOLatin` should convert to `ISO8859-1` - so please feel free to remove any conversions that shouldn't be there (or add some that should, like 0x0400 to 0x00C8 and similar make a lot of sense).

Most Espruino fonts (apart from 4x6 where it doesn't make sense) should now support  ISO8859-1 I believe (I made some changes recently to 6x15 where it didn't handle some chars correctly so I think that doing the first point you suggested would fix most things.

> When we draw the string (e.g DrawString) do the mapping ONLY if the font does not support all iso-8859-1 chars.

[...]

PR #3461, opened 2024-06-19
degree symbol replaced

lawern:
In the newer espruino versions, the degree symbol is not displayed correctly for the temperature in the clock.
I have therefore replaced it with an apostrophe, as shown in the screenshot.


bobrippling:
This might be related to recent font changes: #3312 / espruino/EspruinoAppLoaderCore#60

@jordimas / @gfwilliams what do you think?

gfwilliams:
Just checked into this. The Font used is provided in the file, AntonioMedium - it contains capitals/numbers, but no degree symbol (176) or anything above char code 94 - so I think this is the right solution.

I'm honestly not sure if Espruino ever would have properly displayed this in any version, but maybe Locale used to use `'` rather than the degree symbol?

bobrippling:
Sounds good, I'll merge. Thanks for the fix @lawern!

lawern:
Fixes #3128

PR #3462, opened 2024-06-19
Update Daisy clock to add battery estimate in hs

Poolitzer:
Hi, I saw https://forum.espruino.com/conversations/395838/ and I wanted to implement it on my favorite clock. So I did.

Maybe @hughbarney has a review comment it being your app?

I decided not to check the firmware version when toggling the setting, this could be added if wanted (it being off by default I don't think people enable it if they aren't on the most recent version, or that those people couldn't find out the issue then :D)

Edit: After running around with this I realized that the big hour estimate is really not something I want to convert in my head, so I changed it now that it shows days instead. As soon as we get in the lower region the actual estimate jumps really noticeably, I am unsure if I should "catch" this here and probably just not display them, or if the function should return a more linear estimate. And tbh I personally dont mind the jumping numbers

### DIFF
--- a/apps/daisy/ChangeLog
+++ b/apps/daisy/ChangeLog
@@ -9,3 +9,4 @@
 0.09: Use 'modules/suncalc.js' to avoid it being copied 8 times for different apps
 0.10: Use widget_utils.
 0.11: Minor code improvements
+0.12: Added setting to change Battery estimate to hours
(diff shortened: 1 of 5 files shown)


bobrippling:
LGTM, we'll give @hughbarney a day or so more then I'll merge

bobrippling:
Thanks for the change! If you're interested, I have a similar change in #3477 for a widget too

thyttan:
There's a bug making the app code and settings code default to different things: https://github.com/espruino/BangleApps/issues/3493#issuecomment-2211828151

PR #3470, opened 2024-06-22
New Clock face for measuring time

prefectAtEarth:
## Measure Time

New clock face for literally measuring time. Inspired by a clock face I had on my first Pebble Steel.

- face reloads on every full minute and redraws
- widgets are hidden and can be shown on swipe down


bobrippling:
Nice! Thanks for the tests too 👌

PR #3472, opened 2024-06-23
elapsed_t: use new pickers from the more_pickers module, add seconds display settings

paul-arg:


bobrippling:
Thanks! It looks like [`howManyDaysInMonth`](https://github.com/espruino/BangleApps/blob/e984cb18c0ce0553d9528e9b38cf6860eb319ac4/apps/elapsed_t/app.js#L124-L126) is unused, are you ok to delete it?

thyttan:
Hi @paul-arg! I saw your active again - what do you think of @bobrippling's question above?

thyttan:
OK - thanks!

PR #3473, opened 2024-06-24
New Reply/Canned Responses library for replying to messages

WhelanB:
Adding a new library, "Reply", which allows apps to prompt the user to either select from a set of canned responses or to compose a reply if a ```textinput``` keyboard app is installed. The library handles constructing the reply object for Gadgetbridge and can optionally send this via ```Bluetooth.println``` or handle the reply directly, if they wish to use it for other purposes

It can be used by calling:

```require("reply").reply(options).then(...```

from an app.

Canned responses can be set from the App Loader - though using emojis seems to break the created JSON and I haven't looked at re-encoding the text first before writing yet.

Untitled
image

After it merges, I might go ahead and add it to the default messages app - I've been testing it personally for the past week or so and can reply to messages directly from my Bangle, which is a nice feature to have!

### CODE REVIEW

File: apps/reply/interface.html [hunk 1]
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        function toggleVisibility() {
+            let empty = document.getElementById("empty");
+            let loading = document.getElementById("loading");
+            if (replies.length == 0) {
+                empty.setAttribute("class", "d-block");
+                loading.setAttribute("class", "d-hide");
+            }
+            else if (fetching) {
```
bobrippling, line 68: Should these be the other way round? Testing `fetching`, then `replies.length` ?

apps/reply/interface.html
```diff
@@ -0,0 +1,125 @@
...
+
+        function updateDevice() {
+            fetching = true;
+            renderReplyList();
+            Util.writeStorage("replies.json", JSON.stringify(replies), () => {
+                fetching = false;
+                renderReplyList();
+                return true;
+            });
+            fetching = false;
```
bobrippling, line 119: Should the `fetching = false` above (line 115) be the one to set `fetching` to false, and not this line?

apps/reply/lib.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  return new Promise((resolve, reject) => {
+    var menu = {
+      "": {
+        title: options.title || /*LANG*/ "Reply with:",
+        back: function () {
+          E.showMenu();
+          layout.setUI();
+          layout.render();
```
bobrippling, line 30: Do we need to do these when the user's exiting?

apps/reply/lib.js
```diff
@@ -0,0 +1,75 @@
...
+
+    if (replies.length == 0) {
+      if (!keyboard) {
+        E.showPrompt(
+          /*LANG*/ "Please install a keyboard app, or set a custom reply via the app loader!",
+          {
+            buttons: { Ok: true },
+            remove: function () {
+              layout.setUI();
+              layout.render();
```
bobrippling, line 60: Same here :)

bobrippling (commented):
Looks great! Just a few comments, looks like the linters picked up on `layout` not being declared too - let me know if you want a hand with any of this

apps/reply/lib.js
WhelanB: I had added this to fix a white screen during testing - just removed it as the calling app can redraw after the promise resolves!

apps/reply/interface.html
WhelanB: Yes you're right, fetching should only update when the promise resolves - fixed!

apps/reply/lib.js
bobrippling: Cool, sounds good!

bobrippling:
Thanks, looking forward to seeing it see some action!

soneret:
great job 
 i cant see reply function when massage arrived .i installed your app.what can be problem .thanks in advance

WhelanB:
> great work thanks i cant see reply function when massage arrived .i installed your app.what can be problem .thanks in advance

I still need to open a second PR to use the library from inside the Messages app - I've been testing it myself but haven't had a chance to publish it yet!

PR #3474, opened 2024-06-26
BTHome battery level broadcast for e.g. Home Assistant

deirdreobyrne:
Simple app to broadcast the battery level using the [BTHome protocol](https://bthome.io/) - makes for easy integration with Home Assistant.

I tried the other BTHome app, but it didn't work for me. That other app is more complicated anyway - and goes beyond what my needs are.


bobrippling:
Thanks! Would you mind if we moved the advert code over to using the `ble_advert` module? This'll let it work nicely with other boot apps that might leave `bleAdvert` in a different state (such as openhaystack) or with future changes to the advertising code - it's more or less something like:

```diff
 function advertiseBTHomeBattery() {
   var advert = [0x40, 0x00, btHomeBatterySequence, 0x01, E.getBattery()];
 
-  if(Array.isArray(Bangle.bleAdvert)){
-    var found = false;
-    for(var ad in Bangle.bleAdvert){
-      if(ad[0xFCD2]){
-        ad[0xFCD2] = advert;
-        found = true;
-        break;
-      }
-    }
-    if(!found)
-      Bangle.bleAdvert.push({ 0xFCD2: advert });
-  } else {
-    Bangle.bleAdvert[0xFCD2] = advert;
-  }
-  NRF.setAdvertising(Bangle.bleAdvert);
+  require("ble_advert").set(0xFCD2, advert);
+
   btHomeBatterySequence = (btHomeBatterySequence + 1) & 255;
 }
 
-if (!Bangle.bleAdvert) Bangle.bleAdvert = {};
 setInterval(function() {
   advertiseBTHomeBattery();
 }, 300000); // update every 5 min
```

Out of interest, did you see any errors when using BTHome? Did anything appear in [the IDE](https://www.espruino.com/ide/#)?

deirdreobyrne:
Done.

I wasn't actually looking at the IDE - I was looking at the BLE broadcasts, and I just didn't see the battery level being advertised via the BTHome protocol. I probably did something stupid. But that app is waaay too complicated for what I actually need - I need a BTHome equivalent of the GATT battery broadcasts.

bobrippling:
Great thank you - that makes sense, nice to have a lightweight app!

PR #3477, opened 2024-06-28
hwid battery widget: show power-draw and hours remaining

bobrippling:
This changes the battery widget to show:
- The power currently being drawn in colour (red for high, yellow is medium, green is low)
- The hours remaining at the current power consumption, by shading the battery amount (e.g. 50%) up a certain level:
image

At the moment I mostly see fully red or nothing, @gfwilliams do I have the maths correct for this? I feel like I've got it too sensitive so I never really see medium power consumption


gfwilliams:
It looks ok to me - maybe you can get it to print what it thinks is drawing power as a test?

The LCD backlight draws a ton, so my guess is maybe whenever you have that on (eg the watch is unlocked) it puts the meter red?

bobrippling:
> The LCD backlight draws a ton, so my guess is maybe whenever you have that on (eg the watch is unlocked) it puts the meter red?

Yeah that's a good point - I've altered it to add up power draws not including LCD, to avoid the jolt when the screen turns on. Shall see how that goes

bobrippling:
Change of plan - if `getPowerUsage()` is available, it'll show hours remaining, with the battery percentage as the shading - much easier to understand:

image


@pebl-hank @matijatosic I think you two are the main users of this widget and this is quite a change - would you be ok with it? I'm very happy to move the hours-left display to another widget

pebl-hank:
@bobrippling I'm totally OK with these ideas and happy to have you as a contributor to this. 
Haven't time to check the code though, so I want to add that both options should be selectable to switch on/off by the user in the app settings. Also please keep an eye on battery consumption: If the remaining hours option is not enabled by the user, the calculating code should be avoided. But I bet you already considered all that :) So thanks again 👍

bobrippling:
Sounds good - I'll create a setting for it, cheers for your response!

bobrippling:
On second thoughts, I figured it'd be easier to fork the widget - it was becoming a bit unwieldly!

pebl-hank:
Noooooo! I'm really not a fan of forks. Is there any way to convince you to stay on the path, young padawan?

gfwilliams:
I feel like you're changing what the widget does quite a lot here, checking the power usage, and especially as you've now rewritten in typescript.

Widgets are pretty small bits of code so it's not so much a fork now as a complete new thing - and from a users point of view it's nice if the app name, screenshot and description reflects that it measures power usage. IMO a new app is probably better, and then maybe if you wanted to tweak the way things are rendered later to better show off power usage it wouldn't cause problems for people that like hank's one the way it is

bobrippling:
That's fair, the code is much simpler without having to handle options - I'll test this for a little more and merge if that's ok with you both?

apps/widbattpwr/widget.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		g.setColor(g.theme.fg);
+		drawBar(x, y, 100);
+		g.setColor(pwrColour);
+		drawBar(x, y, batt);
+
+		g.setFontAlign(0, 0);
+		g.setFont("Vector", 16);
+		{
+			const hrs = 200000 / usage;
```
bobrippling, line 45: @gfwilliams thinking about it - does this assume the battery is fully charged? Would the 200k need to be scaled down based on the battery's percentage?

gfwilliams: If it was time until flat then yes, I think so. Also battery is 175mAh so that needs a tweak anyway :)

bobrippling: Cool, I'll linearly scale it based on the battery percentage. I guess we'll want to change this defininition in @Poolitzer's additions to `daisy` too?

https://github.com/espruino/BangleApps/blob/3211cf6955624ea01913581da005106189f1e4e5/apps/daisy/app.js#L119-L130

Poolitzer: Now I do feel though we should somehow export this that not all implementations of getPowerUsage() => remaining hours/days have to redo this, since it is math you can quite easily mess up a bit

bobrippling: Sounds good - would you mind if I merge this as-is and did the consolidation of getPowerUsage --> hours/days in a separate PR? And yes - I got the 200k from that forum post too

bobrippling: PR ready: #3498, gonna merge this

PR #3480, opened 2024-06-30
delaylock: delay lock 5 sec after backlight off

thyttan:
New bootloader app for delaying the locking of the touchscreen to 5 seconds after the backlight turns off. Giving you the chance to interact with the watch without having to press the hardware button again.

Inspired by the conversation between @gfwilliams and user156427 linked here: https://forum.espruino.com/conversations/392219/

Can be installed from my app loader: https://thyttan.github.io/BangleApps/?id=delaylock


bobrippling:
LGTM, thanks!

PR #3483, opened 2024-07-01
A new clock using dithering effects

Chriz76:


bobrippling:
Very cool, great design! Thanks

PR #3489, opened 2024-07-03
swiperclocklaunch: fix detection of clock

bobrippling:
swiperclocklaunch now reliably detects a clock, using existing logic in `setUI` instead of inspecting arguments


bobrippling:
Currently testing - will merge if no issues

PR #3490, opened 2024-07-03
Updes on accrec, pebbleapp and bthome app. 

pinq-:
Changes on the accrec, pebblepp and bthome


apps/accelrec/app.js
```diff
@@ -1,179 +1,251 @@
...
-  var maxAccel = 0;
-  var tStart = SAMPLES, tEnd = 0;
-  var vel = 0, maxVel = 0;
-  for (var i=0;i<SAMPLES;i++) {
-    var a = accely[i]/SCALE;
-    if (a>0.1) {
-      if (i<tStart) tStart=i;
-      if (i>tEnd) tEnd=i;
+    g.clear(1);
+    let csv_files_N = require("Storage").list(/^acc.*.csv$/).length;
```
bobrippling, line 71: You probably want this, so it matches the dot literally:
```suggestion
    let csv_files_N = require("Storage").list(/^acc.*\.csv$/).length;
```

(and below)

apps/bthome/settings.js
```diff
@@ -11,10 +11,24 @@
     require("Storage").writeJSON("bthome.json",settings)
   }
 
+  function getNewIdNumber(){
+    var id_start = 1;
+    var select = false;
+    while(!select){
+      select = true;
+      settings.buttons.forEach((button, idx) => {
+        if (id_start == button.n){
```
bobrippling, line 20: Could the buttons ever go, `2, 1, 3`, for example?

bobrippling (commented):
Looks good! Some things I spotted:

apps/bthome/settings.js
```diff
@@ -11,10 +11,24 @@
...
+          id_start++;
+          select = false;
+        }
+      });
+    }
```
KTibow, line 25: I would personally write this like
```
const id_start = [1, 2, 3].find(id => settings.buttons.every(button => id != button.n));
```

apps/bthome/settings.js
pinq-: Yes it can, if you delete one of the buttons and it re-ranges it.

apps/bthome/settings.js
pinq-: This is not bad suggestion. It would limit the choices to first array options and I liked that it was "limitless". But in practice, you would have max 5-6 buttons ( if you have more that 10, then you have to scroll a lot and I don't think it is really practical) . I will use this code with array off 1-9. This is good one. Thanks!

apps/accelrec/app.js
pinq-: Thanks! I will fix this!

bobrippling:
LGTM - happy to merge when you're happy with how it tests out. Let me know - I'll merge in a day or two otherwise

pinq-:
green flag from my testing also.

bobrippling:
Thanks for the updates!

WhelanB:
Any chance the Pebble++ date at the bottom could be put behind a feature flag/setting? was a bit strange to have it appear after an update as it sticks out quite a bit and doesn't align with the theme (I've got light mode set and a custom colour), especially since the date is already available via clockinfo:
image

PR #3494, opened 2024-07-06
setuichange: refine when to set button to show launcher

thyttan:
Now if some custom btn or btnRelease handler is set the clock button handler won't be added.

@gfwilliams

### DIFF
--- a/apps/setuichange/ChangeLog
+++ b/apps/setuichange/ChangeLog
@@ -1,3 +1,4 @@
 0.01: New App!
 0.02: Fix case where we tried to push to Bangle.btnWatches but it wasn't
 	defined.
+0.03: Tweak when to add button clock action.
--- a/apps/setuichange/boot.js
+++ b/apps/setuichange/boot.js
@@ -72,7 +72,7 @@ Bangle.setUI = (function(mode, cb) {
     ];
   } else if (mode=="clock") {
     Bangle.CLOCK=1;
-    Bangle.btnWatches = [
+    if (!(options.btn||options.btnRelease)) Bangle.btnWatches = [
       setWatch(Bangle.showLauncher, BTN1, {repeat:1,edge:"rising"})
     ];
   } else if (mode=="clockupdown") {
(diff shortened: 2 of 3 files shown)


gfwilliams:
Is this for a specific clock or use case?

I'm just thinking that in the Bangle.js 1 case (or maybe future clocks with >1 button) you can still want to respond to a button press while keeping the handler for the launcher? 

We already have https://github.com/espruino/BangleApps/blob/96a1626b45134e566588305712b63773eb898e2f/apps/setuichange/boot.js#L95C2-L95C37 specifically so we can do `mode:"custom",clock:1` which seems the more readable solution here, and it works already?

thyttan:
Ok, maybe instead for the Bangle 2 we throw an exception if mode==clock or mode==clockupdown similarly to what you suggested re drags and mode==updown or mode==leftright?

But for Bangle 1 we don't throw an exception?

gfwilliams:
Ahh, right - Sorry, I see the issue. It's not that it's unexpected what happens, it's that if it were to be added twice we'd lose track of the first handler?

Actually, maybe we could just move `if (options.clock) Bangle.btnWatches = [` downwards to where we check `options.btn` and then do:

```
if (options.clock) Bangle.btnWatches.push(setWatch(Bangle.showLauncher, BTN1, {repeat:1,edge:"rising"}))
```

in the same way we do with the other stuff?

thyttan:
So my thinking initially was that:
If I set mode clock or clockupdown,
 and if I want some custom button logic,
  then I don't want to also be loading the launcher with the hw button.

But maybe that doesn't make much sense in the bigger picture.

> Ahh, right - Sorry, I see the issue. It's not that it's unexpected what happens, it's that if it were to be added twice we'd lose track of the first handler?

So this is not what I was thinking about but that's also something we should fix now you point it out!

gfwilliams:
> If I set mode clock or clockupdown,
> and if I want some custom button logic,
> then I don't want to also be loading the launcher with the hw button.

Ahh, ok. I guess in that case, if you really want to you can just set `Bangle.CLOCK=1;` manually though, and you can still use `custom` or `updown` handler?

I just think we want to be making things as simple as possible - every new thing we add slows down the loading of apps/clocks and also becomes a maintenance headache later. So really I think it's best to hold off adding features that aren't being used at the moment, and maybe if we find a bunch of apps are having to use `Bangle.CLOCK=1;`  in them then we can always extend `setUI` later and then modify them at that point?

thyttan:
Closing this now :)

PR #3496, opened 2024-07-08
AGPS: Don't unserialize then immediately serialize

KTibow:
Right now we're decoding the base64 for no good reason, as it's immediately encoded afterwards, so this removes that.

Note that the chunk size will change - now less data will be sent each time. I wouldn't expect this to be a problem, but I haven't tested it and I might need to change the chunk size.

### DIFF
--- a/apps/assistedgps/custom.html
+++ b/apps/assistedgps/custom.html
@@ -131,7 +131,6 @@
 // ===================================================
 
       function jsFromBase64(b64) {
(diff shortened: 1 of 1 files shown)


KTibow:
Closed because I initially forgot that atob doesn't stand for ASCII to Base 64 but rather the reverse, and now I understand that it's probably needed to have proper padding and all

PR #3498, opened 2024-07-09
Add power usage module

bobrippling:
A follow-up to #3477, this adds `require("power_usage")` support

Also updates types (based on espruino/Espruino#2533), specifically for `startsWith()`

### DIFF
--- a/apps/daisy/ChangeLog
+++ b/apps/daisy/ChangeLog
@@ -11,3 +11,4 @@
 0.11: Minor code improvements
 0.12: Added setting to change Battery estimate to hours
 0.13: Fixed Battery estimate Default to percentage and improved setting string
+0.14: Use `power_usage` module
--- a/apps/daisy/app.js
+++ b/apps/daisy/app.js
@@ -116,7 +116,7 @@ function updateSunRiseSunSet(now, lat, lon, line){
 function batteryString(){
   let stringToInsert;
   if (settings.batt_hours) {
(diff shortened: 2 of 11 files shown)


bobrippling:
Blocked on a firmware bug - espruino/Espruino#2529

bobrippling:
Firmware bug closed - will hold off on merging until 2v24's released

gfwilliams:
That fix reverted because it breaks parsing `(1)/2` - I think it would be best to modify this code so it can be merged as fixing the issue properly is going to take a while.

What about changing `if(!/^(LCD|LED)/.test(key))` to just `if (!key.startsWith("LCD"))` - I could be wrong but as Bangle.js doesn't have an LED, it seems like there's no need to check it? And this code is likely faster and definitely more readable.

bobrippling:
Sounds good, sorted - I'll test this for a little while and merge in a few days. Bangle.js has both `LED1` and `LED2`, but I doubt anyone's using them since they're just shown on the LCD, so it's probably not important

bobrippling:
@Poolitzer I'll merge later today unless you've seen any issues?

Poolitzer:
Yes looks good to me. For the end user it might be better to have it not that jumpy between screen on/off, but I feel like this is either a bigger addition to the module or smth the apps should take care of since they know in which intervals this gets called.

But thanks for putting it in a module and in the daisy app way better.

PR #3500, opened 2024-07-10
settings: disable fastload

bobrippling:
This prevents issues with not loading settings on exit, or jumping into other apps such as the recorder, while not unloading settings.

See also https://github.com/espruino/Espruino/issues/2530


gfwilliams:
Thanks, but please see my reply on https://github.com/espruino/Espruino/issues/2530 - I really don't want hacks put in all the different apps to deal with `fastload` when the `fastload` itself could fix the issues it causes

bobrippling:
Thanks - I'll close this for now while we discuss over there

PR #3503, opened 2024-07-12
widbattpwr: show percentage if charging

bobrippling:


bobrippling:
Tested, no problems - merging

PR #3504, opened 2024-07-12
splitsw: add new stopwatch app with split timer

jt-nti:


thyttan:
Nice one, thanks!

There are a couple of timer apps now - maybe we should have a library/module for the core functionality eventually?

jt-nti:
A library/module for some shared timer functionality definitely sounds like a good idea.

bobrippling:
Not sure if it's just me - I notice I don't get any backlight in this app, perhaps it's because the timout is set, so if the backlight is off it never turns on?

jt-nti:
That's odd, the backlight stays on for me, which is what I expected from http://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setLCDTimeout

I was already thinking of adding some options to override the Backlight and Lock timeouts separately, so that might fix the issue for you. I'll try to have a look if I get a chance.

bobrippling:
Sounds good, thank you! I'll keep my eyes peeled :)

jt-nti:
@bobrippling 
I've made a start on #3542 which may already fix your backlight issue. I'd be interested if it doesn't anyway!

I'm kind of undecided about having separate options for the screen lock and backlight, especially since my cunning plan to add support for the physical button, even when locked, isn't going too well.

I'm also wondering about better names for the options since I think even I'm confussed byt the current names! Maybe "Keep backlight on" and "Stay unlocked" would be clearer?

bobrippling:
Awesome! Yes that's a fair point, we can see how the cunning plan plays out - I have a few ideas to act on the button press even [when locked](https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_lock):
```javascript
Bangle.on("lock", (locked, reason) => {
  if (!locked && reason === "button") {
    buttonPress();
  }
})
```

Those option names sound good too, let's see how they look though

PR #3505, opened 2024-07-13
New Beep Test App

bb0x88:
Basic beep test app  - Aerobic fitness test created by Léger & Lambert


thyttan:
Hi!

The tests fail so please take a look at that.

I would suggest adding some tags to the metadata to make the app more discoverable  🙂

bb0x88 (commented):
I hope that fixes it

thyttan:
Thanks! Tried the app. But I didn't run! 🏃

PR #3507, opened 2024-07-13
SkySpy v0.03

pavelmachek:
Major rewrite to make it easier to read, add ability to auto-adjust time and altitude.


bobrippling:
Thanks!

PR #3508, opened 2024-07-13
Icon bits: initial import

pavelmachek:
Iconbits is simple bitmap editor, that can be used to edit fonts and icons for B.js2.


thyttan:
Thanks!

I'd suggest that the readme elaborated a little more on how you use it and for what you use the app. E.g. what do you do with an icon once drawn?

Also, I'd prefer if the filename `iconbits.app.js`  was renamed to just `app.js` and renamed when loaded to the watch. Not a requirement, but most other apps and the tutorials/examples do it that way.

pavelmachek:
Could we keep the filename? Having all the apps called app.js is very confusing when working with them in the emulator etc. I did it before, but its.. not ok.

And yes, I plan better README and color support and more for 0.02. If you want to test it, suggested way is to run it in the emulator, and press the button. But it would be easier for me if 0.01 was merged first, pretty please :-).

thyttan:
> Could we keep the filename? Having all the apps called app.js is very confusing when working with them in the emulator etc. I did it before, but its.. not ok.

I think it's easy enough to rename to `iconbits.app.js` when you load/write the app to the emulator, and consistency in the main repository is nice. But I wont force you. 🙃

> **I plan better README and color support and more for 0.02. If you want to test it, suggested way is to run it in the emulator, and press the button.**

Can you just copy/paste that into the readme please. That would clear up much of the confusion I think.

pavelmachek:
Done. Will need to create better cmdline interface and real documentation.

thyttan:
Thanks

PR #3511, opened 2024-07-15
Draft: fix(measuretime): add 3d effect, align numbers correctly

prefectAtEarth:
- add outline for 3d effect
- align numbers so they don't overlap with lines


thyttan:
If the appearance was changed maybe you want to update the screenshot in the readme as well?

PR #3513, opened 2024-07-15
Add support for reply lib to messagegui app

WhelanB:
Adds support for the canned responses/reply lib to the default messagegui app.

By clicking on a notification to head to the notification settings, the user will see a "reply" menu item if the message can be replied to, which will trigger the reply library flow

### DIFF
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -24,7 +24,8 @@ require("messages").pushMessage({"t":"add","id":"call","src":"Phone","title":"Bo
 var Layout = require("Layout");
 var layout; // global var containing the layout for the currently displayed message
 var settings = require('Storage').readJSON("messages.settings.json", true) || {};
-var reply = require("reply");
+var reply;
+try { reply = require("reply"); } catch (e) {}
 var fontSmall = "6x8";
 var fontMedium = g.getFonts().includes("6x15")?"6x15":"6x8:2";
 var fontBig = g.getFonts().includes("12x20")?"12x20":"6x8:2";
@@ -269,7 +270,10 @@ function showMessageSettings(msg) {
     /*LANG*/"View Message" : () => {
       showMessageScroller(msg);
     },
-    /*LANG*/"Reply": () => {
+  };
+
+  if (msg.reply && reply) {
+    menu[/*LANG*/"Reply"] = () => {
       replying = true;
       reply.reply({msg: msg})
         .then(result => {
@@ -281,14 +285,15 @@ function showMessageSettings(msg) {
           replying = false;
           showMessage(msg.id);
         });
-    },
+    };
+  }
+
+  menu = Object.assign(menu, {
     /*LANG*/"Delete" : () => {
       MESSAGES = MESSAGES.filter(m=>m.id!=msg.id);
       checkMessages({clockIfNoMsg:0,clockIfAllRead:0,showMsgIfUnread:0,openMusic:0});
     },
-  };
-
-  if (!msg.reply || !reply) delete menu[/*LANG*/"Reply"];
+  });
 
   if (Bangle.messageIgnore && msg.src)
     menu[/*LANG*/"Ignore"] = () => {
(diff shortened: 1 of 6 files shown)


thyttan:
@soneret was eager to try this in https://github.com/espruino/BangleApps/pull/3473#issuecomment-2227525727.

If I'm correct this update can be installed from https://whelanb.github.io/BangleApps/?id=messagegui , right @WhelanB ? :)

I want to try it some myself and we should wait for @gfwilliams to be back so he can chime in as well before we merge.

WhelanB:
> @soneret was eager to try this in [#3473 (comment)](https://github.com/espruino/BangleApps/pull/3473#issuecomment-2227525727).
> 
> If I'm correct this update can be installed from https://whelanb.github.io/BangleApps/?id=messagegui , right @WhelanB ? :)
> 
> I want to try it some myself and we should wait for @gfwilliams to be back so he can chime in as well before we merge.

Yes! it's hosted there now if you want to point Gadgetbridge to that link :) Let me know if you hit any issues running it!

apps/messagegui/app.js
```diff
@@ -267,11 +269,27 @@ function showMessageSettings(msg) {
     /*LANG*/"View Message" : () => {
       showMessageScroller(msg);
     },
+    /*LANG*/"Reply": () => {
```
gfwilliams, line 272: Please could you add a comment here like `// deleted below if reply not supported` as it might be a bit confusing for others (it was for me!).

apps/messagegui/app.js
```diff
@@ -24,6 +24,7 @@ require("messages").pushMessage({"t":"add","id":"call","src":"Phone","title":"Bo
 var Layout = require("Layout");
 var layout; // global var containing the layout for the currently displayed message
 var settings = require('Storage').readJSON("messages.settings.json", true) || {};
+var reply = require("reply");
```
gfwilliams, line 27: I'm not sure if you need:

```
var reply;
try { reply = require("reply"); } catch (e) {}
```

It might be worth trying without `reply` and seeing what happens as I think it'll throw an exception which will stop the rest of the app loading

WhelanB: I hit this myself while testing yesterday and you're right, it hangs the app if reply isn't installed - will add some safety checks here!

WhelanB:
> This looks great, thanks! Just the 2 comments above. It'll be good to have the ability to do proper replies.
> 
> I guess maybe later we might even want to make swipe right reply to a message if it supports replies?

I've just pushed an update to address the PR comments, and add swipe right to reply! spotted some issues around empty replies as well in the reply lib that are now handled

apps/messagegui/app.js
WhelanB: I've changed this to use an if statement instead just so it's more in line with the surrounding logic

gfwilliams:
Looks great to me - thanks!

RelapsingCertainly:
tested out the new version and found a bug, once you receive a message and reply to it, the message popup again and can not be dismissed (as in it remains in the messages app and can not be removed without disconnecting from the phone). This may just be an android issue, im not sure

RelapsingCertainly:
Update: it seems to be an issue with how google messages handles replies via notifications, signal works fine the notification appears twice but its dismissed once I swipe left

WhelanB:
Sorry for the slow reply on this - @RelapsingCertainly is it that the notification isn't actually dismissed on your phone after replying and so stays present on the Bangle? I've mostly testing on apps like Whatsapp, where the notification is dismissed by Whatsapp after replying and so gets deleted from the Bangle, but I know some apps (like Discord, which doesn't support reply over Bangle atm) keeps the notification around, so maybe I need to handle that case

RelapsingCertainly:
When I get a text message on Google Messages, I can reply to it as usual and it sends successfully. However, the message notification repeats on my watch (and not on the phone), allowing me to reply again. Even after dismissing it on the watch, the notification remains in the Messages app.

I don’t think this is related to the recent pull since a similar issue occurred with music notifications not dismissing.

gfwilliams:
@RelapsingCertainly I'm not sure this is something others are seeing - but perhaps if you could post up a Gadgetbridge log file then we can see what's actually going on.

RelapsingCertainly:
Oh that's weird. When I tried to recreate the bug it has seemingly fixed itself. I'm so confused but looks like no fix is needed anymore :/

RelapsingCertainly:
Also thing to note: the bug also fixed itself on my "dynamic island" app so it might have been a bug from google and now its fixed

RelapsingCertainly:
@WhelanB question: are there any plans re add the ability to dismiss the notification without replying? What I mean is sending an affermative action to open the app on my phone. Kind of how its like it is when you cant reply. 

This is mostly useful when getting a notification but I dont have the time to respond but also dont want to clear the notification or have the flashing icon.

WhelanB:
In future we could add a toggle for whether you want to be able to reply or ack the message from swipe right, but I think for this PR I'll keep the behaviour limited to swipe right to reply for messages that support it

soneret:
email  and other apps reply possible e.g aliexpress 

⁣Android için BlueMail uygulamasını edinin ​

On 13 Ağu 2024 22:40 , at 22:40 , Brian Whelan ***@***.***> yazdı:

RelapsingCertainly:
@gfwilliams is there anything stoping this pull from being merged? I have been testing for a while and it seems to be ready for stable.

WhelanB:
> email  and other apps reply possible e.g aliexpress ⁣Android için BlueMail uygulamasını edinin ​ On 13 Ağu 2024 22:40 , at 22:40 , Brian Whelan ***@***.***> yazdı:
> […](#)
> In future we could add a toggle for whether you want to be able to reply or ack the message from swipe right, but I think for this PR I'll keep the behaviour limited to swipe right to reply for messages that support it -- Reply to this email directly or view it on GitHub: [#3513 (comment)](https://github.com/espruino/BangleApps/pull/3513#issuecomment-2286997152) You are receiving this because you were mentioned. Message ID: ***@***.***>

I think this would need to be handled on the Gadgetbridge side, rather than here. I'd like to take a look at at some point since it has some support for canned responses already, and doesn't support some apps like Discord currently.



> @gfwilliams is there anything stoping this pull from being merged? I have been testing for a while and it seems to be ready for stable.

Yeah, I don't think there's much more on my side for this PR, so if it's looking good would be great to get this merged!

gfwilliams:
Thanks - yes, this looks good to me. Merging now!

Sorry for the delay, I've been on/off holiday a bit the last few weeks

RelapsingCertainly:
is it just me or is the updated version not showing up on app loader?

thyttan:
Check the Development App Loader - accessible from the "More..." tab of the official App Loader. Apps/updates show up there virtually as soon as they are merged here in the repo.

Gordon updates the official App Loader manually with some delay so we get a chance to fix instabilities if they appear on the development one.

PR #3515, opened 2024-07-16
feat(measuretime): visual improvements

prefectAtEarth:
- add 3d effect
- fix number alignment
- update screenshots for ReadMe
- new app icon


thyttan:
Thanks! 

Personally I think I like it a little more without the new vertical lines especially on dark theme as there's then a white line between the black watch body and the black app background. Just my preference!

prefectAtEarth:
@thyttan 
Thanks for your feedback. I plan to include several settings parameters for the watchface. The vertical lines could be one setting (on/off).

PR #3516, opened 2024-07-17
New app: Timestamp Log

ticalc-travis:
This is my first Bangle.js 2 app. Don't know how useful it would be to others, but I personally find it handy for things like keeping track of my wake/bed times.


thyttan:
Thanks! I'll take a look tonight.

It's similar to https://banglejs.com/apps/?id=chronlog which I created, but yours has some niceties mine doesn't. 🙂

thyttan:
Thanks!

PR #3519, opened 2024-07-20
New clock: Quarter Clock

retcurve:
Clock showing the time to nearest 15 minute period using boxes. Settings allow for changing colours, showing/hiding battery level box, showing hiding widgets, and a digital mode replacing the hour box with a number.


bobrippling:
Nice, thanks!

PR #3520, opened 2024-07-24
ffcniftyapp: Fix weather icon on non-English locale

Athemis:
If other languages than English are used, comparing the weather description to determine the appropriate weather icon fails. This pull request alters this behaviour:
  1. Check for the current locale
  2. Use weather code if a language other than English is used instead of description to choose icon

The code in question was borrowed and adapted from the weatherClock app.


bobrippling:
Nice addition, thanks

PR #3524, opened 2024-08-01
Fix size of widget overlay

sir-indy:
The widget overlay was enlarged from 24px to 26px high, to add a contrast bar at the bottom. Most of the '24' in the code were not changed to 26. In use this meant that it did not fully clear off the screen. Replaced the '24' with exports.buffer_h = 26, so they are the same size, and it clears properly.

Not sure if `exports` is the correct thing here, or if it should just be `let`?

### DIFF
--- a/modules/widget_utils.js
+++ b/modules/widget_utils.js
@@ -1,4 +1,5 @@
 exports.offset = 0;
+exports.buffer_h = 26;
 exports.hide = function() {
   exports.cleanup();
   if (!global.WIDGETS) return;
@@ -29,7 +30,7 @@ exports.show = function() {
 
 /// Remove anything not needed if the overlay was removed
 exports.cleanupOverlay = function() {
-  exports.offset = -24;
+  exports.offset = -exports.buffer_h;
   Bangle.setLCDOverlay(undefined, {id: "widget_utils"});
   delete exports.autohide;
   delete Bangle.appRect;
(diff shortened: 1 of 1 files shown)


bobrippling:
The code looks correct but I'm not aware of the 24->26px change, do you have a reference for it?

sir-indy:
It's here, widget_utils.js line 73:
https://github.com/espruino/BangleApps/commit/6bdb7dc0ad5312c0214f0ea7f058b46a007cc2c2

I thought it was something I'd broken at first, but tried it in the emulator and it has the same problem. Screenshots from the watch don't show the overlay, but the emulator does, so, with the pebble clock and module without these changes:
Before:
screenshot (1)
Pull down widgets:
screenshot (2)
Widgets don't fully fold back up:
screenshot

With these changes (tested using the module from my personal app loader in the emulator instead), the widgets fold all the way back up off the screen.

bobrippling:
Gotya, thanks for the details - I'll tag @gfwilliams in with this being a central app change

gfwilliams:
I'm not 100% sure this was the problem - yesterday I noticed this too and put a fix in - please can you check with the latest widget_utils and see if you still get the problem?

It had been caused by a 'fix' put in in may - there was an `if (o>-24) {` added for some reason that wrapped around some other code including another `if (o>-24) { .. } else ...` and meant that the `else` was never executed and the widget overlay was never removed: https://github.com/espruino/BangleApps/commit/9984914a91d4d06fe194d2ff980d4783c7909128

I'd have hoped the linter would have picked that up really.

So it's likely that there was a small 'jump' in the animation caused by the 2px border, but the issue wasn't that the widgets weren't moved offscreen, it was that the overlay was never actually removed.

Leaving the overlay set and even offscreen still really impacts graphics performance - writing to the screen can't take the 'fast path' and I don't think even partial updates are done, so it's likely having it left on would impact battery life/animation speed.

sir-indy:
Thanks Gordon, looks like that did the trick! Just tried in the emulator and it's happy again.

PR #3525, opened 2024-08-01
slopeclockpp: avoid reusing background colour for clkinfo

bobrippling:
@eyecreate @myxor @hughbarney what do you think of this? I feel like there might be a better way to ensure the clock info is always visible


bobrippling:
Improved the fix, much better. Will merge after a few to do a bit of testing and to give you chance to take a look

bobrippling:
Been working great, merging

PR #3526, opened 2024-08-02
fix(measuretime): hour calculation fix on day-end overlapping hours

prefectAtEarth:
- calculate hour number
- make sure to not overflow the max hours/day


bobrippling:
Looks good, thanks!

PR #3527, opened 2024-08-06
Add optional date display to Lato clock

Septolum:
Adds an optional date display above the clock, with the format choices of "DoW, dd MMM" (as seen in the new screenshot), and locale based (short and long)


bobrippling:
Nice, thanks!

PR #3528, opened 2024-08-06
gipy: display track on openstmap

nxdefiant:
adds option to plot openstmap if installed, similar to recorder. Code heavily copied from openstmap.

Notify @wagnerf42


pavelmachek:
Should the code go to openstmap module to make sharing easier? Also fixing or disabling warnings may be good.

nxdefiant:
ok the "heavily copied from openstmap" was not very precise from me, it is actually mixed recorder and openstmap, also with changes, so it is not a 1:1 copy, you will not find many identical lines in succession.

There are also some obvious differences, for example in openstmap there is a call to stop the continous updating of the track which has been left out here. In short: I would not bother sharing code here.

bobrippling:
LGTM but there's some build errors that're blocking this :)

nxdefiant:
I have not changed these lines

thyttan:
I have sometimes just added some let-statements near the beginning of the script as an easy way to make the linter happy with older code. In this case e.g:
```
let direction, c;
```

atjn:
I can fix the warnings if you want me to, but you need to provide me with a sample GPX file. Without it, the app never loads. I tried to download a few from different websites, but they all break the upload script with different errors.

nxdefiant:
You need a GPX-track file, not route. I have created a test file from https://www.bbbike.org/. Both the .gpx and the .gps are in [test.zip](https://github.com/user-attachments/files/16621578/test.zip)

atjn:
Thanks for the file. I made the fix in #3533 so you can merge that first, then merge that afterward.

PR #3529, opened 2024-08-07
locale: Support custom date and time formats, save the last uploaded locale

atjn:
I have been using these features a lot on my local server and I really like them, so I thought I would try to upstream them.

### Feature 1: Support custom date and time formats

I have added  the option "Advanced: Customize the date and time formats.". When you enable it, you get some input boxes that allow you to change the  date and time formats before uploading the locale to your watch. I think this is really useful, as people often have personal preferences that don't align with the default format for their country. I am at least one of those people.

image

As you can see on the screenshot, I have added a few suggestions for each input that appear in a drop down when you click the input. This will hopefully help users a bit, but if this PR is accepted, I also want to write a tutorial that describes the formats in more details.

### Feature 2: Save the last uploaded locale

Most people have a single locale that they will always use, and it can be annoying that every time you open `locale`, it defaults to `en_GB`. This feature remembers which locale you last uploaded to your watch, and then chooses that locale by default for you.

If you uploaded a custom locale, it will create a new special locale so you don't have to recreate it. If I fx created a custom locale based on `fr_FR`, it will be displayed like so:

image

### DIFF
--- a/apps/locale/locale.html
+++ b/apps/locale/locale.html
@@ -352,42 +352,90 @@ ${customizeLocale ? `<tr><td class="table_t">Meridian names</td>
         return getLocaleModule(false);
       }
 
+      const lastUploadedLocaleID = "last-uploaded-locale";
+      let lastUploadedLocale;
+      try{
+        lastUploadedLocale = JSON.parse(localStorage?.getItem(lastUploadedLocaleID));
+      }catch(error){
+        console.warn("Unable to load last uploaded locale", error);
+      }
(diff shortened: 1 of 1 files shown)


bobrippling:
Very cool! I'd be a fan of this, one of the first things I wanted to do was customise the locale. Tagging @gfwilliams since this is a centralish app

gfwilliams:
Looks great! Something I fancied too. I don't have time to test, but as long as you're happy this works ok I'd love to have this in

PR #3531, opened 2024-08-10
Fix promenu behaviour: scroller & image strings

bobrippling:
This fixes promenu to:
- return a `scroller` (permitting it to work with the alarm app)
- avoid calculations on menu items with image strings, assuming the app has figuring this out already


bobrippling:
No problems testing, merging

PR #3537, opened 2024-08-16
antonclk: Handle French time format

atjn:
Now that custom time formats are allowed, you can expect the French time format to be used. Instead of splitting hours and minutes with `:` it uses `h`, sometimes with spaces around it. 

If you try to use the default Anton Clock with this format, it breaks because the Anton font doesn't include "h". The font size is also too big to support anything wider than `:`  between the numbers.

The French format is niche enough that it is probably not worth the extra work and app size to display it correctly, but it is used enough that I think the default app should be able to handle it. My solution is to detect when the French format is being used and replace it with the standard `:` instead.

### DIFF
--- a/apps/antonclk/ChangeLog
+++ b/apps/antonclk/ChangeLog
@@ -14,3 +14,4 @@
 0.10: Use Bangle.setUI({remove:...}) to allow loading the launcher without a full reset on 2v16
 0.11: Moved enhanced Anton clock to 'Anton Clock Plus' and stripped this clock back down to make it faster for new users (270ms -> 170ms)
       Modified to avoid leaving functions defined when using setUI({remove:...})
+0.12: Clock can understand French time formats
--- a/apps/antonclk/app.js
+++ b/apps/antonclk/app.js
@@ -15,6 +15,7 @@ let draw = function() {
   g.reset().clearRect(Bangle.appRect); // clear whole background (w/o widgets)
   var date = new Date();
   var timeStr = require("locale").time(date, 1); // Hour and minute
+  timeStr = timeStr.replace(/ *h */i, ":"); // Don't use French time format, it is not supported by the Anton font
   g.setFontAlign(0, 0).setFont("Anton").drawString(timeStr, x, y);
   // Show date and day of week
   var dateStr = require("locale").date(date, 0).toUpperCase()+"\n"+
--- a/apps/antonclk/metadata.json
+++ b/apps/antonclk/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "antonclk",
   "name": "Anton Clock",
-  "version": "0.11",
+  "version": "0.12",
   "description": "A simple clock using the bold Anton font. See `Anton Clock Plus` for an enhanced version",
   "icon": "app.png",
   "screenshots": [{"url":"screenshot.png"}],
--- a/apps/antonclkplus/ChangeLog
+++ b/apps/antonclkplus/ChangeLog
@@ -14,3 +14,4 @@
 0.10: Use Bangle.setUI({remove:...}) to allow loading the launcher without a full reset on 2v16
       Modified to avoid leaving functions defined when using setUI({remove:...})
 0.11: Minor code improvements
+0.12: Clock can understand French time formats
--- a/apps/antonclkplus/app.js
+++ b/apps/antonclkplus/app.js
@@ -140,6 +140,7 @@ let draw = function() {
   g.clearRect(0, 24, g.getWidth(), g.getHeight() - 24); // clear whole background (w/o widgets)
   var date = new Date(); // Actually the current date, this one is shown
   var timeStr = require("locale").time(date, 1); // Hour and minute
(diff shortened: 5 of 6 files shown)


bobrippling:
Thanks for the fix - going to tag @gfwilliams with this being a central app

bobrippling:
I wonder if there's another way to get around it, rather than having the clock know about specific locales?

atjn:
> I wonder if there's another way to get around it, rather than having the clock know about specific locales?

I would be eager to hear any ideas.

I have considered a more general-purpose regex that would be able to extract the hour and minute number regardless of what text was put before/between/around them. I decided against that because it would be adding a lot of extra compute time for everyone using the clock, and 99% of users would not need it.
The regex in this PR is relatively low in compute time because it won't find a match for space or "h" anywhere in a normal time format, so it can quickly reject and move on.

I also considered adding a fallback that would use a smaller system font if it encountered an unsupported time format. But I eventually decided that the original font and design is more important than having the ":" replaced with an "h". You could still add a fallback to catch other weird time formats if you wanted to, but I don't know if it is worth it.

gfwilliams:
Ok, thanks, I'm not sure I really understand though. According to the locales file, the french time format is:

```
timePattern: { 0: "%HH:%MM:%SS", 1: "%HH:%MM" },
```

In fact I'm not sure I see a single time pattern that isn't `%HH:%MM`? Where is `h` being used?

The regex here doesn't have any real negative impacts, but I'd like to know where `h` comes from - because this'll still fail if someone decides to use `H` or any other character, and I think there are a bunch of other clocks that will have issues too, especially all the clocks that split on ':' : https://github.com/search?q=repo%3Aespruino%2FBangleApps%20split(%22%3A%22)&type=code

Maybe it should actually be doing `timeStr.replace(/[^\d]+/,":")` to just replace any non-digit with ':'.

But I feel like `locate.time(, 1)` should always return a minimal string for time. Having it return `00 h 00` when so far every clock is expecting something else seems like the issue here really...

atjn:
> In fact I'm not sure I see a single time pattern that isn't %HH:%MM? Where is h being used?

I am talking about a French user deciding to customize their time format to use `h` instead of `:`. This is possible now that #3529 has been merged.

I tried this myself because I am learning french, and I thought that at least the default app should be able to deal with it. It is correct that many other apps also don't work with it.

gfwilliams:
> I am talking about a French user deciding to customize their time format to use h instead of :. 

Ok, then I think we definitely shouldn't be merging this I'm afraid. Because if anyone can customise that field to anything, they can break the clock instantly.

At some point if a user makes a change away from default and that breaks it, that is up to them. What if I want to change to `12 OMG!??!?? 34`? Do we have to deal with that too?

Or do we now say that the text from `locale` is untrusted and every clock should just implement it's own method of creating a string from the time and we're in a worse position than before?

You were asking about putting the locale 'sanity check' in a separate file? I feel like the locale editor in https://github.com/espruino/BangleApps/pull/3529 should load up that sanity check and run it on the users' input and display a warning. And maybe we add a check that the 'short' time contains a `:` to that list? I think that would go a long way towards helping

atjn:
Alright that is your call.

> At some point if a user makes a change away from default and that breaks it, that is up to them. What if I want to change to 12 OMG!??!?? 34? Do we have to deal with that too?

No that is ridiculous and you can't expect clocks to work with that. But in my opinion, it would be fair to expect clocks to deal with real and widely used clock formats such as the french "h". After all, isn't that the entire point of the `locale` package?

> feel like the locale editor in https://github.com/espruino/BangleApps/pull/3529 should load up that sanity check and run it on the users' input and display a warning.

Yeah, I've considered that too. Will see what I can do.

gfwilliams:
My point is there's no sanity checking, so at this point someone might expect that whatever they entered would be displayed if it is accepted.

Realistically if you enter `12h34` as the format, but all the clocks then have to have code to detect that and convert back to `12:34` that's not useful either since what was the point of specifying it anyway?

Unfortunately we can't pick and choose exactly what chars go in the fonts right now so to support `h` we'd need all upper and lowercase chars as well as just the numbers - realistically that brings the number of chars from 16 to 64 or so. 4x the memory, and extra slowdown, just in case someone enters `12h34` - I think it's fair to say that disadvantages at least 99% of users just for under the 1% that might do it.

Best bet is add a warning to the locale sanity checker to say `many clock faces may not support 'h' in the 'short time'`, and if you want `h` displayed, make your own clock face (there may be one already?)

atjn:
When I tested it, there were a bunch of clocks that supported the `h` format.

I am a little unsure about how exactly to format that warning. I can see that you added the `wae_CH` locale, which uses a `.` instead of `:`. This also breaks a lot of apps, but it is supported in the deafult anton clock. So should my check warn about that format or not?

gfwilliams:
> I can see that you added the wae_CH locale, which uses a .

Well, not me, but at some point it was added

I think right now, your check probably should warn, yes. Maybe:

```Some clocks may not function correctly if the time separator is not ":"```

Longer term it'd be nice if we could stop clocks using `locale.time(...).split(":")` and similar - maybe adding a module that provided the functionality (since the only reason it's done is to avoid duplicating the 24h check)

PR #3538, opened 2024-08-16
drained: make it clear when charging, via the clock

bobrippling:


pavelmachek:
Would it make sense to add text "charging" somewhere on screen, too? People are not going to see the app too often, and may not know why the clock is now yellow...

bobrippling:
Sounds good, let me know what you think

pavelmachek:
Thank you, looks good to me.

bobrippling:
Great, thanks for the review

PR #3540, opened 2024-08-19
Added OpenLocate Beacon app

jeffyactive:
Tested successfully on Bangle.js 2.


apps/openlocatebeacon/metadata.json
```diff
@@ -0,0 +1,19 @@
+{
+  "id": "openlocatebeacon",
+  "name": "OpenLocate Beacon",
+  "shortName": "OpenLocate Beacon",
+  "version": "0.03",
+  "description": "Advertise GPS geolocation date using the OpenLocate Beacon packet specification.",
```
pavelmachek, line 6: "data"?

apps/openlocatebeacon/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+The advertising packets will be recognised by [Pareto Anywhere](https://www.reelyactive.com/pareto/anywhere/) open source middleware and any other program which observes the standard packet types.  See our [Bangle.js Development Guide](https://reelyactive.github.io/diy/banglejs-dev/) for details.
+
+
+## Features
+
+Advertises packets with the OpenLocate Beacon geolocation element when a GPS fix is available, and packets with the name "Bangle.js" otherwise.
+
+
+## Controls
```
pavelmachek, line 16: You may want to reduce number of blank lines. Empty "controls" section can likely be deleted, too.

apps/openlocatebeacon/metadata.json
jeffyactive: Good catch, corrected. :+1:

apps/openlocatebeacon/README.md
jeffyactive: Removed controls section and extra newlines, as requested.

pavelmachek: Thank you! :-)

apps/openlocatebeacon/openlocatebeacon.js
```diff
@@ -0,0 +1,123 @@
...
+  return { integer: integer, fraction: fraction };
+}
+
+
+// Update barometer
+Bangle.on('pressure', (newBar) => {
+  bar = newBar;
+
+  mainMenu.Altitude.value = bar.altitude.toFixed(1) + 'm';
+  E.showMenu(mainMenu);
```
thyttan, line 96: This redraws the main menu in the top position constantly so it's virtually impossible to scroll down in the menu.

thyttan: Yes, so like this:

> On Bangle.js there are a few additions over the standard graphical_menu:
> 
> The options object can contain:
> - back : function() { } - add a 'back' button, with the function called when it is pressed
> - remove : function() { } - add a handler function to be called when the menu is removed
> - **(Bangle.js 2) scroll : int - an integer specifying how much the initial menu should be scrolled by**
> 
> The object returned by [E.showMenu](https://www.espruino.com/ReferenceBANGLEJS2#l_E_showMenu) contains:
> - (Bangle.js 2) scroller - the object returned by [E.showScroller](https://www.espruino.com/ReferenceBANGLEJS2#l_E_showScroller) - scroller.scroll returns the amount the menu is currently scrolled by

*https://www.espruino.com/ReferenceBANGLEJS2#l_E_showMenu*

jeffyactive: Good points @thyttan and thanks for the reference to the E.showMenu options.  The simplest option was to reduce the number of menu items to avoid the need for scrolling, which I've done.  I've also shortened the app name in the menu title so that it doesn't overflow.

bobrippling:
Nice app! I reckon we can merge this if you're happy @thyttan ?

bobrippling:
Oh no worries, I can wait if you want to give it a whirl :)

thyttan:
Tried it now - seems to work well! 

Two notes:
- The icon background on the watch is solid white so looks a little weird on dark them - but no biggie.
- It's still a little janky with the menu since it's able to scroll a little and then jumps back. Also no biggie.

Merging!

PR #3541, opened 2024-08-23
pomoplus: increase size of fonts and buttons

thyttan:
for legibility and ease of use.

Changes made primarily for the Bangle.js 2. I haven't run them on an actual Bangle.js 1 watch - only the emulator. The update can be tried here (please report back if you try!): https://thyttan.github.io/BangleApps/?id=pomoplus

@bruceblore @notEvil thoughts?

Edit: After measuring ram use it think it still fits on the Bangle.js 1.

### DIFF
--- a/apps/pomoplus/ChangeLog
+++ b/apps/pomoplus/ChangeLog
@@ -2,3 +2,4 @@
 0.02-0.04: Bug fixes
 0.05: Submitted to the app loader
 0.06: Added setting to show clock after start/resume
+0.07: Make fonts and buttons larger for legibility and ease of use.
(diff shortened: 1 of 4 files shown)


pavelmachek:
Unrelated to this, I noticed that I don't know what pomoplus is and what it should be doing. Adding README would be quite welcome.

notEvil:
I tried it on Bjs1 using the emulator and looks good to me, with `const SCALING = W/g.getWidth()`.

thyttan:
> I tried it on Bjs1 using the emulator and looks good to me, with `const SCALING = W/g.getWidth()`.

That will undo the scaling though - it will evaluate to 1 and not do anything. Did it not look good with `= W/176`?

notEvil:
True, silly me. Didn't know what its supposed to look like and preferred a little more spacing between things. But thats just a matter of taste, don't feel obliged to change, especially because I'm bad at visual design in general :)

thyttan:
Thanks @pavelmachek and @notEvil for your feedback. I'll merge this now.

PR #3543, opened 2024-08-24
New App: Binary LED Clock (BLC)

aemkai:
A clean binary watch face with leds showing time and date


thyttan:
Hi!

The app seems to work fine including not leaking memory - although I don't possess the knowledge to verify if the time is right ;)

But we're missing a changelog as the sanitychecker says. So add that and then I think we can merge.

If you could translate the comments in the code to english that would be preferable as well - but I wont force you.

Thanks for the PR! :)

aemkai:
done

thyttan:
Thanks!

Now that the linter gets to run it has some complaints of variables assigned but not used - either delete those or comment them out.

I noticed now that some variables are named in german as well. Same as with the comments it would be nice if those were in english also. Sorry for not noticing before. (won't block a merge though)

aemkai:
Okay, I changed color- and date-names to English, commented unused variables out and did a little simplification at msbits-array.

thyttan:
Nice - thanks! I'll just run it once on my watch and merge after that :)

thyttan:
Thanks again - merging now!

aemkai:
Fine :-)

PR #3548, opened 2024-08-28
tv remote app

Guptilious:
App allows the user to send tv remote input commands to their tv. Currently support is only for panasonic tv's as that's what I had to work with.

I have done my best how to use and navigate app pages. The app would require the user to create a webserver to receive requests. I have detailed and link to my version.

I've done what I can to make the install user friendly but appriciate this may go beyond the scope.


thyttan:
The linter complains about:
- Some variables are never used - either delete them or comment them out.
- some variables are not defined with var, let, or const. Add that where it complains.

If your on a desktop you can see the warnings in the 'files' tab of the PR now that the linter has run.

Guptilious:
@thyttan this is my first time doing a public repo, so let me know if anything I'm doing is wrong for this.

When making changes do I just amend my code and hit another merge request or do I need to do something else as this one is open?

Based on what's being flagged:

Variables and functions I can comment out:
- drawTImeout
- samsungmenu
- prevNumber

The below variables it says are never used, are used by the app, but they have to be uploaded by the user. After that they will only run when a http request is sent. 
- serverPort
- tvIp

The rest state aren't always defined and have checks for that and respond on that.  
- checkValue
- IPASSIGN
- samsIp

There is also an issue stating I'm using a deprecated part of Node but I'm not sure what action I need to take here.

thyttan:
I'm out moving around - I'll check in again in a bit.

But if you make changes on your master branch they will automatically show up here. 🙂👌

Guptilious:
Awesome. 

I've hunted through and think I've manged to resolve the errors that are getting kicked out.

Everything has now been commited back to the master branch, so... 🤞🏼

thyttan:
If I install the app from your app loader and try to run it without doing the setup steps in the readme I get this:

```console
Uncaught Error: Can't read property 'webServerDns' of undefined
 at line 2 col 616 in tvremote.app.js
...me,true);let serverDns=serverData.webServerDns;let serverPort=serverD...
                                       ^
> 
```

This then renders the watch mostly unresponsive until I long press the hardware button to reset to the clock face.

I think we should handle this before merging here. So either some error handling or making sure we don't end up with a breaking error in the first place.

I think if the user didn't do the manual part of the setup detailed in the readme then the app should display a hint message to go back and read the readme.

Another way to solve it could be to add a customiser for the app on the app loader entry where the user types in the required data (https://www.espruino.com/Bangle.js+Custom). A little more involved maybe.

Guptilious:
I've added an 'if' statement to the end of the app, which should provide an Alert to the user if the config file isn't present. Clicking 'ok' should then take you to the main app without an error.

Adding a customiser is definetly a cleaner approach for the user long term but I've never worked with html before, so I'd need a bit more time to figure how to include this. That said, the changes made should solve the problem for now but let me know if something else needs tweaking.

apps/tvremote/app.js
```diff
@@ -0,0 +1,655 @@
...
+
+  if (currentScreen in screenActions) screenActions[currentScreen]();
+}
+Bangle.on('touch', touchHandler);
+
+Bangle.loadWidgets();
+Bangle.drawWidgets();
+
+if (serverData === undefined) {
+  E.showAlert(`${serverDataFile}.json missing.\nSee READ.me`, "Config Error").then(function() {
```
thyttan, line 650: download

The alert body gets a little long. I suggest:


```suggestion
  E.showAlert(`No settings.\nSee READ.me`, "Config Error").then(function() {
```

thyttan:
As seen in the image below some of the buttons in the app loader are misaligned - @gfwilliams do you know what to do to fix that?

IMG_20240912_091054

gfwilliams:
Argh, well, what happens I think is the buttons are in a `tile-action` class which is basically its own column top-bottom. So if the main content is too wide (because of a big URL that's not being split up) it pushes the icons off. I've just added something to main.css to tell it to wrap *any* text to make it fit so that seems to fix it.

... but I feel like probably it'd be better if the buttons were actually part of the title only, so didn't take up real-estate that could be used for wrapping the description.

apps/tvremote/app.js
Guptilious: happy with the amendment. I wasn't sure how detailed to be with the alert. But if I can get the customiser working, this will be sutable for the interim

thyttan:
Thanks for the app and the tweaks!

Guptilious:
No problem :) 

On a side note I realised that I forgot to add my name to the sourcecode.

Is there a preferred way you guys have users add that stuff? And in the metadata etc? I can then submit another 'non-code change' merge request

thyttan:
Ah yes, we usually have an entry at the bottom of the readme stating creator/contrubutors. If  you want to you can add a comment at e.g. the top of the app.js file and similar as well - but that's not very prevalent for other apps. :)

PR #3549, opened 2024-08-29
trail: initial import.

pavelmachek:
Initial import of GPX-following application.


bobrippling:
Nice! Gonna give it a try tomorrow

apps/trail/gpx2egt.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+grep "trkpt lat" | sed 's/.*trkpt.lat=.//' | sed 's/. lon=./ /' | sed 's/".$//'
```
bobrippling, line 2: In case the `<trkpt>` has extra attributes, this would pick out just lat/lon - will leave it up to you though :)

```suggestion
grep "<trkpt" | grep -Eo '(lat|lon)="[^"]*"' | tr -d '"' | sed 's/.*=//' | paste -d' ' - -
```

pavelmachek:
You can try it without gps access by pressing bottom right corner ("demo mode"). If something goes wrong outside, press top right corner ("recover").

apps/trail/gpx2egt.sh
pavelmachek: Yes, better gpx parsing would be cool, I'm sure there's good library somewhere. Ideally GPX would be uploaded from web interface, but command line is just easier for me.

Feel free to merge either version, so I can merge it and continue development.

bobrippling: Not a problem - we can leave this for a potential html version in the future

bobrippling:
Thanks!

PR #3550, opened 2024-08-29
sixths: create version 0.10

pavelmachek:
acknowledge commands by	vibration and message
ui tweaks -- bigger font, compressing uninteresting info
display	meters up/down
display	pressure trend
adjust GPS on/off algorithm for more reliable fix
display	warnings when GPS altitude does	not match baro


bobrippling:
Nice, thanks!

PR #3551, opened 2024-08-29
gbmusic: Upgrade to new translation system

atjn:
This removes the deprecated translation system and adds the new translation system.

I have intentionally removed the translation of the `< Back` string because it will be rendered as a red back button with no text, but if it is translated, the red button disappears and a new entry is added to the menu with the translated version of `< Back`. That is not very nice.


thyttan:
Thanks!

PR #3535, opened 2024-08-15
Improve support for meridians

atjn:
Fixes problem 1 and 4 in #3534.

PR #3552, opened 2024-08-29
locale: Release new version

atjn:
I forgot to release a new version when updating `locale` in #3535 and #3547


thyttan:
Ah - nice. Thanks!

PR #3553, opened 2024-08-31
more_pickers: draw sections independently for faster execution

paul-arg:
this increases execution tremendously and gives the picker a much nicer feel


thyttan:
Looking at the code that seems to make sense - merging. Thanks!

PR #3556, opened 2024-09-03
Support Gadgetbridge Canned Responses

WhelanB:
I realised during development of the "reply" library that Gadgetbridge contains support for setting canned responses/replies directly through the app, which is preferable to customising via ``interface.html`` as it can support emojis/Unicode using text as bitmaps.

This PR consists of 2 main changes:

- Add a "disp(lay)" field to the ``replies.json`` file used by the reply lib. If display is set for a particular canned reply, it will be shown on the menu instead of the default "text" field, which is what is sent to Gadgetbridge as a response. This way, the "disp(lay)" field shown on the watch can contain image representations of emojis and other Unicode characters, while the reply to Gadgetbridge contains the raw Unicode bytes
- Expand the Android integration app to listen for a ``canned_responses_sync`` message, and write the data contained within that message to the ``replies.json`` file used by the reply lib. I'm planning to open an additional PR against Gadgetbridge on CodeBerg to send this message when canned responses are set:

image

### DIFF
--- a/apps/reply/lib.js
+++ b/apps/reply/lib.js
@@ -49,13 +49,12 @@ exports.reply = function (options) {
 
     replies.forEach((reply) => {
       var displayString = reply.disp ?? reply.text;
-      var truncatedDisplayString = g.wrapString(displayString, 120);
+      var wrappedDisplayString = g.wrapString(displayString, 120);
       // Generally handles truncating nicely, but falls down in long runs of emoji since they count as one image
-      if (truncatedDisplayString.length > 1) {
-        truncatedDisplayString = truncatedDisplayString.slice(0,1);
-        truncatedDisplayString[truncatedDisplayString.length-1] += "...";
+      if (wrappedDisplayString.length > 1) {
+        displayString = wrappedDisplayString[0]+"...";
       }
-      menu = Object.defineProperty(menu, truncatedDisplayString, {
+      menu = Object.defineProperty(menu, displayString, {
         value: () => constructReply(options.msg ?? {}, reply.text, resolve, reject),
       });
     });
--- a/apps/reply/lib.js
+++ b/apps/reply/lib.js
@@ -46,11 +46,20 @@ exports.reply = function (options) {
         options.fileOverride || "replies.json",
         true
       ) || [];
+
     replies.forEach((reply) => {
-      menu = Object.defineProperty(menu, reply.text, {
+      var displayString = reply.disp ?? reply.text;
+      var truncatedDisplayString = g.wrapString(displayString, 120);
+      // Generally handles truncating nicely, but falls down in long runs of emoji since they count as one image
+      if (truncatedDisplayString.length > 1) {
+        truncatedDisplayString = truncatedDisplayString.slice(0,1);
+        truncatedDisplayString[truncatedDisplayString.length-1] += "...";
+      }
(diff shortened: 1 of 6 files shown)


thyttan:
Changes to android integration app, calling @gfwilliams!

apps/reply/lib.js
```diff
@@ -46,11 +46,20 @@ exports.reply = function (options) {
...
     replies.forEach((reply) => {
-      menu = Object.defineProperty(menu, reply.text, {
+      var displayString = reply.disp ?? reply.text;
+      var truncatedDisplayString = g.wrapString(displayString, 120);
+      // Generally handles truncating nicely, but falls down in long runs of emoji since they count as one image
+      if (truncatedDisplayString.length > 1) {
+        truncatedDisplayString = truncatedDisplayString.slice(0,1);
+        truncatedDisplayString[truncatedDisplayString.length-1] += "...";
+      }
+      menu = Object.defineProperty(menu, truncatedDisplayString, {
```
gfwilliams, line 58: `truncatedDisplayString` is an array isn't it? presumably you should be doing `truncatedDisplayString.join("\n")`?

gfwilliams:
Thanks! Nice idea about including emoji/etc.

I'm a bit concerned about the `truncatedDisplayString` being an array (see comment) but apart from that this looks good.

apps/reply/lib.js
thyttan: ping @WhelanB

WhelanB: Sorry for the lack of updates on this PR - I was away for a few weeks without access to my PC, and my Bangle isn't functioning correctly atm. I think I was trying a few different ways of truncating here, and ended up with some leftover code - my intention was to just take ``truncatedDisplayString[0] + "..."`` as a hacky workaround for some issues I was seeing with showMenu's ability to handle some strings

for example, the following from Gadgetbridge:

``
this is a looooooooooooooooooooooooooooooooooong message ❤️❤️❤️❤️
``

renders by default like this:
image


vs by taking the first line and adding the ellipsis:
image


Is there a recommended way to handle this kind of overflow? if not I'll update the code to just use ``truncatedDisplayString[0] + "..."``

gfwilliams: I get what you're trying to do - but my point was that `g.wrapString` returns an Array, not a string, so truncatedDisplayString is an Array. You can't define a property on an object that's an array, so Espruino converts it back to a string with commas.

```
a = []; a[[1,2,3]] = 4
=[ "1,2,3": 4 ]
```

In your case it works because you ensure it's a single-element array so no commas are added, but if the linter gets better it'll probably pick this up as an issue, and either way it's a bit confusing and if someone tries to modify it, they'll almost certainly break it. 

You could just do:

```
menu = Object.defineProperty(menu, truncatedDisplayString[0], {
```

But it felt quicker for me to make a change to fix this myself: https://github.com/espruino/BangleApps/pull/3556/commits/648ebd432132867079c6cb44f5077806ead5bd2c

Maybe you could just check that it still works ok after that?

WhelanB: All looks good here - I've also opened the associated PR against Gadgetbridge on Codeberg:
https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4182

gfwilliams:
Great, thanks!

PR #3558, opened 2024-09-05
hadash: initial release

flaparoo:
This new app provides a powerful integration with Home-Assistant.
It allows both retrieving data as well as calling services in a fully configurable menu structure - in other words, almost full control of HA from your wrist!
(I didn't find this level of functionality in existing Bangle.js apps.)


thyttan:
@bobrippling do you have HA stuff at home? I don't 🙃

atjn:
I can test this one, but I can't say exactly when I have time for it.

bobrippling:
Works a treat with my HA setup, I can trigger services such as lights and scenes, and query entity states very well.

I have a few thoughts about improving the app, like returning to the same point in the menu and not requiring the user to tap ok after they've triggered a service, but I don't think they're part of this PR :)

Happy to merge, or I can let you check it out too still @atjn

apps/hadash/jsoneditor.bundlejs
bobrippling, line 1: Do we want such a large file committing in? Perhaps we ignore this too?

apps/hadash/hadash.app.js
```diff
@@ -0,0 +1,227 @@
...
+      for (let i in entry.input[key].options) {
+        if (entry.input[key].value == entry.input[key].options[i]) {
+          idx = i;
+        }
+      }
+      if (idx == -1) {
+        idx = entry.input[key].options.push(entry.input[key].value) - 1;
+      }
+      // the setTimeout method can not be used for the "format" CB since it expects a return value - using eval instead:
+      eval('CBs["'+key+'_format"] = function(v) { return entry.input["'+key+'"].options[v]; }');
```
bobrippling, line 118: Can this be:

```suggestion
      CBs[`${key}_format`] = function(v) { return entry.input[key].options[v]; }
```

bobrippling:
Found an issue - when I trigger a service (for example), returning to the menu no longer displays the back button, leaving me stuck at that menu. The code looks ok so I suspect it may be a problem with the [menu I'm using](https://github.com/espruino/BangleApps/tree/master/apps/promenu).

Will investigate when I have some more time in the next few days

apps/hadash/jsoneditor.bundlejs
flaparoo: jsoneditor.bundlejs is required for the interface.html page.

I suppose it could be linked from a hosted location, but that could create issues when the hosted version changes (or disappears).

My preferred solution (see "To-Do" in the readme) would be a custom editor built into the interface.html page, which builds the menu structure using a UI specific for this app. But that's not a simple fix...

apps/hadash/hadash.app.js
flaparoo: I guess the ```CBs[`${key}_format`]``` part would work, but the anonymous function won't have the "key" variable available. So, we still need eval for that.

flaparoo:
> I have a few thoughts about improving the app, like returning to the same point in the menu and not requiring the user to tap ok after they've triggered a service, but I don't think they're part of this PR :)

I've added your suggestions as To-Dos in the README.

atjn:
@bobrippling I think you should just release it, I plan on trying it out sometime in the next few weeks, and if I find anything I will create an issue/PR for it then :)

bobrippling:
Sounds good, thanks for the app!

apps/hadash/hadash.app.js
bobrippling: Ah gotya! For that we can do something like:
```suggestion
      CBs[`${key}_format`] = ((key) => function(v) { return entry.input[key].options[v]; })(key);
```

apps/hadash/jsoneditor.bundlejs
bobrippling: Gotya, one to remove in the future sounds ok

flaparoo:
Your suggestion works well @bobrippling 
I need to remember that trick.

So long "eval" - nobody seems to like you...

flaparoo:
> I have a few thoughts about improving the app, like returning to the same point in the menu and not requiring the user to tap ok after they've triggered a service, but I don't think they're part of this PR :)

BTW you got me intrigued - there's already a PR in the works for both.

bobrippling:
> BTW you got me intrigued - there's already a PR in the works for both.

Great! I look forward to it :)

PR #3559, opened 2024-09-06
Update boxclk to V0.08

stweedo:
Improves performance, responsiveness, and bug fixes 
- [+] Added box size caching to reduce calculations
- [+] Improved step count with real-time updates
- [+] Improved battery level update logic to reduce unnecessary refreshes
- [+] Fixed optional seconds not displaying in time
- [+] Fixed drag handler by adding E.stopEventPropagation()
- [+] General code optimization and cleanup

### DIFF
--- a/apps/boxclk/app.js
+++ b/apps/boxclk/app.js
@@ -23,84 +23,100 @@
   let h = g.getHeight();
   let drawTimeout;
 
-  // 3. Touch and drag handlers
-  let touchHandler = function(zone, e) {
-    let boxTouched = false;
-    let touchedBox = null;
-  
-    for (let boxKey in boxes) {
-      if (touchInText(e, boxes[boxKey])) {
-        touchedBox = boxKey;
-        boxTouched = true;
-        break;
+  // 3. Event handlers
+  let eventHandlers = {
+    touchHandler: function(zone, e) {
+      let boxTouched = false;
+      let touchedBox = null;
+    
+      for (let boxKey in boxes) {
+        if (touchInText(e, boxes[boxKey])) {
+          touchedBox = boxKey;
+          boxTouched = true;
+          break;
+        }
       }
-    }
-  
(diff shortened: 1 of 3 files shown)


apps/boxclk/app.js
```diff
@@ -207,195 +212,219 @@
...
-      mode : "clock",
-      remove : function() {
-        // Remove event handlers, stop draw timer, remove custom font if used
+      mode: "clock",
+      remove: function() {
+        // Remove event handlers, stop draw timer, remove custom font
         Bangle.removeListener('touch', touchHandler);
         Bangle.removeListener('drag', dragHandler);
+        Bangle.removeListener('step');
+        Bangle.removeAllListeners('lock');
```
thyttan, line 427: This seems a little agressive to me. Could you refactor to just remove this apps listener?

stweedo: I’ve organized all the event handler functions into section 3 and revised the cleanup in setUI. Thank you for catching the lock listener since that wasn’t supposed to be a remove all. And the remove step listener was missing the function to be removed. Do these changes look sufficient?

thyttan:
Thanks! I'll merge if your happy?

stweedo:
Yes, it’s ready to merge then. Thanks!

thyttan:
Out of curiosity - what is the advantage of collecting all event listeners in an object?

stweedo:
I thought about removing them all together in one line, but I'm not sure that would work.

I can change them to individual functions to avoid iterating over the object over and over.

thyttan:
I guess it's a balance between memory usage <-> execution speed <-> minification potential?

stweedo:
I changed the event handlers to individual functions since there should be a small performance benefit with the simpler setup. Is a version bump necessary for this change?

thyttan:
I'm inclined to ask for a bump yes 🙃

PR #3560, opened 2024-09-06
New app: Dutch clock

whenlit:
A clock that tells the time the way a Dutch person might answer you if you asked them what time it is. It is configurable in a number of ways.

This is my first contribution here, please do not hesitate to tell me if I'm doing anything wrong.


apps/dutchclock/app.js
```diff
@@ -0,0 +1,267 @@
...
+    g.drawString(bottomLines.join('\n'), SCREEN_WIDTH/2, SCREEN_HEIGHT - BOTTOM_SPACING, false);
+  }
+
+  /* Show launcher when middle button pressed
+  This should be done *before* Bangle.loadWidgets so that
+  widgets know if they're being loaded into a clock app or not */
+  Bangle.setUI("clock");
+  // Load widgets
+  Bangle.loadWidgets();
+
```
thyttan, line 110: Can this be moved into the initialize function? Or is there a reason for it being here?

apps/dutchclock/app.js
```diff
@@ -0,0 +1,267 @@
...
+    // pad the date - this clears the background if the date were to change length
+    g.drawString(bottomLines.join('\n'), SCREEN_WIDTH/2, SCREEN_HEIGHT - BOTTOM_SPACING, false);
+  }
+
+  /* Show launcher when middle button pressed
+  This should be done *before* Bangle.loadWidgets so that
+  widgets know if they're being loaded into a clock app or not */
+  Bangle.setUI("clock");
+  // Load widgets
+  Bangle.loadWidgets();
```
thyttan, line 109: I'm not sure but it feels like this would fit better in the initialize function?

whenlit: Hey, that was quick! Sounds good, I thought they might reset when the reset function is called but now I see that is just the Graphics object. Thanks.

apps/dutchclock/app.js
```diff
@@ -0,0 +1,266 @@
...
+    // pad the date - this clears the background if the date were to change length
+    g.drawString(bottomLines.join('\n'), SCREEN_WIDTH/2, SCREEN_HEIGHT - BOTTOM_SPACING, false);
+  }
+
+
+  if (settings.showWidgets) {
+    Bangle.drawWidgets();
+  } else {
+    widget_utils.hide();
+  }
```
thyttan, line 114: I'd probably move this to the initialize function as well, just before or after calling your draw function.

Once drawn with `Bangle.drawWidgets` widgets take care of redrawing themselves when they need to. So forcing them to draw here becomes unnecessary overhead. :)

apps/dutchclock/app.js
```diff
@@ -0,0 +1,266 @@
...
+  const m = (date.getHours() * MINS_IN_HOUR + date.getMinutes()) % MINS_IN_DAY;
+
+  if (m !== mins) {
+    mins = m;
+    draw();
+  }
+
+  if (!settings.showWidgets) {
+    widget_utils.hide();
+  }
```
thyttan, line 84: You don't need to hide them on every tick. If you tell them to hide they will stay hidden till you explicitly tell them to show again.

So if you follow my other comment I wrote now you can probably just delete this.

apps/dutchclock/app.js
thyttan: Ah! Yes you're right - better to use clearRect I believe.

whenlit: Ok, that was a nice rabbit hole, thank you. I modified everything to be compatible with top and bottom widgets, which I didn't know about tbh.

pavelmachek:
I guess 1 version in ChangeLog would be enough for now.

More importantly, we have approxclock already. Should we start somehow merging them before we have czech-clock, french-clock, polish-clock ... etc?

pavelmachek:
Hi! The changelog looks good. If it can't be easily merged,  I guess we don't merge, sorry for the noise.

thyttan:
Thanks for the app and the swift communication! Always nice to see new contributors! :)

PR #3562, opened 2024-09-10
calculator: truncate long numbers

flaparoo:
If the resulting number is longer than can be displayed, truncate it so that the (usually) significant part is displayed.
A '-' is append to indicate a truncated number (the font doesn't provide a better character for this than '-').

For example, before this change a simple "5 / 3" resulted in only 6s being displayed - rather than a more useful "1.666666-".


thyttan:
> (the font doesn't provide a better character for this than '-').

Could we go with '...' instead? That seems less ambiguous to me.

flaparoo:
Ah yes, it's not a fixed-width font. "..." is better.

thyttan:
Thanks!

PR #3563, opened 2024-09-10
setuichange: throw if custom drag on some modes

thyttan:
If custom drag handler is added together with mode updown or leftright throw an exception.

See discussion at https://forum.espruino.com/comments/17445657/ for reference.

@gfwilliams


gfwilliams:
Looks good - thanks!

PR #3566, opened 2024-09-10
openhaystack: store battery status in FindMy packet

bobrippling:
This allows a user to know their device's battery, as well as its location

See also:
 https://github.com/seemoo-lab/openhaystack/issues/185
 https://github.com/MatthewKuKanich/FindMyFlipper/blob/e5a678e33dd17af46bb63050be4c9d10e17f6252/FindMyFlipper/findmy_state.c#L126-L137

@olivierbarriere let me know what you think - I'll be testing this over the next week or so


gfwilliams:
This looks cool. I'd be fine with merging it anyway.

Interestingly it's nice to have a way to send some data back. For instance if you had a remote letterbox or something it could report battery level as full if there was something in it, or nearly empty if there wasn't.

bobrippling:
Yes that would be very cool to repurpose the 4 values as another form of communication - or it'd be interesting to see if "invalid" battery values made it through the network too.

I've had this running but my watch hasn't appeared on the network in a while, so can't confirm yet

thyttan:
@bobrippling should we merge this?

bobrippling:
Yeah I think so - I'll see if there's any news on the OpenHaystack front for why the messages haven't been propagated, but I think this is a low risk change and we've shown it's not the cause of the drop in messages, so I'll sort the conflicts when I get a mo and we can merge

bobrippling:
Ready to go :)

PR #3568, opened 2024-09-12
Update boxclk to v0.10

stweedo:
Revised suffix formatting in getDate function. The version of this function introduced in v0.07 did not work correctly.
Also centered most of the boxes in the example boxclk-2.json configuration file.


thyttan:
Thanks!

PR #3569, opened 2024-09-12
Updated Binary LED Clock (BLC)

aemkai:
added settings page and functionality for customizing LED colors


apps/blc/blc.js
```diff
@@ -2,135 +2,181 @@
...
+            col_off = black;
         }
-        
+        else
+        {
+            bord_col = black;   
+            col_off = white;
+        }
+
+        let pwr2 = [1, 2, 4, 8, 16, 32];        // array with powers of 2, because poweroperator '**' doesnt work
```
bobrippling, line 66: Instead of this you can use bitshifting if you like:

```diff
-pwr2[bit_cnt]
+1 << bit_cnt
```

backup.js
```diff
@@ -55,7 +55,7 @@ function bangleDownload() {
     showToast('Backup complete!', 'success');
     if (typeof Android !== "undefined" && typeof Android.saveFile === 'function') {
       // Recent Gadgetbridge version that provides the saveFile interface
-      Android.saveFile("Banglejs backup.zip", "application/zip", content);
+      Android.saveFile("Banglejs backup.zip", "application/zip", btoa(content));
```
bobrippling, line 58: Does this fix a bug? Let's pull it out to a separate PR if so

bobrippling (commented):
Thanks for the PR - had a few comments

apps/blc/blc.js
aemkai: Yes, this is also a good solution. In my opinion the the "LUT"-way may be a bit faster an with only six values not much bigger regarding code size.

backup.js
aemkai: I have no clue what this is? I didn't create a backup.js

bobrippling: No problem - looks like it's a github thing, seems fine locally

apps/blc/blc.js
bobrippling: Sure, we can keep it as-is for now (the bit shift will be faster as there's no memory indirection or bounds checking)

PR #3573, opened 2024-09-14
Improve sanity check for locales

atjn:
I have removed the old sanitycheck which ran in the frontend. I have added a more comprehensive sanitycheck library that is used in two places:
- It runs as part of the CI sanitycheck, which means it should catch all issues before the code is merged.
- It runs in the frontend if the user chooses to customize their locale. If it detects an issue, the user is warned about the errors, but is still allowed to upload the locale if they wish. Example:
image

The upper limits on text length corresponds to the limits I defined in #3530

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -112,6 +112,7 @@ module.exports = {
         "getSerial": "readonly",
         "getTime": "readonly",
         "global": "readonly",
+        "globalThis": "readonly",
         "HIGH": "readonly",
         "I2C1": "readonly",
         "Infinity": "readonly",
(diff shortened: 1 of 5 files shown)


atjn:
Tagging @gfwilliams so @bobrippling doesn't have to do it :)

gfwilliams:
Looks good, thanks!

I forget, but what's the deal with meridian now? It used to be that we left it undefined for countries that didn't usually include it in the time, but I guess now we're just including for everyone and leaving it up to whatever clock there is?

atjn:
> I forget, but what's the deal with meridian now? It used to be that we left it undefined for countries that didn't usually include it in the time, but I guess now we're just including for everyone and leaving it up to whatever clock there is?

Yes, it should be included in all locales, as defined in #3534

PR #3575, opened 2024-09-16
locale: Remove meridians from time patterns

atjn:
Related to #3534

These formatting symbols are not used, and they should not be used because most apps only work with the time string `HH:MM:SS` and they expect the meridian to be accessible from the separate `meridian` function. I suggest removing these symbols.

The removal changes nothing for existing users of the app.


atjn:
Requesting review from @gfwilliams

gfwilliams (approved):

PR #3576, opened 2024-09-16
counter: Support translations

atjn:
Add support for translations with the `/*LANG*/` comment. I have split the strings into separate sentences to make the translation process a bit nicer, and also make it more likely that translations can be reused across apps.


thyttan:
Thanks!

PR #3577, opened 2024-09-16
counter: Add danish translations

atjn:
Add danish translations for counter on banglejs2. These are app-specific because I shortened them quite a lot to make them fit on the screen, so they might not be applicable for other apps.

This needs  #3576 to work.


thyttan:
Thanks!

PR #3582, opened 2024-09-22
edgeclk: live updates of step count

thyttan:
Add option to update step number with every step event (off by default).

### DIFF
--- a/apps/edgeclk/app.js
+++ b/apps/edgeclk/app.js
@@ -281,9 +281,8 @@
   };
 
   const onStep = function () {
-    if (settings.redrawOnStep) drawSteps();
+    drawSteps();
   }
-
 
   /* Lifecycle Functions
   ------------------------------------------------------------------------------*/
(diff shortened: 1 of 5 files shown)


apps/edgeclk/app.js
```diff
@@ -279,6 +280,10 @@
     drawLower();
   };
 
+  const onStep = function () {
+    if (settings.redrawOnStep) drawSteps();
```
thyttan, line 284: Maybe I should do this check before registering the onStep listener instead? Better for battery, but don't know if significant. 

I did this style since I thought it matched the coding style of the app better.

bobrippling: Yeah I think we break tradition and do the check before adding/removing the listener, but happy to merge as-is if you prefer

thyttan: I refactored it 👍

bobrippling: Nice!

PR #3586, opened 2024-09-24
[mtnclock] Fix widgets disappearing on weather update

### LINKED ISSUE #3441: [mtnclock] Widgets disappear on Mountain Pass Clock
### Affected hardware version

Bangle 2

### Your firmware version

~~2v21 107~~
2v22 Stable

### The bug

When using the Mountain Pass Clock app, the widget bar disappears  when the screen is locked. When unlocking the device, the widget bar shows up. After opening the launcher, then going back to the clock the widget bar disappears. When going to the app settings before going back to the clock, the Show Widgets button is no longer checked. The only consistent way I can see this bug happening is when Bluetooth connection is lost, though it happens even while still connected to Bluetooth.

### Installed apps

Clock: Mountain Pass Clock
Launcher: Icon Launcher & Pattern Launcher
Widgets:
- Message Widget
- Lock Widget
- Moon Phase
- Message Icons
- Simple Heart Rate Widget
- Weather
- Quiet Mode Schedule and Widget
- Bluetooth Widget with Notification
- A Battery Widget (with percentage)
- Widget Editor

gloriousDan:
Fixes  #3441

I found the issue for why the widget bar keeps disappearing in the mountain clock app.
It's related to the weather update but the bug is within mtnclock's `app.js` file.

The widget setting is written within the `mtnclock.json` file. The weather data is written to the same file. When a weather update comes in,  the file is overwritten and the `showWidgets` setting consequently deleted from the file which leads to the widget bar disappearing. 

This should fix it. We could also read the current state of data before writing it but in my testing it worked fine like this.


gloriousDan:
You can try it out on my app loader: https://gloriousdan.github.io/BangleApps/?q=mtnclock

bobrippling:
Sounds good, thanks for the fix - looks like there's two lint errors if you're ok to sort those out? What do you think to merging the whole file each time, like a sort of:
```javascript
writeJson(
  "mtnclock.json",
  Object.assign(
    readJson("mtnclock.json"),
    newSettings,
  )
);
```

gloriousDan:
Good point, I thought about that too but at first implemented the simpler fix. But now I implemented your recommendations.

The linter issues should also be fixed now. Unfortunately I didn't get to set up the tests locally, so let's see if the workflows succeed this time :)

gloriousDan:
Can I somehow get the CI to run within in my fork?

bobrippling:
Sure, I've kicked off CI here if that helps, it was because the PR was in draft. Those changes look great btw, I'll merge if you're happy with everything?

gloriousDan:
Thanks, I marked it as draft because I thought of a potential error I still wanted to check/ fix. Maybe you could also help me determine wether it's a problem or not.

Since I load set the updated weather data not to the `data` object anymore, but to the `newSettings` and then write it to the `mtnclock.json`, I suspect that the clock only shows the updated data when the app is reloaded again. In my code the `data` object is only ever read from `mtnclock.json` at the start of the app and not updated later. 

Is my understanding correct, that line 1 is only run once at the start?

I probably get around to fixing this soon, and afterwards we can merge it.

bobrippling:
Yeah that's right, line 1 only runs at the start so the settings are read once - we can reassign to `data` instead so it remains the most up-to-date settings. Hope you don't mind - I've popped this in, let me know what you think (09bea87)

gloriousDan:
That looks good, that's pretty much how I would have done it as well. 

I just have one more Question concerning how the app's code is run. 

When I exit the clock while opening the launcher, is the app paused in the background and resumes running or does it exit and start again from scratch after exiting the launcher?

If the former is the case, the current code might not pick up a change of the mtnclock showWidgets setting and then overwrite it with the outdated value again.

bobrippling:
> When I exit the clock while opening the launcher, is the app paused in the background and resumes running or does it exit and start again from scratch after exiting the launcher?

Sweet, and no you're safe - the app is restarted if you go from the launcher back to the clock, so that's all good. I'll merge - thanks for the fix!

gloriousDan:
Great! Thank you for merging and your help 👍

bobrippling:
Thanks for the time to put together the fix!

PR #3587, opened 2024-09-25
clkinfomsg: clockinfo for displaying messages count

kamilkrz:
Read and unread. This is proof of concept but mostly working.
Only nice to have would be handling messages that was read on phone (I believe if keep msgs is not set in android, messages read on phone trigger some kind of event for removing it from watch)

Nothing fancy


apps/clkinfomsg/clkinfo.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  var debug = function(o) {
+    //console.log(o);
+  }
+  var msgUnread;
+  var msgAll;
+  var msgs = require("messages");
+
+  var getAllMGS = function() {
```
bobrippling, line 17: Minor typo - should this be `getAllMSGs` ?

apps/clkinfomsg/clkinfo.js
```diff
@@ -0,0 +1,88 @@
...
+    if (msgUnread === undefined) {
+      debug("msgUnread is undefined");
+      msgUnread = msgs.getMessages().filter(m => m.new && !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length;
+    }
+    return msgUnread; 
+  }
+
+  var msgCounter = function(type, msg) {
+    var msgsNow = msgs.getMessages(msg);
+    msgUnread = msgsNow.filter(m => m.new && !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length;
```
bobrippling, line 36: Do we need the `.map()` if we then just find the length of the array? (Similarly for a few other instances)

apps/clkinfomsg/clkinfo.js
```diff
@@ -0,0 +1,88 @@
...
+    var msgsNow = msgs.getMessages(msg);
+    msgUnread = msgsNow.filter(m => m.new && !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length; 
+    if (msgUnread === undefined) {
+      msgUnread = "?";
+    }
+    msgAll = msgsNow.filter(m => !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length; 
+    if (msgAll === undefined) {
+      msgAll = "?";
+    }
+    info.items[0].emit("redraw");
```
bobrippling, line 44: What happens here if `items[1]` is being displayed?

apps/clkinfomsg/clkinfo.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  var msgCounter = function(type, msg) {
+    var msgsNow = msgs.getMessages(msg);
+    msgUnread = msgsNow.filter(m => m.new && !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length; 
+    if (msgUnread === undefined) {
+      msgUnread = "?";
+    }
+    msgAll = msgsNow.filter(m => !['call', 'map', 'music'].includes(m.id)).map(m => m.src).length; 
+    if (msgAll === undefined) {
```
bobrippling, line 41: I don't think the length can be `undefined`, perhaps you mean `0`?

bobrippling (commented):
Thanks for the app! Had a few questions :)

apps/clkinfomsg/clkinfo.js
kamilkrz: Good catch - fixing

apps/clkinfomsg/clkinfo.js
kamilkrz: Oh totally, I was testing different approach.

apps/clkinfomsg/clkinfo.js
kamilkrz: Yup, leftover from old approach - will fix and test.

apps/clkinfomsg/clkinfo.js
kamilkrz: I need some guidance - is there a chance to have context here about current CI is being shown? I don't like idea of refreshing all CI modules but in my opinion its best place to do it here.

bobrippling: I'll merge, but feel free to open another PR if you want to tweak it :)

PR #3588, opened 2024-09-25
Update 0.02

mjearlb:
0.02: 
- Removed HRV functionality
- Added milliseconds to the timestamps

### DIFF
--- a/apps/phystrax/ChangeLog
+++ b/apps/phystrax/ChangeLog
@@ -1,1 +1,4 @@
 0.01: New App.
+0.02: 
+- Removed HRV functionality
+- Added milliseconds to the timestamps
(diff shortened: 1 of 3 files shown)


thyttan:
Calling @ekwawu for feedback.

mjearlb:
Hi @thyttan - I am working on the continuation of the Capstone project which @ekwawu started. The HRV function is not accurate, so that's why I have removed it. I totally understand why you need approval first though. Thanks!

thyttan:
Ok! When will you want to be using this in classes again? I.e. when would you like to see this merged? Do you have direct contact with @ekwawu and do you think he will respond here? If not I think we can go ahead.

mjearlb:
@thyttan I apologize, I must have missed your reply. @ekwawu has graduated, so might not be active on their account any more. I would like it merged soon so that we can get started programming the batch of watches we ordered. 

Thank you for your help.

thyttan:
I merged it now. It will be available on the [development app loader](https://espruino.github.io/BangleApps) right away. It will probably be pushed to the official app loader in a week or so.

PR #3591, opened 2024-09-27
fix: widbtstates is not clkinfo

kamilkrz:
Change for sorting on apploader.
Not sure if version bump is needed for this.


bobrippling:
Nice spot, no need for the version bump - thanks!

PR #3592, opened 2024-09-27
lcdclock/lcdclockplus: fix special characters in clockinfo menus

Freubert:
When syncing e.g. weather info on the watch, the display/layout for temperature is broken/misplaced/invisible on `lcdclock` and `lcdclockplus` inside the clockinfo menus.
screenshot_20240927-105228
screenshot_20240927-105021

This PR fixed that issue by removing special characters from the clockinfo item text that can't be displayed by `7seg` font. The width calculation for the size definition becomes more accurate, which leads to a correct placement of the text.
screenshot_20240927-102959
screenshot_20240927-103607


kamilkrz:
This looks like need for version bump to be addressed in updater.

bobrippling:
Nice find, thanks!

PR #3594, opened 2024-09-29
recorder: list tracks in reverse chronological order

thyttan:


gfwilliams:
Thanks! Yes, this makes a lot of sense :)

PR #3598, opened 2024-09-30
widminbt is now correct bt widget

kamilkrz:
This is minor change only to ability to be replaced by another bt widget and vice versa

### DIFF
--- a/apps/widminbt/widget.js
+++ b/apps/widminbt/widget.js
@@ -1,15 +1,13 @@
-(()=> {
-  WIDGETS.minbt={area:"tr",width:NRF.getSecurityStatus().connected?0:15,draw:function() {
-    if(this.width<15)return;
+WIDGETS["bluetooth"] = {area: "tr", width: NRF.getSecurityStatus().connected ? 0 : 15, draw: function () {
+    if (WIDGETS.bluetooth.width < 15) return;
     g.reset();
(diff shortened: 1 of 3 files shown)


apps/widminbt/widget.js
```diff
@@ -1,15 +1,13 @@
-(()=> {
-  WIDGETS.minbt={area:"tr",width:NRF.getSecurityStatus().connected?0:15,draw:function() {
-    if(this.width<15)return;
+WIDGETS["bluetooth"] = {area: "tr", width: NRF.getSecurityStatus().connected ? 0 : 15, draw: function () {
+    if (WIDGETS.bluetooth.width < 15) return;
```
bobrippling, line 2: You could keep `this` here if you prefer:
```suggestion
    if (this.width < 15) return;
```

bobrippling (commented):
thanks, looks great - i'll merge tomorrow to give you chance to change that suggestion, happy either way

apps/widminbt/widget.js
kamilkrz: I used this example to mimic same way widbthide is using.
In fact I found widbthide is exact opposite of widminbt. 
It seems nice to keep it simple for both so one way or another seems fit but changes needs to be done on both apps.

bobrippling:
Thanks for the PR!

PR #3601, opened 2024-10-02
promenu: don't modify the menu object

bobrippling:
This fixes `hadash`'s use of `showMenu`, which uses `menu[""]` (previously `delete`'d)


thyttan:
Seems innocent enough but I'd like a version bump I think :upside_down_face:

bobrippling:
Ah good save, thanks!

thyttan:
Thanks!

PR #3604, opened 2024-10-07
alarm: toggle alarms from main menu

thyttan:
This will let the user toggle existing alarms, timers and events from the main menu of the app. Either by long touch of the menu entries or by touch on the icon on the right.

This change requires to send the touch event used in showScroller through to `alarm` app. Both `showScroller` and `showMenu` needs changes to make it work. So in this draft state of this PR I add two boot files for modifying those two functions. Have a look at the individual commits to see the diff from firmware version, which I copied over and committed unchanged before tweaking them.

Test it via: https://thyttan.github.io/BangleApps/?id=alarm

The tests fail for the showMenu boot file, but it works on the watch. Enough for testing purpouses.

### DIFF
--- a/apps/alarm/ChangeLog
+++ b/apps/alarm/ChangeLog
@@ -53,3 +53,5 @@
 0.48: Use datetimeinput for Events, if available. Scroll back when getting out of group.  Menu date format setting for shorter dates on current year.
 0.49: fix uncaught error if no scroller (Bangle 1). Would happen when trying
 	to select an alarm in the main menu.
+0.50: Bangle.js 2: Long touch of alarm in main menu toggle it on/off. Touching the icon on
+	the right will do the same.
--- a/apps/alarm/README.md
+++ b/apps/alarm/README.md
@@ -20,6 +20,8 @@ It uses the [`sched` library](https://github.com/espruino/BangleApps/blob/master
   - `Disable All` &rarr; Disable _all_ enabled alarms & timers
   - `Delete All` &rarr; Delete _all_ alarms & timers
 
+On Bangle.js 2 it's possible to toggle alarms, timers and events from the main menu. This is done by clicking the indicator icons of corresponding entries. Or long pressing anywhere on them.
+
 ## Creator
 
 - [Gordon Williams](https://github.com/gfwilliams)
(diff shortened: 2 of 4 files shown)


bobrippling:
Alarm app diff looks good to me, and so do the `showMenu` & `showScroller` ones. Given it a trial run and it works a treat

thyttan:
> Alarm app diff looks good to me, and so do the `showMenu` & `showScroller` ones. Given it a trial run and it works a treat

Thanks!

thyttan:
@gfwilliams what would you say about tweaking `showMenu_Q3` and `showScroller_Q3` in firmware like I propose on this PR and then merge this without the boot modifications?

When I tested just now, the `alarm` code seems to work as before if using the standard fw versions of showScroller/Menu. I don't think this should break anything - but maybe I'm overlooking something.

thyttan:
Just tried on emulated Bangle.js 1 and got:

```console
 at line 20 col 1070 in alarm.app.js
...mMenu,10,e,index,undefined,scroller.scroll,group)
                                      ^
in function "onchange" called from line 3 col 106
...onchange)a.onchange(a.value);Bangle.touchHandler==u&&d.draw(...
                              ^
a?d.move(a):d.select()
                     ^
d()
  ^
> 
```

Looking at the git blame it seems to have been around since 2023-11-14.

Will open a separate PR with a fix.

gfwilliams:
Personally, I don't think we should be including different versions of showMenu/etc in apps as it's a maintenance nightmare, but I guess you did this just so we could see what it was like?

I'm up for modifying the built-in ones to pass the extra values through - those changes you made are pretty tidy, although some docs on that would be good, so it's not yet another hidden feature that gets used in only one app ;)

Same for the Alarm app really - it's a nice addition, but it'd also need to be documented in the README as it's not obvious it's something that can be done, especially as different to the way the menu works in all other apps

thyttan:
Thanks!

> Personally, I don't think we should be including different versions of showMenu/etc in apps as it's a maintenance nightmare, but I guess you did this just so we could see what it was like?

Yes - exactly!

> I'm up for modifying the built-in ones to pass the extra values through - those changes you made are pretty tidy, although some docs on that would be good, so it's not yet another hidden feature that gets used in only one app ;)

Ok - will look at that 👍

> Same for the Alarm app really - it's a nice addition, but it'd also need to be documented in the README as it's not obvious it's something that can be done, especially as different to the way the menu works in all other apps

... and this to!

thyttan:
I think this is ready for merge together with https://github.com/espruino/Espruino/pull/2565.

gfwilliams:
Looks good, thanks! Should be compatible with older firmwares too - it just won't work?

thyttan:
Yes, on older firmwares `touch` parameter will become undefined and cause the else statement to run.

PR #3605, opened 2024-10-07
alarm: fix uncaught error if no scroller (Bangle 1)

thyttan:
This PR seems to fix the following. Tested in the Bangle.js 1 emulator on the Web IDE.

Just tried on emulated Bangle.js 1 and got:

```console
 at line 20 col 1070 in alarm.app.js
...mMenu,10,e,index,undefined,scroller.scroll,group)
                                      ^
in function "onchange" called from line 3 col 106
...onchange)a.onchange(a.value);Bangle.touchHandler==u&&d.draw(...
                              ^
a?d.move(a):d.select()
                     ^
d()
  ^
> 
```

Looking at the git blame it seems to have been around since 2023-11-14.

_Originally posted by @thyttan in https://github.com/espruino/BangleApps/issues/3604#issuecomment-2397608781_


gfwilliams:
Thanks! Good spot - yes, we don't use a scroller on Bangle.js1 because we can't scroll just part of the screen

PR #3607, opened 2024-10-09
recorder: cache downloaded tracks

bobrippling:
... to allow a user to save as GPX, then instantly as CSV, rather than waiting to download the same data again.


thyttan:
Nice one!

PR #3609, opened 2024-10-11
Fix custom scrollers so they forward type of touch

thyttan:
... like the firmware version now does.


bobrippling:
LGTM, think you got them all too - thanks!

PR #3611, opened 2024-10-14
Tinyheads

retcurve:
New clock


thyttan:
Fun one - thanks!

Just a thought, personally I think it looks a little off when on light theme the widget field will be white. But it also made me think maybe `tinyheads` could randomize the background color, so it looked a bit like the thiny head wore a headband.

PR #3613, opened 2024-10-16
quicklaunch: do fewer storage interactions

thyttan:


bobrippling:
LGTM, let me know when you're happy

thyttan:
🙏 Not looking to finish this just yet

apps/quicklaunch/boot.js
```diff
@@ -1,22 +1,23 @@
 {
+setTimeout(() => { // wrap in zero ms timeout so the foreground app loads first.
```
thyttan, line 2: @gfwilliams I'm interested what you think of postponing loading quicklaunch to after the forground app, in particular?

thyttan: Ping @gfwilliams

gfwilliams: Sorry, missed this at the time. Well, in theory it's a good idea since it's unlikely to be needed right away, but I wonder how much time this really takes to load? Apart from loading the JSON it all looks quite fast - it's probably not going to be much slower executing that parsing over it at the start.

What I'd do is to load it right away (without timeout), but then just load the settings on demand:

```
let settings;
...
// in the 2 places you use it:
if (!settings) settings = require("Storage").readJSON("quicklaunch.json", true) || {};
```

Then it should be very fast...

thyttan:
@bobrippling ready for merge!

bobrippling:
Sweet, thanks!

thyttan:
Yes thanks - I should get into the habit of doing that.

PR #3570, opened 2024-09-12
promenu: scroll long titles

bobrippling:
Currently titles are truncated, e.g. `"Settings for cl..."`. This change scrolls the item when focused, so each frame you see:
```
"Settings for cl..."
"ettings for clo..."
"ttings for cloc..."
"tings for clock..."
"ings for clockf..."
"ngs for clockfa..."
"gs for clockfac..."
```

PR #3614, opened 2024-10-16
promenu: don't show "..." when a string is fully scrolled in

bobrippling:
Following on from #3570, truncated titles which are scrolled in don't have "..." appended.

```
"Settings for cl..."
"ettings for clo..."
"ttings for cloc..."
"tings for clock   "
"ings for clock    "
"ngs for clock     "
```


thyttan:
Thanks!

PR #3616, opened 2024-10-17
trail 0.10: Redesign to make screen updates fast

pavelmachek:
Got down from 2 seconds per update to < 0.5 sec.


thyttan:
Cool! Thanks!

PR #3618, opened 2024-10-21
EdgeWrite keyboard

retcurve:
Keyboard that implements the EdgeWrite system https://depts.washington.edu/ewrite/


thyttan:
This is really cool - thanks!

I have some suggestions:
- use `require("widget_utils").hide()/.show()` to hide the widget field when displaying the keyboard (https://www.espruino.com/Bangle.js+Hideable+Widgets).
  - will make it easier to input on the top part of the screen when the app calling the keyboard is showing widgets.
- add one/some screenshots to the readme or metadata.

These are not a requirement. I can merge now if you want? or wait if you want to add anything.

retcurve:
@thyttan I did originally use `require("widget_utils").hide()/.show()` but decided against it in the end in case the calling app had hidden the widget bar - didn't want the keyboard to force it back on if it shouldn't be there. But thinking about it reading `Bangle.appRect` before hiding the bar will indicate if the bar needs to be shown on exit so I'll add that now.

thyttan:
The readme says word level stroking doesn't work. I wonder if that's outdated? I can do e.g. word forward/backward, home, end.

Please just say if you want me to merge!

retcurve:
Word level stroking is an extension to the basic edgewrite system that does a kind of word autocompletion. The idea is that after a letter stroke is completed the 4 most likely words based on frequency are shown and some kind of extension to the stroke selects one of the words. I'm not sure yet if this will be practical on a small screen like the Bangle which is why I haven't implemented it. 

This PR is good to merge now, thanks.

thyttan:
Thanks again!

thyttan:
P.s. would love accents so umlauts became possible!

PR #3620, opened 2024-10-23
sched/clkinfo: reset `last` when activating an alarm

bobrippling:
Oddly, my watch didn't go off this morning - I have a nightly backup which ran before the alarm time, so I had this snapshot of my alarm before it went off:

```json
{
  "t": 23940000,
  "del": false,
  "on": true,
  "rp": false,
  "as": false,
  "dow": 127,
  "last": 23,
  "vibrate": "::",
  "msg": ""
}
```

I see two possible reasons the alarm didn't trigger:
- The scheduling code missed the window (unlikely, since this alarm has fired in the past)
- Somehow, `alarm.last` was already set to 23 (i.e. today), so the alarm was skipped. I can't see how it will have been set to 23, but it clearly was at the time of the snapshot.

This PR fixes the possible issue where a user has dismissed the alarm but then re-activates it via clkinfo. This shouldn't normally be possible for that alarm to retrigger in the same day (without time travel), but clearly somehow the `.last` property has been set on this alarm already.


thyttan:
Thanks!

thyttan:
I haven't tested this yet. But if you rebase on the changes Gordon made I can give it a try 👍

bobrippling:
All sorted, and thanks - I would deploy to my gh-pages but I've got that in-use already I'm afraid! (#3626)

thyttan:
Sorry for not coming back sooner. But I tried setting an alarm, then toggling it off then on again. It then fired at the set time a minute later.

Not an exhaustive test, but at least doesn't seem to break on my watch.

bobrippling:
No worries, thanks for the check - I've been using this for a little while too with no issues. I could merge at the weekend if you've no objections?

thyttan:
Yes I think that's fine 👍

PR #3621, opened 2024-10-23
sched interface: handle app-based alarms

bobrippling:
This displays if an alarm is from an app (`appid`) and sorts it to the bottom if so


bobrippling:
Works well, will merge shortly

PR #3627, opened 2024-10-28
chore: rm semicolons at end of clkinfo + settings

thyttan:
This was done in response to https://github.com/espruino/BangleApps/commit/9185793042c7bfddba13333d33f678c2ca246e5d#commitcomment-148439450 .


apps/clkinfostopw/clkinfo.js
```diff
@@ -74,4 +74,4 @@
             }
         ]
     };
-});
+}); // FIXME: semi-colon added automatically when Typescript generates Javascript file?
```
thyttan, line 77: @bobrippling Since you know TS, can the generated JS be made to not add a semi-colon here?

apps/gpsrec/settings.js
```diff
@@ -1,4 +1,4 @@
 (function(back) {
   // just go right to our app - we need all the memory
   load("gpsrec.app.js");
-})();
+})() // FIXME: Since this is called directly, should it still have a semi-colon?
```
thyttan, line 4: @gfwilliams what do you say regarding the code comment above?

apps/gpssetup/settings.js
```diff
@@ -1,4 +1,4 @@
 (function(back) {
   // just go right to our app
   load("gpssetup.app.js");
-})();
+})() // FIXME: Since this is called directly, should it still have a semi-colon?
```
thyttan, line 4: ...also here.

thyttan:
Shouldn't need version bumps for most of the apps right?

apps/clkinfostopw/clkinfo.js
bobrippling: Good question - no not via ts itself but we could add a post-processing step. That would also potentially help with the variable names, [such as `_1`](https://github.com/espruino/BangleApps/pull/3565#issuecomment-2342930224).

If that sounds good to you, I'll pop something onto the end of this PR

bobrippling:
> Shouldn't need version bumps for most of the apps right?

I suppose it depends if old versions (with semi-colons) will work with the [newer code](https://github.com/espruino/BangleApps/commit/9185793042c7bfddba13333d33f678c2ca246e5d#commitcomment-148439450)?

apps/clkinfostopw/clkinfo.js
thyttan: Super - please do!

thyttan:
> > Shouldn't need version bumps for most of the apps right?
> 
> I suppose it depends if old versions (with semi-colons) will work with the [newer code](https://github.com/espruino/BangleApps/commit/9185793042c7bfddba13333d33f678c2ca246e5d#commitcomment-148439450)?

Yes - I suppose you're right 😬😅

bobrippling:
> I'll get a chance to practice vim macros I guess!

If it helps, one approach might be:

```bash
vim $(
  git diff --numstat origin/master... |
  awk '{print $3}' |
  cut -d/ -f2 |
  uniq |
  sed 's;.*;apps/&/metadata.json;' |
  head
) 
```

Then in vim:
```
:set hidden<CR>qa/version":<CR>f.<C-A>yi":e %:h/CHA*<CR>:$pu<CR>A: Remove semi-colon from settings & clkinfo<Esc>:n<CR>@aq@a
```

apps/gpsrec/settings.js
gfwilliams: Good point. Maybe make it `/* ... */` just in case we decide to add this inline? I guess in most cases it won't be an issue as pretokenisation is on anyway

gfwilliams:
Both versions should work with the either code, so I wouldn't worry about the version bump. I don't think it's going to be a problem either way

apps/gpsrec/settings.js
thyttan: I meant the FIXME comment re the semicolon. Was your comment in response to that?

thyttan: I wonder if I should leave the semi-colon in instead of removing it?

gfwilliams: Ahh. Well, yes, it was - but the content of the comment didn't go in. I'd remove it :)

gfwilliams: And no, sorry. The semicolon should be removed - it's still executed the same way as everything else I think? Either way having no `;` on the end won't cause problems

thyttan: Ok - thanks!

apps/clkinfostopw/clkinfo.js
thyttan: This is ready now. So when you get the time please add on the post-processing step 🙏

bobrippling: Sorted - see 3e67088354e0 for the workings 🔍

gfwilliams (approved):

PR #3628, opened 2024-10-28
messagegui: stop buzzing for a message when it's removed

bobrippling:
e.g. a call is answered, and the user has `auto-open` disabled. This allows a user to have infinite buzz (`Vibrate timer`) while the phone is ringing, but the buzz will stop when the call is answered (or declined)


bobrippling:
@gfwilliams this is a reasonably central app change, what do you think?

gfwilliams:
Looks good to me!

bobrippling:
Thanks!

PR #3629, opened 2024-10-29
dtlaunch: postpone load icons not initially needed

thyttan:
Draw the first screen, then load in the rest of the icons from storage.


thyttan:
I guess it's better to do as little as possible with storage for the sake of its longevity. But other than that I kind of prefer to keep the canges as is - should be slightly snappier when switching pages I'd think.

bobrippling:
Sounds good, thanks again!

PR #3630, opened 2024-10-30
settings: restore previous menu's scroll position

bobrippling:
To improve settings usability

- [x] Confirm no memory leaks
- [x] Confirm all menus still work
- [x] Find max memory usage (for banglejs 1)
- [x] Document `{push,pop,restore}Menu`

### DIFF
--- a/apps/setting/ChangeLog
+++ b/apps/setting/ChangeLog
@@ -83,3 +83,4 @@ of 'Select Clock'
 0.72: Add setting for configuring BLE privacy
 0.73: Fix `const` bug / work with fastload
 0.74: Add extra layer of checks before allowing a factory reset (fix #3476)
+0.75: Restore previous menu's scroll positions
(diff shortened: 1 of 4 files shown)


thyttan:
Tried it - works good on my watch!

I'll leave the merging to @gfwilliams.

bobrippling:
> This is a pretty big change - how sure are you it's not going to break something, and that there aren't memory leaks with the push/pop menus?

I've been through most menus, particularly the ones that don't fit the typical pattern, but I'll do an exhaustive check for memory leaks and thus confirm there's no hidden bugs

> And do you know if it noticeably increases memory usage? It looks like maybe it wouldn't by anything noticeable but it'd be really bad if it stopped working on Bangle.js 1.

Will add my findings from the above check

> It might be nice to have some kind of docs on when push/pop/restoreMenu should be used so that if someone works on this in the future we don't end up with it getting broken?

Sounds good, yes I'll add that

bobrippling:
Exhaustively checked all the menu items, memory usage stabilises on returning out of the menus and I've [noted down](https://github.com/espruino/BangleApps/pull/3630/commits/308956b17b2cdd298650145535391cfc11cc0271) the usage in the readme. Give me a shout if there's anything I've missed

gfwilliams:
That's awesome - thanks! Let's merge then!

bobrippling:
Thanks!

PR #3632, opened 2024-10-31
EdgeWrite: Accented characters

retcurve:
@thyttan as requested :)

Accented characters, some extended mode characters (limited by vector font), variable font size.


apps/kbedgewrite/lib.js
```diff
@@ -220,6 +355,7 @@ exports.input = function(options) {
   // Draw initial string
   require("widget_utils").hide();
   g.setBgColor(g.theme.bg);
+  wrapText();
   draw();
```
thyttan, line 359: From using the `podadrem` app I noticed that if a menu isn't cleared (by calling `E.showMenu()`, w/o argument) before going into `edgewrite` with `require("textinput").input({text:"some text"})` the top left corner will be overwritten by a square in solid background colour, matching the size+position of the previous menus red back button. Subsequent draws are all good though.

`draw()` on L359 here could be wrapped in `setTimeout(draw,0)` to get around this.

But probably the calling app should just make sure to clear the menu properly before calling `edgewrite`. (I will tweak `podadrem` and `spotrem` accordingly.

thyttan:
Thanks again - this is so cool! 😀🙏🚀

PR #3635, opened 2024-11-01
tinyhead: change widget bg and fg colour so that widgets blend in with the rest of the clockface

micheledaros:
as from title.

I did not like the visual effect of the widget bar having a completely different bg color from the rest of the clockface.
So I changed it, IMHO it looks much better now.

please let me know what you think @retcurve

### DIFF
--- a/apps/tinyheads/lib.js
+++ b/apps/tinyheads/lib.js
@@ -135,7 +135,7 @@ exports.drawFace = function(scale, eyesNum, mouthNum, peek, offset) {
   // Draw face
   let xOffset = (g.getWidth() - (exports.faceW * scale)) / 2;
   let yOffset = (offset ? offset : 0) + ((g.getHeight() - (exports.faceH * scale)) / 2);
-  g.setBgColor(settings.hairColour);
+  g.setBgColor(exports.settings.hairColour);
   g.clearRect(Bangle.appRect);
   g.setClipRect(Bangle.appRect.x, Bangle.appRect.y, Bangle.appRect.x2, Bangle.appRect.y2);
 
--- a/apps/tinyheads/lib.js
+++ b/apps/tinyheads/lib.js
@@ -135,7 +135,7 @@ exports.drawFace = function(scale, eyesNum, mouthNum, peek, offset) {
   // Draw face
   let xOffset = (g.getWidth() - (exports.faceW * scale)) / 2;
   let yOffset = (offset ? offset : 0) + ((g.getHeight() - (exports.faceH * scale)) / 2);
-  g.setBgColor(0, 0, 0);
+  g.setBgColor(settings.hairColour);
   g.clearRect(Bangle.appRect);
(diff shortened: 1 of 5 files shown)


retcurve:
Thanks for the update! I agree, the widget bar looks much better when it blends in with the hair. 

However `Bangle.setUI("clock", {` can't be used because setting the clock input mode in that way unfortunately stops the touch handler from working. This is why I used a custom mode with `clock: true`. It looks like you've done that as a way of resetting the theme but I think this will probably need to be done directly in the clock when exiting in both the remove handler and the touch handler.

micheledaros:
I see, thanks for reviewing, it should be fixed now. Please have a look if it makes sense.

BTW how do you test your changes to lib.js? If I upload it directly to the watch's storage via the espruino IDE, I get an error 

`Uncaught ReferenceError: "exports" is not defined`

retcurve:
The background colour shouldn't be set to the hair colour, especially if the widget bar is disabled. It's set to black to hide the fact that the graphics were originally created for the Pebble which has a slightly different screen size, so the slight bars down the sides of the face blend in with the black of the watch.

You'll always get the error about exports not being defined, but as long as you're uploading to the correct filename (tinyheads.lib.js) you can ignore the error and everything should work fine.

micheledaros:
I've pushed another commit, to set the bg colour to the hair colour only when the widget bar is showing.

I find it looks better this way, as it avoids the strange "T" effect (see screenshots).

same color of hair



black


Anyway, if you think it causes more hassle than good, I can revert the changes to lib and leave the bg always black

apps/tinyheads/lib.js
```diff
@@ -135,7 +135,13 @@ exports.drawFace = function(scale, eyesNum, mouthNum, peek, offset) {
   // Draw face
   let xOffset = (g.getWidth() - (exports.faceW * scale)) / 2;
   let yOffset = (offset ? offset : 0) + ((g.getHeight() - (exports.faceH * scale)) / 2);
-  g.setBgColor(0, 0, 0);
+  
+  if (exports.settings.showWidgets == 'on' || (exports.settings.showWidgets == 'unlock' && !Bangle.isLocked())) {
```
retcurve, line 139: This will also cause the background to change to the hair colour when editing the tinyhead, but we can work around  this by checking the scale value so we still use a black background for the smaller editor scale. Bit of a hack but it'll do.

`if (scale == 9 && (exports.settings.showWidgets == 'on' || (exports.settings.showWidgets == 'unlock' && !Bangle.isLocked()))) {`

micheledaros: good point, it should be fixed now

retcurve:
:+1: This is ready to be merged now

gfwilliams:
Looks good - you just need to bump the version and add a note in the ChangeLog

gfwilliams:
Thanks! merging now

PR #3637, opened 2024-11-02
setuichange: Bangle.js 1 support

thyttan:
In step with changes up to https://github.com/espruino/Espruino/pull/2571/commits/12dacead0d02fb3671fb9c7b7b01e18ff0c0119d, on PR https://github.com/espruino/Espruino/pull/2571

This is currently untested since I don't have a Bangle.js 1 myself.


thyttan:
@gfwilliams this my be superfluous now that the firmware was changed. But ready to merge here as well.

gfwilliams:
Thanks! Just merging.

Yes, I think when we push out 2v25 we can probably remove this app?

PR #3638, opened 2024-11-02
Score: bug fixes and some usability and performance improvements

AleixDev:
You can see the improved app on my [own copy of the App Loader](https://aleixdev.github.io/BangleApps/?id=score). 

I only have a Bangle2 so I couldn't test it on Bangle1 (and I couldn't get it to work properly on the Bangle1 emulator either). This is the first app I modify so let me know if I missed anything, thanks!

_One of the changes is the correction of the tiebreak alignment problem, I attach the before and after screenshots:_
image

image


bobrippling:
Nice, thank you!

PR #3640, opened 2024-11-04
Update reply lib metadata.json to support Bangle JS 1

WhelanB:
The reply lib metadata.json file currently does not list Bangle JS 1 support, which then blocks newer versions of the Messages app from installing on Bangle JS 1 devices as the lib can't be found. This adds support for Bangle JS 1, as I've tested in the emulator and there doesn't appear to be anything blocking this running on both devices


gfwilliams:
Great, thanks!

PR #3641, opened 2024-11-05
EdgeWrite: Draw initial text after setUI in case it removes a back button

retcurve:
As discussed in https://github.com/espruino/BangleApps/pull/3633 setUI removes the back button by clearing a rectangle, even if the widget bar is hidden, so this PR moves hiding the widget bar and the draw of any initial text until after this has happened.


thyttan:
Thanks!

PR #3642, opened 2024-11-05
dtlaunch: remember page between dtlaunch instances

thyttan:
@bobrippling interested what you think about this!

Available on https://thyttan.github.io/BangleApps/?id=dtlaunch

- [x] test for ram leaks

### DIFF
--- a/apps/dtlaunch/app-b2.js
+++ b/apps/dtlaunch/app-b2.js
@@ -41,7 +41,7 @@
     page = (global.dtlaunch&&dtlaunch.handlePagePersist()) ??
       (parseInt(s.read("dtlaunch.page")) ?? 0);
     initPageAppZeroth = page*4;
-    initPageAppLast = (page*4+3<apps.length-1)?page*4+3:apps.length-1;//Math.min(page*4+3, apps.length-1); // FIXME:What is fastest?
+    initPageAppLast = Math.min(page*4+3, apps.length-1);
   }
 
   for (let i = initPageAppZeroth; i <= initPageAppLast; i++) { // Initially only load icons for the current page.
@@ -50,7 +50,7 @@
   }
 
   let Napps = apps.length;
-  let Npages = parseInt(Napps/4+1); //Math.ceil(Napps/4); // FIXME: What is fastest?
+  let Npages = Math.ceil(Napps/4);
   let maxPage = Npages-1;
   let selected = -1;
   //let oldselected = -1;
--- a/apps/dtlaunch/app-b2.js
+++ b/apps/dtlaunch/app-b2.js
@@ -33,8 +33,12 @@
     s.writeJSON("launch.cache.json", launchCache);
   }
   let apps = launchCache.apps;
-  let page = (Bangle.dtHandlePagePersist&&Bangle.dtHandlePagePersist()) ?? (parseInt(s.read("dtlaunch.page")) ?? 0);
-  for (let i = page*4; i < Math.min(page*4+4, apps.length); i++) { // Initially only load icons for the current page.
+  let page = (global.dtlaunch&&dtlaunch.handlePagePersist()) ??
+    (parseInt(s.read("dtlaunch.page")) ?? 0);
+
+  const INIT_PAGE_APP_ZEROTH = page*4;
+  const INIT_PAGE_APP_LAST = Math.min(page*4+3, apps.length-1);
+  for (let i = INIT_PAGE_APP_ZEROTH; i <= INIT_PAGE_APP_LAST; i++) { // Initially only load icons for the current page.
     if (apps[i].icon)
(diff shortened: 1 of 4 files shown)


thyttan:
hm, the linter doesn't seem to handle `??` well.

thyttan:
It feels like this adds just enough milliseconds to be noticable when switching back and forth to the clock. 

Might do some timings to compare.

apps/dtlaunch/app-b2.js
```diff
@@ -99,13 +99,31 @@
...
+  for (let i = 0; i < apps.length; i++) { // Load the rest of the app icons that were not initially.
+    if (i >= page*4 && i < Math.min(page*4+4, apps.length)) continue;
     if (apps[i].icon)
       apps[i].icon = s.read(apps[i].icon); // should just be a link to a memory area
   }
 
+  if (!Bangle.dtHandlePagePersist) {
+    Bangle.dtHandlePagePersist = (page) => {
+      // Function for persisting the active page when leaving dtlaunch.
+      if (page===undefined) {return this.page||0;}
```
bobrippling, line 113: Here, `this` will be the global object, is that what we want?

apps/dtlaunch/app-b2.js
```diff
@@ -99,13 +99,31 @@
...
+      if (!this.killHandler) { // Only register kill listener once.
+        this.killHandler = () => {
+          s.write("dtlaunch.page", this.page.toString());
+        };
+        E.on("kill", this.killHandler); // This is intentionally left around after fastloading into other apps. I.e. not removed in uiRemove.
+      }
+
+      this.page = page;
+    };
+    Bangle.dtHandlePagePersist(page);
```
bobrippling, line 124: Similarly to the `this` thing above, do we want to chuck this onto `Bangle`, or perhaps we just have an unrelated global object we use instead?

apps/dtlaunch/app-b2.js
```diff
@@ -33,7 +33,8 @@
     s.writeJSON("launch.cache.json", launchCache);
   }
   let apps = launchCache.apps;
-  for (let i = 0; i < 4; i++) { // Initially only load icons for the current page.
+  let page = (Bangle.dtHandlePagePersist&&Bangle.dtHandlePagePersist()) ?? (parseInt(s.read("dtlaunch.page")) ?? 0);
```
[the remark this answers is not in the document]
bobrippling: Yes, I think the linter doesn't know how to handle `??`, I'm sure @gfwilliams pointed out why/what changes would need to be made previously but I can't find the comment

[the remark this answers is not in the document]
gfwilliams: It'll be the acorn parse in sanitycheck - it would be a nice one to fix I guess

apps/dtlaunch/app-b2.js
thyttan: No that wasn't my intention. I thought I somehow attached the value to the function but I see that's not the case.

apps/dtlaunch/app-b2.js
thyttan: Maybe that's better!

Should I do it by e.g. `global.dtlaunch = {}; global.dtlaunch.handlePagePersist = ()=>{...}`?

bobrippling: Nice, yeah I think that'll work great

bobrippling: Nice - perhaps we want to camelCase the function to keep with the existing pattern? (`handlePagePersist` instead of `HandlePagePersist`)

thyttan: Yes - just noticed that as well 👍

thyttan: > Nice, yeah I think that'll work great - and if you make it
> ```javascript
> global.dtlaunch.handlePagePersist = function() {...}
> //                                  ^~~~~~~~
> ```
> 
> then when you call it as `dtlaunch.handlePagePersist()`, then you can use `this` exactly as you intended (as `this == dtlaunch`)

Interesting! So that's a difference between `function() {...} ` and `() => {...} `?

bobrippling: Yes exactly - `() => ...` won't have its own `this`, it just uses one from up a level, whereas `function`s always have their own `this`. Also no, the `global` at the start of my suggestion isn't needed, nice spot

What you have now looks great btw!

thyttan:
Measuring performance on non-pretokenized code:

Time to first draw: 438 ms -> 441 ms

Time to run whole app: 600 ms -> 615 ms

apps/dtlaunch/app-b2.js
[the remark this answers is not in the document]
bobrippling: (merged)

bobrippling:
Here we go!

PR #3643, opened 2024-11-05
backswipe: change to prepend swipe listener

thyttan:
Seems to mitigate some conflicts when switching between `edgeclk` (with seconds active) and `dtlaunch`. Sometimes the seconds redraw timeout of `edgeclk` would draw over `dtlaunch`. I don't really understand the conflict but it seems to be resolved by this change..

Can be tried via https://thyttan.github.io/BangleApps/?id=backswipe

- [x] test for ram leaks


bobrippling:
Yeah looks like a timing issue, let me know if it does ever happen after this change and we can debug more

PR #3644, opened 2024-11-05
smpltmr: adapt to new setui

thyttan:
btn -> btnRelease as to not fire unintentionally when long pressing the button to reset the watch.

### DIFF
--- a/apps/smpltmr/ChangeLog
+++ b/apps/smpltmr/ChangeLog
@@ -8,3 +8,4 @@
 0.08: Timer ClockInfo now updates once a minute
 0.09: Timer ClockInfo resets to timer menu when blurred
 0.10: Timer ClockInfo now uses +- icons, and changes timer from 'T-5 min' to just '5 min' to aid readability
+0.11: Fix to handle updated firmware on fw 2v25 (or cutting edge >2v24.164), btn -> btnRelease. The timer would fire on long press before the watch reset.
--- a/apps/smpltmr/app.js
+++ b/apps/smpltmr/app.js
@@ -177,7 +177,7 @@ Bangle.setUI({
   mode : "custom",
   touch : function(n,e) {onTouch(n,e);},
   drag : function(e) {onDrag(e);},
-  btn : function(n) {onButton();},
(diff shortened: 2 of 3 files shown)


bobrippling:
Nice idea - is this something we might want to change in other apps too? And if so, would you say we might want a shorter way of doing that?

bobrippling:
Yeah, it's a little unfortunate that we'd need the version checking code in other files, which is a few lines of code

```javascript
if (parseFloat(process.env.VERSION.replace("v","")) < 224.164) {
  uiOptions.btn = ...
} else {
  uiOptions.btnRelease = ...
}
```

I guess we could trim it down to something like this:
```javascript
uiOptions[parseFloat(process.env.VERSION.replace("v","")) < 224.164 ? "btn" : "btnRelease"] = ...
```

and going off a search, apply it to:
```
apps/angles/app.js
apps/bikespeedo/app.js
apps/calendar/calendar.js
apps/calibration/app.js
apps/clockbg/settings.js
apps/drained/app.js
apps/drained/app.ts
apps/espruinoterm/app.js
apps/folderlaunch/app.js
apps/folderlaunch/app.ts
apps/gipy/app.js
apps/gpstrek/app.js
apps/kbedgewrite/lib.js
apps/messagelist/app.js
apps/mitherm/app.js
apps/openstmap/app.js
apps/pomoplus/app.js
apps/rebbleagenda/app.js
apps/scicalc/app.js
apps/smpltmr/app.js
apps/spacew/app.js
apps/spacew/bench.js
apps/stopwatch/stopwatch.app.js
apps/sudoku/app.js
apps/tempmonitor/tempmonitor.app.js
apps/tetris/tetris.app.js
apps/tinyheads/settings.js
apps/torch/app.js
apps/wrkmem/app.js
```

But perhaps as another PR, wdyt? Reckon we flag it to Gordon as something that we might want to change in `setUI` instead?

bobrippling:
Yeah sounds good, we can sort them when they crop up I reckon 👍

bobrippling:
And thanks for the PR!

PR #3645, opened 2024-11-06
Poker Timer App

ki9us:
Is it ok to add an app as a git submodule?


thyttan:
> Is it ok to add an app as a git submodule?

@gfwilliams probably knows best here.

gfwilliams:
No, I'm afraid not. Because then we're unable to check on any changes - you could push an update, accidentally mess up the metadata.json and break the app loader for everyone, and we couldn't do anything apart from remove your app.

Sometimes we may need to make quite broad changes (maybe some API changes or we become aware of something your app is doing that's a problem) and we want to be able to make changes to a bunch of apps at once. If we have a bunch of submodules we're going to have to submit PRs to all the app owners who may/may not merge them and the whole thing becomes onworkable.

gfwilliams:
Sounds good. Please can you change the version number style to `0.01` (not `0.0.1`) - I believe parts of the app loader expect it to be a number

gfwilliams:
Yes - it's because of:

```
 {"name":"timer.app.js","url":"app.js"},
    {"name":"timer.img","url":"app-icon.js","evaluate":true}
```

They should be `{"name":"pokertimer...` since the app is called pokertimer

gfwilliams:
Looks good to me - thanks!

pavelmachek:
App seems to have its own copy of LICENSE agreement. Is that reasonable?

gfwilliams:
I'd prefer not - the whole idea is everything is licensed the same so everyone knows where they stand.

The license is MIT, the same as the repo anyway, so it seems kind of redundant

PR #3647, opened 2024-11-06
drained: restore menu position

bobrippling:
To aid in usability when toggling options, such as BLE enabled


bobrippling:
See also espruino/Espruino#2574

bobrippling:
No issues in testing, will merge soon

PR #3650, opened 2024-11-10
Bttfclock

NoobEjby:
Made some changes to the status line.
Made it possible to change step goal with the Health Tracking app.


bobrippling:
Nice, thank you

PR #3653, opened 2024-11-11
sanitycheck: parse ecma2020 constructs (`??`)

bobrippling:
To help with PRs such as #3642

### DIFF
--- a/bin/sanitycheck.js
+++ b/bin/sanitycheck.js
@@ -11,10 +11,9 @@ apploader.init({
 });*/
 
 var jsparse = (() => {
-  var acorn, stage3;
+  var acorn;
   try {
     acorn = require("acorn");
-    stage3 = require("acorn-stage3");
   } catch (e) {
     console.log("=====================================================");
     console.log("  ACORN NOT FOUND");
@@ -25,8 +24,7 @@ var jsparse = (() => {
     return str => {throw new Error("no acorn")};
   }
 
-  var parser = acorn.Parser.extend(stage3)
-  return str => parser.parse(str, { ecmaVersion: 2020 });
+  return str => acorn.parse(str, { ecmaVersion: 2020 });
 })();
 
 var BASEDIR = __dirname+"/../";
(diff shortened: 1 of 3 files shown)


bobrippling:
@gfwilliams this PR assumes we can parse all syntaxes from the 2020 edition

https://github.com/espruino/BangleApps/blob/a8f1aa97e51bf9fa3d2b3777c63248b243cc0199/bin/sanitycheck.js#L29

... but I don't think that's the case (e.g. destructuring isn't supported), so I have two questions:

1. Is there a list anywhere of what's supported and what isn't?
2. Does this matter or do we just accept that we might pass some things through `sanitycheck` which then fail to parse on-device?

gfwilliams:
> Is there a list anywhere of what's supported and what isn't?

https://www.espruino.com/Features

I guess ideally it'd be really nice to throw errors for unsupported stuff. Is there a 'visitor' for acorn maybe, so we can iterate over everything in the syntax tree and check for unsupported stuff that gets parsed (eg spread I guess)?

bobrippling:
I've gone with upgrading to 8.0 which lets acorn parse `??` without an extra dependency, at the price of having to specify an ecma version. A visitor would be ideal but I'm afraid I don't have capacity atm (happy to close this for now to keep the PRs clear)

gfwilliams:
Thanks! No I'm fine if you want to merge this - I guess the Lint config (or something in espruinotools) should ideally be the one to catch it anyway.

bobrippling:
Cool, sounds good - I suppose there's existing constructs that are parsed but not warned about too, so the lint config does sound like the right approach (e.g. destructuring)

gfwilliams:
Thanks! I think this should probably be in https://github.com/espruino/EspruinoTools (which the IDE as well as BangleApps uses). We could then use the tools themselves (there is some commented out code in https://github.com/espruino/BangleApps/blob/master/bin/sanitycheck.js#L565-L573) to parse which could check for supported constructs as well as potentially some other things (like libraries that don't exist)

bobrippling:
Ah, so the tools can accurately tell if the JS will be parsed by espruino? That's cool

gfwilliams:
I think that'd be best - we already do attempt to parse with Acorn so adding some code to run over that wouldn't be too hard I hope

PR #3657, opened 2024-11-11
widadjust: Add application to automatically determine PPM fixup from GPS data

pavelmachek:


bobrippling:
Nice idea, thank you

vilhalmer:
Sorry to comment on such an old PR, but how is this app intended to be used? I've tried reading the code but it's incredibly terse. Does it automatically configure the widget? Does it ever lock onto a particular value? I've left it open for about 15 minutes at a time but it never seems to settle on an error value.

PR #3659, opened 2024-11-13
waypoints 0.05: implement goto functionality

pavelmachek:
Update libraries, refactor and implement goto functionality


apps/waypoints/waypoints.app.js
```diff
@@ -219,10 +490,85 @@ function showNumpad(text, key_, callback) {
...
+  g.clear();
+  la.render();
+}
+
+function showCard() {
+  var menu = {
+    "" : {title : "Select WP"},
+    "< Back" : mainMenu
+  };
+  if (Object.keys(wp).length==0) Object.assign(menu, {"No WPs":""});
```
bobrippling, line 543: Is `wp` always an array? If so we could:
```suggestion
  if (wp.length==0) Object.assign(menu, {"No WPs":""});
```

apps/waypoints/waypoints.app.js
```diff
@@ -219,10 +490,85 @@ function showNumpad(text, key_, callback) {
...
+  else {
+    wp.forEach((val, card) => {
+      const name = wp[card].name;
```
bobrippling, line 546: Here we could save a bit of code:
```suggestion
      const name = val.name;
```

bobrippling (commented):
Looks good, two minor suggestions

pavelmachek:
Yeah, there's ton more optimalizations/cleanups that could be done. I'd prefer this to be applied, then I can take your patches as a separate commit, do more work on top of that, and submit it as new version after some testing?

pavelmachek:
Thank you, it makes my patch flow easier.

PR #3669, opened 2024-11-22
new app : Pushups

wagnerf42:
hi, so i have a new app for exercising with the accelerometer.

i didn't test it with someone else but it works for me.


bobrippling:
Nice idea! i notice the screen is initially black and i have to tap to get the images, do you see that? I don't seem to be able to trigger the numbers to change, perhaps I'm doing it wrong?

wagnerf42:
hi, i never had a black screen. is it pure black or do you see the image ?
sometimes the widgets can interfere with the colors. maybe i should set the color before clearing.

for the number changing you can test by pivoting your wrist left and right while seated.
i'm not actually sure the accelerometer will always give the same numbers on every watch.

could you do a few pushups under the "acceleration recorder" app and send me the file ?
do the other activities work ?
(you can swipe to skip)

bobrippling:
Yes - pure black, could it be because I'm using the dark theme? I'll have another go with the exercise and see about sending you the record

bobrippling:
So I've confirmed - if a user has a dark theme they can't see the app, some colour values would need altering for it to work (e.g. using `g.theme.fg`, etc). With light theme though, I can see what's happening and can use the app

wagnerf42:
hi, sorry for the delay. the theme issue is fixed

bobrippling:
No worries, and thanks for the fix

PR #3674, opened 2024-11-27
spacer: Compare GPS with Baido and Glonass

pavelmachek:


bobrippling:
Looks cool - will give it a text next I'm outside

apps/spacer/spacer.app.js
```diff
@@ -0,0 +1,210 @@
...
+  fmtSys: function(sys) {
+    return sys.sent + "." + sys.d23 + "D "+ this.tof(sys.pdop) + " " + this.tof(sys.vdop) + "\n";
+  },
+  display: function() {
+    let m = this.old_msg;
+    let msg = "" + this.tof(m.time) + "\n" + 
+        "q" + m.quality + " " + m.in_view + " " + m.hdop + "\n" +
+        "gp"+ this.fmtSys(m.gp) +
+        "bd" + this.fmtSys(m.bd)  +
+        "gl" + this.fmtSys(m.gl);
```
bobrippling, line 102: These cause an exception the first time round, looks like (at least) `gp` isn't defined:

```
Uncaught Error: Can't read property 'sent' of undefined
 at line 94 col 15 in spacer.app.js
sys.sent + "." + sys.d23 + "D "+ this.tof(sys.pdop) + " " + ...
   ^
in function "fmtSys" called from line 100 col 31 in spacer.app.js
        "gp"+ this.fmtSys(m.gp) +
                              ^
in function "display" called from line 200 col 41 in spacer.app.js
graw.display();
             ^
```

pavelmachek:
I must admit I did development in the meantime. Now it has satellite view, too, touch bottom part of screen to switch between views. And you'll still sometimes get "undefined", due to races, but it should demonstrate the idea. Unfortunately hw does not report separate pdop (etc) values, even when there's space in the protocol for that, so you'll have to switch systems to play with that (top left corner).

GPS fix is usual here, but I was able to get Beidou fix, too. Yay!

bobrippling:
Cool, nice additions! There's just the [`undefined` issue above](https://github.com/espruino/BangleApps/pull/3674#discussion_r1865019987) and I notice if a user has the dark theme set then they can't see what's going on in the app - some colour values would need altering for it to work (e.g. using `g.theme.fg`, etc)

pavelmachek:
U replaced undefined with (no data). For themes, that would be more work, and looking at code... there's more work to be done on the code. Could I get you to apply this and do theme fixes in the next version (along with other cleanups)?

bobrippling:
Yes not a problem :)

Your change for the `undefined` problem will still see the exception btw:
```javascript
if (sys.sent !== undefined && sys.d23 !== undefined)
```

`sys` is `undefined` here so we still get the error - your check would want to be `if (sys) { ... }`

pavelmachek:
Ok, this should only happen in the first cycle or so. Now should be fixed.

bobrippling:
Great, thanks!

Issue #3444, opened 2024-06-06
[OpenWeatherMap] com.android.volley AuthFailureError


doma9on:
### Affected hardware version

Bangle 2

### Your firmware version

2v21

### The bug

OpenWeatherMap force refresh no longer works as I am met with 'com.android.volley AuthFailureError'. It seems to have been happening since they upgraded to API v3, so I created a new API key and saved it but I still get the error. Nothing was changed on the Bangle which could indicate why it no longer works.

bobrippling:
Yes, this is a v3 issue, there was a PR (#3427), but it wasn't merged. I'm sure @devsnd wouldn't mind if you wanted to bring it forward

doma9on:
> Yes, this is a v3 issue, there was a PR (#3427), but it wasn't merged. I'm sure @devsnd wouldn't mind if you wanted to bring it forward

Hi, thanks for reply. Will the PR be merged? If possible I would like to be able to use v2.5 if it means I don't have to enter payment details. If not, is there any other option for free weather data like this on Bangle? :)

devsnd:
@doma9on I closed the PR because the 2.5 API was discontinued in June. You either need to find another weather provider or add your payment details. I chose the latter, as it's practically free to use with the bangle still.



> [...] One Call API 2.5 has been finally closed in June 2024. 
https://openweathermap.org/one-call-transfer

zwiglm:
Sorry for repeating here!

I also got a paid subscription. From eg Postman works fine. Within the app, I still get the error message "com.android.volley..." when trying to force-update the weather.
Any suggestions?

zwiglm:
Hmm. Sorry. Now working

PR #3675, opened 2024-11-28
Undo change to One Call API 3.0

gsuarezc:
After see the issue #3444 and the pull request #3427, I re-read the documentation I referenced in my pull request #3387 and  realize I misinterpreted the information.

The post says they upgraded the __One Call API 2.5__ to __One Call API 3.0__, but this app uses the [__Current weather data API__](https://openweathermap.org/current), which is different and free, and it does not require a payment method.

This pull request undoes that change.

I apologize for any inconvenience caused.


bobrippling:
No worries, thanks for the restore!

PR #3679, opened 2024-11-30
owmweather: Fix infinite loop when settings.updated is not defined

gsuarezc:
Fix an infinite loop when settings.updated is not defined. This issue was draining the entire battery within a day.
Fix some missing semicolons.


bobrippling:
Thanks!

PR #3681, opened 2024-12-03
widbtstates: generate better js indexing

bobrippling:
with a slight typescript hack - avoid generating a new string when we can use a boolean


bobrippling:
No problems, merging

PR #3685, opened 2024-12-03
messagegui: refactor the messages scroller and use it more (WIP)

thyttan:
WIP

Initializes the scroller with ~~three~~ all messages loaded, scrolled to the chosen message. You can scroll up/down to see the rest of the messages.

~~This is far from merge ready and I'm not sure it should ever be merged. Should be implemented as a setting if to be merged I think.~~ I'm inching closer to something that could be considered for merging.

With that said it can be tried from my app loader: https://thyttan.github.io/BangleApps/?id=messagegui

### DIFF
--- a/apps/messagegui/ChangeLog
+++ b/apps/messagegui/ChangeLog
@@ -114,4 +114,5 @@
 0.83: Add option to not open the first unread message
 0.84: Fix: Assign show message entry to the settings menu and not the message itself.
 0.85: Use new Rebble fonts if available
-      Remove workaround for 2v10 (>3 years ago) - assume everyone is on never firmware now
+0.86: Refactor to display a scroller with all messages loaded. 
+      Initial scroll position is at the chosen message.
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -340,6 +340,9 @@ function showMessagesScroller(msg) {
           setColor("#f00").drawImage(footerImgNeg,r.x+5+3,r.y).
           setColor("#0f0").drawImage(footerImgPos,r.w-64-5,r.y);
       }
+      if (0===scrollIdx) {
+        g.setColor("#f00").drawImage(atob("GBiBAAAYAAH/gAf/4A//8B//+D///D///H/P/n+H/n8P/n4f/vwAP/wAP34f/n8P/n+H/n/P/j///D///B//+A//8Af/4AH/gAAYAA=="), r.x, r.y);
+      }
       if (scrollIdx<shownScrollIdxFirst) {shownScrollIdxFirst = scrollIdx;}
       if (scrollIdx>shownScrollIdxLast) {shownScrollIdxLast = scrollIdx;}
     },
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -78,7 +78,7 @@ var onMessagesModified = function(type,msg) {
   }
   if (msg && msg.id=="nav" && msg.t=="modify" && active!="map")
     return; // don't show an updated nav message if we're just in the menu
-  showMessageOverview(msg&&msg.id, false);
+  showMessageRouter(msg, persist, "dependsOnActive");
 };
 Bangle.on("message", onMessagesModified);
 
@@ -86,6 +86,39 @@ function saveMessages() {
   require("messages").write(MESSAGES);
 }
 E.on("kill", saveMessages);
+
+function showMessageRouter(msg, persist, explicitDestnation) {
+  //explicitDestnation (undefined/"scroller"/"overview"/"dependsOnActive")
+
+  ////var active; // active screen (undefined/"list"/"music"/"map"/"overview"/"scroller"/"settings")
+  //if (active==undefined) { } else if (active=="list") ... //and so on.
+
+  if (persist) {cancelReloadTimeout();} else if (Bangle.isLocked()) {resetReloadTimeout();}
+
+  if (msg.id=="music") {
+    cancelReloadTimeout(); // don't auto-reload to clock now
+    return showMusicMessage(msg);
+  }
+  if (msg.id=="nav") {
+    cancelReloadTimeout(); // don't auto-reload to clock now
+    return showMapMessage(msg);
+  }
+  if (msg.id=="call") {
+    return showMessageOverview(msg.id);
+  }
+  if ("scroller"===explicitDestnation) {
+    return showMessageScroller(msg);
+  }
+  if ("overview"===explicitDestnation) {
+    return showMessageOverview(msg.id);
+  }
+  if ("dependsOnActive"===explicitDestnation) {
+    if ("scroller"===active) {return;}
+    if ("list"===active) {return returnToMain();}
+    if ("settings"===active || "overview"===active) {return;}
+  }
+  //if (false) {showMessageSettings(msg);}
+}
 
 function showMapMessage(msg) {
   active = "map";
@@ -552,7 +585,7 @@ function checkMessages(options) {
     },
     select : idx => {
       if (idx < MESSAGES.length)
-        showMessageOverview(MESSAGES[idx].id, true);
+        showMessageRouter(MESSAGES[idx], true, "overview");
     },
     back : () => load()
   });
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -383,16 +383,17 @@ function showMessagesScroller(msg) {
     }
   });
 
-  // helper function for message selection
-  let findSelectedMsg = function(scrollIdx) {
-    for (let i = firstTitleLinePerMsg.length - 1; i >= 0; i--) {
-      if (scrollIdx >= firstTitleLinePerMsg[i]) {
-        return MESSAGES[i];
-      }
-    }
-  }
+  //// Additional input handling to extend that of E.showScroller ////
+
+  // Input handling for positive and negative response via swipes.
+  // Emits a touch event that in turn triggers the select function of E.showScroller above.
+  Bangle.swipeHandler = (lr) => {
+    if (lr) {Bangle.emit("touch", 1, {x:Math.floor(APP_RECT.x2/2), y:Math.floor(APP_RECT.y2/2), type:{swipeLR:lr}});}
+  };
+  Bangle.on("swipe", Bangle.swipeHandler);
 
   // Add an external back touch handler.
+  // Emits a new touch event that in turn triggers the select function of E.showScroller above.
   let touchHandler = (button, xy)=>{
     // if ((left side of Banlge 1 screen) || (top left corner of Bangle 2 screen))
     if ((!xy && 1===button) || (xy && xy.type===0 && xy.x<30 && xy.y<30)) {
@@ -409,6 +410,18 @@ function showMessagesScroller(msg) {
   };
   Bangle.prependListener("touch", touchHandler);
 
+  //// Helper functions ////
+
+  // helper function for message selection
+  let findSelectedMsg = function(scrollIdx) {
+    for (let i = firstTitleLinePerMsg.length - 1; i >= 0; i--) {
+      if (scrollIdx >= firstTitleLinePerMsg[i]) {
+        return MESSAGES[i];
+      }
+    }
+  }
+
+  // Helper function to update new status of messages
   function updateReadMessages() {
     let shownMsgIdxFirst, shownMsgIdxLast;
     const LINES_PER_SCREEN = APP_RECT.h/FONT_HEIGHT;
@@ -446,6 +459,8 @@ function showMessagesScroller(msg) {
     }
     //print(MESSAGES)
   }
+
+  // Helper functions to handle positive and negative responses to messages
   var negHandler,posHandler = [ ];
   if (msg.negative) {
     negHandler = (msg)=>{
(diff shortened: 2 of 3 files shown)


thyttan:
I'm testing/developing this by generating test messages from the WebIDE console field:

```javascript
let i = 0;
let msgInterval = setInterval(
  ()=>{  
    print(i);
    GB({"t":"notify","id":1575479849+i,"src":"Hangouts","sender":"sender"+i,"title":"A Name"+i,"body":"message contents that are very similar to one another. in that they are somewhat long, and exactly the same. Let's have it go on and on and so forth for a while to make sure we overflow properly."});
    i++;
    if (i == 9) {clearInterval(msgInterval); delete i; delete msgInterval;}
  }, 500
);
```

thyttan:
@bobrippling If you'd help me by testing this a bit I'd be interested to hear your thoughts!

bobrippling:
> @bobrippling If you'd help me by testing this a bit I'd be interested to hear your thoughts!

Sorry I missed this the first time round - taken a look, I like it, very nifty feature to have! And it's quite easily discoverable too. Means I have less cause to dig my phone out, works really well.

I do have two thoughts:
- The scrolling direction makes sense on the message overview but then on returning to the router the direction is then the opposite. Maybe we swap that around, or add an option for it? (happy to leave this to another PR)
- When scrolling down there's a bit of a yank/lag as it loads in the next message, but oddly not if I go up and load in the previous. Do you see that?

thyttan:
> very nifty feature to have! And it's quite easily discoverable too. Means I have less cause to dig my phone out, works really well.

Thanks! I wanted the bigger font of the scroller for what pops up for legibility's sake.
 
> I do have two thoughts:
> 
> - The scrolling direction makes sense on the message overview but then on returning to the router the direction is then the opposite. Maybe we swap that around, or add an option for it? (happy to leave this to another PR)

Yeah, this feels a bit unfortunate to me as well.
- For the scroller it feels right to me how I've made it so the newest message is the one furthest to the top (similar to the list menu screen).
- For the overview screen (the function renamed from just "showMessage" to "showMessageOverview") that comes to the discussion of what is up/and down motions w r t swipes. People have different interpretations based on previous experiences as discussed [here](https://forum.espruino.com/conversations/388188/). When [implementing the swipes](https://github.com/espruino/BangleApps/pull/3215) I went with the style I believe Gordon uses for swipe interactions.
  - My thinking is that a relatively good solution to the different intuitions people have regarding this is to develop [the Swipe inversion app](https://banglejs.com/apps/?id=swipeinv) more.

>   - When scrolling down there's a bit of a yank/lag as it loads in the next message, but oddly not if I go up and load in the previous. Do you see that?

Yes, that is partly a limitation of the implementation. But I also find it a bit convenient that the scrolling stops at the end (or start, depending of if you scroll up or down) of the current message. So I don't scroll past the message by accident (I use [kinetic scroll](https://banglejs.com/apps/?id=kineticscroll) for my scrollers). The yank happens when another message is added to the array of wrapped strings and the scroller is reinited with that updated array. The yank (currently actually a 40ms timeout) also makes sure there are not other graphical glitches while loading in the next/previous message.

The yank doesn't happen when you go back because those messages are already part of the array of wrapped strings.

bobrippling:
> Thanks! I wanted the bigger font of the scroller for what pops up for legibility's sake.

Nice idea, I've not been using my watch with GB for a little while but this might bring it back. And your podcast controls I've recently spotted too - in that other thread

> > - The scrolling direction makes sense on the message overview but then on returning to the router the direction is then the opposite.
> Yeah, this feels a bit unfortunate to me as well.
> 
> - For the scroller it feels right to me how I've made it so the newest message is the one furthest to the top (similar to the list menu screen).
> - For the overview screen (the function renamed from just "showMessage" to "showMessageOverview") that comes to the discussion of what is up/and down motions w r t swipes. People have different interpretations based on previous experiences as discussed [here](https://forum.espruino.com/conversations/388188/). When [implementing the swipes](https://github.com/espruino/BangleApps/pull/3215) I went with the style I believe Gordon uses for swipe interactions.

Yeah that's reasonable, I hadn't actually clocked the order of the messages, I think that works well. I suppose we could always try it for a while, perhaps invert it ourselves and see if that feels more "natural". Or yes - try with [swipeinv](https://banglejs.com/apps/?id=swipeinv)

> > - When scrolling down there's a bit of a yank/lag as it loads in the next message, but oddly not if I go up and load in the previous. Do you see that?
> Yes, that is partly a limitation of the implementation. But I also find it a bit convenient that the scrolling stops at the end (or start, depending of if you scroll up or down) of the current message. So I don't scroll past the message by accident (I use [kinetic scroll](https://banglejs.com/apps/?id=kineticscroll) for my scrollers). The yank happens when another message is added to the array of wrapped strings and the scroller is reinited with that updated array. The yank (currently actually a 40ms timeout) also makes sure there are not other graphical glitches while loading in the next/previous message.

That's a good point about halting in fact - I think with `kineticscroll` especially it'll stand out as a clear and intentional stop. I think like above, we could try this for a while and see how it feels

> On another note, I'm a bit unhappy with how involved the code for processing, passing around and updating the `alreadyProcessed` data in `showMessagesScroller` is now. It feels like it's not super easy to read at first glance?

Yes I haven't fully got my head round it - I jumped in with a live test of it instead! I can have a go at suggesting some alterations (if I can come up with any), if you would find it useful to have another point of view?

thyttan:
> That's a good point about halting in fact - I think with `kineticscroll` especially it'll stand out as a clear and intentional stop. I think like above, we could try this for a while and see how it feels
> 
> > On another note, I'm a bit unhappy with how involved the code for processing, passing around and updating the `alreadyProcessed` data in `showMessagesScroller` is now. It feels like it's not super easy to read at first glance?
> 
> Yes I haven't fully got my head round it - I jumped in with a live test of it instead! I can have a go at suggesting some alterations (if I can come up with any), if you would find it useful to have another point of view?

I've refactored the scroller to just load in all the messages from the start. Makes the logic much simpler. The halting is gone by that token, not sure if it's a plus or minus.

But that meant I had to implement some way of deciding what messages should be deemed read by the user (i.e. `msg.new=false`). If there was a better way to do that I'd be happy - but I can't think of one now.

I'm now at a point where I don't think I'll go much further with the code by myself. I'd be thankful for you to give it a review @bobrippling!

bobrippling:
Not forgot - will get round to this soon :)

bobrippling:
> I've refactored the scroller to just load in all the messages from the start. Makes the logic much simpler. The halting is gone by that token, not sure if it's a plus or minus.

Nice, yeah this works very well. I do see the messages in a different order now (after running your [setup code](https://github.com/espruino/BangleApps/pull/3685#issuecomment-2570021424)) - zoomed out to the summary view:

image

Is this expected?

> But that meant I had to implement some way of deciding what messages should be deemed read by the user (i.e. `msg.new=false`). If there was a better way to do that I'd be happy - but I can't think of one now.

No I think this works for now - just to check, we're not off by one anywhere are we? Maybe we go up to just `shownMsgIdxLast`? And the user has to go to the end of it to mark it as read?

https://github.com/espruino/BangleApps/blob/447f1503c90c493185d333bf826f42be4e4eb1c8/apps/messagegui/app.js#L396-L398

> I'm now at a point where I don't think I'll go much further with the code by myself. I'd be thankful for you to give it a review @bobrippling! But no rush if you'd want to live test it some first, again.

Thanks - live tested and it looks good to me, just the two questions I have above :)

thyttan:
>> I've refactored the scroller to just load in all the messages from the start. Makes the logic much simpler. The halting is gone by that token, not sure if it's a plus or minus.
>
>Nice, yeah this works very well. I do see the messages in a different order now (after running your [setup code](https://github.com/espruino/BangleApps/pull/3685#issuecomment-2570021424)) - zoomed out to the summary view:
>
>image
>
>Is this expected?

I never understood why they ended up not entirely in order from 0 up to 8 before either. Or why the order changed with this latest change. Something about code execution time in different parts of the app? Maybe there's some logic somewhere that is a little off. Or not. But I *think* it is probably not a problem in normal use...

>> But that meant I had to implement some way of deciding what messages should be deemed read by the user (i.e. `msg.new=false`). If there was a better way to do that I'd be happy - but I can't think of one now.
>
>No I think this works for now - just to check, we're not off by one anywhere are we? Maybe we go up to just `shownMsgIdxLast`? And the user has to go to the end of it to mark it as read?
>
>https://github.com/espruino/BangleApps/blob/447f1503c90c493185d333bf826f42be4e4eb1c8/apps/messagegui/app.js#L396-L398

Hm. I looked at the logic again and I think it's not wrong. But maybe it should always mark the message we open to as read even if the user didn't scroll down to see all of that message's contents?

What is it that makes you suspect an off by one miss? So I know what I'm looking for.

bobrippling:
> I never understood why they ended up not entirely in order from 0 up to 8 (or maybe the other way around) before either. Or why the order changed with this latest change. Something about code execution time in different parts of the app? Maybe there's some logic somewhere that is a little off. Or not. But I _think_ it is probably not a problem in normal use...

Strange - perhaps a timing issue with the `GB()` callback and scheduling. Yes I don't think it's a problem though

> Hm. I looked at the logic again and I think it's not wrong. But maybe it should always mark the message we open to as read even if the user didn't scroll down to see all of that message's contents?
> 
> What is it that makes you suspect an off by one miss? So I know what I'm looking for. **EDIT:** Did a change in [1c19b13](https://github.com/espruino/BangleApps/commit/1c19b13b96415ab5d246e8b4e7ebd2e914b6447f) possibly fixing an off by one.

No you're right, I double checked and all is good - I was thinking it might trigger too early as the user scrolls not a full page down but you'd got it right already :)

thyttan:
Tried the updates here on Bangle.js 1 emulator. I found two problems so far:

1. Widget Utils seems to have become incompatible with Bangle.js 1 because of it's reliance on `Bangle.setLCDOverlay`.
```
>eval(require("Storage").read("messagegui.app.js"))
=undefined
Uncaught Error: Function "setLCDOverlay" not found!
 at line 16 col 549 in messagegui.app.js
exports.offset=-24,Bangle.setLCDOverlay(undefined,{id:'widge...
                          ^
in function "cleanupOverlay" called from line 16 col 863 in messagegui.app.js
exports.cleanupOverlay(),delete exports.offset,exports.swipe...
                       ^
in function "cleanup" called from line 16 col 105 in messagegui.app.js
if(exports.cleanup(),!global.WIDGETS)return;g.reset();for(va...
                   ^
in function "hide" called from line 308 col 11 in messagegui.app.js
  WU.hide();
          ^
in function "showMessagesScroller" called from line 640 col 40 in messagegui.app.js
    showMessagesScroller(newMessages[0]);
                                       ^
in function "checkMessages" called from line 745 col 22 in messagegui.app.js
    dontStopBuzz: 1 });
                     ^
```

[Note: [Bangle.setLCDOverlay] is only available in Bangle.js 2 smartwatches and DICKENS](https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setLCDOverlay)


2. The scroller works with my test messages, but when I try to exit to the message overview I get memory full error:
```
>
New interpreter error: LOW_MEMORY,MEMORY
```

bobrippling:
Nice spots - I don't have a Bangle1 but can give it another test on the 2 if you like?

thyttan:
Force pushed a restructuring of the commits so they look neater.

Before that, since my last comment I have fixed the messages scroller so exiting it doesn't cause low memory error on Bangle.js 1 emulator.

> Nice spots - I don't have a Bangle1 but can give it another test on the 2 if you like?

@bobrippling If you don't particularly feel like it I don't think it's super necessary - but thanks either way! Do we know someone other than Gordon who could test it on an actual Bangle.js 1 though?

bobrippling:
Nice, yeah I don't I'm afraid - nice idea with that discussion though

thyttan:
Made some changes. I have a memory leak when navigating back and forth in the app. I think especially for the showMessagesScroller.

bobrippling:
Thanks - shall take a look and let you know

bobrippling:
I've had a look through and seems ok, but I'm struggling to tell what's changed since my last review, particularly as github doesn't show the old commit names or ids any more, once the changes have been rebased. If they were merges it might be easier to spot new commits or have a history to diff against.

Of course I understand that rebases keep the history simpler, we could perhaps rebase once we're happy for the PR to be merged, to flatten out the merges and bring it in, what do you think?

---

For the changes, I think the new stuff that I don't recognise from before is:
- `showMessageRouter(msg, persist, explicitDestnation)`
- header and footer images
- updating of read messages

if that seems right to you?

bobrippling:
Re-tested and this works very nicely! One small thing that was initially confusing - on dismiss/swipe to the left, I'd expect the message to disappear from the scroller (as well as scrolling to the next message, which is neat), what do you think?

thyttan:
Thanks for the feedback! It should be possible to add that behaviour 👍

thyttan:
... Sorry for force pushing again. I'll try to get into the habit of not rebasing and merge in upstream/master instead...

bobrippling:
Ah no worries, that works perfectly well if you're able to remember the earliest new commits :) Shall take a look through if you're happy with the latest changes?

thyttan:
Only look through if you feel like it - I'm not particularly looking for more feedback right now. Still have some stuff tl work through.

thyttan:
I'll close this for now - doesn't mean I wont finish it. But I remember someone wanted to contribute to `messagegui` a while back. So by doing this I indicate that I don't wont to hold others off from doing that.

I'm also considering a slightly different approach - I'll see.

PR #3688, opened 2024-12-08
assistedgps: Fix combination of all systems

pavelmachek:
There was typo preventing all systems to work together.


gfwilliams:
Great - thanks!

PR #3692, opened 2024-12-11
andark: Restore redraw on charging and small fixups

buzden:
PR #3678 removed immediate redrawing on charging, which is inconvenient, since when you connect the charger, watches may be locked and thus not redrawing, not showing the charging status. Additionally, I did a small correction regarding to the cleanup order to make fast loading work well, since one timeout was removed before a callback that could reset it again. Also, I did some (I hope) reasonable refactoring. I did all in separate commits to make the review easier.

### DIFF
--- a/apps/andark/app.js
+++ b/apps/andark/app.js
@@ -12,9 +12,9 @@ const zahlpos=(function() {
   let z=[];
   let sk=1;
   for(let i=-10;i<50;i+=5){
-     let win=i*2*Math.PI/60;
-     let xsk =c.x+2+Math.cos(win)*(c.x-10),
-         ysk =c.y+2+Math.sin(win)*(c.x-10);
+    let win=i*2*Math.PI/60;
+    let xsk =c.x+2+Math.cos(win)*(c.x-10),
+        ysk =c.y+2+Math.sin(win)*(c.x-10);
     if(sk==3){xsk-=10;}
     if(sk==6){ysk-=10;}
     if(sk==9){xsk+=10;}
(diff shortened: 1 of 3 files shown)


apps/andark/app.js
```diff
@@ -62,41 +61,42 @@ let drawHands = function(d) {
...
   g.setColor(1,1,1);
   g.setBgColor(0,0,0);
   for(let i = 0;i<12;i++){
-     g.drawString(zahlpos[i][0],zahlpos[i][1],zahlpos[i][2],true);
+    g.drawString(zahlpos[i][0],zahlpos[i][1],zahlpos[i][2],true);
   }
 };
 
-let drawTimeout;
+let drawTimeout = undefined;
```
thyttan, line 87: Out of curiosity, what's the reason for being explicit here. Does the behavior change?

buzden: I just read in description of `let` that unlike `var` reading of uninitialised `let` may lead to an error instead of just an `undefined` value. I don't know if espruino interpreter does this, but I decided to be on the safe compatible side

bobrippling: You're thinking of hoisting and the [temporal dead zone](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#temporal_dead_zone_tdz) - this is fine:
```javascript
let a;

console.log(a);
```

This is not:
```javascript
console.log(a);

let a;
```

The code you have works either way though :)

buzden: Oh, yes, I see that I misread the specification and did useless thing. I deleted this piece and pushed the change.

bobrippling: No problem - ok I think we can merge if you're happy @thyttan ?

thyttan: I'll go ahead then!

PR #3695, opened 2024-12-12
fix(agenda): all day event offsets, locale display

tnyeanderson:
Currently, allDay events are off by one day for negative timezones. Per the CalendarContract:

    If allDay is set to 1 eventTimezone must be "UTC" and the time must
    correspond to a midnight boundary.

For example, in GMT-2:00, an all day event on December 2nd (beginning at 00:00:00) will be wrongly displayed as starting on December 1st, since the locale will determine that the event's start time is actually 22:00:00 on Dec 1.

Source:

https://developer.android.com/reference/android/provider/CalendarContract.Events.html

This commit:

* Corrects the offset back to UTC 00:00:00 for allDay events

* Fixes the conditional for single-day all day events in showEvent()

* Fixes the display of formatDateShort() for some English locales by also removing any trailing commas or whitespace when the year is removed

### DIFF
--- a/apps/agenda/agenda.js
+++ b/apps/agenda/agenda.js
@@ -30,11 +30,16 @@ var settings = require("Storage").readJSON("agenda.settings.json",true)||{};
 
 CALENDAR=CALENDAR.sort((a,b)=>a.timestamp - b.timestamp);
 
-function getDate(timestamp) {
-  return new Date(timestamp*1000);
+function getDate(timestamp, allDay) {
+  // All day events are always in UTC and always start at 00:00:00, so we
+  // need to "undo" the timezone offsetting to make sure that the day is
+  // correct.
+  var offset = allDay ? new Date().getTimezoneOffset() * 60 : 0
+  return new Date((timestamp+offset)*1000);
 }
+
 function formatDay(date) {
-  let formattedDate = Locale.dow(date,1) + " " + Locale.date(date).replace(/\d\d\d\d/,"");
+  let formattedDate = Locale.dow(date,1) + " " + Locale.date(date).replace(/,*\s*\d\d\d\d/,"");
   if (!settings.useToday) {
     return formattedDate;
   }
@@ -57,8 +62,9 @@ function formatDateLong(date, includeDay, allDay) {
   }
   return shortTime;
 }
+
 function formatDateShort(date, allDay) {
-  return formatDay(date)+(allDay?"":Locale.time(date,1)+Locale.meridian(date));
+  return formatDay(date)+(allDay?"":" "+Locale.time(date,1)+Locale.meridian(date));
 }
 
 var lines = [];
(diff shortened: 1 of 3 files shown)


tnyeanderson:
@glemco it seems like you're the de facto owner of the agenda app, so mentioning you here :)

apps/agenda/agenda.js
```diff
@@ -69,16 +75,18 @@ function showEvent(ev) {
   //var lines = [];
   if (ev.title) lines = g.wrapString(ev.title, g.getWidth()-10);
   var titleCnt = lines.length;
-  var start = getDate(ev.timestamp);
-  var end = getDate((+ev.timestamp) + (+ev.durationInSeconds));
+  var start = getDate(ev.timestamp, ev.allDay);
+  // All day events end at the midnight boundary of the following day. Here, we
+  // subtract one second for all day events so the days display correctly.
+  var end = getDate((+ev.timestamp) + (+ev.durationInSeconds) - ev.allDay, ev.allDay);
```
glemco, line 81: While this looks clever, requires the reader to check your comment to understand why you'd subtract what's basically a boolean value from a timestamp. I'd be a bit more verbose just for the sake of clarity, what do you think?
```suggestion
  // All day events end at the midnight boundary of the following day. Here, we
  // subtract one second for all day events so the days display correctly.
  const allDayOffset = ev.allDay ? 1 : 0;
  var end = getDate((+ev.timestamp) + (+ev.durationInSeconds) - allDayOffset, ev.allDay);
```

apps/agenda/agenda.js
```diff
@@ -30,11 +30,16 @@ var settings = require("Storage").readJSON("agenda.settings.json",true)||{};
...
+  // All day events are always in UTC and always start at 00:00:00, so we
+  // need to "undo" the timezone offsetting to make sure that the day is
+  // correct.
+  var offset = allDay ? new Date().getTimezoneOffset() * 60 : 0
+  return new Date((timestamp+offset)*1000);
 }
-function formatDay(date) {
-  let formattedDate = Locale.dow(date,1) + " " + Locale.date(date).replace(/\d\d\d\d/,"");
+
+function formatDay(date, allDay) {
```
glemco, line 41: `allDay` seems unused in this function, if you really don't need it, let's remove it from the call as well
```suggestion
function formatDay(date) {
```

apps/agenda/agenda.js
```diff
@@ -69,16 +75,18 @@ function showEvent(ev) {
...
+  // subtract one second for all day events so the days display correctly.
+  var end = getDate((+ev.timestamp) + (+ev.durationInSeconds) - ev.allDay, ev.allDay);
   var includeDay = true;
   if (titleCnt) lines.push(""); // add blank line after title
   if(start.getDay() == end.getDay() && start.getMonth() == end.getMonth())
     includeDay = false;
-  if(includeDay && ev.allDay) {
-    //single day all day (average to avoid getting previous day)
+  if(!includeDay && ev.allDay) {
+    //single day all day
```
glemco, line 87: Good catch, that was actually not too readable and I got myself confused.. I wonder if we should change it to something like `sameDay`, that'd probably be more meaningful

glemco (approved):
Thanks for taking care of this! I remember testing this across time-zones was quite convoluted and I didn't know about this Contract.
Added some minor comments but it looks good to me.

apps/agenda/agenda.js
tnyeanderson: I agree that this should be renamed, but `sameDay` would reverse the current boolean's values, and that variable is used in quite a few places that wouldhave to be fixed.

I was already planning on doing a second pass with some cleanup like this to help make the code a little more clear, so I'll take care of the variable renaming then to avoid having too many changes in this PR, which might make it harder to understand.

apps/agenda/agenda.js
tnyeanderson: Hilariously, that's exactly what I had wanted to do, but decided to be "clever" since it's an embedded device with minimal resources, etc (in the name of efficiency). Even told the wife how I already hated what I'd done :)

Every time I do something "clever", I regret it, so I'm glad we agree! lol

tnyeanderson: To avoid colliding with the "offset" concept in `getDate()`, I chose a slightly different name than you did above. Let me know if this is alright with you.

apps/agenda/agenda.js
tnyeanderson: Thanks for catching this, I changed where the logic was a couple times and this is a remnant of that!

tnyeanderson:
Should be ready for a final review. Thanks!

apps/agenda/agenda.js
glemco: Yes good idea, feel free to add me as a reviewer

apps/agenda/agenda.js
glemco: Ahahah nice one. As a side note, I'm not sure what the jit compiler would do but I'd bet a constant like this would have negligible effect.
The name you put is even clearer, go ahead with it, thanks

tnyeanderson:
Forgot to increment version and add the changelog entry... NOW we are good :)

glemco:
@tnyeanderson did you forget to push? There's still only one changed file in this PR

tnyeanderson:
Sorry, was in a rush and didn't `git add` properly :)

glemco (approved):

bobrippling:
Nice work, thanks for the PR @tnyeanderson and the review @glemco!

PR #3697, opened 2024-12-17
[gpssetup] Bangle.js2 support

nravanelli:
This is an update to the gpssetup app/module to emulate the PSMOO feature in the Bangle.js1 on the Bangle.js2.

### DIFF
--- a/apps/gpssetup/ChangeLog
+++ b/apps/gpssetup/ChangeLog
@@ -1,2 +1,3 @@
 0.01: First version of GPS Setup app
 0.02: Created gppsetup module
+0.03: Added support for Bangle.js2
--- a/apps/gpssetup/README.md
+++ b/apps/gpssetup/README.md
@@ -48,12 +48,16 @@ used. These settings will remain for all apps that use the GPS.
    the interval the more time the GPS will spend sleeping in low
    power mode (7mA) between obtaining fixes (35mA).  For walking in
    open country an update once every 60 seconds is adequate to put
-   you within a 6 digit grid refernce sqaure.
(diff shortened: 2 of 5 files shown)


pavelmachek:
I guess warnings should be fixed :-).

Plus.. what does it do? Am I reading it right that it simply turns off GPS just after acquiring a fix? Some documentation would be nice.

(And that approach may be too simplistic. I played with powersaving a lot in sixths. You can take a look or I can try to explain.)

nravanelli:
@pavelmachek thanks for the feedback - the errors are github interpretations of the code, worked fine in my development, but I have refined code to pass the checks. Took a couple of times, wanted to test out githubs codebase.

You are right, it is a basic on/off timeout for the GPS to try and emulate the PSMOO that was available on the GPS module of the Bangle.js1. What do you mean by sixths? Eg, in sixths such as 1:6? Would be happy to try and implement/refine.

pavelmachek:
Yes, there seem to be warning checks on commit, and it does not play well with my workflow, either.

sixths is clock application in apps/sixths/ , it tries to do similar powersaving logic, and I had to play with it a lot to get something reasonably working.

In particular, if you get a fix, you know you have good reception, and you want to keep receiving for a while, to grab ephemeris data etc. I keep running it for 20 seconds minimum and 6 minutes maximum, depending on how long it took to take a fix in the first place. Logic is in gpsHandle() function.

bobrippling:
Nice extra feature @nravanelli  - and thanks for the extra details/review @pavelmachek

PR #3701, opened 2024-12-23
Contacts 0.05: Nicer UI, Refactoring, new icon

devsnd:


apps/contacts/contacts.app.js
```diff
@@ -17,35 +17,49 @@ function callNumber (number) {
...
     menu['No Contacts'] = () => {};
-  } else {
-    for (const e of wp) {
-      const closureE = e;
-      menu[e.name] = () => showContact(closureE);
-    }
   }
-  menu["Add"] = addContact;
-  menu["Remove"] = removeContact;
+  contacts.map((e, idx) => {
```
bobrippling, line 32: You can use `.forEach` if you don't use the result of `map`, so we avoid creating a new array :)

```suggestion
  contacts.forEach((e, idx) => {
```

bobrippling (commented):
Looks good, I'll give it a test soon (there's one linter problem outstanding too)

devsnd:
Thanks @bobrippling ! I'll try to fix it asap, but it would be a Christmas miracle if I make it today ;)

Thanks for your suggestions and happy Christmas holidays :)

devsnd:
Oh cool, I didn't realize I could just fix it on the phone in the browser...! The linter caught a mistake I made during refactoring, so that check was well worth it. It should be all ready now 🤞

bobrippling:
Nice job, thanks for the quick fixes!

PR #3705, opened 2024-12-31
Fixed taglauncher crashing with certain apps installed

BlueFox4:
If an app has no tags (e.g. [Drained](https://banglejs.com/apps/?id=drained&readme), in my case), it caused taglauncher to crash because of a missing check (and in turn running .split on undefined). This missing check was added and apps without the tag are now added to the Misc category automatically.
I also fixed some code indentation issues.

Real changes to app.js only from lines 51-61 in the old app.js (or ll. 75-90 in the new app.js).

### DIFF
--- a/apps/taglaunch/ChangeLog
+++ b/apps/taglaunch/ChangeLog
@@ -4,3 +4,4 @@
       Add tag 'health' for apps like Heart Rate Monitor
 0.04: Fix remove handler
 0.05: Make the "App source not found" warning less buggy
+0.06: Fixed a crash if an app has no tags (app.tags is undefined)
--- a/apps/taglaunch/README.md
+++ b/apps/taglaunch/README.md
@@ -26,5 +26,5 @@ Creator
 Contributors
 --------
 
-[atjn](https://github.com/atjn)
-[BlueFox4](https://github.com/BlueFox4)
+- [atjn](https://github.com/atjn)
+- [BlueFox4](https://github.com/BlueFox4)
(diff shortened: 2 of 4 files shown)


thyttan:
Do you think you could break that first commit up in one for the bugfix and one for the indentation change if it's not too much trouble?

BlueFox4:
> Do you think you could break that first commit up in one for the bugfix and one for the indentation change if it's not too much trouble?

Yes, I'm going to do that now. (It's my first PR, so hopefully i wont break things... Otherwise, I'll open another PR and close this, which shouldn't be a problem, right? =)

BlueFox4:
I believe I got it... Please tell me if something's still missing 😉

thyttan:
Tried it on my watch - seems to work well (I didn't try to reproduce the bug though). Thanks for the fix!

Thinking again, lets not do the indentation change at all.  I think it already follows the style guide pretty well (e.g. 2 spaces for indentation): https://www.espruino.com/Code+Style

If you can drop the commit entirely that's neat. But reverting it is also fine :)

Thanks for doing your first contribution! 🚀

BlueFox4:
Done.

thyttan:
Nice - thanks for doing that! Merging now 🎉

thyttan:
The readme changes are good 👌

thyttan:
It's nice that you put the readme changes in separate commit/s when they didn't have to do with the other changes directly.

PR #3715, opened 2025-01-12
[weatherClock] Add yellow to error icon to show on dark theme

thinkpoop:
Link: https://thinkpoop.github.io/BangleApps/?id=weatherClock

The error icon in weatherClock is all black, so it doesn't show up in the dark theme. This adds a yellow fill to fix that. 

It does increase the size from 259 to 287 characters. I used "Optimal 2 bit" for the encoding. There are probably other options to minimize the size if needed, like using white instead of yellow.

### Before
icon-before-light icon-before-dark

### After
icon-after-light icon-after-dark


bobrippling:
Nice, thanks!

PR #3717, opened 2025-01-17
Add Reloader App

Sleuth56:
A very basic app to reload the Banglejs2 after an amount of time has passed. The goal of this app is to prevent crashing and memory leaks.


thyttan:
Hi - thanks for your first PR!

I'm personally a little hesitant towards this approach. Could you please provide some background to why you needed this in the first place? Both as a rationale for the PR and so we can maybe fix the original problem.

Ideally we should strive to make the code on the watch not crash and not have memory leaks. If we can identify problems we should track and fix them.

When/if we pull this in I think we should add a paragraph in the readme asking for the user to report the problem that caused them to install the Reloader app.

From the wiki *(doesn't apply entirely to this reloader app though but is part of the basis of my skepticism)*:
```markdown
# For modifications to existing apps/modules:
[...]
- Avoid hacks to work around particular problems. If something is broken in some other part of Bangle.js then ideally we should raise a bug for that and fix it, rather than adding a hack to every app that wants to use that functionality.
```
*https://github.com/espruino/BangleApps/wiki/App-Contribution#for-modifications-to-existing-appsmodules*

Sleuth56:
I've had my Banglejs2 for over a year now and I've ran several different combinations of software on it. I used to have memory leak problems than I switched out what I was running on it and now I'm not sure if it's memory related but my watch stops connecting to my phone over night. This happens every night. When I get up I have to reload the watch to get it to connect again. 

Troubleshooting it would be the normal way to resolve these kind of problems. They problem is troubleshooting a problem that takes hours to show up is very hard if not impossible. Because the watch will eventually no longer connect to a device over bluetooth you can't debug it once it gets into that state.

Sleuth56:
I agree that putting a line in the README to have people report the problems instead of bandaging them with this app is a good idea.

thyttan:
> I switched out what I was running on it and now I'm not sure if it's memory related but my watch stops connecting to my phone over night. This happens every night. When I get up I have to reload the watch to get it to connect again.

I'd like to put this with a bug report, either as extra info if it already exists or as a new one. Could you do it by clicking the link in the App Loader 'More...' tab? That way most of the relevant info will be auto-filled, I think.

thyttan:
I'm closing this for now. Feel free to reopen this or open a new PR if you want to discuss more. 

Thanks again for this PR and for filing that issue!

thyttan:
I'm closing this for now. Feel free to reopen this or open a new PR if you want to discuss more.

Thanks again for this PR and for filing that issue!

PR #3719, opened 2025-01-25
Add Dense Clock app

yogsoy:
(Available at https://yogsoy.github.io/BangleApps/?id=denseclock)
This clockface puts battery level, power usage, air pressure, step count, and bluetooth status in plaintext all on one screen.

I designed and built a lot of this a couple of years ago and gutted and rewrote it to make this version. Used the example clock and antonclk among other apps to bring it up to date.


thyttan:
Thanks! The checks fail, missing ChangeLog

thyttan:
Thank you for your first addition to the app loader! 🥳

thyttan:
Sorry, just noticed that the app doesn't seem to have an icon on my watch. Could you add that in another PR please?

thyttan:
Ok! I tried both dark and light theme with same result.

PR #3720, opened 2025-01-25
Denseclock: app icon update

yogsoy:
Looks like the app icon for denseclock was having some issues in my initial version. I gave it a bit of background so its easier to see in different themes, and re-encoded it so it'll now display properly.


thyttan:
Nice, thanks for the quick fix!

PR #3721, opened 2025-01-26
Home Assistant Control app

yogsoy:
( Available at https://yogsoy.github.io/BangleApps/?id=hasscontrol )

_Control your Home Assistant instance from your Bangle. Uses GadgetBridge's HTTP capability to send API requests directly from your watch._

Another rewrite of an old app I wrote. Missing some documentation but there's enough (an example config and a config editor) for an enterprising person to start using.

Plans for the future might include some app settings, a more graphical config editor (rather than just writing JSON), and of course a README.


thyttan:
I have no ha stuff to try this. I believe @bobrippling and @gfwilliams do?

bobrippling:
Works a treat, tested the `number` type with a light and the on/off/toggle feature too 👍

bobrippling:
There is already [`hadash`](https://yogsoy.github.io/BangleApps/?id=hadash) which is very similar - perhaps we look at merging functionality into it? @flaparoo what would you think?

bobrippling:
Yeah I like your number input - the instant feedback is very cool. I don't think @flaparoo would mind about contributions either, and they normally respond within a few days.

Thanks, I'll be ready for your next PR :)

flaparoo:
Go ahead - please improve hadash :-)

A number selector would be a nice addition.
And as the read-me mentions, a better way to create the config would be nice too (rather than editing JSON manually).

Issue #3707, opened 2025-01-05
[calendar][agenda] Feature request: calendar view of Gadgetbridge-synchronized events


randyheydon:
`calendar` is a great app, having a nice calendar view and a nice agenda view.  But it only shows events that have been explicitly added to the app; it does not show any events from my regular calendar on my phone (it appears to read from "calendar.json").

`agenda` is a great app, having a nice agenda view of events synchronized from my phone via Gadgetbridge (it appears to read from "android.calendar.json"), and a detail view for each event.  But it can't show those events in or alongside a calendar view.

I'm using both apps, but I find myself switching back and forth between them to make sense of when things are happening.  It would be nice to have one app that can show synchronized events in a calendar view.  I see three possible approaches for this:
1. Enhance `calendar` to also read events from the synchronized calendar.
2. Enhance `agenda` to add a calendar view.
3. Create a new app that does all of this.

For reference, I have a Bangle 2 on firmware 2v25, with `calendar` 0.18 and `agenda` 0.15.  I also gave a quick try to `rebbleagenda` 0.02 and `calclock` 0.09, but like `agenda`, those do not have calendar views.

nxdefiant:
I would prefer option 4. Have calendar and agenda use the same appointment source and only this. I see no point in having multiple apps using their own data source. This should be handled similar to waypoints or scheduler.

I was not aware of agenda when I added the holidays to the simple calendar, this is why it uses its own. Agenda seems to read android.calendar.json - does anyone have a description of this file? Does it cover all features of calendar? Basically it is only
- day
- title
- type holiday/other
- repeat yearly

nxdefiant:
> The "type" entries aren't the same between calendar.days.json and android.calendar.json (int vs. string)

as long as we can identify a mapping this should be ok

> there's no repeat information at all in android.calendar.json

can we just add a field? Non Gadgetbridge users will just ignore it and it just won't be set or needed for Gadgetbridge user.

> Also, I suspect that android.calendar.json is effectively read-only, since it's synced from Gadgetbridge. It might be possible to write to android.calendar.json, but those writes won't be sent back to the phone and may be overwritten by Gadgetbridge (I'm guessing, this _might_ be avoidable if id values don't conflict). If you want `calendar` to drop use of calendar.days.json entirely, then it might have to give up the ability to add entries, making it unable to work well without a paired phone. That wouldn't bother me, but other users might not like that.

This is not good. However see answer to point 5 below

[...]

PR #3722, opened 2025-01-29
calendar: read events synchronized from Gadgetbridge

randyheydon:
This is a partial fix for feature request #3707.  This allows synchronized events (stored in "android.calendar.json") to be shown in the app, alongside the manually entered events and alarm events that are already shown.

This is a sort of minimum viable implementation of the feature request, as it does not address most of what @nxdefiant and I discussed.  I hope we can work toward those in the future.  Specifically:

- The file "calendar.days.json" is still read from and written to when creating events in the app.
- All events from "android.calendar.json" are assumed to not be holidays, just regular events.
- No UI is added to expose the extra information available in the synchronized events (e.g. duration, description).

### DIFF
--- a/apps/calendar/ChangeLog
+++ b/apps/calendar/ChangeLog
@@ -19,3 +19,4 @@
       Display Widgets in menus
 0.17: Load holidays before events so the latter is not overpainted
 0.18: Minor code improvements
+0.19: Read events synchronized from Gadgetbridge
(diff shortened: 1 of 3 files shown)


bobrippling:
LGTM, happy to merge - we can either ignore those warnings or if you wanted to pop in the `let`s, go for it, it's a pretty low-risk change

randyheydon:
Looks like the `let`s did the trick with linting, and calendar still runs on my watch.  So I'm happy with this code if everyone else is.

bobrippling:
Nice, thanks!

PR #3723, opened 2025-01-29
confthyttan: (new) Install a preconfiguration of apps and settings

thyttan:
This is the configuration I use on my watch. App Loader link: https://thyttan.github.io/BangleApps/?id=anotherconf

Maybe we could have a section/tag for stuff like this? Maybe named "configs", "preconfigs", "compilations", "collections" or something. What do you think @gfwilliams @bobrippling @halemmerich?

Reference this forum thread: https://forum.espruino.com/conversations/371248/?offset=25#16964620

Some warnings I'm not sure if I should handle or add to known warnings.


gfwilliams:
When you say 'Another Default Config` - you mean as opposed to the actual default, or is there another app that does this already?

It might make more sense to call this `Thyttan's Default config` though - as you say it's your ideas for defaults, and I don't want 'Another Default Config++`, 'Yet Another Default Config`, and so on if everyone else jumps on.

> Maybe we could have a section/tag for stuff like this? 

Definitely tag it as `defaultconfig` so it can be searched but I wouldn't add another section for it in the App Loader until we get 3+ configs - it's busy enough as-is and I'd rather not have sections with just one item in them.

> Boot code that deletes the app

Maybe just define the app type as `RAM` and then the app loader won't try and install an `app.info` or expect a JS file - but it should install all the other stuff.

> App anotherconf uses clock_info but doesn't have clkinfo tag

I wouldn't depend on clkinfo at all - other apps you're installing will depend on clockinfo and so they should install it automatically.

> App anotherconf storage file widbt_notify.json is also listed as data file for app widbt_notify

We're always going to have that as an issue - but perhaps there should be a new app type `defaultconfig` that does the same thing as `RAM` but that automatically erases everything on the watch first, and doesn't show these overlap errors? That would make this whole thing a lot tidier.

## OR

The current 'Install Default Apps' button uses https://github.com/espruino/BangleApps/blob/master/defaultapps_banglejs2.json (there's one for Bangle.js 1 as well), and I wonder whether actually we should just extend that file to include multiple defaults that can be chosen from a popup list, so instead of `[array_of_apps]` it's:

```
[
 { name : "Default",
   description : "Bangle.js apps as installed from the Factory",
   apps : [array_of_apps]
 },
 { name : "Thyttan's suggested apps",
   description : "...",
   apps : [array_of_apps],
   override : { "my_settings.json" : "new_contents" , ... } 
 }
]
```

... but maybe the custom app is easier - it's a lot less modification (we can always put a link next to 'Install Default Apps' that shows all the `defaultconfig` tags)

pavelmachek:
There's https://github.com/espruino/BangleApps/wiki -- with "recommended apps" section. I believe that may be worth extending with other useful tips.

As for default config, I believe folderlauncher would be good substitute for default launcher, for example, but... I kept adding tips to the wiki instead of trying to change the defaults.

bobrippling:
I like the sound of a custom app or install-default-apps choice, maybe using the custom loader

thyttan:
# In response to gfwilliams
> It might make more sense to call this Thyttan's Default config though

Yes that makes sense - changed now!

> Definitely tag it as defaultconfig so it can be searched

Did that now. 👍

> Maybe just define the app type as RAM

Done.

> I wouldn't depend on clkinfo at all

Removed it.

> We're always going to have that as an issue - but perhaps there should be a new app type defaultconfig that does the same thing as RAM but that automatically erases everything on the watch first, and doesn't show these overlap errors? That would make this whole thing a lot tidier.
> OR
>
>The current 'Install Default Apps' button uses https://github.com/espruino/BangleApps/blob/master/defaultapps_banglejs2.json (there's one for Bangle.js 1 as well), and I wonder whether actually we should just extend that file to include multiple defaults that can be chosen from a popup list, so instead of [array_of_apps] it's:
> [...]
> ... but maybe the custom app is easier - it's a lot less modification (we can always put a link next to 'Install Default Apps' that shows all the defaultconfig tags)

Yeah, both works for me. This path with an app is maybe more approachable for developers? While your idea seems a bit more elegant for a user, easier to stumble upon an so on.

# In response to pavelmachek

That's neat - thanks for reminding me!

# In response to bobrippling

> maybe using the custom loader

I don't get this part - what do you mean? 🤔🙂

bobrippling:
> > maybe using the custom loader
> 
> I don't get this part - what do you mean? 🤔🙂

Sorry, I meant this part of Gordon's response:

> ... but maybe the custom app is easier - it's a lot less modification (we can always put a link next to 'Install Default Apps' that shows all the defaultconfig tags)

gfwilliams:
I'm just checking it now - looks like `msgtwscr` doesn't exist, so I'll remove that for now?

thyttan:
Ah, sorry. Didn't mean to leave that in there now. Yes delete that line - or I can do it and push if you want?

thyttan:
@gfwilliams There - msgtwscr is a goner. (for now)

gfwilliams:
Ok, I'm pretty sure this is now all merged. The `defaultconfig` app type now automatically erases the bangle, and prompts the user twice just to make sure!

thyttan:
> As @pavelmachek says it'd be great to mention this on https://github.com/espruino/BangleApps/wiki too as a different option - maybe others will want to create new apps for default installs.

Did this now: https://github.com/espruino/BangleApps/wiki#alternative-default-configurations

PR #3724, opened 2025-01-30
Grandfather Clock Widget

yogsoy:
( Available at: https://yogsoy.github.io/BangleApps/?id=grandfather-clock )

A widget that uses vibrations to count fractions of the hour and the hour of the day, much like an old grandfather clock.

I did try to work on this as a set of options for Chimer, but found that it was easier and more efficient to make a separate widget. Keeps both apps lighter for people that need them - there's quite a bit of logic that went into the counting function.

Goals:
- Recreate(?) the _[Taptic Time](https://support.apple.com/en-au/guide/watch/apd2c755c294/watchos)_ accessibility feature from devices such as the Apple Watch, reminiscent of the chimes of old clocks.

Current considerations:
- Is the name descriptive enough? Happy to hear any other options.
- App setting configuration isn't there yet. It's in the todo for v0.02 (arbitrary). I've fiddled a lot and I think the default configuration will be familiar for most people that would be interested.
- I'd like a proper icon for both the app loader and on-screen widget. The one included in this PR is just an icon from material.io. (I am not very good at images)
- On-screen icon is to be made optional, I just found it useful to double-check the widget was running and to remind me why my wrist was vibrating. I'd like to make a tap on the widget buzz out the whole time.


thyttan:
> * Is the name descriptive enough? Happy to hear any other options

Works for me! I'd prefer though if the app id was just "grandfatherclock" since most app ids only use letters.

> * I'd like a proper icon for both the app loader and on-screen widget. The one included in this PR is just an icon from material.io. (I am not very good at images)

Gordon pays icon8.com for a license to use the icons they provide - maybe you can find one you like there? https://www.espruino.com/Bangle.js+First+App#icon-timer-img

> * On-screen icon is to be made optional, I just found it useful to double-check the widget was running and to remind me why my wrist was vibrating. I'd like to make a tap on the widget buzz out the whole time.

👍

thyttan:
So what you do is set the widgets draw method to `undefined` if it shouldn't draw. 

I guess if you tested it and it works we're good - what do you think @bobrippling ? 

What many other widgets seem to do is change the width to zero and the redraw. But that is maybe only needed if it's not statically either shown or not.

bobrippling:
Yeah looks good to me - I think perhaps we change to a zero width + check in the widget draw method to match the others, so if some future code assumes `draw` is there, the future code won't bail at that point, wdyt?

thyttan:
Or just change `: undefined` to `:  ()=>{}`. That way draw will always be a function as well.

thyttan:
Just say when you tested it again and I will merge 🙏☺️

thyttan:
Thank you 🎆

PR #3725, opened 2025-02-01
added a settings page to multidice

Le-Kat:
I added a settings page for multidice, allowing users to change what the app considers a 'roll', alongside some timers to change how the watch vibrates.


Le-Kat:
oh my god I've been commiting to the wrong branch

PR #3726, opened 2025-02-01
bwclk & bwclklite: Support 12-hour time format.

randyheydon:
This modifies how the time string is generated so it respects the system's locale settings, specifically for 12-hour time.  This is done with the locale package, which also simplifies some of the string handling since the package creates a complete time string.  Note this does not add any display of AM/PM, but neither do any other clock faces.


thyttan:
This makes sense - thanks! I take it you tested both apps with this change and they worked? After you answer this I can merge.

randyheydon:
Yes, of course.  I installed both, confirmed both will show 12- and 24-hour time, and confirmed both respect the "Hide Colon" setting.  I'm now using bwclk as my daily driver and haven't run into any issues.

If you or anyone else wants to test it before it's merged, I switched my Github Pages to pull from this branch, so you can get it at https://randyheydon.github.io/BangleApps/.

thyttan:
Thank you once again! 🚀🌸

PR #3730, opened 2025-02-02
Fix with counter2's default values in settings

voloved:
Prior to this change, the default counter values were blank in settings.
It was referencing `settings` as an array, rather than the attributes of the object.

Here's an Issue Ticket referring to the problem: https://github.com/Stuff-etc/BangleApps/issues/2


thyttan:
Thanks!

PR #3732, opened 2025-02-03
New app: txtreader

tonykakuuu:
A simple txt file reader. Does not yet include a way to upload files, though this can be done via the IDE.


pavelmachek:
Perhaps squash the commits? Anyway, looks interesting, I recently put notes onto watch by just doing E.showMessage("...") and this is certainly better.

PR #3734, opened 2025-02-04
New app: txtreader

tonykakuuu:
A simple txt file reader. Does not yet include a way to upload files, though this can be done via the IDE.
My apologies for the closed PRs, I accidentally closed the last one when trying to rebase my commits.


pavelmachek:
Thanks, it works for me (somehow). Font is very small and I can't advance past page 2 on file I tried. I guess it may have something to do with line that has 450 characters. I guess it would be best to merge this and then I can submit patches...

thyttan:
> I guess it would be best to merge this and then I can submit patches...

You could also submit patches as a PR directly to @tonykakuuu's master branch tonykakuuu/BangleApps. Always nice if the app is polished ahead of merging into espruino/BangleApps 😺🙏

pavelmachek:
I still need to find time, plus I'm not sure if my git-fu is good enough for that. Please don't wait for me.

pavelmachek:
The "won't go after page 2" issue can be fixed like this:

```
         y += 10; // Move down for the next line
         linesDisplayed++;
         if (y >= g.getHeight() - 10) {
           // If we run out of space, stop drawing
+          if (i == startLine)
+            // If we did not move single line, force it
+            i += 1;
           return { nextStartLine: i , linesDisplayed: linesDisplayed };
         }
       }

```

tonykakuuu:
Thanks for the feedback! After testing I've noticed that rendering new pages doesn't work on actual hardware beyond a certain file size. A potential solution would be to use offsets/chunks when reading the file (luckily storage.read already allows for that). I'll try that later and report back.

pavelmachek:
Yeah, getting arbitrary sized display on hardware with limited amount of RAM is always going to be tricky. I guess you could just do `var text = require("Storage").read(file, 0, 32768);` and be done with that for now. [also note there are two types of files, and this currently only works with one type.]

Better solution would be to scan file in a single pass, and store offsets for start of each page...

thyttan:
At one point I was trying to do something like this app you create @tonykakuuu: https://forum.espruino.com/conversations/380524/

Code here if it would be interesting: https://github.com/thyttan/BangleApps/tree/books/apps/books

thyttan:
It seems to work really well!

One thing to fix - it works on light theme but not on dark theme where the text is black on black background. If you want to hardcode the background white or make the text color change with the theme is up to you.

Once that's sorted I think we should merge 👍

pavelmachek:
Thank you, I briefly tested it and it works ok.

thyttan:
Thank you!

thyttan:
A heads up, doesn't seem to break the app but:

```
 ____                 _
         |_| espruino.com
 2v25.40 (c) 2024 G.Williams
Uncaught Error: Unhandled promise rejection: Error: Can't read property 'linesDisplayed' of null
 at line 12 col 57 in txtreader.app.js
...ffset,linesDisplayed:result.linesDisplayed});
                              ^
in function "onFileSelected" called from line 1 col 213 in txtreader.app.js
if(confirm){onFileSelected(file);}else{showFileSelector();}
                                  ^
> 
```

PR #3735, opened 2025-02-06
widbt_notify: redraw widget on NRF dis-/connect

thyttan:


thyttan:
Looking more closely it seems this change shouldn't be needed...

PR #3738, opened 2025-02-07
New app "Shortcuts"

WhelanB:
Adds a new "Shortcuts" app which allows you to create shortcuts to apps that act as ``clock_info`` items, can be placed on Clocks that support ``clock_info``, and launch into the app when interacted with.

You can add a shortcut from the app:

Select an app:
image

Select an icon (since ``clock_info`` icons are 1bpp 24x24, unlike app icons):
image

If you have a ``text_input`` keyboard installed, you can optionally rename the shortcut.

It'll then appear on your Clock like so:
example_shortcuts_screenshot

I've excluded Bangle.JS 1 support as I don't have a Bangle.JS 1 to test on and I'm not quite sure how ``clock_info`` acts outside of the Bangle.JS 2. I also only noticed that the name and icon are similar to a Bangle.JS 1 app after going to open this PR, but the functionality is different enough to warrant a new app.


thyttan:
Neat!

If the name is too long it seems "..." is displayed instead of the string. If that is something you'd want to change let's do it in another PR - merging now. Thanks!

WhelanB:
Thanks! I did notice the "..." which is what led me to add the ability to rename if a keyboard is installed, but I might take a look at using the ``short`` value on the clock_info items and truncate the name down, since it might be a more suitable fallback than the triple dots that Pebble++ displays!

PR #3739, opened 2025-02-08
terminalclock: Add option for ISO 8601 date format

JimMadge:
This PR adds a boolean setting for the terminalclock app. When true the date line will display render in the ISO 8601 format (like, 2025-02-08) rather than using `locale`.

The setting is false by default to preserve backwards compatibility with existing settings files.

I've tested from my fork using my Bangle.js 2.


apps/terminalclock/app.js
```diff
@@ -113,47 +114,57 @@
...
     let time = ">" + (""+h).substr(-2) + ":" + ("0"+m).substr(-2);
     drawLine(time, pos);
   };
 
-  let drawDate = function(now, pos){
-    let dow = locale.dow(now, 1);
-    let date = locale.date(now, 1).substr(0,6) + locale.date(now, 1).substr(-2);
-    let locale_date = ">" + dow + " " + date;
-    drawLine(locale_date, pos);
+  let drawDate = function(now, isoDate, pos) {
```
bobrippling, line 124: We don't use this parameter - let's delete it (and update call sites)

bobrippling:
Thanks for the PR - all of the whitespace changes are in with code changes, can you split them into separate commits so it's easy to view functional changes to the code? (and if we need to come back later, we won't have conflicts due to whitespace if we need to cherry-pick or otherwise move the commit around)

JimMadge:
> Thanks for the PR - all of the whitespace changes are in with code changes, can you split them into separate commits so it's easy to view functional changes to the code? (and if we need to come back later, we won't have conflicts due to whitespace if we need to cherry-pick or otherwise move the commit around)

Sure, no problem. I'll get on to that soon. (I read the note in the contributing guide about whitespace changes too late 😅)

apps/terminalclock/app.js
JimMadge: Thanks, I think this must be left over from when I was trying to pass it before realising `clock` was in scope :+1:.

bobrippling:
No worries, and thanks!

PR #3740, opened 2025-02-08
[gbmusic] Fix #3737 and a handful of other fixes

### LINKED ISSUE #3737: [gbmusic] scrolling track titles cause track numbers to not clear
### Affected hardware version

Bangle 2

### Your firmware version

2v25

### The bug

Documenting/reporting this since I haven't been able to figure out if the cause is in `gbmusic` or in `Layout`. (I'm assuming gbmusic, but just in case.)

## Issue

If a track title is long enough to scroll, the track number doesn't clear properly when going from a longer to a shorter string. (e.g. from `#10/12` to `#9/12`, it leaves `##9/12`)

This only happens if function `rScroller` is hit. Shorter track titles that don't scroll don't cause the issue.

## To reproduce

1. Open `gbmusic` in the emulator: https://banglejs.com/apps/?id=gbmusic
2. Select Bangle 2 
3. Run the following in the console:

```js
info({"artist": "", "album": "", "track": "The Track Title", "c": 12, "n": 12});
info({"artist": "", "album": "", "track": "The Track Title", "c": 12, "n": 2});
```

This simulates changing from track `#12/12` to `#2/12`. 

Image

If you shorten the "track" value to just "The", it won't scroll and will clear the value correctly:

```js
info({"artist": "", "album": "", "track": "The", "c": 12, "n": 12});
info({"artist": "", "album": "", "track": "The", "c": 12, "n": 2});
```

## Workaround

I have a rough workaround, but the actual bug is beyond me so it isn't actually a "fix".

Adding the following in `function info(info)` (see *** comments):

```js
function info(info) {
  scrollStop();
  layout.title.label = info.track || "";
  layout.album.label = info.album || "";
  layout.artist.label = info.artist || "";

  // *** ADDED ***
  const newNum = formatNum(info);

  if (layout.num.label.length > newNum.length) {
    layout.clear(layout.num);
  }
  // *** /ADDED ***

  // color depends on all labels
  layout.title.col = infoColor("title");
  layout.album.col = infoColor("album");
  layout.artist.col = infoColor("artist");
  layout.num.label = newNum;                 // *** CHANGED ***
```

### Installed apps

_No response_

thinkpoop:
App Link: https://thinkpoop.github.io/BangleApps/?id=gbmusic

1) Fix for #3737 

_I'll try to save way to recreate this, in case anyone wants to work on figuring out the root cause._

2) Fix `saved.state` being a string instead of an object

`boot.js` around line 31 saves the `e.state` instead of `e`. 
`app.js` function `state(e)` expects an object though, so it was saving `stat=e.state`, and `stat` was ending up undefined.

3) Send "playpause" if stat is not "play". 

This way, if `stat` is unknown/undefined, gadgetbridge can figure out what to do. `gbmusic` could always send `playpause` and that would work too.

4) Ignore touches if they're in the widget area.

Opening message notifications was pausing music on me. I'm not sure if I should add the bottom widget bar too? (I'm assuming it'd be `pos.y <= Bangle.appRect.y2`?)

5) Fix lint warnings (and use `globalThis` instead of `global` in the boot.js).


thinkpoop:
This PR to fix existing issues before I submit a couple possible enhancements (config setting for pause-timeout, and adding thumbs up/down commands).

Let me know if you'd prefer I combine any of these into this PR to cut down on updates.

thyttan:
@rigrig if you'd like to comment since you wrote most of this app.

rigrig:
Thanks for pinging me.

I haven't used/looked at this app for quite some time though. It's pretty old, so I'm not surprised it bitrotted. (Does it even play nice with e.g. the Message UI app? Otherwise it might be time to mark it as deprecated.)

One general comment: please delete unused code instead of commenting it out: it clutters up the file, and we have git to keep track of history.

thinkpoop:
Deleted the commented code.

I can't remember why I ended up on gbmusic.. a mix of testing random apps, and maybe it was easier to modify with my own tweaks? 

It's been my go-to for a few months though. 

The pause/inactive timeouts being so long have been my only remaining complaint -- I was going to add a settings options to tweak them. But it's been buggy so I'm holding off on that.

apps/gbmusic/boot.js
```diff
@@ -1,6 +1,6 @@
 setTimeout( // make other boot code run first, so we override e.g. android.boot.js GB
   () => {
-    const APP = global.__FILE__==="gbmusic.app.js",
+    const APP = globalThis.__FILE__==="gbmusic.app.js",
```
thyttan, line 3: Out of curiosity, what is the be benefit if using the `globalThis` keyword instead of `global`?

thinkpoop: I'm not 100% sure... 

I assume just future-proofing, meeting Ecmascript standards? I just stumbled onto the recommendation in the docs: https://www.espruino.com/ReferenceBANGLEJS2#l__global_global

thyttan: Ok - thanks!

apps/gbmusic/app.js
```diff
@@ -433,15 +383,21 @@ function sendCommand(command) {
...
 
+function handleTouch(btn, pos) {
+  if (pos === undefined || pos.y >= Bangle.appRect.y) {
+    togglePlay();
+  }
+}
+
 function togglePlay() {
-  sendCommand(stat==="play" ? "pause" : "play");
+  sendCommand(stat==="play" ? "pause" : "playpause");
```
thyttan, line 393: Is it better to use `"playpause"` than `"play"`, I believe that's the reason for the change?

I think this goes to `.../GBMusicControlReceiver.java` in gadgetbridge, am I correct?

"playpause" should maybe be added to possible value for the `command` parameter for the `sendCommand` function?

thyttan:
Ok - thanks!

PR #3745, opened 2025-02-11
pacer - run with a virtual partner

smulrine:
Running app. Run with a virtual partner at your chosen pace, and export the GPX data


pavelmachek:
If you find bangle.js2 GPS performance to be poor, take a look at "skyspy". If you wait for a good fix (speed and altitude match reality) it should be usable in open areas.

bobrippling:
Thanks for the screenshot! (and app!)

PR #3747, opened 2025-02-15
txtreader: Make font size selectable

pavelmachek:
This makes font a bit bigger, and allows user to select font size.

### DIFF
--- a/apps/txtreader/ChangeLog
+++ b/apps/txtreader/ChangeLog
@@ -1,1 +1,2 @@
 0.01: New App!
+0.02: Make font size selectable
--- a/apps/txtreader/README.md
+++ b/apps/txtreader/README.md
@@ -11,8 +11,9 @@ Very basic text reader with an integrated file selector.
 ## Controls
 
 Bangle.js 2
-- tap the right side of the screen to flip to the next page
-- tap the left side of the screen to flip to the previous page
+- tap the bottom right corner of the screen to flip to the next page
+- tap the bottom left corner of the screen to flip to the previous page
+- tap the top left corner of the screen to toggle font size and go back to page 1
 - exit by pressing the physical button 
 
 ## Creator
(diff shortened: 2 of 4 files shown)


tonykakuuu:
Very cool! I think it would work better though if this was an option in the settings menu. 
Also, I dislike that both available fonts are way bigger than the original. This makes it annoying to read longer text files because you need to flip pages so frequently.

thyttan:
I liked that the screen wasn't cleared before displaying next/preivous page in the initial version. Maybe this could be a setting (if there are actually different preferences between us - otherwise it makes more sense to just keep the v0.01 way).

I think I prefer this PR's way of switching font size to having to navigate to a settings page. But maybe it could cycle through three sizes instead of two? Not a strong opinion though - leaving it up to you two.

pavelmachek:
I'd also prefer not having settings page for this. Lets do the original and then the big one for now, and I'll likely do three sizes in future.

I added screen clearing as a way to provide feedback that touch was registered. I guess I can disable it for now. Would vibration be appropriate feedback?

thyttan:
> Would vibration be appropriate feedback?

I was thinking maybe just flash or display a little green dot in the top info field when triggering loading in the next page?

pavelmachek:
Ok, so I pushed version without controversial changes, so it can be applied.

Green dot.. dunno, may be a bit too subtle. I don't really know how to do flashing without slowing things down.

My other idea was to get rid of top status line, and just display page # and filename when moving between pages -- solving feedback and putting more text on display. Anyway, that's ideas for next version.

thyttan:
Ok - thanks again! I think it's time to merge now 🙂

pavelmachek:
Thank you! What do you think about moving from status line on top to providing feedback by displaying page number and filename on page flip?

thyttan:
Personally I want almost as small feedback as possible between turning pages since for me it feels like an interruption to the reading experience.

I'm generally sympathetic to making interactions feel more responsive. But in this case I think the effect of an intermediate, immediate update makes it feel *less* responsive - since it gives the brain more sensory input to process and also highlights the time it takes to load the next page.

PR #3748, opened 2025-02-16
skyspy 0.10: add screens with time to fixes and estimate of time to fix

pavelmachek:
Internals changed a lot to share code with space race, we now get more statistics.


thyttan:
I looked at the code - didn't test. Looks good - thanks!

Issue #3752, opened 2025-02-22
Want to Respond to Alarm with Button Click


voloved:
Hello,

My goal is to be able to press the physical button to end an alarm.
My use-case is that I often wake up before my alarm, so I have a buzzing watch on my wrist while driving to work and I don't want to navigate with the touchscreen while driving to silence it.

Plan is to add a function called `E.showPromptBtnCancel` in `apps/sched/sched.js`. From looking at the code, this is the function that runs when an alarm is called. I am modifying it according to [this doc](https://www.espruino.com/Bangle.js+Modification).

I am running into multiple issues:
1. ~How do I modify `E.showPrompt` to treat `BTN1` the same way as cancel? 
  a.  I see this logic:`Bangle.setUI({mode: "leftright", remove: options.remove, redraw:draw, back:options.back}, dir=>{`, but I cannot find the definition of `leftright` or what `dir` returns in the [References](https://www.espruino.com/ReferenceBANGLEJS2#t_l_Bangle_setUI). My hope is that `BTN1` is also a callback coming from `leftright`, but I want to make sure.~
2. ~I am unit-testing  `E.showPromptBtnCancel` in a simple app [here](https://github.com/voloved/BangleApps/blob/devolov/apps/devolov/app.js) and whenever I run this code on my watch, it hangs on a loading screen forever and a reset of the watch is needed. I tested this code on the Web IDE emulator and got no issues. Is there something I'm doing wrong when moving from Simulating to testing on hardware?~
3. ~When copying the default `showPrompt` logic, it always returns that I pressed the first option. The snippet below is how I call `showPrompt` in my testing. I first use the default function and `Stop pressed` prints when I press `Stop`. But when I copy the [default logic](https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showPrompt.js), the screen appears the same, but `Snooze Pressed` diplays when I press `Stop`. Is the default function I'm referencing in the repo up-to-date?~
``` Please let me know if there's anything I'm misunderstanding or if there are resources I missed.
E.showPrompt("the", Also, { if there is a forum or Discord that I can join, please let me know!
title: "ALARM!",
buttons: { /*LANG*/"Snooze": true, /*LANG*/"Stop": false }
}).then(ok => {
if (ok) print("snooze pressed");
else print("stop pressed");
});
```

[...]

PR #3753, opened 2025-02-22
Physical Button Can Stop an Alarm

voloved:
Currently, if an alarm is going off, the lockscreen needs to be awaken with the physical button and then `Stop` must be touched.

This can be tedious, especially if the alarm goes off while driving.

This commit adds the ability to turn off the alarm by pressing the physical button twice (once to wake the watch and once to stop the alarm).

An issue is that this does replicate the [showPrompt_Q3](https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showPrompt_Q3.js) logic for the Bangle 2, which means that this functionality will not work on the Bangle1.

If this is not an option, one nice-to-have is to allow the behavior of the physical button to be mapped to either the `Yes` or `No` when calling `showPrompt`, but that is outside of the scope of this codebase.


bobrippling:
Thanks - I suppose this is something we should consider merging into `E.showPrompt`, like you say. Let's see - tagging @gfwilliams  with this being a core app

Closes #3752

gfwilliams:
Yes - changing showPrompt in the firmware would be preferable. However it does support using the button press even without your changes, because supplying `back` as a callback argument works fine, it's just not as nice:

```
E.showPrompt("Test", {back:function() { 
  Bangle.setUI(); // remove prompt
  g.clear();
  print("Cancel"); 
}}).then(r => {
  print("Button "+r); 
});
```

Although it's still nicer than redefining showPrompt, and it should work on Bangle.js 1 too. So I'd say if you want this, I'd add it that way for now.

Also I can't really see a reason not to have this behaviour by default? It just makes sense to me, and I can't see anyone's really going to complain about it. Not sure what others think, but I'd be very tempted not to have the option for it at all and just add the behaviour by default.

dashavoo:
> Also I can't really see a reason not to have this behaviour by default? It just makes sense to me, and I can't see anyone's really going to complain about it. Not sure what others think

A flyby comment because the title of this PR caught my eye: I'd suggest that the default action on button press should be to snooze the alarm rather than stop it. The consequences of doing it by accident are less drastic.

nxdefiant:
I disagree on the default to snooze, but honestly I never understood what the usecase for snooze is. Never needed it once.

+1 for pysical button to "ok", but maybe we can do a short press/long press for snooze/ok?

gfwilliams:
Yes, I agree with not snoozing by default. But there's no reason it shouldn't be an option

thyttan:
Ping! @voloved

Issue #3755, opened 2025-02-22
cutelauncher has 2 changelogs


voloved:
@kkayam,

It looks as though you added two changelog files in the cutelauncher.
Was that on purpose?

[Path with 2 changelogs.](https://github.com/espruino/BangleApps/tree/master/apps/cutelauncher)

thyttan:
Fixed with https://github.com/espruino/BangleApps/pull/3757

PR #3759, opened 2025-02-23
Daisy Clock Display 12hrs, 5% ring Increments and Settings to Change the Ring Behavior

voloved:
This commit modifies the Daisy watch behavior by:
* Making it so the ring goes up in increments of 5%, rather than 10%
* Allow changing the behavior of the ring; before this commit, it'd only display steps taken throughout the day. The new options are:
  * Show minutes, like the minute hand of an analog watch
  * Show seconds (the watch will update every 5 seconds)
  * Show battery percentage.
* Added toggle to show time in 12 hr mode.

The ring options and the 12hr mode are both options in the Settings.

The only new concerns from this change are:
* The `Seconds` option will drain the watch faster, as it'll update the face every 5 seconds, rather than every minutes
* The `Battery` option may cause the Ring to pop due to lack of deadband, As in, if the battery jumps between 50% and 51% continuously, the ring will jump between the two each minute. This is unlikely to be a concern, but I wanted to point it out.

I tested all of the new settings on a Bangle2, and they all work well. The `Seconds` option confirmed that all of the 5% ring steps looked good as well.


bobrippling:
Thanks, LGTM! @hughbarney are you happy for this to be merged?

voloved:
Actually, hang on, I want to make the Sunrise and Sunset time remaining an option on the wheel and make  those times show as 12 hours as well.

bobrippling:
I'll pop this into draft - if you put it back to ready when you're happy I can look at merging :)

apps/daisy/settings.js
```diff
@@ -6,53 +6,95 @@
...
+      'Ring Display': {
+        value: 0 | ring_options.indexOf(s.ring),
+        min: 0, max: ring_options.length - 1,
+        format: v => ring_options[v],
+        onchange: v => {
+          let prev = s.ring;
+          s.ring = ring_options[v];
+          save();
+          if (prev != s.ring && (prev === 'Steps' || s.ring === 'Steps')) {
+            setTimeout(showMainMenu, 0);  // Reference https://github.com/orgs/espruino/discussions/7697
```
bobrippling, line 80: Perhaps we inline the reason here too, wdyt?

```suggestion
            // redisplay the menu with/without ring setting
            // Reference https://github.com/orgs/espruino/discussions/7697
            setTimeout(showMainMenu, 0);
```

apps/daisy/app.js
```diff
@@ -540,14 +686,15 @@ function buzzer(n) {
 // timeout used to update every minute
 var drawTimeout;
 
-// schedule a draw for the next minute
+// schedule a draw for the next minute or every 5 seconds
```
bobrippling, line 689: Should this be 3? Or perhaps we just say N seconds to avoid the comment becoming stale

```suggestion
// schedule a draw for the next minute or every N seconds
```

apps/daisy/app.js
```diff
@@ -540,14 +686,15 @@ function buzzer(n) {
...
 function queueDraw() {
+  let delay = (settings.ring == 'Seconds') ? (3000 - (Date.now() % 3000)) : (60000 - (Date.now() % 60000));
```
bobrippling, line 691: Very minor, but we might want to avoid parens around the condition for a ternary as it might be misleading, for example:
```javascript
x == 3 || (x == 4) ? "4" : "not 4"
```

evaluates to `"4"` even if `x` is 3, because it's parsed as:
```javascript
(x == 3 || (x == 4)) ? "4" : "not 4"
```

so we might as well say
```javascript
x == 3 || x == 4 ? "4" : "not 4"
```

bobrippling (commented):
This looks really good, I especially like the sunset/sunrise circles, great job! Just a few little questions

bobrippling:
And I think your idea of reversing the ring-fill direction for dusk til dawn is nice, I think we go with it

apps/cutelauncher/ChangeLog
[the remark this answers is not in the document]
voloved: It's related to [this](https://github.com/espruino/BangleApps/issues/3755). Cutelauncher had a file called `Changelog` and another called `ChangeLog`. It seemed to have heavily confused Git.

This may have been the wrong one to delete. It can be undeleted, so long as it does not match the one that `master` deleted.

apps/daisy/settings.js
voloved: I like it. I'll add it into my next commit

apps/daisy/app.js
voloved: I'd either call it out to 3, or have the 3second amount be a const in the code. I don't think the 3 seconds will change, but having this info in 3 seperate places makes me consider making is a const.

apps/daisy/app.js
voloved: I think I'll disagree with you on this, be have no issue with following your request.
In your example, I'd expect it to be written:
```
(x == 3 || ((x == 4)) ? "4" : "not 4")
```

Where the parenthesis are always well-defined, but I can understand the issue and will include in my next commit.

voloved:
Sorry for the forced push. To clean up the issue with `Changelog` in `cutelauncher`, I needed to unmerge and remerge with master to keep the repo clean.
For whatever reason, `stash`, `reset`, and `gitignore` did not stop `Changelog` from just coming back.

It does have the change you suggested.

apps/daisy/app.js
voloved: Changed [here[(https://github.com/espruino/BangleApps/pull/3759/commits/f431258eafca3167610b295b625fda24fa87ea38).

apps/daisy/app.js
voloved: Changed [here](https://github.com/espruino/BangleApps/pull/3759/commits/f431258eafca3167610b295b625fda24fa87ea38).

apps/daisy/settings.js
voloved: Changed [here](https://github.com/espruino/BangleApps/pull/3759/commits/f431258eafca3167610b295b625fda24fa87ea38).

bobrippling: Smashing!

apps/daisy/app.js
bobrippling: Const is a great idea, thanks!

apps/daisy/app.js
bobrippling: Yeah that's a fair point, parenthesising the entire thing - thanks for making the change even so :)

PR #3762, opened 2025-02-26
noteify: make notes available to clock info.

randyheydon:
This adds a clock info list for Noteify, with each note as an entry in that list.  This gives a way for a user to put a note directly on any clock with a clock info area.  Loosely inspired by the scratchpad of the Advanced Casio Clock (advcasio).

Tested on my Bangle2 on multiple clocks (bwclk 0.34, circlesclock 0.26, slopeclockpp 0.11, lcdclock 0.07, lcdclockplus 0.03, lato 0.04) with zero notes, one note, and multiple notes.  Notes don't always show up well on all clocks, depending on line count and line length.  This code does not attempt to shorten the notes to display better because the clocks are all so different in how they display the text.

### DIFF
--- a/apps/noteify/ChangeLog
+++ b/apps/noteify/ChangeLog
@@ -1,3 +1,4 @@
 0.01: Initial version
 0.02: Use default Bangle formatter for booleans
 0.03: Drop duplicate alarm widget
+0.04: Make notes available to clock info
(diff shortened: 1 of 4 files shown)


thyttan:
Thanks! Could you test this with the Bangle.js 1 emulator in the Web IDE? 

Alternatively, we could drop the Bangle.js 1 support in the metadata for now.

randyheydon:
I can't quite figure out how to test this in the emulator.  It looks like I would have to manually upload all files for Noteify, a clock, clock_info, and a keyboard.  Is there a way to connect the emulator to an app loader yet to simplify this?

I woudn't want to drop Bangle1 support, since Noteify has supported it up to now.  If anyone else wants to test it on physical hardware, I have this version of Noteify served from https://randyheydon.github.io/BangleApps/?q=noteify

randyheydon:
I just realized that [clock_info only supports Bangle2](https://github.com/espruino/BangleApps/blob/master/apps/clock_info/metadata.json#L9), so the feature I added in this pull request would not work on Bangle1.  However, that shouldn't limit Noteify's availability on Bangle1; for example, [agenda](https://github.com/espruino/BangleApps/blob/master/apps/agenda/metadata.json#L9) supports both Bangles while providing clock info entries like this.  Without clock_info, I assume noteify.clkinfo.js is just never executed.

randyheydon:
@thyttan what should the next steps be here?  I've been using this feature happily for a while, but I still can't figure out how to test it in the Bangle1 emulator, especially considering that I would be trying to test that this new file is just never executed.  Or at this point should the app just drop Bangle1 support?

thyttan:
Sorry for not getting back to you! I'll merge this now and maybe test it in the emulator at a later date.

Thank you for you solution focused replies 🙏

PR #3763, opened 2025-02-26
all apps: refactor `global.` -> `globalThis.`

thyttan:
As per:
[global is used in Node.js. Consider using the identical globalThis as it was introduced in the ECMAScript spec.](https://www.espruino.com/ReferenceBANGLEJS2#l__global_global)

If we decide it's worth merging this we should probably bump versions of the apps as well?

@bobrippling @gfwilliams Does this make sense?


gfwilliams:
Thanks, but I'm not really sure if this adds anything? It'll make all these apps very slightly larger and slower and introduces the possibility of breakage.

But otherwise it's not fixing anything, just following what the ECMA spec team randomly changed. Looking at Node.js specs it says `Stability: 3` which means they have no plans to get rid of `global` so it's not like this JS code is going to be un-runnable in other JS engines if we keep `global`.

... perhaps the wording in `global` in the docs should be changed actually?

thyttan:
> ... perhaps the wording in global in the docs should be changed actually?

Yes that makes sense to me in that case 👍

PR #3766, opened 2025-03-06
terminalclock: Correct ISO date format

JimMadge:
This PR adds leading zeros to the day and month when using the ISO date format.

It also corrects the date appearing in the wrong position due to the position variable not being passed correctly.


bobrippling:
Thanks!

PR #3767, opened 2025-03-06
agenda: Fixed bug where "Today" and "Tomorrow" labels display in non-current weeks.

JamesClarke01:
This fixes a bug where the "Today" and "Tomorrow" labels display on future weeks due to matching on day of the week instead of date, e.g. if today is Thursday an event on Thursday next week will be labelled "Today".


bobrippling:
Nice spot, thank you

PR #3776, opened 2025-03-17
android: handle android STREAM_MUSIC volume level forwarding

thyttan:
@gfwilliams

More context on companion PR to Gadgetbridge https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4658

### DIFF
--- a/apps/android/lib.js
+++ b/apps/android/lib.js
@@ -44,6 +44,10 @@ exports.gbHandler = (event) => {
     "musicinfo" : function() {
       require("messages").pushMessage(Object.assign(event, {t:"modify",id:"music",title:"Music"}));
     },
+    // {t:"audio", v:(percentage of max volume for android STREAM_MUSIC)}
+    "audio" : function() {
+      Bangle.emit("musicVolume", event.v);
+    },
     // {"t":"call","cmd":"incoming/end/start/outgoing","name":"Bob","number":"12421312"})
     "call" : function() {
       Object.assign(event, {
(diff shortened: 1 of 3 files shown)


apps/android/lib.js
```diff
@@ -44,6 +44,17 @@ exports.gbHandler = (event) => {
     "musicinfo" : function() {
       require("messages").pushMessage(Object.assign(event, {t:"modify",id:"music",title:"Music"}));
     },
+    // {t:"musicvolume", v:(percentage of max volume)}
+    "audio" : function() {
+      if (Bangle.musicVolume === undefined) {
+        E.on("kill", ()=>{
+          require("Storage").writeJSON("setting.json", Object.assign({}, require("Storage").readJSON("setting.json"), {musicVolume:Bangle.musicVolume}));
```
thyttan, line 51: This is untested still.

gfwilliams:
Thanks - Gadgetbridge PR looks good.

Personally I think if we're going to save the volume level we should probably not do it in the global `settings.json` file as if everyone does that it's going to end up getting pretty bloated. There's an `android.settings.json` we could use.

However maybe for now I'd just skip doing that and the stored variable and just do `Bangle.emit("musicVolume", v)` - I'm trying to imagine use cases and I feel like the Bangle would probably only want to show volume when it changed anyway to save on screen real estate - and since we're not sending the device volume on connection there's really no guarantee that the stored volume would be correct anyway, in which case it's probably more of a hinderance than a help?

thyttan:
We should maybe wait at least until the Gadgetbridge PR is merged before doing the same here.

PR #3783, opened 2025-03-25
promenu: fix settings overwriting

bobrippling:
promenu declares and uses a global variable, `settings`. This is common in apps such as `settings` and `alarms`, which will come along and overwrite said variable.

This has the effect of all settings in promenu being falsy (unless a setting name overlaps with one from bangle settings). Renaming the variable resolve this problem.


thyttan:
Seems innocent enough to me - thanks! 👍

bobrippling:
Thanks!

devsnd:
Oohhh that bug was driving me nuts! Thanks bob

Issue #3785, opened 2025-03-28
[launcher] fullscreen setting doesn't work


onebitboy:
### Affected hardware version

Bangle 2

### Your firmware version

2v25 and 2v25.366

### The bug

1. Launcher seems to completely ignore the ~~"Show clocks" and~~ "Fullscreen" settings. No matter what it's set to, it ~~always shows clocks and~~ never goes fullscreen. I've tried it with v0.21 and v0.22 (stable and dev branches) on both the 2v25 and the 2v25.366 firmware.

~~2. Some launcher font sizes cause a complete breakage at least on 2v25.366. Setting it to (non-vector) 17 for example, the launcher doesn't display anything any more at all, so you can't even go back to the settings to change the font size again.~~

Edit: I'll create separate issues for the other two bugs.

dashavoo:
Aside: This should probably be two separate issues.

> Launcher seems to completely ignore the "Show clocks" and "Fullscreen" settings. No matter what it's set to, it always shows clocks and never goes fullscreen. I've tried it with v0.21 and v0.22 (stable and dev branches) on both the 2v25 and the 2v25.366 firmware.

I didn't know about the fullscreen option, so when I saw this bug open I was interested to try it.
Like for @onebitboy, on my watch the setting seemed to make no difference. However, if I run the launcher from the IDE (Connect to watch, press the Storage icon , find `launch.app.js` and press the "Run File" icon) the launcher loads and correctly does not display widgets. It continues to work for as long as the watch doesn't reset (so I can move between launcher and fast-load supporting clock as many times as I like and the launcher doesn't show widgets, but if I load another app such as 
"Settings" the launcher no longer seems to respect the settings.

dashavoo:
@onebitboy which Clock are you using please?

dashavoo:
Aha.

The clock I use supports fast loading. I switched to one that doesn't and the fullscreen option worked correctly. So I checked the instructions for supporting fast loading.

I think [this](https://www.espruino.com/Bangle.js+Fast+Load#taking-advantage-of-fast-loading) is the paragraph that explains this behaviour:

> You can also use Bangle.showClock() (if your app is not a clock - eg a launcher). Or, you can use Bangle.load("load_me.app.js") - but again, **the app you load must use widgets, since your Clock would have loaded with widgets and they cannot be unloaded.**

onebitboy:
[...]

PR #3788, opened 2025-03-29
launch: fix fullscreen and showClocks setting

### LINKED ISSUE #3786: [launcher] "show clocks" setting doesn't work
### Affected hardware version

Bangle 2

### Your firmware version

2v25.369

### The bug

Launcher completely ignores the "show clocks" setting. The installed clocks are always listed, no matter whether the setting is enabled or not.

### Installed apps

Tested with launcher 0.21, 0.22 and 0.23 on firmware 2v25, 2v25.366 and 2v25.369.

thyttan:
0.24: Fix fullscreen when fastloading the launcher. (TODO:fix back btn flicker)
      Fix respecting the showClocks setting.

Try it via my app loader: https://thyttan.github.io/BangleApps/?id=launch

closes #3786 and #3785

@gfwilliams because it is a core app.

### DIFF
--- a/apps/launch/app.js
+++ b/apps/launch/app.js
@@ -85,13 +85,9 @@
         if (lockTimeout) clearTimeout(lockTimeout);
         Bangle.removeListener("lock", lockHandler);
         // Restore widgets if they were hidden by fullscreen setting
-        if (global.WIDGETS) require("widget_utils").show();
+        if (global.WIDGETS && settings.fullscreen) require("widget_utils").show();
       }
     });
-    if (settings.fullscreen) {
-      require("widget_utils").hide();
-      scroller.draw(); // FIX: The red back button will flicker before the widget is hidden and scroller redrawn.
-    }
     g.flip(); // force a render before widgets have finished drawing
 
     // 10s of inactivity goes back to clock
--- a/apps/launch/app.js
+++ b/apps/launch/app.js
@@ -22,8 +22,11 @@
   let height = 50*scaleval;
 
   // Now apps list is loaded - render
-  if (!settings.fullscreen)
+  if (!settings.fullscreen) {
     Bangle.loadWidgets();
+  } else if (global.WIDGETS) {
+    require("widget_utils").hide();
+  }
   let R = Bangle.appRect;
   g.reset().clearRect(R).setColor("#888");
   for (var y=R.y;y<R.y2;y+=height) {
@@ -54,7 +57,7 @@
 
 
   const drawMenu = () => {
-    E.showScroller({
+    let scroller = E.showScroller({
       h : height, c : apps.length,
       draw : (i, r) => {
         var app = apps[i];
@@ -81,8 +84,14 @@
         // cleanup the timeout to not leave anything behind after being removed from ram
         if (lockTimeout) clearTimeout(lockTimeout);
         Bangle.removeListener("lock", lockHandler);
+        // Restore widgets if they were hidden by fullscreen setting
+        if (global.WIDGETS) require("widget_utils").show();
       }
     });
+    if (settings.fullscreen) {
+      require("widget_utils").hide();
+      scroller.draw(); // FIX: The red back button will flicker before the widget is hidden and scroller redrawn.
+    }
     g.flip(); // force a render before widgets have finished drawing
 
     // 10s of inactivity goes back to clock
--- a/apps/launch/app.js
+++ b/apps/launch/app.js
@@ -57,7 +57,7 @@
 
 
   const drawMenu = () => {
-    let scroller = E.showScroller({
+    E.showScroller({
       h : height, c : apps.length,
       draw : (i, r) => {
         var app = apps[i];
--- a/apps/launch/ChangeLog
+++ b/apps/launch/ChangeLog
@@ -23,3 +23,4 @@
 0.21: Make the "App source not found" warning less buggy
 0.22: Add less padding between launcher items, use new font if available in 2v26+
 0.23: Draw a placeholder screen right at the start to speed up apparent boot time
+0.24: Fix fullscreen when fastloading the launcher. (TODO:fix back btn flicker)
--- a/apps/launch/ChangeLog
+++ b/apps/launch/ChangeLog
@@ -24,3 +24,5 @@
 0.22: Add less padding between launcher items, use new font if available in 2v26+
 0.23: Draw a placeholder screen right at the start to speed up apparent boot time
 0.24: Fix fullscreen when fastloading the launcher. (TODO:fix back btn flicker)
+      Fix showClocks setting not taking effect by now clearing cache when
+      changing those settings.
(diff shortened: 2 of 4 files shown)


gfwilliams:
Thanks! This looks great! 

For the flicker, I guess there are some options:

* if when we weren't showing widgets we didn't add a `back` handler but instead just add a setWatch on the button?
* Maybe Bangle.setUI shouldn't display a 'back' widget if there are no widgets currently loaded? 
* It looks like if you defined `WIDGETS.back = {remove:()=>{}};` beforehand when fullscreen then it might stop a redraw? Bit of a hack though! https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_setUI_Q3.js#L9-L12

apps/launch/app.js
```diff
@@ -81,8 +84,14 @@
         // cleanup the timeout to not leave anything behind after being removed from ram
         if (lockTimeout) clearTimeout(lockTimeout);
         Bangle.removeListener("lock", lockHandler);
+        // Restore widgets if they were hidden by fullscreen setting
+        if (global.WIDGETS) require("widget_utils").show();
```
bobrippling, line 88: Do we also need:
```suggestion
        if (global.WIDGETS || settings.fullscreen) require("widget_utils").show();
```

or perhaps:
```suggestion
        if (hiddenWidgets) require("widget_utils").show();
```
(with that bool set when we call widget utils' `hide()`)

gfwilliams: I *think* we might be ok: https://github.com/espruino/BangleApps/blob/master/modules/widget_utils.js#L21

But it feels like a good idea to check anyway as otherwise it may force a redraw? Did you mean `if (global.WIDGETS && settings.fullscreen)` instead?

bobrippling: Ah no - we're looking at it from different sides. Your side is whether it's safe to call (so I can see why you're thinking `&&` and the `widget._draw` check), mine is if we need to undo our changes, which are either if we've done a `.hide()` from having widgets:

https://github.com/espruino/BangleApps/blob/5dd73bbea83c535568a179d56b7478b7d489885c/apps/launch/app.js#L27-L28

or if we've had the fullscreen setting:

https://github.com/espruino/BangleApps/blob/5dd73bbea83c535568a179d56b7478b7d489885c/apps/launch/app.js#L91-L92

If either of those are true, we need to `.show()`, hence `global.WIDGETS || settings.fullscreen`, right?

thyttan: But if there are no widgets, there's no need to update their properties with the `.show()` call - right? There's nothing for `widget-utils` to act on.

bobrippling: Yes, but then [the for loop](https://github.com/espruino/BangleApps/blob/ebbe4206e604660571020d712e4abea4d01c6e1a/modules/widget_utils.js#L19-L20) wouldn't be entered in that case so we're ok, still. And we're balancing our [above call to `.hide()`](https://github.com/espruino/BangleApps/blob/5dd73bbea83c535568a179d56b7478b7d489885c/apps/launch/app.js#L27-L28), which I think is as far as this code needs to go, and it can then trust that `widget_utils` will do the right thing

apps/launch/app.js
```diff
@@ -81,8 +84,14 @@
...
       }
     });
+    if (settings.fullscreen) {
+      require("widget_utils").hide();
```
thyttan, line 92: @bobrippling this makes me think you may be right regarding [L88](https://github.com/espruino/BangleApps/pull/3788/files#r2022414408)

thyttan: But we also have this consideration about maybe changing the back-function behaviour: https://github.com/espruino/BangleApps/pull/3788#issuecomment-2765635283

bobrippling: That's true, it is a bit of a special case I suppose. Perhaps we try out a few different approaches?

gfwilliams: Ok - so in this case the widgets would have been hidden before (if there were any), but because we're using `back` a widget is being added anyway and so we're having to hide them again?

I feel we should probably just fix the `back` issue rather than having these lines in there?

apps/launch/app.js
gfwilliams: Ahh, right - I see what you mean - but actually we only call `require("widget_utils").hide();` when we're supposed to be in fullscreen mode (line 27 is actually `if (!fullscreen) {..} else if (...WIDGETS) ...`. 

If we're not in fullscreen mode we almost certainly do have widgets but they're not hidden.

thyttan:
Without thinking too much about it - yes I like that solution!

thyttan:
I'm debating in my head if the touch in top left corner should still work though even when the widget is not displayed. But probably not?

gfwilliams:
Yeah, although I wonder whether someone might tap in the launcher and expect to launch the app?

I was considering making setUI use an overlay to display the back button in that case, but it just feels like overkill!

thyttan:
> Yeah, although I wonder whether someone might tap in the launcher and expect to launch the app?

It's such a small area that it shouldn't happen very often. But I'm not sure which way is best.

> I was considering making setUI use an overlay to display the back button in that case, but it just feels like overkill!

Agreed. It could be cool - but probably not worth the overhead?

I'd like it in the context of my messagegui changes - but not if it impacts scrolling drawing performance/input latency.

apps/launch/app.js
bobrippling: Ah I hadn't spotted the fullscreen check, yeah I'm with you

bobrippling:
I suppose one question would be, for apps without widgets do users regularly use the back button anyway? Or are they more likely to soft-reset to escape? Or are we talking menus?

gfwilliams:
> For messagegui it's E.showScroller and not Bangle.setUI though

Ahh - yes, I looked at that. It seems showScroller calls setUI under the hood :)

Having the overlay definitely would hurt scroll performance, so maybe we forget that for now.

> for apps without widgets do users regularly use the back button anyway? Or are they more likely to soft-reset to escape? Or are we talking menus?

Yes, I think they'd almost certainly use the button. I reckon if the whole screen is being used there's a high likelihood that UI elements might end up being over that top left area anyway - imagine if you have a menu with no title - the top menu item would then be forever half obscured by the back button.

I'll go fix setUI then...

gfwilliams:
Huh, well I did that - and I guess as expected it wasn't that easy! Because when widgets are there but hidden WIDGETS is still defined - so I had to check if the widget bar was there. Fixed now though 453def81645c08e46195ad0627c1b77c8a59ee2a

I've updated the launcher code - does this look ok now? Obv it'll need new firmware but I should probably do a release soon anyway

thyttan:
Nice!

I believe so - didn't test now though

gfwilliams:
Just FYI I added another change which allows the item height to be adjusted independently of the font.

PR #3789, opened 2025-03-30
dtlaunch: fix caching preventing settings taking effect and use own chache file

thyttan:


thyttan:
I overlooked the showLaunchers setting - that  needs a tweak as well. Will do that  first.

PR #3797, opened 2025-04-10
agenda: fix all-day event offsets in clock info.

randyheydon:
All-day events synchronized from Gadgetbridge always start at midnight UTC, which causes them to display as different days depending on time zone offset (specifically negative offsets display the date as one day early).  This was previously fixed in the agenda app in pull request #3695 (commit bcf88c3). This commit replicates that fix in the clockinfo that agenda provides.

Tested this on my Bangle2 with time zone offsets that are negative, positive, and zero.


bobrippling:
LGTM, thanks!

PR #3805, opened 2025-04-17
sched: prevent timers from firing immediately

### LINKED ISSUE #3804: [alarm/sched] Timer doesn't fire (or fires immediately) when enabling from list
### Affected hardware version

Bangle 2

### Your firmware version

2v25

### Description
On the Bangle.js 2, a timer can be quickly enabled by pressing its icon or long pressing the item in the list. However, this causes some unexpected behaviour. Most of the time the timer never fires, but sometimes it fires immediately after enabling it this way.

### Expected Behaviour
Enabling the timer via this shortcut should have the same effect as opening its menu and setting it to be Enabled.

### Suspected cause
When enabling via the main menu, neither `prepareTimerForSave` (or `prepareAlarmForSave` for alarms) is called, so an old expiration time is used for the timer instead of it being updated when it is enabled.

### Installed apps

- alarm (0.50)
- sched (0.31)
- boot (0.65)

bobrippling:
fixes #3804


thyttan:
@gfwilliams on this one :)

bobrippling:
ah yes, thanks!

bobrippling:
This doesn't fix the bug, marking as draft (see #3804)

bobrippling:
done after in 0a67619541383950dc0b6be1332948196025c6fd

bobrippling:
Re-opening to keep the fix for `sched` - @gfwilliams this look ok to you?

gfwilliams:
Thanks! Yes, that looks great.

PR #3806, opened 2025-04-18
Moon phase hide

voloved:
Added ability to hide the moon phase widget.
Hiding occurs whenever the moon color setting would also take effect.
Tested on Bangle 2.


voloved:
Need to rebase from the master branch.... will re-PR

PR #3807, opened 2025-04-18
Added ability to hide moon phase widget

voloved:
Added ability to hide the moon phase widget.
Hiding occurs whenever the moon color setting would also take effect.
Tested on Bangle 2.


bobrippling:
LGTM, thanks!

PR #3808, opened 2025-04-18
runplus: vibrate on start/stop

bobrippling:


thyttan:
Looks good! Settings could be deafaulted to true in my taste. But merging now.

PR #3809, opened 2025-04-18
exstats: make time active-time (rather than elapsed time)

### LINKED ISSUE #3802: [run+ (or exstats)] Time recording is inaccurate
### Affected hardware version

Bangle 2

### Your firmware version

2v25

### Description

When running with my watch, I notice the time is a little over what I'd expect, using run+. For instance, I recently ran for 18 mins 30 seconds and the time reported was over 20 mins.

### Expected

The time to be accurately recorded and presented to me in run+

All I can think is that the exstats addition doesn't account for time taken to do the addition, but if anything this would cause the time to be less, rather than more:

https://github.com/espruino/BangleApps/blob/cc051726ccd43ca079cf7e006d9571d87e5676f3/modules/exstats.js#L323-L324=

I'll see if I can get a reasonable reproduction

### Installed apps

_No response_

bobrippling:
fixes #3802 

Also fixes a bug where the user stops recording outside the app, and exstats doesn't match this on load (meaning we have stats ticking over when the recording is stopped)


bobrippling:
Testing in progress

gfwilliams:
If this works ok, it looks good to me!

bobrippling:
Testing done, working fine for me - happy for this to be merged

PR #3810, opened 2025-04-18
runplus: optimise UI rendering

bobrippling:
Only redraw layout if we might've shown a setRecording prompt


thyttan:
Seems to work ok when I try it! Not sure what the functional change is just from a user perspective?

It's not this PR's fault but i get in Web IDE: 
`WIDGETS["recorder"].setRecording is deprecated. Use require("recorder").setRecording instead`. 

Maybe we change that in another PR later on.

bobrippling:
Thanks! Yes no functional change, just a slightly faster response when the user wants to start their run (if they don't have the recorder app installed).

I'd notice it takes almost a second to draw, which doesn't give me confidence that runplus has started recording, but at the same time I don't want to be hanging about at the start line! So I've been attempting to speed things up

PR #3813, opened 2025-04-20
edgeclk: Reset graphics before initial clearing of the screen.

thyttan:
Helps in some situations if using fastload utils where the graphics from the previous app would not be cleared.

What triggered the issue for me was:

1. Use fasload utils
2. have a message notification pending showing with `widmsggrid`
3. exit `spotrem` by clicking the HW button on the main ui.
4. edgeclk is displayed but the `spotrem` graphics are showing behind the clock graphics.

I didn't figure out what exactly caused it, but maybe somehow the theme bg color ended up transparent or not set for the `g.clear()` call. This change seems to fix it. But don't know if we should be looking to fix something somewhere else.

### DIFF
--- a/apps/edgeclk/ChangeLog
+++ b/apps/edgeclk/ChangeLog
@@ -2,3 +2,5 @@
 0.02: Fix reset of progress bars on midnight. Fix display of 100k+ steps.
 0.03: Added option to display weather.
 0.04: Added option to display live updates of step count.
+0.05: Reset graphics before initial clearing of the screen. Helps in some
+      situations if using fastload utils.
--- a/apps/edgeclk/app.js
+++ b/apps/edgeclk/app.js
@@ -336,7 +336,7 @@
   /* Startup Process
   ------------------------------------------------------------------------------*/
 
-  g.clear();
+  g.reset().clear();
   drawAll();
   startTimers();
   registerEvents();
--- a/apps/edgeclk/app.js
+++ b/apps/edgeclk/app.js
@@ -336,7 +336,7 @@
   /* Startup Process
   ------------------------------------------------------------------------------*/
 
-  g.reset().clear();
+  g.clear(1);
   drawAll();
   startTimers();
   registerEvents();
--- a/apps/edgeclk/metadata.json
+++ b/apps/edgeclk/metadata.json
@@ -2,7 +2,7 @@
   "id": "edgeclk",
   "name": "Edge Clock",
   "shortName": "Edge Clock",
-  "version": "0.04",
+  "version": "0.05",
   "description": "Crisp clock with perfect readability.",
   "readme": "README.md",
   "icon": "app.png",


bobrippling:
Yeah very odd, I can't think how the graphics would remain - like you say, I'd expect (at worst) a different BG colour to be used instead. LGTM anyway, happy to merge if you are

thyttan:
@gfwilliams I think we probably don't have to look into why `g.clear()` sometimes fails to clear the the screen withot a `g.reset()` first? (Especially since this problem mostly shows with fastload utils present)

If so I'll just merge this.

gfwilliams:
You can always just do `g.clear(1)` which has the same effect as `g.reset().clear()` - it's probably good practice to do reset anyway as the colors/etc could have been changed anyway.

>  but maybe somehow the theme bg color ended up transparent or not set for the g.clear() call.

I don't believe you can set a transparent color, but about the only thing I can think of is `g.setClipRect` could have been called. If so it'd be really good to figure out *why* though as anything that sets clipRect should also reset it.

Maybe try `g.setClipRect = () => { throw new Error("Called here"); };` and see if you can figure out what's doing it and not resetting it after?

thyttan:
> Maybe try g.setClipRect = () => { throw new Error(\"Called here\"); }; and see if you can figure out what's doing it and not resetting it after?

I'll try this 🙏👍

thyttan:
I think we found the culprit then at https://github.com/espruino/BangleApps/blob/a243ac6c881a29b877ff91cdbf4828e39741e810/apps/widmsggrid/widget.js#L33 As that's the only `setClipRect` I find between leaving `spotrem` for `edgeclk`. So it's not a bug with the firmware which is nice :)

So I can look at adding a `g.reset()` there somewhere.

But what is the policy really? That:
1. One should only reset graphics state before setting up new draws?
2. One should only reset graphics state after done drawing?
3. One should reset graphics state before setting up new draws, but it's also good to reset after drawing?

1. Should be best for reliability + performance.
2. Doesn't make much sense to me.
3. May be nice so stuff like this issue with `edgeclk` is less likely to surface. On the other hand it feels a bit like a devil in disguise 😛

So, I prefer to just fix `edgeclk` to properly reset state before clearing the screen. And not add a `g.reset()` to `widmsggrid` (but am open to do it anyway).

Writing this out it feels a bit bikeshedding-y 😂

gfwilliams:
> But what is the policy really? 

> One should only reset graphics state before setting up new draws?

Yes, this one - assume anyone could have been messing with Graphics color/font/etc when your execution starts - and if that calls other functions you don't want to add .reset to all of those.

**But an exception really is setClipRect should always be reset to point to the whole screen, because otherwise you end up with really annoying hard to track down bugs** - so `widmsggrid` should really be fixed as well. It seems `if (w.total > 1) ... g.reset()` so I guess this only really happens if you have less than 2 messages showing.

Also, there are cases like `g.setColor(..).fillRect(...)` where you're overwriting all the state that you're using (hopefully!) so `g.reset()` doesn't need to be called though. And sometimes (like `E.showScroller`) we don't want to call `reset` because the clipRect will have already been set up to the rect that we're supposed to draw and we don't want to get rid of it.

PR #3815, opened 2025-04-20
swipeinv: (WIP) Add inversion of drag events. Refactor to hijack and modify the

thyttan:
events instead of the handling of them. This should add compatibility with
	most apps, even if Bangle.setUI is not used.

Ping @nxdefiant and @jt-nti , what do you think of this approach?

I'm also thinking of adding settings so horizontal and vertical inversion can be set individually, also per app.

### DIFF
--- a/apps/swipeinv/ChangeLog
+++ b/apps/swipeinv/ChangeLog
@@ -1,2 +1,5 @@
 0.01: New app!
 0.02: Minor code improvements
+0.03: (WIP) Add inversion of drag events. Refactor to hijack and modify the
+	events instead of the handling of them. This should add compatibility with
+	most apps, even if Bangle.setUI is not used.
--- a/apps/swipeinv/boot.js
+++ b/apps/swipeinv/boot.js
@@ -1,4 +1,7 @@
 {
+  // TODO: Should the global case just hijack the swipe and drag events, modifying them before passing on to other listeners?
+  //   - That could be a new separate bootloader app instead though?
+
   const settings = Object.assign({
     global: false,
     apps: []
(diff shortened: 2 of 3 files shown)


thyttan:
I think this works now, can be tried at https://thyttan.github.io/BangleApps/?id=swipeinv

jt-nti:
@thyttan looks interesting, I'll give it a try when I get a chance, thanks

thyttan:
I think I notice this approach making the watch feel a bit slower than the old one. So that's a minus in that case. But not conclusive, just a feeling.

thyttan:
I also had the idea to do the inversions by wrapping the listener functions at their registration. That way one would not depend on being first in the listener call change.

It's here: https://github.com/thyttan/BangleApps/commit/5cbb72ee55f7fb7d335ffba228575a862a0ae612

But I didn't get that to work yet.

thyttan:
Doesn't feel too much slower (if at all) now since I added ram keyword to functions.

gfwilliams:
So you're happy to merge this now?

As it's just an optional app I feel like there's no problem merging this in?

thyttan:
Yes I believe it can go in 🙂 

If I find a better approach I can do another PR. But on that note, do you think wrapping listener functions (`Bangle.on/prependListener`) could be better? I suspect it may not be possible without significantly slowing things down though. 

> I also had the idea to do the inversions by wrapping the listener functions at their registration. That way `swipeinv` would not depend on being first in the listener call chain to work.
> 
> It's here: https://github.com/thyttan/BangleApps/commit/5cbb72ee55f7fb7d335ffba228575a862a0ae612
> 
> But I didn't get that to work yet.

gfwilliams:
Thanks! I reckon the current method is good enough. Overriding `.on/etc` feels like it's going to cause issues (removeListener likely won't work then).

Although looking at it, because `drag` passes an object, you don't need all the re-emit stuff that `dragInverter` does. Just:

```
let dragInverter = (e) => {
        "ram"
        const APP_ID = getAppIdFromCurrentFile();
        if (SETTINGS.global ^ Object.keys(SETTINGS.apps).includes(APP_ID)) {
            e.inverted = true;
            if (SETTINGS.global ^ (SETTINGS.apps[APP_ID]&&SETTINGS.apps[APP_ID].dragH)) {e.dx *= -1;}
            if (SETTINGS.global ^ (SETTINGS.apps[APP_ID]&&SETTINGS.apps[APP_ID].dragV)) {e.dy *= -1;}
        }
      }
```

Because subsequent functions in that chain will get the same (modified) object. That'll speed it up.

swipeInverter needs to be as-is

thyttan:
Did that change you suggested - haven't tested yet so changed to draft.

thyttan:
Tried it some - seems to work ok! Thanks!

gfwilliams:
Great, thanks!

PR #3819, opened 2025-04-22
Fix settings altitude calibration saving

### LINKED ISSUE #3814: [Altimeter] Altitude resets itself
### Affected hardware version

Bangle 2

### Your firmware version

2v25

### The Bug
The altitude for where I live is 52m. If I go to Settings>System>Altitude in settings and adjust altitude using "up" button to correct altitude, that works. Then If I load Altimeter app, it shows correct altitude. But when pressing button on side to exit app, it resets it back to what it was before (in my case -47m).

I also tried pressing the "Default" Button in Altitude settings, but that just resets it to the original number.

### Expected Behavior
Altitude settings should persist when closing altitude app.

### Steps to Reproduce

1. Go to Settings>System>Altitude in settings. Adjust altitude offset.
2. Load "Altimeter" app, then press side button to exit. Altitude offset will reset.

### Installed apps

_No response_

bobrippling:
By also saving on reset

Fixes #3814

### DIFF
--- a/apps/setting/settings.js
+++ b/apps/setting/settings.js
@@ -1054,17 +1054,17 @@ function showAltitude() {
     /*LANG*/"Adjust up" : function() {
       Bangle.buzz(80);
       seaLevelPressure++;
-      Bangle.setOptions({seaLevelPressure:seaLevelPressure});
+      Bangle.setOptions({seaLevelPressure});
     },
     /*LANG*/"Adjust down" : function() {
       Bangle.buzz(80);
       seaLevelPressure--;
-      Bangle.setOptions({seaLevelPressure:seaLevelPressure});
+      Bangle.setOptions({seaLevelPressure});
     },
     /*LANG*/"Set Default" : function() {
       Bangle.buzz();
       seaLevelPressure=1013.25;
-      Bangle.setOptions({seaLevelPressure:seaLevelPressure});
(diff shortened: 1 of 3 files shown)


apps/setting/settings.js
```diff
@@ -1036,35 +1036,41 @@ function showAltitude() {
...
+      altitudeDone();
       popMenu(systemMenu());
     }},
     /*LANG*/"Pressure (hPa)" : menuPressure,
     /*LANG*/"Altitude (m)" : menuAltitude,
     /*LANG*/"Adjust up" : function() {
       Bangle.buzz(80);
       seaLevelPressure++;
-      Bangle.setOptions({seaLevelPressure:seaLevelPressure});
+      Bangle.setOptions({seaLevelPressure});
```
thyttan, line 1063: Is it a typo dropping the key:value pairing?

bobrippling: Ah this is [shorthand](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#property_definitions) - `{ x: x }` is equivalent to `{ x }`

thyttan: Whoa - trippy!

thyttan:
Looks good to me then! I did not try it though.

bobrippling:
Thanks!

PR #3821, opened 2025-04-24
Jsonclock v2

voloved:
After testing the watch on hardware for the past few days, I made a few changes:

* There was an array that would get elements added to it without the array resetting. It's been refactored so this cannto happen.
* The color scheme more matches VSCode.
* The icons and screenshots have been redone.
* Changed ordering of JSON to show the sun info last.


bobrippling:
Nice, thanks

PR #3822, opened 2025-04-24
android: pass HTTP request timeout to Gadgetbridge

flaparoo:
Currently, Gadgetbridge (Android) does not use the timeout value specified in a HTTP request. This is not according to the current documentation, and also not expected behaviour.

This PR fixes this - on the Bangle.js side. I've also created a PR in the [Gadgetbridge repo](https://codeberg.org/Freeyourgadget/Gadgetbridge). Both changes can be rolled-out independently.

### DIFF
--- a/apps/android/lib.js
+++ b/apps/android/lib.js
@@ -321,6 +321,7 @@ exports.httpHandler = (url,options) => {
   if (options.method) req.method = options.method;
   if (options.body) req.body = options.body;
   if (options.headers) req.headers = options.headers;
+  req.timeout = options.timeout || 30000;
   exports.gbSend(req);
   //create the promise
   var promise = new Promise(function(resolve,reject) {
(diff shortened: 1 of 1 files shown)


bobrippling:
Sounds good - tagging @gfwilliams with this being a central app

gfwilliams:
Just for reference, this is the PR: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4769

Looks good in theory, but I wonder if we want to wait 500ms *more* on the Bangle.js side than on the Gadgetbridge side?  Because if you specify a timeout, ideally you want Gadgetbridge to timeout first, then send the error message, and then have Bangle.js parse that and report the error back?

But once that's done I don't see a problem with merging this ahead of the Gadgetbridge merge, you'll just need to increase the version number and add to the changelog too.

flaparoo:
Adding an extra 500ms on the watch side can't hurt - that's done.

And I've also included the missing updates to the Changelog and version.

PR #3823, opened 2025-04-24
Updates to 2 apps, including 1 new one (HeatSuite)

nravanelli:
**coretemp**: Revised the application to be more like the BTHRM app. Added the ability to pair an ANT+ heart rate strap directly to the CORE Sensor so it can use the exertional algorithm when HR > 120 BPM.

**gpssetup**: Added adaptive mode where it can scan for GPS signals more frequently if a signal was found. Otherwise, reverts back to normal scan interval.

**heatsuite**: New app.


thyttan:
Would you feel comfortable resetting and redoing your commits so you end up with just three commits here, one for each app?

nravanelli:
Sure. 

I am going to close this pull request, and open a new one with 3 commits, one for each app.

PR #3825, opened 2025-04-24
widmsggrid: reset graphics state after setClipRect

thyttan:
as per https://github.com/espruino/BangleApps/pull/3813#issuecomment-2826952155.


bobrippling:
Nice one tracking that down too :)

PR #3828, opened 2025-04-25
settings: handle a `undefined` scroller (BJS1)

### LINKED ISSUE #3824: [Settings] Error compacting storage: Can't read property 'scroll' of undefined
### Affected hardware version

Bangle 1

### Your firmware version

2v25

### The bug

In the Settings app on Bangle 1 (but not Bangle 2), using Utils | Compact Storage always results in the following error appearing in the IDE:
```
>Compacting...\
 at line 3 col 2909 in setting.app.js
...enu[""].scroll=menuScroller.scroll;menuScroller=E.showMenu(m...
                              ^
in function "restoreMenu" called from line 18 col 69 in setting.app.js
...ct();restoreMenu(utilMenu());
                              ^
in function "a" called from line 1 col 187
...ected))}else if("function"==typeof a)a();else if("object"==typeof a){if("number"==typeof a.v...
                                                              ^
a?d.move(a):d.select()
                     ^
d()
  ^
>

```
The Bangle 1 appears to hang because the display with the "Compacting..." message remains, but in fact after compacting storage finishes, pressing BTN1 or BTN3 overwrites the display with selections of the Utils menu.

The same behavior/error also occurs when doing Utils | Turn Off | No

### Installed apps

_No response_

bobrippling:
Fixes #3824

### DIFF
--- a/apps/setting/ChangeLog
+++ b/apps/setting/ChangeLog
@@ -86,3 +86,4 @@ of 'Select Clock'
 0.75: Restore previous menu's scroll positions
 0.76: Add altitude calibration menu (and update README after menu changed)
 0.77: Save altitude calibration when user exits via reset
+0.78: Fix menu scroll restore on BangleJS1
--- a/apps/setting/metadata.json
+++ b/apps/setting/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "setting",
   "name": "Settings",
-  "version": "0.77",
+  "version": "0.78",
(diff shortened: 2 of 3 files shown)


bobrippling:
@thyttan I reckon we can merge this to fix #3824, going off [this comment](https://github.com/espruino/BangleApps/issues/3824#issuecomment-2831462674):

> Thanks for fixing the original problem I wrote up!

wdyt?

bobrippling:
Cheers!

PR #3835, opened 2025-04-28
score: fix bangle 2 compat and add squash preset

bobrippling:


bobrippling:
Tested and works well, will merge soon if no problems appear

bobrippling:
No issues in test - seems stable, merging

PR #3836, opened 2025-04-28
promenu: forward on touch event

bobrippling:
See #3804 and:
- https://github.com/espruino/BangleApps/pull/3604
- https://github.com/espruino/Espruino/pull/2565
- https://github.com/espruino/BangleApps/pull/3609
- https://github.com/espruino/Espruino/pull/2569

Also update types in folderlaunch (no need for new version)

- [x] Requires espruino/Espruino#2629
- [x] Requires espruino/Espruino#2630


bobrippling:
tested, forwarding works. alarms can also be enabled via a long-press

PR #3837, opened 2025-04-28
qrcode: permit custom app names

bobrippling:
... for storing multiple QR codes.

Also tidied up `custom.html`


bobrippling:
Tested and works well

PR #3840, opened 2025-05-03
clock_info: add BLE icon

bobrippling:
this allows a user to quickly toggle whether BLE is on/off

- [x] complete testing

### DIFF
--- a/apps/clock_info/ChangeLog
+++ b/apps/clock_info/ChangeLog
@@ -14,3 +14,4 @@
 0.13: Cache loaded ClockInfos so if we have clockInfoWidget and a clock, we don't load them twice (saves ~300ms)
 0.14: Check for .clkinfocache and use that if exists (from boot 0.64)
 0.15: Fix error when displaying a category with only one clockinfo (fix #3728)
+0.16: Add BLE clkinfo entry
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -138,7 +138,7 @@ exports.load = function() {
           NRF.wake();
           Bluetooth.setConsole(1);
         }
-        this.emit("redraw");
+        setTimeout(() => this.emit("redraw"), 250);
       },
       show: function(){},
       hide: function(){},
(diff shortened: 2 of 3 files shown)


bobrippling:
ready for review/merge

thyttan:
Just tried it - it's a nice one! 

One note:
- If I'm connected to the Web IDE and then toggle this clock info entry it disconnects alright. But the clock info graphic still says "ON" until the clock info is unfocused. I had expected the entry to say "OFF" no later than when the watch disconnects.

bobrippling:
Nice spot! Sorted in the latest push

thyttan:
Perfect - thanks!

gfwilliams:
I just updated and this appeared on my Bangle :) Just FYI when I run this on a clock face that attempts to fill the inside of the icons, it looks a bit strange.

image

The icon is also off-center and when drawn with a border on for some reason the border doesn't appear - I think clocks and clockinfo.drawBorderedImage are often expecting a 24x24px transparent image, and so the 11x20 non-transparent one caused issues. I just pushed a change to the image and it appears to have fixed that.

I don't know how you feel about the filled areas - we could just leave it, or I could add a small gaps to the BLE icon so the filled areas (which are calculated by doing a flood fill from the outside) are left out.

I think in general I'd avoid adding too many more things directly to the clockinfo library - we can always just have apps that add them, and the way clockinfo works now there's no performance drop compared to having them built into the library.

bobrippling:
Icons are definitely my weak point! Thanks for that adjustment, yeah I think we remove the filled areas if that's easy to add those gaps? And noted - I guess a `clkinfo_extra` or similar would be a nice place to put things like this in the future

gfwilliams:
> if that's easy to add those gaps?

Just done! Turns out it's not quite that nice. Without a border drawn, it all works great.

With the border, we draw the image border first, then flood fill - so any gap would have to be 3 pixels - but when we have a white border the middle filled isn't that noticeable anyway

bobrippling:
Ah I get you - the new one looks good, so thanks!

PR #3841, opened 2025-05-04
Zambretti Weather Forecaster

nxdefiant:
Zambretti Forecaster, uses the Barometer for empirical weather forecast in the Northern Hemisphere (see https://web.archive.org/web/20110610213848/http://www.meteormetrics.com/zambretti.htm), similar to weather stations. Watch must be stationary and its height above sea level set.

Uses code from widbaroalarm for measurements (boot.js)


bobrippling:
LGTM, thanks!

PR #3844, opened 2025-05-05
Daisy Ring Dynamically Created

voloved:
Currently, the Daisy rings can show only in 5% increments and must start at zero.
I rewrote that logic to be able to render the ring to be able to fill it  at any point of the ring.
The function is `drawRing(start, end, max)`.
So now,  we can start the ring and end it wherever we want with whatever increment we want. The ring looks the same (color and thickness) as previous versions, but now, if we' want to show the ring as 32% filled, it'll fill 32% instead of 30%.

This also does allow the seconds to update once a second, rather than once every 3, and the ring will fill  1/60 each second. Though it'll drastically run through the battery given the once-a-second update.
Along with dynamic rendering, I also added the toggle for an hourly buzz, where the clock will vibrate at the top of every hour if it's daytime.

### drawRing(85,15,100)
drawrRing(85,15,100)

### drawRing(15,85,100)
drawring(15,85,100)

### drawRing(0,15,60)
drawring(0,15,60)


apps/daisy/settings.js
```diff
@@ -3,40 +3,40 @@
...
+s[key] = saved[key];
+}
 
-  function save() {
-    settings = s;
-    storage.write(SETTINGS_FILE, settings);
-  }
+function save() {
+settings = s;
+storage.write(SETTINGS_FILE, settings);
```
bobrippling, line 26: The indent here and in the above `for` loop seems to have been lost in a reformat:
```suggestion
  settings = s;
  storage.write(SETTINGS_FILE, settings);
```

apps/daisy/app.js
```diff
@@ -318,9 +364,20 @@ function drawClock() {
...
   // recalc sunrise / sunset every hour
-  if (drawCount % 60 == 0)
-    updateSunRiseSunSet(date, location.lat, location.lon, settings.ring == 'Sun');
+  if (drawCount % 60 == 0) {
+    let recalcSunLeft = (settings.ring == 'Sun' || settings.hourly_buzz);
+    updateSunRiseSunSet(date, location.lat, location.lon, recalcSunLeft);
+  }
   drawCount++;
+
+  if (settings.hourly_buzz) {
```
bobrippling, line 373: The [`chimer` app](https://espruino.github.io/BangleApps/?id=chimer) could be used for this, to save on code-size and features

bobrippling:
Nice code-size reduction, looks good. Has it been tested much?

apps/daisy/settings.js
voloved: Cleaned up

apps/daisy/app.js
voloved: I personally like the logic being contained in my app, but I can gut it from this PR and leave it in my personal repo.

voloved:
> Nice code-size reduction, looks good. Added two little comments. Has it been tested much yet btw?

It has! My only concerns when testing it so far is the battery life for the seconds.
What I just did in a new commit is make the seconds update every 2 seconds when we're are 50% battery, and 10 seconds at 20% battery. I added a 5% deadband, so if we drop to 50%, we'll need to come up to at least 55% before we tick every second again. Battery is only checked once a minute to avoid a ton of analogReads.

apps/daisy/app.js
bobrippling: Not a problem, just in case you wanted!

bobrippling:
Sounds good to me, nice tweaks to the battery handling too - thanks!

PR #3845, opened 2025-05-07
bwclk: Adjust clkinfo positions for long text.

randyheydon:
Previously, all text in a clock info entry was centered.  This caused long lines to spill off both sides of the screen, and entries with more than two lines to cover the time.  With this change, clock info text is arranged so it only ever spills off the right and bottom. This allows the start of any text to always be visible, and the time to never be covered, regardless of the amount of text.  To achieve this, multiple lines are now left-aligned instead of center-aligned.

This fixes some of my own issues, but I am open to feedback on this because it might not be the best way to handle arbitrary clock info text.  Also, I have not copied this over to bwclklite yet, but I can do that after everyone is happy with the change.

Here are comparative screenshots for reference.

One long line switches to small font and left alignment:

| bwclk 0.34 | this PR |
|------------|----------|
| onelong | onelong-new |

Two lines change to left alignment:

| bwclk 0.34 | this PR |
|------------|----------|
| twoline | twoline-new |
| vic | vic-new |

More than two lines run off the bottom instead of into the time:

| bwclk 0.34 | this PR |
|------------|----------|
| threeline | threeline-new |
| fourline | fourline-new |

One short line is unchanged:

| bwclk 0.34 | this PR |
|------------|----------|
| oneline | oneline-new |
| steps | steps-new |

### DIFF
--- a/apps/bwclk/ChangeLog
+++ b/apps/bwclk/ChangeLog
@@ -34,3 +34,4 @@ clkinfo.addInteractive that would cause ReferenceError.
 0.32: Make the border of the clock_info box extend all the way to the right of the screen.
 0.33: Fix issue rendering ClockInfos with for fg+bg color set to the same (#2749)
 0.34: Support 12-hour time format
+0.35: Adjust clock info positions to better fit long text
--- a/apps/bwclk/app.js
+++ b/apps/bwclk/app.js
@@ -171,6 +171,7 @@ let clockInfoMenu = clock_info.addInteractive(clockInfoItems, {
       strWidth = g.stringWidth(text);
       if (strWidth+imgWidthClear > options.w) {
         g.setMiniFont();
+        text = g.wrapString(text, options.w-imgWidthClear).join("\n");
         strWidth = g.stringWidth(text);
       }
     }
--- a/apps/bwclk/app.js
+++ b/apps/bwclk/app.js
@@ -141,7 +141,7 @@ let clockInfoMenu = clock_info.addInteractive(clockInfoItems, {
     var hideClkInfo = info.text == null;
 
     g.reset().setBgColor(g.theme.fg).clearRect(options.x, options.y, options.x+options.w, options.y+options.h);
-    g.setFontAlign(0,0).setColor(g.theme.bg);
+    g.setFontAlign(-1,-1).setColor(g.theme.bg);
 
     if (options.focus){
       var y = hideClkInfo ? options.y+20 : options.y+2;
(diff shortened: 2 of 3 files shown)


thyttan:
This certainly looks better - thanks!

I could merge this now. But just want to hear your thoughts on if using some combination of reformatting strings by merging short lines and adding wrap to long lines. And also use the new function for selecting an appropriate font for a given string on a given area?

If interesting I can add links to documentation.

randyheydon:
Now that you mention it, I think that `g.wrapString` would be a good idea for one-line strings that are too long.  Give me a couple days and I'll add a commit.

I had considered wrapping everything, but it would mess up flow for some things that are deliberately on two lines.  The key example is agenda items, like the "Victoria day" screenshot above; agenda always puts the date/time on a second line, and wrapping long event names would force that date off the screen.  Well, agenda currently truncates event names to 12 characters so wrapping wouldn't happen, but part of my motivation here is that I'd like to change agenda to show more of the name.

I had not considered merging short lines.  I'm a bit against it for the same reason as above: some things are deliberately on multiple lines.  Maybe it could be a rule that two lines stays unchanged, but three or more lines get merged and re-wrapped?

Finally, I would be interested in a function to select font size.  Currently, bwclk just sets a large size or small size, but it would look much nicer if size could change smoothly with content.  I couldn't find such a function in the docs, but if you point me there, I'll try to use it.

thyttan:
> Finally, I would be interested in a function to select font size. [...] I couldn't find such a function in the docs, but if you point me there, I'll try to use it.

My bad, it's not in stable firmware yet so lets maybe wait a bit. Here's some examples from cutting edge firmware:


Do you want me to merge what you have so far? :)

randyheydon:
Don't merge yet, I still want to add line wrapping (at least for single lines, I'm still on the fence about what to do with 2+ lines).  I just struggle to find time.  I'll get to it in the next few days.

thyttan:
No need to stress at all 🙂

randyheydon:
@thyttan now I'm happy with it and ready to merge.  This commit adds wrapping only to single-line text.  I'm still on the fence about wrapping multi-line text, but this was ultimately easier.  Here's a visual comparison of wrapping a single line, with all other screenshots being unchanged.

| bwclk 0.34 | this PR initial | this PR now |
|------------|----------|--------------|
| onelong | onelong-new | oneline-newer |

thyttan:
Very nice - thanks! If you'd want to add that to bwclk lite as well when you have the time that would be very welcome 🙂

PR #3846, opened 2025-05-07
[lightswitch] Add setLCDPower

storm64:
Updated lightswitch to setLCDPower on actions initiated by the app:
 - on flash
 - on unlock
 - on switching the light via widget


thyttan:
Thanks!

PR #3847, opened 2025-05-09
spotrem: add volume knob gesture

thyttan:


bobrippling:
Very cool! Works well when I tested it. Do you think it's worth mentioning that the clockwise/anticlockwise gesture detection is stopped after a brief period of no-interaction?

Not for this PR, but if you wanted ideas on going further with this feature, it'd be cool to pull it out to a library and let other apps opt into the clockwise gesture, and during the clockwise gesture, display a circle on screen that follows the user's gesture. But obviously those are much bigger changes!

thyttan:
Thanks :)

> Do you think it's worth mentioning that the clockwise/anticlockwise gesture detection is stopped after a brief period of no-interaction?

Yes! Will add that.

> [...] it'd be cool to pull it out to a library and let other apps opt into the clockwise gesture, and during the clockwise gesture, display a circle on screen that follows the user's gesture. But obviously those are much bigger changes!

Yes - that's where I started with this. But I switched to this focused implementation first to dial in the UX (pun intended).

thyttan:
@bobrippling this is maybe done now! Please comment any feedback you may have :) 

It now adds two new modules for dial functionality, one for touch interaction and one for graphics. This is used in `spotrem` to provide the volume knob.

modules/Dial_Display.js
```diff
@@ -0,0 +1,59 @@
...
+
+    let drawCircle = (value, R, G, B, rad, isFill)=>{
+      let x = CENTER.x+27*Math.sin(value*(2*Math.PI/options.stepsPerWholeTurn));
+      let y = CENTER.y-27*Math.cos(value*(2*Math.PI/options.stepsPerWholeTurn));
+      g.setColor(R,G,B)
+      if (!isFill) g.drawCircle(x, y, rad);
+      if (isFill) g.fillCircle(x, y, rad);
+    }
+    if (this.isFirstDraw) {
+      g.setColor(0,0,0).fillCircle(CENTER.x, CENTER.y, 25);
```
thyttan: I guess the radii for the drawing operations should not be hardcoded to 25, 23, 9 etc but depend on the `DIAL_RECT`...

thyttan: Maybe adding a additional `displayRect` setting to the options object that if is not set defaults to what the `dialRect` option is.

modules/Dial.md
```diff
@@ -0,0 +1,60 @@
...
+======================
+
+> Take a look at README.md for hints on developing with this library.
+
+Usage
+-----
+
+```JS
+var Dial = require("Dial");
+var dial = new Dial(cb, options)
```
bobrippling: We don't use `Dial` here as a constructor (no `this` and it returns an object explicitly) so the `new` (while it works) might confuse users. What about:

```suggestion
var dial = dial(cb, options)
```

(suggested lowercase `dial` as it's no longer a class too, don't mind if you prefer it Titlecase though)

modules/Dial_Display.js
```diff
@@ -0,0 +1,59 @@
...
+    this.value += step;
+    //g.setFont("Vector:30");
+    //g.drawString(this.value);
+
+    let drawCircle = (value, R, G, B, rad, isFill)=>{
+      let x = CENTER.x+27*Math.sin(value*(2*Math.PI/options.stepsPerWholeTurn));
+      let y = CENTER.y-27*Math.cos(value*(2*Math.PI/options.stepsPerWholeTurn));
+      g.setColor(R,G,B)
+      if (!isFill) g.drawCircle(x, y, rad);
+      if (isFill) g.fillCircle(x, y, rad);
```
bobrippling: Perhaps:

```suggestion
      if (!isFill) g.drawCircle(x, y, rad);
      else g.fillCircle(x, y, rad);
```

modules/Dial_Display.js
bobrippling: Yeah, I think this is ok for now but we could have a follow-up PR which infers a radius from the rect? Or like you say, adds a setting

modules/Dial_Display.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+> Take a look at README.md for hints on developing with this library.
+
+Usage
+-----
+
+```JS
+var DialDisplay = require("Dial_Display");
+var dialDisplay = new DialDisplay(options);
```
bobrippling, line 12: Same again (as the [other comment](https://github.com/espruino/BangleApps/pull/3847#discussion_r2280978797)), but here I think we'll want to change to make it more class-like, or alter how `this` is used within the `dialDisplay` function. At the moment it'll be writing to the global object.

For example, in the IDE:
```javascript
> cb
=function (step) { ... }
>cb(1)
=undefined
>cb(1)
=undefined

// the dial renders successfully
// `this` variables are written to the global:

>isFirstDraw
=false
>value
=14
```

Happy to go through a few suggestions for it, if you like

bobrippling (commented):
> @bobrippling this is maybe done now! Please comment any feedback you may have :)
> 
> It now adds two new modules for dial functionality, one for touch interaction and one for graphics. This is used in `spotrem` to provide the volume knob.

Great, nice separation of functionality! I spotted a few things, let me know if I've been unclear or if you've any questions :)

modules/Dial.md
thyttan: OK - if that makes more sense I'm all for it. 

I looked at the Layout module for inspiration when developing ~~and taking a quick look now I fail to see how they differ in this regard. Out of interest, do they?~~

modules/Dial_Display.js
thyttan: Yes, lets leave it as is for now :)

modules/Dial.md
thyttan: Sorry, I think I get what you mean now!

modules/Dial_Display.md
thyttan: > Happy to go through a few suggestions for it, if you like

Please do! Fun with a learning opportunity :) Feel free to commit to this PR.

modules/Dial_Display.js
thyttan: This whole `drawCircle` should be removed I guess, when I think about it. I intended to use it more but now it's only used to make the small marks around the dial. (I had another design in mind at first where there would be small hollow circles around the dial, except for the one the indicator pointed to which would be filled.)

modules/Dial_Display.md
bobrippling: Of course, I'll pop something together

modules/Dial_Display.js
bobrippling: Ah gotya - I'll let you decide on `drawCircle`'s fate just in case you do find a use

modules/Dial.md
bobrippling: Cool, ok :)

modules/Dial_Display.md
bobrippling: [Here you go](https://github.com/espruino/BangleApps/pull/3847/commits/91a6b9d78638c7ea3f7737c877a18aa7c144d4b6?diff=split&w=1) - I've summarised the usage in the updated readme (inline below), let me know what you think. Happy to revert if you'd prefer a different approach :)


https://github.com/espruino/BangleApps/blob/91a6b9d78638c7ea3f7737c877a18aa7c144d4b6/modules/Dial_Display.md#L11-L20

thyttan: Thanks - looks nice I think! Wont claim to understand it completely yet -  I'll try to test it tomorrow or the day after. The only thing I got a little hung up on is if the name `reset` may be ambiguous - it could be interpreted as reset graphics, values or both. I'll think about that.

Thanks again :D

thyttan: I really like getting rid of those extra parameters and moving to methods instead. That's something I'll try to take with me going forward :)

I'm curious if using `set` twice before drawing can lead to not clearing the previous visual indicator position. I'll look at that when I get to testing.

bobrippling: Yes very true, `reset` is a little ambiguous! (feel free to change it, delete any of the methods or anything of course)

But yes - I found exactly what you said, `set()` makes it forget the previous value, so I altered the code slightly to deal with that too

thyttan: > so I altered the code slightly to deal with that too

I don't think you pushed this? 🙂

bobrippling: Ah, sorry I should've said - it was all in the same commit, the code keeping `prevValue` in sync is here :)

https://github.com/espruino/BangleApps/blob/91a6b9d78638c7ea3f7737c877a18aa7c144d4b6/modules/Dial_Display.js#L24-L32

thyttan:
@bobrippling would you mind having a look again? :) 

I fixed an edge case where `Dial_Display` could lose track of the previous position and not clear the indicator there. Also renamed the `reset` method to `queueRedraw`.

These are the commits:
https://github.com/espruino/BangleApps/pull/3847/commits/681ce21546bd5815c2b8c6ef1814a8148acaf999
https://github.com/espruino/BangleApps/pull/3847/commits/251da4fa684a0cd56ab91e2141bfedbc4a1a664f

bobrippling:
Of course - nice fix! That code's a lot more robust feeling too, LGTM :)

thyttan:
Thanks 🌞🙏

I'll go ahead and merge then! 

@gfwilliams FYI re adding new modules for rotary dial functionality.

gfwilliams:
Nice - thanks!

PR #3855, opened 2025-05-18
bwclklite: Adjust clkinfo positions for long text.

randyheydon:
This copies changes from bwclk into bwclklite.  For reference, see PR #3845 and commits 83bedcd and 04f3475.

Previously, all text in a clock info entry was centered.  This caused long lines to spill off both sides of the screen, and entries with more than two lines to cover the time.  With this change, clock info text is arranged so it only ever spills off the right and bottom. This allows the start of any text to always be visible, and the time to never be covered, regardless of the amount of text.  To achieve this, multiple lines are now left-aligned instead of center-aligned, and single lines are wrapped.

Here's a screenshot showing one before-and-after, example, but see PR #3845 for more thorough comparisons with bwclk.
| bwclklite 0.35 | this PR |
|--------|-------|
| bwlite-35 | bwlite-36 |


thyttan:
Thank you for adding that here as well! 🙂

PR #3860, opened 2025-05-24
Daisy Multiple Rings

voloved:
In the previous version for the Daisy watch, the rings are dynamically generated.
This made it easy to resize and modify them.

With that change, I added 2 new ways to display the rings, and the ability to display a max of 3 rings at once.
By default, the watch face will look exactly the same as the previous version's, but int he settings, you can add other rings.
See the images below. All of the text resizes based on the ring count.


bobrippling:
Very cool, thanks!

PR #3861, opened 2025-05-25
Chess: Use button rising edge for Menu

nxdefiant:
With falling edge and current firmware menu will always reopen on close.


thyttan:
Thanks!

PR #3862, opened 2025-05-25
Added ability to display only one counter and fast-scrolling

voloved:
Added fast-scrolling and ability to display only one counter.


bobrippling:
Nifty, thanks!

bobrippling:
@voloved I notice there's some adjustment to `lockTimeout`, is this intended to keep the screen on? How would you configure `counter2` to have the screen always unlocked?

PR #3866, opened 2025-05-26
msgtwscr: Temporarily activate scroll on twist function when a new message triggers the message app.

thyttan:
Can be installed from my app loader: https://thyttan.github.io/BangleApps/?id=msgtwscr

previous now closed PR: https://github.com/espruino/BangleApps/pull/3714


thyttan:
I think this works well enough. Logic could be made more sophisticated in the future.

apps/msgtwscr/boot.js
```diff
@@ -0,0 +1,62 @@
+{
+  // If doing regular loads, not Bangle.load, this is used:
+  if (global.__FILE__=="messagegui.new.js") {
+    onTwistEmitDrag();
```
gfwilliams: I think this won't work? Bangle.js doesn't pull function decls to the head of the file (to avoid parsing the file twice) so if this ever happens you'll just get a `ReferenceError` for `onTwistEmitDrag`

thyttan: Yes you're right! Thanks 👍

gfwilliams:
Sorry - just noticed one more thing. It's probably best to change `function onTwistEmitDrag` to `let onTwistEmitDrag = function...` because otherwise the function will end up being global - while `{}` hide `let/const` it doesn't work for `var or function`

thyttan:
I wonder why I didn't do that in the first place 🤔👍

thyttan:
Fixed 👌

gfwilliams:
Thanks!

PR #3870, opened 2025-06-02
Feat/timer widget

leonweber:
This PR adds a timer widget that can be controlled via gestures: Swipe a T (for timer) to start the timer and unlock the controls. Then single swipes adjust the time: right/left ±1min, up/down ±10min. Will buzz upon timer completion.


bobrippling:
Thanks!

PR #3869, opened 2025-05-29
Square Analog clock face

elcste:
https://elcste.github.io/BangleApps/?id=square

I made a clock face I'd like to share. It's a fork of Analog Dark (andark).

The biggest change is that it follows the system theme instead of always dark. I find the light theme much easier to read, especially when the screen is locked.

I also removed the battery display, tweaked the date to show the weekday instead of the year and changed the default settings.

Thanks!

PR #3871, opened 2025-06-02
andark: three new features

elcste:
I added some new features to andark. I first made a [new fork but was suggested to add them as settings](https://github.com/espruino/BangleApps/pull/3869).

* Show the weekday and not the year. (I rarely wonder what year it is but can often use a reminder of the weekday.)
* Hide the battery percentage. (I find this distracting on the clock and can check it using the swipable widgets. When this setting is enabled the font for the date is increased since there is more space.)
* Dark theme. With this setting disabled the clock will follow the system theme, light or dark. The setting sets dark theme all the time so it is enabled by default to match the current behavior. (I find the light theme more readable, especially in the sun with the screen locked.)

I updated the readme to list these settings and added a second screenshot showing some settings options enabled to contrast with the default options. I also enabled the link to open the emulator from the app loader, which I find convenient when trying our and developing apps.

Edit: Forgot [my app loader link](https://elcste.github.io/BangleApps/?id=andark)

### DIFF
--- a/apps/andark/README.md
+++ b/apps/andark/README.md
@@ -21,5 +21,5 @@
 * date and battery can be printed both below hands (as if hands were physical) and above (more readable)
 * hour hand can be made slighly shorter to improve readability when minute hand is behind a number
 * show the weekday and not the year
-* hide the battery percentage
+* hide the battery percentage; the font for the date is increased since there is more space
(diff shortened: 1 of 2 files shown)


thyttan:
Thanks - always fun with new contributors 🌼🎉😃

PR #3873, opened 2025-06-07
skyspy 0.11: Display glyph of GPS status

pavelmachek:
This adds another screen with graphical representation of GPS status, that can be used to quickly estimate when fix will be available and if user needs to move to get it

### DIFF
--- a/apps/skyspy/skyspy.app.js
+++ b/apps/skyspy/skyspy.app.js
@@ -200,10 +200,16 @@ let gps = {
   },
 };
 
-/* ui library 0.1.5 */
+/* ui library 0.1.5, dirty, bad, revert! */
 let ui = {
   display: 0,
   numScreens: 2,
+  clear: function() {
+    g.reset()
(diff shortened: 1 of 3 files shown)


apps/skyspy/skyspy.app.js
```diff
@@ -200,10 +200,16 @@ let gps = {
   },
 };
 
-/* ui library 0.1.5 */
+/* ui library 0.1.5, dirty, bad, revert! */
```
thyttan, line 203: "dirty, bad, revert!" ??

pavelmachek:
Sorry for a scary comment. I'm actually developing ui library version
0.2 elsewhere, and this is just a note that "the other" version is
superior and this is dead end. It will get sorted next time I update
the library.

Still this is what I tested with and should be ok (despite comment).

-- 
I don't work for Nazis and criminals, and neither should you.
Boycott Putin, Trump, and Musk!

thyttan:
Then - thanks! 😅

PR #3874, opened 2025-06-07
fileman 0.04: Add submenu to display free space and to allow quick navigation with many files

pavelmachek:


thyttan:
I think code looks good. I'll trust you tested this yourself. Thanks! 😉

pavelmachek:
Yes, I did. Thanks for merging (both this and skyspy stuff).

There's more work todo here, I guess we should display both free and
"freeable" space, and jumping would make more sense in by names than
by percents, but I'm not sure when I'll get to that and this is
already and improvement.

-- 
I don't work for Nazis and criminals, and neither should you.
Boycott Putin, Trump, and Musk!

PR #3876, opened 2025-06-10
[owmweather] prevent updating settings.updated when call to OWM fails

thinkpoop:
Link: https://thinkpoop.github.io/BangleApps/?id=owmweather

In `boot.js` function `onCompleted`, the `settings.updated` value is always updated, even if it's being called during error handing. 

Example:
GB gets disconnected.
OWMWeather tries to refresh, gets a timeout, sets settings.updated to now + 1hr.
GB gets reconnected.
OWMWeather checks settings.updated and thinks it doesn't need , even though the data is bad.

### DIFF
--- a/apps/owmweather/boot.js
+++ b/apps/owmweather/boot.js
@@ -10,10 +10,14 @@
     return settings.refresh * 1000 * 60 + 1;  // +1 <- leave some slack
   };
 
-  let onCompleted = function () {
+  let onCompleted = function (result) {
     loading = false;
-    settings.updated = Date.now();
-    require('Storage').writeJSON("owmweather.json", settings);
+    if(typeof result === "string") {
+      console.log("owmweather error: " + result);
+    } else {
+      settings.updated = Date.now();
+      require('Storage').writeJSON("owmweather.json", settings);
+    }
     if (timeoutRef) clearTimeout(timeoutRef);
     timeoutRef = setTimeout(loadIfDueAndReschedule, refreshMillis());
   };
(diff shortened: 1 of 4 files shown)


apps/owmweather/boot.js
```diff
@@ -10,14 +10,21 @@
     return settings.refresh * 1000 * 60 + 1;  // +1 <- leave some slack
   };
 
-  let onCompleted = function () {
+  let onCompleted = function (result) {
```
bobrippling: Is `result` used?

bobrippling:
LGTM, one small comment

apps/owmweather/boot.js
kidneyhex: Nope. But that question made me realize I missed an error situation.

In `lib.js`, the `parseWeather` method returns `undefined` on success, or a string on error. 

I've added checking that result and calling the appropriate callback based on it.

bobrippling: Nice! I'm happy to merge but if you wanted there is the option of trimming down a bit too:
```diff
       if (result === undefined) {
         if (completionCallback) completionCallback();
       } else {
-        if (errorCallback) errorCallback(result);
+        throw result;
       }
     }).catch((e)=>{
       if (errorCallback) errorCallback(e);
```

kidneyhex: I went one step further and put the throw into the `parseWeather` and got rid of the `result === undefined` stuff. (commit bd047f31f855e41a3a643adddb33e44c44bb615d)

bobrippling: Nice, thanks!

bobrippling:
Looks grand!

PR #3877, opened 2025-06-11
agenda: Change clockinfo title truncation to preserve images

randyheydon:
Previously, all event titles would be truncated to 12 characters when displayed in a clock info.  This hides some information, and completely breaks event titles containing embedded images (i.e. converted emojis).  This change simply removes the truncation, allowing the full title and full images to be shown.

I am open to feedback on this change, because I know it can make things look worse on some clocks.  With PRs #3845 and #3855, long agenda items look great on bwclk and bwclklite, but they won't look great on clocks that center clockinfo text (e.g. lato).  I would be open to keeping the truncating behaviour as long as we can find a way to avoid truncating embedded images; I have autogenerated birthday events that all start with a birthday cake emoji, so truncating to 12 bytes always results in nothing but a few dots displayed.

Here are some screenshots to compare:

| demo | agenda 0.18 | this PR |
|--|----------------------|---------|
| long text | screenshot_20250610-085120 | screenshot_20250610-090015 |
| emoji | screenshot_20250610-085230 | screenshot_20250610-085950 |
| long text in lato | - | screenshot_20250610-092120 |

### DIFF
--- a/apps/agenda/ChangeLog
+++ b/apps/agenda/ChangeLog
@@ -18,4 +18,4 @@
 0.16: Correct date for all day events in negative timezones, improve locale display
 0.17: Fixed "Today" and "Tomorrow" labels displaying in non-current weeks
 0.18: Correct date in clockinfo for all-day events in negative timezones
-0.19: Show full title in clockinfo instead of truncated
+0.19: Change clockinfo title truncation to preserve images
--- a/apps/agenda/agenda.clkinfo.js
+++ b/apps/agenda/agenda.clkinfo.js
@@ -64,7 +64,7 @@
 
   agenda.forEach((entry, i) => {
 
-    var title = entry.title;
+    var title = g.setFont("6x8").wrapString(entry.title,100)[0];
     // All day events are always in UTC and always start at 00:00:00, so we
     // need to "undo" the timezone offsetting to make sure that the day is
     // correct.


thyttan:
@gfwilliams do you have some thoughts regarding this?

gfwilliams:
Looking at Pebble++ that tries to render this stuff in quite a small area, it actually already does crop the text:

https://github.com/espruino/BangleApps/blob/master/apps/pebblepp/app.js#L103-L106

It does this using wrapString which is handy because that *will* preserve images... so I think removing the crop before that happens in that case could be good.

But before putting this in it'd be handy if you could go through the clocks that use clockinfo (https://github.com/search?q=repo%3Aespruino%2FBangleApps+.addInteractive+path%3A%2F%5Eapps%5C%2F%2F&type=code) and verify that they do in fact all implement some kind of cropping of text and won't get broken by long text titles? No need to install, but just check it's not just blindly rendering `info.text` in each case

randyheydon:
Here are screenshots of eight different clocks with long agenda titles.  I created an event titled "Test of very long event title that takes up space" and saw how that fit on each clock.  The results are... not encouraging.

| clock | screenshot | comments |
|---|---|---|
| bwclk (& bwclklite) | screenshot_20250612-063028 | Text is left-aligned, so excess is drawn off-screen.  Allows at least the start of all agenda items to be shown. |
| pebblepp | screenshot_20250612-065303 | Long text is replaced with "...", making most agenda items invisible.  Not sure why that's the result of the code linked above... |
| circlesclock | screenshot_20250612-063232 | [Uses the short text](https://github.com/espruino/BangleApps/blob/master/apps/circlesclock/app.js#L194), which is only the date, so it is unaffected by this PR.|
| lcdclock | screenshot_20250612-064831 | Text spills across to other clockinfo area, then spilled text gets covered up when other clockinfo area updates. |
| lcdclockplus | screenshot_20250612-064926 | Similar to above. |
| supaclk | screenshot_20250612-065557 | Also similar to above. |
| slopeclockpp | screenshot_20250612-070717 | Text spills across to time area, then spilled text gets covered up when time updates.  But given the small size of the clockinfo areas, it was likely spilling even with 12 characters. |
| lato | screenshot_20250612-065012 | Text is centered on the screen, causing it to spill off both sides and only show the middle. |

gfwilliams:
Thanks for all that testing! I just put a fix in for Pebble++ so it now crops text properly at least.

It's a bit tricky though - so right now, if we put this change in, over half the clocks you tested would probably be worse after it (even after I fixed Pebble++)?

I think in an ideal world we'd fix all the clocks and merge this, but I wonder if we'll really manage that?

I guess replacing:

```
var title = entry.title.slice(0,12);
```

with 

```
var title = g.setFont("6x8").wrapString(entry.title,100)[0];
```

should fix the icon issue at least?

randyheydon:
I added a commit changing to @gfwilliams logic above (set a representative font, wrap to a specified size, use only the first line).  This fixes the image issue without resulting in agenda item titles of unlimited length.  The only downside is that some titles end up shorter than they were before, because of wrapping by word rather than by individual character, but I think that's still a reasonable tradeoff, since I'd rather read whole words anyway.

Here's how a bunch of the clocks look now.  Most of them still don't handle agenda items very well, though I'd still say they handle this length better than an unlimited length.  Ultimately, the issue looks to be that they were designed around very short text strings (e.g. step count, heart rate) and don't include any special logic to handle longer strings cleanly.

| clock | screenshot | comments |
|---|---|---|
| bwclk (& bwclklite) | screenshot_20250619-033428 | Pretty much the same as 0.18, except that images work now. |
| lcdclock (& lcdclockplus) | screenshot_20250619-033534 | Still enough text to spill off the screen and to adjacent areas, but not as extreme as before. |
| supaclk | screenshot_20250619-033754 | Similar to above. |
| slopeclockpp | screenshot_20250619-033914 | Not much different from before: still too much text for the small area, and having two lines overlaps the image. |
| lato | screenshot_20250619-033643 | Also pretty much the same as 0.18, which is much better than the full-length titles were. |

I expect many clock developers simply don't think about longer text because the most common clock info items are quite short.  Perhaps this could be helped with a comment in the documentation about limits to clock infos (or lack of limits, if that's what you prefer).  And/or perhaps this could be improved with options in the Graphics class that would make it easier to handle long strings in limited space (rather than working such logic into each clock).  Two things that come to mind:
- Drawing to an area such that no drawing happens outside of that area (so text would be cut off at the edge of the box rather than spilling over to the surroundings).
- Centered drawing, but constrained so that the left side (or right side with right-to-left text) never leaves the designated area.

gfwilliams:
Thanks for all the testing! Looks great - just merging.

> Drawing to an area such that no drawing happens outside of that area (so text would be cut off at the edge of the box rather than spilling over to the surroundings).

We do have this (`g.setClipRect`) and clocks could use it. There's also `g.findFont` now which is better at finding a suitable font to render in the given area.

As you say I think the issue is that the vast majority of clockinfos use very short text, so clock face developers don't see any issues when testing. I guess maybe a 'developer' clockinfo could be made that contained long text (and maybe other things that are tricky to render?).

PR #3880, opened 2025-06-14
tileclk: Initial release v0.01

stweedo:
This is a new watch app designed to display time using tiles with simple color transition animations and optional borders for each tile. The app updates only the tiles that need changing (either clearing or drawing), making it somewhat efficient for normal background use when not always showing seconds.

Currently there is one clock info display at the lower part of the screen. This can be accessed by tapping that area or the seconds if they are visible. In the future I plan to add a second optional clock info display at the top where space allows.

There's a small issue with state persistence when using pattern launch versus the button. The app should remember the user’s preference for what was last displayed if it changed since the app was last exited. Apart from this, the app is ready to go.

Feedback is welcome.


stweedo:
I think I've squashed most of the bugs that were still lingering and most importantly discovered memory leaks that have now been resolved. It can be tested at https://stweedo.github.io/BangleApps/?q=tile and should now be ready to merge.

thyttan:
That's cool - thanks! :)

PR #3881, opened 2025-06-14
Replace non-ASCII dashes with ASCII hyphens to fix loader parsing and display issues

spycat111:
This PR addresses two critical bugs that occur only when installing the app via the development App Loader (they do not appear when uploading the script directly through the Web IDE):

Malformed regex / syntax error

Cause: The source uses the Unicode minus sign (U+2212) in the BT calibration menu, which the Espruino minifier mis-parses and embeds into crsclock.app.js as an unfinished regex.

Symptom:

Uncaught SyntaxError: Got UNFINISHED REGEX expected ']'
Location: crsclock.js ≈ lines 612–614.

Garbled on-device text

Cause: Use of the EN DASH (U+2013) in prompt strings for the sleep window and range selections.

Symptom: Dash characters render as â (mojibake) on Bangle.js.

Locations:

Sleep window label (lines 84–86)

Range prompts (lines 714 & 723)

All non-ASCII dashes have been normalised to the standard ASCII hyphen-minus (-) so that both the Espruino minifier and the Bangle.js firmware handle them correctly.

Changes:

crsclock.js

Replaced all occurrences of U+2013 (EN DASH) and U+2212 (MINUS SIGN) with -

Updated prompt strings and menu labels accordingly

crsclock.app.js (auto-generated; should now minify without errors)

Verification:

Uploaded via App Loader—BT calibration menu works, no syntax errors in console
Sleep-window settings now display 08:00–22:00 (with ASCII hyphen) correctly on the device
Uploaded via Web IDE—unchanged behaviour, as expected

References:

Sleep window string initially using EN DASH (lines 84–86)
Unicode minus sign in BT calibration menu (lines 612–614)
Range prompts using EN DASH (lines 714 & 723)

### DIFF
--- a/apps/crsclock/crsclock.js
+++ b/apps/crsclock/crsclock.js
@@ -83,7 +83,7 @@ const STATS_FONT_SIZE = 16;
 function getSleepWindowStr() {
   let a = ("0"+S.sleepStart).substr(-2) + ":00";
   let b = ("0"+S.sleepEnd).substr(-2)   + ":00";
-  return a + "–" + b;
+  return a + "-" + b;
 }
 
 function stability(arr, key, hours) {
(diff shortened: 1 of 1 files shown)


bobrippling:
Thanks - looks like there's a hyphen in a comment on line 521 and a unicode copyright on line 746 which might trip things up!

spycat111:
I created another PR #3883 for this - Thank you!

Additionally, PR #3882 is also relevant to this.

bobrippling:
Thanks!

PR #3889, opened 2025-06-17
Add Modern Clock (Reworked)

RKBoss6:
Fixed (hopefully) fast loading


RKBoss6:
How do I remove the last commit? I do not have a local repo on my computer to use git with.

PR #3893, opened 2025-06-19
Create Modern Clock App 

RKBoss6:
Removed commits for other projects

### DIFF
--- a/apps/modclock/app.js
+++ b/apps/modclock/app.js
@@ -165,7 +165,7 @@ let draw = function() {
     
   }
   
-
+  
   // draw the date, in a normal font
   g.setFont("Vector",18);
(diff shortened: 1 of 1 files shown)


thyttan:
I quite like this one, it looks neat. Will probably use it for a while!

But the time is off center for me: 



Do you want to fix that now or in an update? :)

(But also I think the off-center look is a bit cool 😅)

thyttan:
Also maybe an icon that doesn't make it look like it's an analog clock would make more sense?

RKBoss6:
Ah, the off center is caused because you don't have meridians enabled. I'm fixing it right now!

RKBoss6:
Should be fixed now!

thyttan:
Thanks! :)

PR #3895, opened 2025-06-19
Add icons for Bereal, Nextcloud, Thunderbird (Blue "mail" icon), Davx⁵, Kleinanzeigen, Element X

f-elle:
- Added icons (.png) for BeReal, Nextcloud, Davx⁵ (sync.png) and Kleinanzeigen.
- Changed icon_names.json and generate.js accordingly.
- Also added new entries for Thunderbird (blue mail.png icon) and Element X (matrix element icon) which were already available.

### DIFF
--- a/apps/recorder/ChangeLog
+++ b/apps/recorder/ChangeLog
@@ -56,3 +56,8 @@
 0.44: List tracks in reverse chronological order.
 0.45: Move recorder from widget into library
       Improve recorder ClockInfo icons
+0.46: Ensure altitude graph draws properly (or any graph using the last column of CSV data)
+      Lower accuracy of barometer data to ~1cm (saves about 15b/record)
+0.47: Fix 'blip' on speed map on some recordings
+      Ensure Battery voltage is only stored to 0.01v
+      Add graphs for Steps+Battery
--- a/apps/recorder/app.js
+++ b/apps/recorder/app.js
@@ -197,6 +197,14 @@ function viewTrack(filename, info) {
     menu[/*LANG*/'Plot HRM'] = function() {
       plotGraph(info, "Heartrate");
     };
+  if (info.fields.includes("Steps"))
+    menu[/*LANG*/'Plot Steps'] = function() {
+      plotGraph(info, "Steps");
+    };
+  if (info.fields.includes("Battery Percentage"))
+    menu[/*LANG*/'Plot Battery'] = function() {
+      plotGraph(info, "Battery");
+    };
   // TODO: steps, heart rate?
   menu[/*LANG*/'Erase'] = function() {
     E.showPrompt(/*LANG*/"Delete Track?").then(function(v) {
@@ -325,6 +333,7 @@ function plotGraph(info, style) { "ram"
   var lt = 0; // last time
   //var tn = 0; // count for each time period
   var strt, dur = info.duration;
+  if (dur<1) dur=1;
   var f = require("Storage").open(filename,"r");
   if (f===undefined) return;
   var l = f.readLine(f);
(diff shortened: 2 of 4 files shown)


f-elle:
This is my first pull request ever, so please be extra gentle... :smiley_cat:

gfwilliams:
Thanks! This looks really good - I'm just not sure about the Unicode for davx⁵ - are you sure those app names properly match what the apps are reporting themselves as to Bangle.js? If you're using Gadgetbridge you should hopefully be able to look in the Gadgetbridge logs and see

f-elle:
Thanks for your feedback! I've just checked the logs and tested the handling of messages coming in from Davx⁵. 
First of all, here's a screenshot of a message being displayed:
screenshot_20250624-121257-obfuscated
As we can see, the new icon is used, even though the app name is displayed as "DAVx". So apparently the Message UI can't display the Unicode '⁵' character, but internally the reported app name and the name in "my" new version correlate just fine.  

Here's another screenshot, showing an unopened Davx⁵ message waiting and the new icon being displayed properly.
screenshot_20250624-120454
  
Finally, for the sake of completeness, here's the relevant line from the Gadgetbridge log file:
`06-24 10:02:46.304 20106 20106 I n.freeyourgadget.gadgetbridge.service.devices.banglejs.BangleJSDeviceSupport: UART TX: \x10GB({"t":"notify","id":1750592403,"src":"DAVx\\u2075","title":"[redacted]","subject":"","body":"Netzwerk- oder E/A-Fehler - Unable to resolve host \\"[redacted]\\": No address associated with hostname","sender":""})`
... which also shows that the unicode character is included in the Gadgetbridge request.

gfwilliams:
Excellent - thanks for checking! As you spot Espruino doesn't have space to store all the Unicode chars by default. You can upload a smaller font that does contain them if you need though, but it doesn't look as nice.

It's just unfortunate that the Unicode will cause Espruino's pretokenisation optimisations to bail out, so calls to getImage might be a bit slower than before. Probably not too bad though.

gfwilliams:
Thanks - just merged (although next time please can you bump the version number and ChangeLog so folks know there's an update? like https://github.com/espruino/BangleApps/commit/d0e4d3455b1ca439c41c3c5f66711a41d9e16b05 )

Checked on timings and it goes from 45ms to 55ms to get an icon, so it's not the end of the world :)

f-elle:
Thanks for pointing this out! I will do that. I wasn't sure if the change log was perhaps auto-generated...
I've been using the new version since some days and didn't notice that it takes longer to load the icons so it really doesn't make much of a difference...

paul-arg:
hi @gfwilliams 

I reopen this discussion because it looks like the problem I had last year when running the generation script occured again here.
I updated the `messageicons` app and the icons for messages is now blank.

My MR: https://github.com/espruino/BangleApps/pull/3554
My forum post about it: https://github.com/orgs/espruino/discussions/6993

If this helps, I made the PNGs for these icons on MacOS with Photoshop, maybe there is something related to file encoding or something that messes with the generation script?

paul-arg:
I ran the PNGs through a metadata analyzer and indeed there is a difference between my PNGs (`agenda`, `gmail`, `leboncoin`, `messages`) and yours:

Mine: 
```
bit_depth  1
color_type Palette
```

Yours: 
```
bit_depth  8
color_type RGB with Alpha
```

Maybe that's worth exploring

gfwilliams:
> I updated the messageicons app and the icons for messages is now blank.

All icons? Because I can see in the `icons.img` file there's definitely a bunch of data (it's not all zeros)

Is this after installing from the App Loader, or do you mean that you re-ran the generation script and uploaded that and it broke?

paul-arg:
@gfwilliams I found the problem. Photopea was the culprit. I exported the broken PNGs with Gimp and it now works.
See my MR: https://github.com/espruino/BangleApps/pull/3975

PR #3897, opened 2025-06-20
Add settings to Charge Animation [Show Time, Show Battery Percent]

RKBoss6:
Adds app settings menu for showing time, and battery. There is a bit of flicker on the text, if anyone can get into it and help me find out why, that would be much appreciated. I know it has something to do with the text refreshing on top of the animation, but I would like to fix this issue.


apps/chargeanim/app.js
```diff
@@ -27,9 +77,23 @@ function anim() {
   });
   g.drawImage(imgbat, mx,my,{scale:b2scale, rotate:Math.sin(getTime()*2)*0.5-Math.PI/2 + b2rot});
   g.flip();
+  
+  if(settings.showTime==true){
+    g.drawString(clockStr,x,cy);
+  }
+  if(settings.showBatPercent==true){
+    g.drawString(batteryPercentStr,x,by,true);
```
gfwilliams, line 83: Can you put these 2 drawStrings *before* `g.flip()` please, so it's at the end? The flip is what force-pushes the graphics to the screen, so once you do that all the flicker should be gone!

RKBoss6:
Fixed! Thank you!

gfwilliams:
Looks great - thank you!

PR #3900, opened 2025-06-22
feat: japanese walking

Koell:
This app is based on the [NYT article](https://archive.nytimes.com/well.blogs.nytimes.com/2015/02/19/walk-hard-walk-easy-repeat/?_r=0) promoting alternating brisk and easy walking to boost fitness.

It provides:

    Configurable interval and total time

    Choice of start mode (Relax/Intense)

    Vibration alerts for mode change & completion

    Pause/resume with a tap

    Displays current time and time left


pavelmachek:
This really needs to display/compute distances from the step counter :-). [Ok, maybe that can be in next version, but...]

Koell:
> This really needs to display/compute distances from the step counter :-). [Ok, maybe that can be in next version, but...]

Not a bad idea, but i wanted to keep this simple. Maybe adding that optionaly would be good.
But probably just a stepcounter as distance is at best an aproximation

Koell:
Widgets are not cleared anymore, so if a stepcounter is desired, just add a widget

Ps: it is quite efficient to use a real device to test the code

thyttan:
I wonder now that you let widgets show there are widgets for displaying the current time - so maybe you don't need it in the app code anymore? But you decide :)

One reason I mention it is that when the screen locks the screen stops updating. If you want to keep the current time displayed on the app rectangle maybe it would make sense to redraw everytime the clock ticks a minute at least, as to not fool the user they are in the past! ;)

Koell:
Good points!

The clock can now be hidden. Additionaly the view will also be updated everytime the clock changes.
The update behavior can also be changed in the settings now.

thyttan:
Looks good - thanks!

PR #3901, opened 2025-06-23
ios 0.19: Convert numeric weather values to int from BangleDumpWeather shortcut

stweedo:
This ensures the numeric values are not being treated as strings when received from iOS shortcut BangleDumpWeather. It allows the weather codes to be parsed correctly to show the correct icons. Other numerical values were added for good measure. 

This was in regards to discussion [7791](https://github.com/orgs/espruino/discussions/7791).


gfwilliams:
Thanks!

PR #3902, opened 2025-06-23
weather: Add UV index display

stweedo:
This uses the already available UV index data from weather updates and displays it centered below the icon where there was available space. 

This was in regards to discussion [7791](https://github.com/orgs/espruino/discussions/7791).

### DIFF
--- a/apps/weather/app.js
+++ b/apps/weather/app.js
@@ -5,11 +5,43 @@ let current = weather.get();
 
 Bangle.loadWidgets();
 
+
 var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
   {filly: 1},
   {type: "h", filly: 0, c: [
-    {type: "custom", width: g.getWidth()/2, height: g.getWidth()/2, valign: -1, txt: "unknown", id: "icon",
-      render: l => weather.drawIcon(l, l.x+l.w/2, l.y+l.h/2, l.w/2-5)},
+    {type: "v", width: g.getWidth()/2, c: [  // Vertical container for icon + UV
+      {type: "custom", fillx: 1, height: g.getHeight()/2 - 30, valign: -1, txt: "unknown", id: "icon",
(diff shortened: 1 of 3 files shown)


RKBoss6:
Hey! One thing I noticed in the app was that the icon for weather was almost at the UV bar, and was clipped by the widget bar. Could you maybe reduce the size of the icon slightly to fix this? 

Screenshot 2025-06-24 at 9 33 25 PM

Thanks.

apps/weather/app.js
```diff
@@ -5,11 +5,43 @@ let current = weather.get();
...
 
+
 var layout = new Layout({type:"v", bgCol: g.theme.bg, c: [
   {filly: 1},
   {type: "h", filly: 0, c: [
-    {type: "custom", width: g.getWidth()/2, height: g.getWidth()/2, valign: -1, txt: "unknown", id: "icon",
-      render: l => weather.drawIcon(l, l.x+l.w/2, l.y+l.h/2, l.w/2-5)},
+    {type: "v", width: g.getWidth()/2, c: [  // Vertical container for icon + UV
+      {type: "custom", fillx: 1, height: g.getHeight()/2 - 30, valign: -1, txt: "unknown", id: "icon",
+        render: l => weather.drawIcon(l, l.x+l.w/2, l.y+l.h/2, l.w/2-10)},
```
stweedo, line 14: @RKBoss6 I had changed the -5 at the end of this line to -10 to fix the large icon issue, can you see if the version you are using has that change?

RKBoss6: Yep, that fixed it! Thanks!

thyttan:
I don't have weather set up myself. But it sounds like this works for you both @stweedo and @RKBoss6 ? If so I'll merge :)

RKBoss6:
@thyttan It works very well :)

thyttan:
Nice! Thanks @stweedo :)

PR #3904, opened 2025-06-24
[Power Manager] Round battery percentage

RKBoss6:


apps/powermanager/boot.js
```diff
@@ -74,8 +75,8 @@
         return result;
       })(Bangle[functionName]);
     }
-
-    let functions = {};
+    /* exported functions */
+    var functions = {};
```
Koell: 'functions' is assigned a value but never used
functions is never used, so remove it and lint should not complain

RKBoss6:
Thanks!

RKBoss6:
I also have noticed that monotonic percentage is not truly monotonic, it fluctuates, instead of only going down when not charging, and up when charging.

RKBoss6:
I will see if I can fix it, but anyone else is welcome to take a look as well

RKBoss6:
I've renamed the monotonic percentage and voltage to be stable percentage, because it is not truly monotonic, and in my mind, it makes much more sense this way. To not break existing user's settings, I only changed the display name. Under the hood, it's still saved as monotonic.

RKBoss6:
@bobrippling @stweedo What do you guys think, should it be monotonic, or does stable make sense here? Just making sure everyone else is fine with the change, and it doesn't just make sense to me...

apps/powermanager/boot.js
```diff
@@ -131,19 +133,34 @@
...
+        }
+      }else{
+        if(newPercent<savedBatPercent){
+          percentToReturn=newPercent;
+          savedBatPercent=newPercent;
+        }else{
+          percentToReturn=savedBatPercent;
+        }
+      }
+      return percentToReturn;
```
bobrippling: This is perhaps more readable like so:

```javascript
      if(Bangle.isCharging()){
        if(newPercent > savedBatPercent)
          savedBatPercent = newPercent;
      }else{
        if(newPercent < savedBatPercent)
          savedBatPercent = newPercent;
      }

      return savedBatPercent;
```

bobrippling:
> @bobrippling @stweedo What do you guys think, should it be monotonic, or does stable make sense here? Just making sure everyone else is fine with the change, and it doesn't just make sense to me...

Gonna be honest with you, I don't know what monotonic means and wouldn't expect the average bangle user to either, so I would vote for stable or bounded too

RKBoss6:
@bobrippling Yeah, I had to look it up when I came across it as well 🤣, stable seems much more user-friendly to me. Monotonic is how smartphones' battery meters work, only going in one direction unless charging, and while this stabilizes it, it doesn't prevent fluctuations, so stable seems like a much better choice here..

bobrippling:
Nice, yeah that makes sense to me - happy to merge if you are (looks like there's a small linting thing left to sort)

RKBoss6:
@bobrippling Yep, just fixed it :)

bobrippling:
Nice, thanks!

PR #3905, opened 2025-06-26
Add new fun facts to Fact list

RKBoss6:
Added many more fun facts, removed redundant ones. (Koalas are native to australia, Koalas are only found in australia)

### DIFF
--- a/apps/textfacts/facts.txt
+++ b/apps/textfacts/facts.txt
@@ -1,163 +1,125 @@
-This list of facts was created by asking the Google Gemini AI
+These facts were created with Gemini AI
 Black holes are regions of spacetime where gravity is so strong that nothing can escape
-Blue whales are filter feeders that eat krill
-Blue whales are the largest animals on Earth
 Blue whales can weigh as much as 33 elephants
-Cardiac muscle is found only in the heart
-Cartilage is a soft, flexible tissue that is found in the joints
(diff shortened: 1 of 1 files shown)


gfwilliams:
Looks great to me, thanks!

Just checking - this is tested and works?

RKBoss6:
I'm actually not sure what the issue is, but each fact when pulled is a mashup of other facts (ippos. A group of ravens i)

gfwilliams:
That would have been handy to know in the PR!

You need to run this code and use the `idx` and `count` values it prints:

https://github.com/espruino/BangleApps/blob/4bb0e47660cf38b0b396ad9fa50ed376f4042b8d/apps/textfacts/lib.js#L6-L21

I guess you did run them before, but count is 237 and yet there are 316 lines so I guess maybe it was run on out of date data?

gfwilliams:
@RKBoss6 any chance of fixing this? If you run the code again with the new data and update the library file it should be fine

RKBoss6:
Oh yeah, I had put this aside for a while, I'll get this working :)

RKBoss6:
I think it's fixed now!

gfwilliams:
Thanks for dealing with this - Copilot is very picky!

RKBoss6:
Hehe no worries! Glad this 6 month old PR was finally ready to merge :)

PR #3907, opened 2025-06-26
[Settings] Add new theme colors

RKBoss6:
Added Dark purple, maroon, dark green, dark blue.
Removed check for bangle.js 1 for: Orange, purple, grey.

### DIFF
--- a/apps/setting/ChangeLog
+++ b/apps/setting/ChangeLog
@@ -88,3 +88,4 @@ of 'Select Clock'
 0.77: Save altitude calibration when user exits via reset
 0.78: Fix menu scroll restore on BangleJS1
 0.79: Ensure that tapping on pressure/altitude doesn't cause a menu to display temporarily
+0.80: Add new theme colors (Indigo, Grey, Orange, Purple, Navy, Dark Green)
--- a/apps/setting/README.md
+++ b/apps/setting/README.md
@@ -12,7 +12,7 @@ See below for options under each heading:
 
 ## System - System settings
 
-* **Theme** Adjust the colour scheme
+* **Theme** Adjust the colour scheme. Choose between light mode, dark mode, or a custom theme.
 * **LCD** Configure settings about the screen. How long it stays on, how bright it is, and when it turns on - see below.
 * **Locale** set time zone, the time format (12/24h, for supported clocks) and the first day of the week
 * **Clock** if you have more than one clock face, select the default one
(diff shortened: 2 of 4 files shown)


thyttan:
Hi! Are you trying this on the actual watch to see that the colors, contrast and dithering looks good with these? :)

RKBoss6:
Yes, I have tried all of them. They all look good on backgrounds, used right. (No dithered colors for text). I am not sure if we need a 'purple' and a 'magenta' both, as they look pretty similar...
If you want, and are able to, try them on your end as well, and see if any changes need to be made from your view!

RKBoss6:
If we remove purple or magenta, for users with that selected, we then have to default to the other option though...
EDIT: Bangle saves the color, not index in menu, so that should be fine. Users cannot change it back to those colors if they select a different color.

thyttan:
I'm not 100% sure on this, sorry. Although at least some of the colors look good (I didn't test all of them yet, but especially the dark green goes well together with my wrist band) I'm cautious that adding many colors could make it somewhat overwhelming when choosing colors for all different options. I like a somewhat minimal and focused approach. Maybe I'm overly conservative here though.

Would one option be to have the settings app look for a color library and if a user installed it that can add extra colors to choose from?

@gfwilliams what do you think about this?

thyttan:
There seems to be two dedicated theme apps, "Theme setter" and "Themes": https://espruino.github.io/BangleApps/?q=themes

Maybe adding to or tweaking those would make more sense?

RKBoss6:
Ah, I did not realize you can do that! That is a much more elegant solution that flooding settings with colors. Thanks so much!

gfwilliams:
Thanks! Yes, it feels like adding to a separate app would be good.

Actually if you could do a PR with your changes to the README that'd be great (maybe link to those themes apps as well?)

Not sure if it helps (it feels like Theme Setter does some kind of demo image) but for Clock Backgrounds I had a bit of code that allowed the colors to be put into an image for the menu item: https://github.com/espruino/BangleApps/blob/master/apps/clockbg/settings.js#L17-L29

As an aside, if you're up for adding color combinations, the Clock Backgrounds app is really crying out for it: https://github.com/espruino/BangleApps/blob/master/apps/clockbg/settings.js#L97

There's even some code to help out with generating the palettes (https://github.com/espruino/BangleApps/blob/master/apps/clockbg/settings.js#L190-L209)

PR #3908, opened 2025-06-27
weather 0.28: Fix UV positioning, hide when 0

stweedo:
Separated the UV index display from the icon container as it made the icons smaller and interfered with the lazy rendering so that it wouldn't clear properly when drawing a new icon. It's now positioned under the wind speed and hides when the UV index is 0.

I also added "haze" in the BangleDumpWeather shortcut that maps to weather code "741" along with mist and fog and "shower" that maps to code "521" that was previously not used by the shortcut. I put the new link in the README. There is still one icon that this shortcut is not utilizing yet called drawBrokenClouds when it's in the 800's but not 800, 801, or 802. Any ideas of what we could filter for to show that? We already have "wind" and "part" map to "801" for drawFewClouds, and "cloud" maps to "802" for drawCloud.
https://www.icloud.com/shortcuts/ae5f3d7d6ed3460c98a3396b267aa1c5

### DIFF
--- a/apps/weather/ChangeLog
+++ b/apps/weather/ChangeLog
@@ -24,3 +24,4 @@
 0.25: Added monochrome parameter to drawIcon in lib
 0.26: Expose update function (for use by iOS integration)
 0.27: Add UV index display
+0.28: Fix UV positioning, hide when 0
--- a/apps/weather/README.md
+++ b/apps/weather/README.md
@@ -7,8 +7,11 @@ It also adds a ClockInfo list to Bangle.js.
 You can view the full report through the app:
 
 ![Screenshot](screenshot.png)
+
 ## iOS Setup
-Use the iOS shortcut [here](https://www.icloud.com/shortcuts/dbf7159200d945179e0938c15e64f102). The shortcut uses Apple Weather for weather updates, and sends a notification, which is read by Bangle.js. To push weather every hour, or interval, you will need to create a shortcut automation for every time you want to push the weather.
+
+Use the iOS shortcut [here](https://www.icloud.com/shortcuts/ae5f3d7d6ed3460c98a3396b267aa1c5). The shortcut uses Apple Weather for weather updates, and sends a notification, which is read by Bangle.js. To push weather every hour, or interval, you will need to create a shortcut automation for every time you want to push the weather.
+
 ## Android Setup
 
 1. Install [Gadgetbridge for Android](https://f-droid.org/packages/nodomain.freeyourgadget.gadgetbridge/) on your phone.
@@ -51,8 +54,8 @@ When you first load QuickWeather, it will take you through the setup process. Yo
 
 **Note:** at one time, the Weather Notification app also worked with Gadgetbridge. However, many users are reporting it's no longer seeing the OpenWeatherMap API key as valid. The app has not received any updates since August of 2020, and may be unmaintained. 
 
-
 ## Clock Infos
+
 Tap on any clockInfo when focused to directly open the weather app.
 Adds:
 * Condition ClockInfo with condition icon
@@ -60,6 +63,7 @@ Adds:
 * Wind speed ClockInfo.
 * Chance of rain ClockInfo.
 * Temperature ClockInfo without condition icon.
+
 ## Settings
 
 * Expiration timespan can be set after which the local weather data is considered as invalid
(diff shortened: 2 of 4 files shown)


stweedo:
The newest iOS shortcut utilizes all the icons that are available from the weather app and has been updated in the README.

RKBoss6:
Hey! I added a feels like display to the weather app, and updated the shortcut from here to send feels like as well. I am adding the readme changes to my PR. Is that alright?

thyttan:
> There is still one icon that this shortcut is not utilizing yet called drawBrokenClouds when it's in the 800's but not 800, 801, or 802. Any ideas of what we could filter for to show that? We already have \"wind\" and \"part\" map to \"801\" for drawFewClouds, and \"cloud\" maps to \"802\" for drawCloud.\nhttps://www.icloud.com/shortcuts/ae5f3d7d6ed3460c98a3396b267aa1c5

@gfwilliams do you have any insight for this?

stweedo:
@RKBoss6 That would be nice to have. The problem I ran into with the UV is there was not a lot of space and combining it with the icon container actually broke the icon clearing and also made them smaller. Maybe it could be switched from “real temp” to “feels like” when touching the temp and a little indicator showing if it’s the feels like temp. Maybe a small blue dot next to it if it’s colder than the real temp and red if it’s warmer? 

Another nice one to display would be precipitation chance since it’s already being sent. Perhaps an entire second page that could be swiped to with forecast information like the other weather PR out there. We might be able to get that from iOS too.

stweedo:
> > There is still one icon that this shortcut is not utilizing yet called drawBrokenClouds when it's in the 800's but not 800, 801, or 802. Any ideas of what we could filter for to show that? We already have "wind" and "part" map to "801" for drawFewClouds, and "cloud" maps to "802" for drawCloud.\nhttps://www.icloud.com/shortcuts/ae5f3d7d6ed3460c98a3396b267aa1c5
> 
> @gfwilliams do you have any insight for this?

@thyttan I actually mapped that one to mostly cloudy in the iOS shortcut so it is being used now. Thanks!

RKBoss6:
@stweedo I was debating creating a second page, so it wouldn't be as cramped. Unfortunately, I am pretty new at JS, and have no clue how to do that. ;) If you have any insight on how to do that, that would be very much appreciated!

RKBoss6:
As for the icon breaking, for right now, the only issue I can see is the icon does not fully erase when a new icon is drawn, like you mentioned. Do you know a way in Layout to increase the clearing size, or do we just use g.clearRect()?
For everything else, I think there is enough space that we can get away with it for now, but I do agree with you on another page. I know there is another PR right now for gadgetbridge forecast support, for daily forecasts perhaps. It would be pretty nifty if we could build another page for future forecasts as well.

thyttan:
@Rengyr @stweedo @RKBoss6 What do you think of the merge order of these weather PRs?

It seems to me this order would make sense:
#3908 
#3910 
#3911 
#3912 

There will probably need to be some tweaks to subsequent PRs as preceding ones are merged.

@stweedo Do I read you right that this PR is ready to go with regards weather codes you wrote about?

stweedo:
Yes. This version was ready to go, but it might conflict with the positioning @RKBoss6 had in his PR. And the shortcut in this readme does not have the feels like temp. But I just commented on #3912 with a shortcut link that does everything including feels like.

stweedo:
Okay, one more change to the iOS shortcut. Added a check for “mostly sunny” so it doesn’t show the full sun, but the sun with a cloud in front of it. 

https://www.icloud.com/shortcuts/73be0ce1076446f3bdc45a5707de5c4d

RKBoss6:
@stweedo did you add the new shortcut to the Readme, or should I?

RKBoss6:
Honestly, now might not need it, but we could do with a seperate PR for readme and shortcut updates...

stweedo:
Yeah I did update the PR I had with the new link, which includes the Feels Like temperature.

Rengyr:
> @Rengyr @\stweedo @\RKBoss6 What do you think of the merge order of these weather PRs?
> 
> It seems to me this order would make sense: #3908 #3910 #3911 #3912

Looks good to me, just one thing

Currently, while the "feels like" is contained in the gadgetbridge v2 weather data, it's not really usable without parsing/requesting forecast data as it's not in v1 data. Though it could be added to v1 in gadgetbridge to fix this issue. Otherwise the field will always have unknown value in android.

Though I might have something more to discus on my PR, so it can be done last so it doesn't block other.

thyttan:
Ok - lets go with this one first then 🙂

PR #3910, opened 2025-06-28
Weather: Add v2 weather parsing in Weather App (forecast support)

Rengyr:
This PR adds:
- Adjust code to unified and more modern javascript style
  - The main reasoning for fixing formatting of the code is that at first I thought I would fork the Weather App to a new one with forecast, but after discussion on GadgetBrige PR with weather data it was decided to keep it in the current Weather App. So I reused the code I had and tried to keep the improvements to readability.
- Split settings and weather data into two files in storage
  - It's cleaner this way to keep settings separate from weather data
- Add support for new version of weather data and forecast from GadgetBridge (requires version 0.86.0 or higher)
- Add support to automatically fetch weather at time interval defined in settings (requires GadgetBridge version 0.86.0 or higher)
- Add button to settings to force fetch weather data (requires GadgetBridge version 0.86.0 or higher)
- Add new API to get weather from Weather App for other Apps
  - This is described more in readme with examples

The forecast data are parsed only when requested by `weather.getWeather(true)` to minimize processing time used. 

PS: I haven't tested it on BangleJS 1 and I haven't tested how it behaves with older version of GadgetBridge, but I don't see reason why it should break.

PSS: I wanted to add the support for forecast parsing/api separately, as I don't know how long it will take me to make the App to show the forecast data (mostly figuring out the design of it as there is not much screen space). However, with this change, someone else can make their own in the meantime or use the data for other purpose.

### DIFF
--- a/apps/weather/clkinfo.js
+++ b/apps/weather/clkinfo.js
@@ -1,5 +1,4 @@
-
-(function() {
+(() => {
     var weather;
     var weatherLib = require("weather");
 
@@ -8,9 +7,9 @@
       if(weather){
         weather.temp = require("locale").temp(weather.temp-273.15);
         weather.feels = require("locale").temp(weather.feels-273.15);
-        weather.hum = weather.hum + "%";
+        weather.hum = `${weather.hum}%`;
         weather.wind = require("locale").speed(weather.wind).match(/^(\D*\d*)(.*)$/);
-        weather.wind = Math.round(weather.wind[1]) + "kph";
+        weather.wind = `${Math.round(weather.wind[1])}kph`;
       } else {
         weather = {
           temp: "?",
--- a/apps/weather/lib.js
+++ b/apps/weather/lib.js
@@ -1,5 +1,5 @@
-const storage = require('Storage');
-const B2 = process.env.HWVERSION===2;
+const storage = require("Storage");
+const B2 = process.env.HWVERSION === 2;
 
 let expiryTimeout;
 function scheduleExpiry(json) {
@@ -7,7 +7,7 @@ function scheduleExpiry(json) {
     clearTimeout(expiryTimeout);
     expiryTimeout = undefined;
   }
-  let expiry = "expiry" in json ? json.expiry : 2*3600000;
+  let expiry = "expiry" in json ? json.expiry : 2 * 3600000;
   if (json.weather && json.weather.time && expiry) {
     let t = json.weather.time + expiry - Date.now();
     expiryTimeout = setTimeout(update, t);
@@ -15,7 +15,7 @@ function scheduleExpiry(json) {
 }
 
 function update(weatherEvent) {
-  let json = storage.readJSON('weather.json')||{};
+  let json = storage.readJSON("weather.json") || {};
 
   if (weatherEvent) {
     let weather = weatherEvent.clone();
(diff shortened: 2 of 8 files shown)


Rengyr:
Thinking about the other weather PRs, there are two open things from my side.

1. With iOS support, we might want to conditionally show the settings based on what "weather provider" App is installed. Mostly to not confuse users with settings that won't do anything.
2. It's not used in this PR, but there are some extra data in the v2 even without forecast data and GB supports fetching only that. Extending the api shouldn't be an issue, just would be yet another settings item to fetch "extended data without forecast" or something.

Rengyr:
I've added two things (both need the new GB).
- If we have at least extended version 2 data (even without forecast), it adds the "feels like temperature" to stored v1 weather. Still wouldn't rely that users will have the field, but can be used if it exists.
- Added to settings option to request extended data without forecast, maybe it will be useful for someone that wants extra weather data, but don't need the forecast (e.g. cloud cover, visibility, etc.).

Open points still are:
- Do we want to show all settings even if their installed weather provider (iOS in this case) can't support the settings (e.g. forecast)?
- I will rebase this PR to get rid of conflicts after all things are finalized, when we will be ready to merge it.

RKBoss6:
I would say for all android or iOS specific settings, show only if the app is there 'ios integration, gadgetbridge integration.' That way it won't clutter up settings for an iOS user, and vice versa. I was thinking about integrating forecasts into weather for iOS, but that probably won't be right away.

thyttan:
@Rengyr just a FYI that the other weather app PR's have been merged, if you want to adjust this to get it in as well 😇

RKBoss6:
@Rengyr I was planning on adding different screens to the weather app, to make it feel less cluttered. If you want, it would be good if you could add wind gust parsing, or other niche data that might come in handy, as we now might get different screens to do that. It won't come in use until the screens are done, but always nice to have, as I know nothing about bangle.js development for android XD...

apps/weather/settings.js
```diff
@@ -1,31 +1,87 @@
-(function(back) {
-  const storage = require('Storage');
-  let settings = storage.readJSON('weather.json', 1) || {};
+(back) => {
```
thyttan: As per the checker, this function needs to be wrapped with parentheses and exclude the finishing semi-colon:

```
((back) => {
....
})
```

(like it was before)

thyttan:
@gfwilliams Can I leave this to you? :) I know you were involved on the gadgetbridge PR and I don't use weather myself.

https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4824

Rengyr:
@RKBoss6 
I think I've added nearly all data that GadgetBridge supports (though of course that can probably be extended). Looking on the GB cached weather data there is only wind speed and direction, no gust speed.

Rengyr:
@thyttan You were faster than me to write reply on this PR :p

It should be ready to merge from my side.

However, I would appreciate if someone could double check if everything works as it should and no hidden bug made it through before it's merged for people (although I run this for few weeks on my bg2 watch, minus the few recent commits for iOS, and it works fine so far).

gfwilliams:
I don't actually use weather myself - I can take a look but I won't be working much the next 2 weeks I'm afraid so I can't promise I get to merge it that soon.

@glemco you did a bit of work on weather in the past - are you still using it?

thyttan:
> so I can't promise I get to merge it that soon.

Gotcha 👍

> glemco you did a bit of work on weather in the past - are you still using it?

Or maybe @rigrig as well then :)

cbkerr:
I've been running it from @Rengyr's app loader for a month or so to try out the forecast api and haven't noticed any bugs. I mainly just use the widget though. Thanks for all your work with forecasts!

Bangle.js 2, Android 14, QuickWeather

RKBoss6:
@Rengyr Is the forecast API exposed for potential iOS development? I've been wondering whether we could add forecasts to iOS as well, and looks like it should work if accessible...

Rengyr:
@RKBoss6
It mostly depends on what route you will take. The data from GB has custom encoding to save space, as sending it in JSON is slow for bangle (you can check the discussion on GB PR: https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4824). So getting the data to Bangle is probably one question, but if you want to send it in different format, we can always just add some tag to the data to differentiate iOS vs android and parse it in the Weather app based on that.

The API itself should be universal enough to support iOS. For non extended data (no forecast) we read the "old" weather.json because reading forecast data take a lot of time to parse/read and thus we try to delay parsing of the forecast until necessary. If forecast is requested via the API, we just check if weather2.json exists (if not, it falls back to weather.json without forecast). There is a check for field 'd' in the weather2.json that would signify it needs to be parsed from its binary encoding, otherwise it's stored decoded for faster loading.

Bottom line is, you just need to parse the extended day data and forecast (with the same fields as are currently defined) into weather2.json and it will work fine with the API (see [getWeather(...)](https://github.com/Rengyr/BangleApps/blob/6f9b5ad0b4bd348a17a243982ef681aabb606e62/apps/weather/lib.js#L125) in this PR).

apps/weather/clkinfo.js
```diff
@@ -8,9 +7,9 @@
       if(weather){
         weather.temp = require("locale").temp(weather.temp-273.15);
         weather.feels = require("locale").temp(weather.feels-273.15);
```
thyttan, line 9: My watch says "Feels: -262 C". I suspect it's because we convert to celcius both here and on Gadgetbridge side?

Gadgetbridge, Bangle support class:
image

But if that's the case I'm not sure why the same doesn't happen with the current temp value as the code would make it seem it's also twice converted to celcius. 🤔

cbkerr: Same here, showing large negative feels like Fahrenheit temp in the app. Under both extended and forecast data type setting.

apps/weather/lib.js
```diff
@@ -1,126 +1,368 @@
...
+    hum: weather2.hum,
+    rain: weather2.rain,
+    uv: weather2.uv,
+    code: weather2.code,
+    txt: weather2.txt,
+    wind: weather2.wind,
+    wdir: weather2.wdir,
+    wrose: weather2.wrose,
+    loc: weather2.loc,
+    feels: weather2.feels,
```
thyttan: Referring back to my previous comment, I guess this may be the difference. `feels` data does not get converted back to Kelvin here, but the other temp data above is. So that's the bug I think? :)

apps/weather/clkinfo.js
thyttan: Found the bug at https://github.com/espruino/BangleApps/pull/3910#discussion_r2280578552

apps/weather/lib.js
thyttan:
```suggestion
    feels: weather2.feels + 273,
```

Rengyr: You are right, thanks for pointing that out. I somehow blanked when I added the feels like (as it was way later than I wrote the rest of the code) and added only due to iOS using it in "basic" weather data. There was as well one other place where it was wrong, but that should be fixed with last PR as well.

thyttan: Nice one spotting that other place as well!

glemco:
> @glemco you did a bit of work on weather in the past - are you still using it?

Unfortunately I fried my bangle with a broken cable a few months ago..

thyttan:
@glemco sad to hear that!

@Rengyr I've tried the update now and seems to work as intended. I didn't see UV presentation since the provider I used doesn't seem to have it. But I feel comfortable merging this now - thanks a bunch for your work and contribution both here and to Gadgetbridge! :)

PR #3911, opened 2025-06-28
[iOS] Add feels-like value parsing for weather app support

RKBoss6:
I am adding feels-like temp as a display in the weather app, this is needed for correct parsing from iOS notification.


RKBoss6:
@gfwilliams Just bumped it :)

gfwilliams:
Thanks!

PR #3912, opened 2025-06-28
[Weather] Add feels-like temperature display

RKBoss6:


stweedo:
I just checked your version of the shortcut and it was missing the “mostly cloudy” filter that shows multiple clouds on the bangle. I added your changes as well and this version contains everything. https://www.icloud.com/shortcuts/71f49664b5a44b6cb80703b73af2d8f0

RKBoss6:
Alright @stweedo thanks!

thyttan:
>> @Rengyr @\stweedo @\RKBoss6 What do you think of the merge order of these weather PRs?
>> 
>> It seems to me this order would make sense: #3908 #3910 #3911 #3912
>
>Looks good to me, just one thing
>
>Currently, while the "feels like" is contained in the gadgetbridge v2 weather data, it's not really usable without parsing/requesting forecast data as it's not in v1 data. Though it could be added to v1 in gadgetbridge to fix this issue. Otherwise the field will always have unknown value in android.
>
>Though I might have something more to discus on my PR, so it can be done last so it doesn't block other.

_Originally posted by @Rengyr in https://github.com/espruino/BangleApps/issues/3908#issuecomment-3016306120_

@RKBoss6 I think we should not merge something that adds an non functional field for android users. What do you think we tweak the logic so it only shows for ios users for now?

Rengyr:
There might be support for "feels like temp" for android users later (and if users will use the new GadgetBridge extended weather data, which is optional setting). So best would be to limit not it based on the platform, but just check if the field exists in the weather data we get. Either we will have value or the field will be undefined, in which case we could just hide it?

RKBoss6:
@Rengyr I agree. We should hide the field when not in use, or no data is available.

RKBoss6:
Before merging this, a bug needs to be fix where the icon does not fully clear itself. To achieve proper clearing, do we use g.clearRect?

thyttan:
> Before merging this, a bug needs to be fix where the icon does not fully clear itself. To achieve proper clearing, do we use g.clearRect?

Since it uses the layout module it feels like there should be some way native to that for clearing correctly. If not maybe it should be added. I'm unfortunately not good at that module. I should learn it though...

But I'll pass the question along to @bobrippling, do you know?

apps/weather/app.js
```diff
@@ -9,59 +9,67 @@ Bangle.loadWidgets();
...
-      {filly: 1},
-      {type: "txt", font: "6x8", pad: 2, halign: 1, label: /*LANG*/"Humidity"},
-      {type: "txt", font: "9%", pad: 2, halign: 1, id: "hum", label: "000%"},
-      {type: "txt", font: "6x8", pad: [2, 2, 2, 2], halign: -1, label: /*LANG*/"Wind"},
-      {type: "h", pad: [0, 2, 2, 2], halign: -1, c: [
-        {type: "txt", font: "9%", pad: 2, id: "wind",  label: "00"},
-        {type: "txt", font: "6x8", pad: 2, valign: -1, id: "windUnit", label: "km/h"},
-      ]},
-      {type: "custom", fillx: 1, height: 15, id: "uvDisplay",
+        render: l => weather.drawIcon(l, l.x+l.w/2, l.y+l.h/2.1, l.w/2.1-10)},
```
bobrippling: I guess this is the icon that's not clearing? The `custom` layout will [clear it for you if you set `bgCol`](https://www.espruino.com/Bangle.js+Layout#layout-positioning), then so long as you draw in the bounds I'd expect you to be alright

RKBoss6: @bobrippling  I tried that, but it didn't look like it was being drawn in the bounds. How would you change the bounds for the icon? I'm asking because if we fit the icon in the bounds, there is a lot of wasted space that can be minimized to get a larger icon.

bobrippling: Interesting - I'm afraid i'm not great on icons, @gfwilliams do you know?

gfwilliams: I think they're treated like any other element in layout, it's just that the icon is drawn in the middle. So if you set other elements in the line to fillx:1 then they will fill any excess space

gfwilliams: Ideally you just get the actual icon and pass it in then it's automatic. Since you are supplying a render function not an icon it has no idea how big it should be, so you need to specify and width and height. 

https://www.espruino.com/Bangle.js+Layout#layout-positioning
https://www.espruino.com/Bangle.js+Layout#custom-elements

RKBoss6:
@bobrippling @thyttan @stweedo Alright, this should be good to go now. I fixed the icon clearing issue, by making it slightly smaller to fit inside the bounds, and changed the expiry text to be ambiguous to both iOS and Android.

bobrippling:
LGTM, @thyttan happy for you to merge when you're happy too, unless @stweedo you have anything in the meanwhile

thyttan:
Thanks @RKBoss6 and everyone for chiming in! 🙏🌞

thyttan:
@gfwilliams FYI

PR #3913, opened 2025-06-29
[widalarmeta] Add listeners to draw when unlocked

cbkerr:
I have this widget set to show seconds only while unlocked, but before this change it wouldn't redraw the widget to show seconds when the display wakes up.

I've been testing it for the past few weeks.


bobrippling:
Nice! Thanks

PR #3916, opened 2025-07-01
[Modern Clock] Add inline clockInfo

RKBoss6:
Added an inline clockInfo next to the date, in addition to the 2 clockInfos from before


RKBoss6:
The commit was badly named, I did nothing with espruino #2, just the second iteration of commit de91fbf.

bobrippling:
Thanks!

PR #3917, opened 2025-07-01
[Modern Clock] Fix clearing of date

RKBoss6:
Throughout the night, I realized that since the date was right aligned, it would not erase a long date to a short one. I added extra string date padding to fix this.


thyttan:
Thanks!

PR #3918, opened 2025-07-01
poweroff app: Use E.showMessage() instead of g.drawString().

f-elle:
Changed to use E.showMessage() instead of g.drawString() to display power off message. Looks better and is more coherent.

### DIFF
--- a/apps/poweroff/app.js
+++ b/apps/poweroff/app.js
@@ -28,6 +28,10 @@ if (showPrompt) {
  });
 } else {
   E.showMessage("Powering off...");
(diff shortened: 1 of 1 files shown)


bobrippling:
Sounds good - does the main functionality still work / tested ok?

f-elle:
It was good that you asked. I tested everything myself and figured that I took away some lines of code that I shouldn't have. Fixed it and now everything works as intended.

bobrippling:
Great stuff, looks good to me :)

PR #3919, opened 2025-07-03
Adding a Breakout clone called BrickBreaker to the games

IsReal8a:
Simple but effective mini-game that it took like forever for me to complete, plus the few changes happened after I started, but now is there, out in the open and ready to be part of the official Espruino BangleApps.


thyttan:
Ah a Bangle.js 1 app! I'll leave this to you @gfwilliams 🙂

gfwilliams:
Looks great! I'm happy to pull this in! 

Although @IsReal8a  did you do this on a real Bangle? Real Bangle.js 1 have the corners of the screen slightly covered by a curved panel, so I'd be surprised if the text in the top left/right corners could be read...

gfwilliams:
Can confirm - just tried on a real Bangle.js 1 and you can't see the corner text, and the graphics are very flickery as it's redrawing the whole screen each time without a buffer - but it's still playable so merging :)

I can see you've made it work on a variable screen width, so maybe you could take a look at getting it running on Bangle.js 2 (using the drag events to position the cursor?). That has square edges and an offscreen buffer so no flicker.

It's likely to get a lot more use too as the vast majority of people are now using Bangle.js 2 - there aren't that many v1 users around

PR #3925, opened 2025-07-04
Fixed some part of the flickering and the position of the HUD

IsReal8a:
Now it looks better, playable and tested on my BangleJS (found it!)


thyttan:
Thanks! 🙏🙂

PR #3926, opened 2025-07-05
[A Battery Widget] Background color matches Bangle’s theme

RKBoss6:
Rather than defaulting to black or white, this uses the background and foreground colors that the rest of the bangle uses, for greater cohesion with users’ themes.


thyttan:
Thanks!

PR #3931, opened 2025-07-09
[Storage Analyser] Add pie charts

RKBoss6:
From this [Issue](https://github.com/espruino/EspruinoAppLoaderCore/issues/68), @gfwilliams created a storage analyser app, to see what apps were taking up what storage. I added a pie chart visualization, from google-charts, and implemented an interface to show them.


RKBoss6:
I do feel as though in terms of sorting order, this app should come after firmware update in the app loader under the 'online' tab, as it is a system-level app. However, when I changed sortOrder in metadata, it did not affect anything, even though it should have. Does anyone know why this is?

gfwilliams:
Hi - this looks great!

I'm not sure why it's not going - maybe now the sortorder ends up only being used when the 'normal' sort method (eg favourited/new/etc) doesn't push things to the front.

Please can you take the sortorder line out for now? I think in general it's best not to have the sort order anyway - otherwise everyone thinks their app is important and should be ahead of everyone else's and we just have a race to see who can use the highest number.

RKBoss6:
@gfwilliams Ok, thanks for the help!

RKBoss6:
Added another pie chart for total storage taken

gfwilliams:
Looks great, thank you!

PR #3932, opened 2025-07-10
[Tag Launcher] Add haptic feedback in menu when selecting

RKBoss6:
Added haptic feedback when you select a category or an app, and changed the maximum vector font size, to allow for larger text and easier readability


bobrippling:
Tagging the app creator: @nxdefiant

nxdefiant:
Can you please make the buzzing optional (Add an entry to settings). I would like the option to disable it, or better have this disabled by default since the default launcher "launch" does not have this behavour.

@bobrippling Thank you for the notification.

RKBoss6:
@nxdefiant @bobrippling  Of course! Just updated it :)

nxdefiant:
Thanks, fine with me

apps/taglaunch/settings.js
```diff
@@ -21,9 +22,16 @@
...
       value: settings.vectorsize || 10,
-      min:10, max: 20,step:1,wrap:true,
+      min:10, max: 25,step:1,wrap:true,
       onchange: (m) => {save("vectorsize", m)}
     },
+    /*LANG*/"Haptic Feedback": {
+      value: settings.buzz == true,
+      onchange: (m) => {
+        save("buzz", m);
+        require("Storage").erase("taglaunch.cache.json"); //delete the cache app list
```
bobrippling: Does the cache need clearing in this case?

apps/taglaunch/app.js
```diff
@@ -108,14 +109,32 @@ let showTagMenu = (tag) => {
...
+        let app = appsByTag[tag][i];
+      
+        if (!app) return;
+        if (!app.src || require("Storage").read(app.src)===undefined) {
+          Bangle.setUI();
+          E.showMessage(/*LANG*/"App Source\nNot found");
+          setTimeout(showMainMenu, 2000);
+        } else {
+          load(app.src);
+        }
```
bobrippling: Perhaps we have this code in a function so it's not duplicated? (and the app becomes more lightweight as a bonus)

apps/taglaunch/settings.js
RKBoss6: Ah it really shouldn't need it. That was from something else I was testing...

RKBoss6:
@bobrippling Just changed it :)

apps/taglaunch/app.js
bobrippling: Ah sorry, I was thinking something like this, so we have the whole of the app loading factored out - wdyt?

```diff
     select : i => {
-      if(settings.buzz){
-        buzz();
-        //let the buzz have effect
-        setTimeout(() => {
-          let app = appsByTag[tag][i];
-
-          if (!app) return;
-          if (!app.src || require("Storage").read(app.src)===undefined) {
-            Bangle.setUI();
-            E.showMessage(/*LANG*/"App Source\nNot found");
-            setTimeout(showMainMenu, 2000);
-          } else {
-            load(app.src);
-          }
-        }, 27);
-      }else{
+      const loadApp = () => {
         let app = appsByTag[tag][i];

         if (!app) return;
         if (!app.src || require("Storage").read(app.src)===undefined) {
           Bangle.setUI();
@@ -137,6 +122,13 @@ let showTagMenu = (tag) => {
         } else {
           load(app.src);
         }
+      };
+      if(settings.buzz){
+        buzz();
+        //let the buzz have effect
+        setTimeout(loadApp, 27);
+      }else{
+        loadApp();
       }
     },
     back : showMainMenu,
```

RKBoss6: @bobrippling That makes a lot more sense! Just committed!

bobrippling:
Great!

PR #3933, opened 2025-07-10
Create BackLite App

RKBoss6:
This app saves battery by mimicking newer modern smartwatches that have a dedicated backlight button - unlock the bangle normally, and the backlight remains off, for cases where you do not need the backlight. Long press the button to unlock it, and the light will turn on for 5 seconds. Just enough for you to see what you need to see.

### DIFF
--- a/apps/backlite/metadata.json
+++ b/apps/backlite/metadata.json
@@ -2,7 +2,7 @@
   "id": "backlite",
   "name": "BackLite",
   "version": "0.01",
-  "description": "Conserves battery life by turning the backlight on only on a long press of the button from a locked state. ### Requires the latest settings update (v0.80)",
+  "description": "Conserves battery life by turning the backlight on only on a long press of the button from a locked state. **Requires the latest settings update (v0.80)**",
   "icon": "icon.png",
   "type": "bootloader",
   "tags": "system",


apps/backlite/boot.js
```diff
@@ -0,0 +1,33 @@
+
+//Set LCD to zero every reboot
+let s = require("Storage").readJSON("setting.json", 1) || {};
```
bobrippling: Here `s` should be `delete`d or live in a `{ ... }` block to free the memory

(same for `longPressTime`)

pavelmachek:
Can you add some power estimates? AFAICT backlight does not use that much power on b.js2.

RKBoss6:
@pavelmachek It does, especially if you have a high brightness set. This app tries to eliminate time spent with the backlight on. I'll see if I can get some data for you, do you know how to? Thanks :)

apps/backlite/boot.js
RKBoss6: @bobrippling I've kept the entire boot.js inside a `{....}` block, would that work?

bobrippling: Nice, yeah that'll do it

RKBoss6:
Added settings, readme updates as well, ready for merge!

apps/backlite/ChangeLog
```diff
@@ -1 +1 @@
-0.01: New app
+0.01: New app! (settings, boot.js) TODO: Add a setting for long press time?
```
bobrippling: Let's move the TODO into the readme (or its own todo file)

bobrippling:
Looks good - shall merge, just a small tweak to the changelog if you don't mind

apps/backlite/ChangeLog
RKBoss6: @bobrippling Of course, just changed it :)

bobrippling: Nice

PR #3934, opened 2025-07-11
[Settings] Add 0 level brightness setting for LCD settings

RKBoss6:
In most modern smartwatches, there is a way to disable the backlight, and bangle.js does not have that. This update adds the option for the brightness to be set to zero, allowing for more customization over light and power usage. This also changes the default LCD setting from full brightness to zero brightness, allowing for more battery life right off the bat.


apps/setting/settings.js
```diff
@@ -326,20 +326,24 @@ function showThemeMenu(pop) {
...
-    rgb[/*LANG*/'black'] = "#000";
-    rgb[/*LANG*/'white'] = "#fff";
-    rgb[/*LANG*/'red'] = "#f00";
-    rgb[/*LANG*/'green'] = "#0f0";
-    rgb[/*LANG*/'blue'] = "#00f";
-    rgb[/*LANG*/'cyan'] = "#0ff";
-    rgb[/*LANG*/'magenta'] = "#f0f";
-    rgb[/*LANG*/'yellow'] = "#ff0";
-    if (!BANGLEJS2) {
-      // these would cause dithering, which is not great for e.g. text
```
bobrippling: I think we want to keep this for just Bangle1 (?)

bobrippling (commented):
I might be mixed up but I don't see a brightness change here, just colours!

RKBoss6:
@bobrippling You're completely right! That was something else I was working on, glad you caught my mistake before it was too late!

bobrippling:
;)

Tagging @gfwilliams with this being a central app. Also @RKBoss6 looks like the modules lines from the compiled code have made it in - can you remove?

RKBoss6:
@bobrippling Thanks for the catch!

apps/setting/settings.js
```diff
@@ -72,7 +72,7 @@ function resetSettings() {
     // clockHasWidgets: false,      // Does the clock in 'clock' contain the string 'Bangle.loadWidgets'
     "12hour" : false,               // 12 or 24 hour clock?
     firstDayOfWeek: 0,              // 0 -> Sunday (default), 1 -> Monday
-    brightness: 1,                  // LCD brightness from 0 to 1
+    brightness: 0,                  // LCD brightness from 0 to 1
```
gfwilliams: Sorry, but I'm not happy with changing the defaults here - this is going to result in me getting hundreds of complaints that there is no or a broken backlight.

apps/setting/settings.js
```diff
@@ -474,11 +474,12 @@ function LCDMenu() {
   Object.assign(lcdMenu, {
     /*LANG*/'LCD Brightness': {
       value: settings.brightness,
-      min: 0.1,
+      min: 0,
       max: 1,
       step: 0.1,
       onchange: v => {
-        settings.brightness = v || 1;
+        settings.brightness = v || 0;
```
gfwilliams: Same with the default change here - and why is there a print statement left in?

gfwilliams:
Hi - thanks for the PR. Sorry it took a while to review - I was away last week.

Please can you leave the defaults set as 1? As mentioned above it'll just cause a massive support headache when everyone asks why their backlight is broken.

Also, please can you set min to : `min : BANGLEJS2 ? 0 : 0.1`

For people with the old Bangle, turning off the backlight will make the screen basically unreadable and then they won't be able to see to put it back

RKBoss6:
@gfwilliams Makes total sense! Just changed it, thanks :)

gfwilliams:
Looks good - thanks! Just merging

PR #3943, opened 2025-07-23
counter2: always-unlocked option

bobrippling:


bobrippling:
Tested, no problems - merging

PR #3945, opened 2025-07-23
[Settings] Update ChangeLog

RKBoss6:
I had recently added the option to set brightness to 0 in settings, but I had put incorrect defaults in the changelog, so I'm fixing that now :)


bobrippling:
Good spot!

PR #3946, opened 2025-07-25
[BackLite] Fix settings bug

RKBoss6:
Fixes bug where settings operator defaults to 0.


RKBoss6:
Does anyone know why in settings menu, when the app settings are opened, the back button is not shown? I have absolutely no clue as to why it is happening, and I've tried numerous ways around it, to no avail.

RKBoss6:
Ok, it's fixed now. Something was up during the null check, so I moved that to its own line.

apps/backlite/settings.js
```diff
@@ -1,19 +1,20 @@
...
-    require('Storage').writeJSON(FILE, settings);
+    require("Storage").writeJSON(FILE, settings);
   }
 
-  // Show the menu
   E.showMenu({
     "" : { "title" : "BackLite" },
-    'Brightness': {
-      value: 0.3|settings.brightness,  
+    "< Back": () => (back()), // fallback if run standalone
```
bobrippling: This can save a bit of memory with:

```suggestion
    "< Back": back, // fallback if run standalone
```

apps/backlite/settings.js
```diff
@@ -1,19 +1,20 @@
 (function(back) {
   var FILE = "BackLite.settings.json";
-  // Load settings
-  var settings = Object.assign({
-      brightness: 0.3,
-  }, require('Storage').readJSON(FILE, true) || {});
+
+  // Load settings safely (avoid crashes on bad/missing JSON)
```
bobrippling: Do you have some setup to reproduce the crash? I'd like to see what's going on there

RKBoss6: @bobrippling That was a comment from an older version that I left in by mistake... The crash is gone, but I suspect it was because of the null check that was there before, the `0.3|settings.brightness` , which would make the whole thing 0. No clue why that happened, but the `isFinite` seems to have fixed it now.

apps/backlite/boot.js
```diff
@@ -11,9 +11,9 @@
   //Set LCD to zero every reboot
   let s = require("Storage").readJSON("setting.json", 1) || {};
   s.brightness = 0;
-  if (!("lcdTimeout" in s)) s.lcdTimeout = 5; // fallback so  logic doesn't break
   require("Storage").writeJSON("setting.json", s);
-
+  //remove large settings object from memory
+  s=null;
```
bobrippling: If you like we can even free up the variable - this isn't standard JS bit a common pattern in espruino:

```suggestion
  delete s;
```

apps/backlite/settings.js
bobrippling: Ah makes sense, thanks for the details

apps/backlite/boot.js
bobrippling: Nice!

PR #3944, opened 2025-07-23
Create Smart Battery Module

RKBoss6:
This provides a module an app can use for battery. It learns your usage patterns over time by averaging out drainage rates, and extrapolates that into a pretty accurate prediction.

PR #3947, opened 2025-07-25
Create Smart Battery Widget

RKBoss6:
Uses `smartbatt` module to provide accurate predictions tailored to you.


RKBoss6:
I'm getting an error saying `widsmartbatt` has no entrypoint, anyone know why this is? Thanks.

sonicityV:
hey quick question, 

i was just browsing the source code. Whats the reason for updating once every 30 seconds instead of like once a minute or even 10 minutes (when not charging) and once a minute (when charging)?

bobrippling:
> I'm getting an error saying `widsmartbatt` has no entrypoint, anyone know why this is? Thanks.

The checker's expecting your widget to be named a particular way:

https://github.com/espruino/BangleApps/blob/76754841494be528a26a06b6fd98f7ad8749b778/bin/sanitycheck.js#L458-L458=

RKBoss6:
@bobrippling Thanks for the help!

RKBoss6:
@RelapsingCertainly As for the update times, this was modified from `A_Battery_Widget`, and still has the update time changing. As it uses the `smartbatt` module, it really doesn't make sense anymore, so I can remove it. Thanks for the catch!

RKBoss6:
This looks good to go :)

apps/widsmartbatt/widget.js
```diff
@@ -0,0 +1,90 @@
...
+}
+  WIDGETS["widsmartbatt"] = {
+    area: "tr",
+    width: 30,
+    draw: draw
+  };
+
+  // Touch to temporarily show battery percent
+  Bangle.on("touch", function (_btn, xy) {
+    if (WIDGETS["back"] || !xy) return;
```
bobrippling, line 64: Do we need to check the back widget? This widget'll be in the top-right so touches won't be near the back widget

RKBoss6: @bobrippling True, but if they use the widget editor app, that could change, so I think I'll just keep it in for now

RKBoss6:
This PR is good to go now, along with #3944

apps/widsmartbatt/widget.js
bobrippling: Fair!

bobrippling:
Thanks!

PR #3950, opened 2025-07-27
[Modern Clock] Add settings to change color when clockInfo is focused

RKBoss6:
Created a settings page for changing the accent color.


apps/modclock/settings.js
```diff
@@ -0,0 +1,48 @@
+(function(back) {
+  const FILE = "modclock.settings.json";
```
bobrippling, line 2: Does `modclock` need modifying to read this?

RKBoss6: @bobrippling I accidentally pushed my commits to a different branch! I'm so sorry, thanks for the catch.

RKBoss6:
As of now, this should be good to go

thyttan:
Thanks @RKBoss6 and @bobrippling!

apps/modclock/settings.js
bobrippling: Ah cool!

PR #3951, opened 2025-07-28
cload: Initial import.

pavelmachek:
This allows running binary code on Bangle.js2.


apps/cload/Makefile
```diff
@@ -0,0 +1,45 @@
...
+ARCH_FLAGS=
+CROSS=arm-linux-gnueabihf-
+CC=$(CROSS)gcc
+#  -nostdinc -- ouch, olive needs stddef.h
+#  -mfloat-abi=soft -- breaks build with __aeabi_fsub references
+CFLAGS=-ffreestanding -Os -nostdlib -fPIC $(ARCH_FLAGS) -mlittle-endian -mthumb -mcpu=cortex-m3  -mfix-cortex-m3-ldrd  -mthumb-interwork
+#  --emit-relocs
+LDFLAGS=-Ttext=0x0100 $(ARCH_FLAGS)
+
+all: $(TARGET).bin
```
bobrippling, line 12: We probably want `.PHONY: all` and the same for `dump`, etc below here

apps/cload/cload.app.js
```diff
@@ -0,0 +1,118 @@
...
+  let a3 = E.getAddressOf(img.buffer, true);
+  print(a1, a2, a3);
+  return ccode(0x8a61c, a1, a2, a3);
+}
+
+var binData;
+
+function load() {
+  bin = new Uint8Array(12*1024);
+  binData = Uint8Array(require("Storage").readArrayBuffer("cload.bin"));
```
bobrippling, line 54: Can this just be `binData = bin` (or make them the same) ?

apps/cload/cload.app.js
```diff
@@ -0,0 +1,118 @@
...
+  //j_test(img.buffer);
+  
+  print("Buffers");
+  ios = 128;
+  io = new Uint32Array(ios);
+  rams = 2*1024;
+  ram = new Uint8Array(rams);
+  
+  msg("Reloc");
+  io[3] = 0x51a87;
```
bobrippling, line 78: Might it be worth noting down the source of these addresses?

apps/cload/cload.c
```diff
@@ -0,0 +1,263 @@
...
+        "str sp, [r0, #0]       \n"
+        "str lr, [r0, #4]       \n"
+        "str r4, [r0, #8]       \n"
+        "str r5, [r0, #12]      \n"
+        "str r6, [r0, #16]      \n"
+        "str r7, [r0, #20]      \n"
+        "str r8, [r0, #24]      \n"
+        "str r9, [r0, #28]      \n"
+        "str r10, [r0, #32]     \n"
+        "str r11, [r0, #36]     \n"
```
bobrippling, line 106: Is there a `stm` equivalent for the target arch?

apps/cload/cload.c
```diff
@@ -0,0 +1,263 @@
...
+        "ldr r5, [r1, #12]      \n"
+        "ldr r6, [r1, #16]      \n"
+        "ldr r7, [r1, #20]      \n"
+        "ldr r8, [r1, #24]      \n"
+        "ldr r9, [r1, #28]      \n"
+        "ldr r10, [r1, #32]     \n"
+        "ldr r11, [r1, #36]     \n"
+
+        "bx lr                  \n"
+    );
```
bobrippling, line 121: Do we want clobbers on this (and the other `asm`s) ?

apps/cload/cload.c
```diff
@@ -0,0 +1,263 @@
...
+}
+
+volatile char welcome[] = "Hello\nwatch\n:-)";
```
bobrippling, line 124: If you make these (and all functions except public ones) `static` you'll get better loads/stores and potentially less memory use with the GOT depending on how the linker fares

apps/cload/cload.c
```diff
@@ -0,0 +1,263 @@
...
+}
+
+void flip(void) {
+		g_io[1] = 2;
+		switch_stack(&alt_ctx, &main_ctx);
+}
+
+// Save 32 FPU registers (S0–S31) and FPSCR (Floating Point Status and Control Register)
+void call_with_fpu_preserved(void (*func)(void)) {
+    __asm volatile (
```
bobrippling, line 150: Are the `asm`s really `volatile` ? There's nothing non-deterministic going on

apps/cload/cload.c
```diff
@@ -0,0 +1,263 @@
...
+		}
+		msg("Done");		
+	}
+}
+
+void prepare_alt_context(void) {
+    uint32_t *stack_top = (uint32_t *)(alt_stack + sizeof(alt_stack));
+
+    for (int i=0; i<sizeof(alt_stack); i++)
+	    alt_stack[i] = 0xbe;
```
bobrippling, line 214: Would the `memset` above do the trick here?

apps/cload/lib.c
```diff
@@ -0,0 +1,154 @@
...
+float cosf(float x) {
+    return sinf(x + 1.57079632679489661923f); // + π/2
+}
+
+typedef struct {
+    int quotient;
+    int remainder;
+} divmod_result;
+
+// __aeabi_idivmod: FIXME this likely needs to return values in registers
```
bobrippling, line 62: I believe the gcc attribute is `pcc_reg_return` or something like that (from memory)

bobrippling (commented):
Nice - tagging @gfwilliams for interest

gfwilliams:
It's great that you got this working, but I feel like surely this should just be added as `_example_compiled_c` or something? It's not just an app that someone can install from the app loader and magically have C.

There's also all the code for relocation you're doing, but I think if you use `-fpic` to ask for position independent code, the code will 'just work' when running from RAM without having to have a relocation step? The compiler we use for Inline C uses a bunch of flags to create its code that might help: https://github.com/gfwilliams/EspruinoCompiler/blob/master/src/compile.js#L457-L469

It might also be worth mentioning https://www.espruino.com/InlineC in the README. That's built in and while it's just for code defined inline, it 'just works' in the IDE and app loader I believe, so you don't have to worry about hard-coding addresses or anything.

pavelmachek:
gfw: I experimented a bit some time ago, and could not find reliable way to get code that did not need relocations.

I know about InlineC, but it relies on server magic somewhere, and is not really meant for, say, 2000 lines of C being imported that way. I'll add pointer to documentation.

gfwilliams:
This is the code I've used quite recently to compile C code to something that can be embedded. It worked great for me, and the code mostly comes from EspriunoCompiler. Hopefully it will help? Hopefully with it, we can do the without hard-coded addresses in the JS.

This is for one function, but I believe you should be able to handle others in a similar way - we already do `return E.nativeCall(new DataView(E.toArrayBuffer(bin)).getUint32(0,1),...` which uses the pointer in the first word as the function pointer, so you just need to ensure the second function pointer goes in 2nd word and so on

Example c file:

```
int main(uint32_t a) {
}

void *entryPoint[]  __attribute__ ((section (".entrypoint"))) = {(void*)main};
```

linker.ld:

```
ENTRY(entryPoint)

MEMORY
{
  RAM (xrw)       : ORIGIN = 0, LENGTH = 100K
}

SECTIONS
{
  .entrypoint :
  {
    . = ALIGN(4);
    *(.entryPoint)
  }
  /* We put everything in .text section as we load everything into RAM as one binary string */

  .text :
  {
    . = ALIGN(4);
    *(.text)
    *(.text*)
    *(.rodata)
    *(.rodata*)
    . = ALIGN(4);
    *(.data)           /* .data sections */
    *(.data*)          /* .data* sections */
    . = ALIGN(4);
    *(.bss)
    *(.bss*)
    *(COMMON)
    . = ALIGN(4);
  } >RAM

  /* Remove stuff we don't want */
  /DISCARD/ :
  {
    libc.a ( * )
    libm.a ( * )
    libgcc.a ( * )
    *(.ARM.*)
    *(.comment)
  }
}
```

Commands to create base64 code that can be run as-is:

```
INCLUDES="-I... "
CFLAGS="-mlittle-endian -mthumb  -mthumb-interwork -mcpu=cortex-m4 -mfloat-abi=softfp -mfpu=fpv4-sp-d16 -std=gnu11 -nostdlib -fno-common -fno-exceptions -fdata-sections -ffunction-sections -flto -fno-fat-lto-objects -Wl,--allow-multiple-definition -fpic -fpie -fpermissive -Os -Tlinker.ld "
arm-none-eabi-gcc $INCLUDES $CFLAGS compiledcode.c -o compiledcode.elf || exit 1
# arm-none-eabi-objdump -x -S compiledcode.elf # debug print
arm-none-eabi-objcopy -O binary compiledcode.elf compiledcode.bin || exit 1
echo "(function(){var bin=E.toFlatString(atob(\"`base64 compiledcode.bin  | tr -d '\n\r'`\"));return E.nativeCall(new DataView(E.toArrayBuffer(bin)).getUint32(0,1),'int(int)', bin);})"
```

PR #3952, opened 2025-07-28
drained: set type to clock, for launchers

bobrippling:


RKBoss6:
Hey! Not sure if this should be implemented in this app, or a seperate one, but I would really love to see a warning notification, eg. 'Battery is lower than 10%' before it switches to low power mode. Wdyt? Would that be a feasible implementation in this app? I could handle that in a separate PR, but then we might need to deal with some merge issues, so just running that by you @bobrippling. Thanks!

thyttan:
@RKBoss6 My suggestion would be to do that as a stand alone bootloader app instead.

RKBoss6:
@thyttan That's what I was thinking as well...

bobrippling:
Yeah I'd be happy with either, I don't mind :)

I'll merge this anyway - had no problems with it

PR #3953, opened 2025-07-31
[widalarmeta] Add option to only show when on clock

thinkpoop:
Link: https://thinkpoop.github.io/BangleApps/?id=widalarmeta

Adds an option to the Alarm ETA widget to only show when on the clock face.

I started using a widget to show the time when not on a clock. This way I see the alarm ETA when on the clock and the current time when not on the clock.

### DIFF
--- a/apps/widalarmeta/settings.js
+++ b/apps/widalarmeta/settings.js
@@ -7,6 +7,7 @@
     padHours: true,
     showSeconds: 0, // 0=never, 1=only when display is unlocked, 2=for less than a minute
     font: 1, // 0=segment style font, 1=teletext font, 2=6x8:1x2
+    whenToShow: 0, // 0=always, 1=on clock only
   }, require("Storage").readJSON(CONFIGFILE,1) || {});
 
   function writeSettings() {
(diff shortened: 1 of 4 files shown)


apps/widalarmeta/widget.js
```diff
@@ -55,44 +56,48 @@
...
-      calcWidth = g.stringWidth(text) + 2; // One pixel on each side
-      this.bellVisible = false;
-    } else if (config.drawBell && this.numActiveAlarms > 0) {
-      calcWidth = 24;
-      // next alarm too far in future, draw only widalarm bell
-      if (this.bellVisible !== true || fromInterval !== true) {
-        g.reset().drawImage(atob("GBgBAAAAAAAAABgADhhwDDwwGP8YGf+YMf+MM//MM//MA//AA//AA//AA//AA//AA//AB//gD//wD//wAAAAADwAABgAAAAAAAAA"),this.x,this.y);
-        this.bellVisible = true;
+    // If always showing, or the clock is visible
+    if (config.whenToShow === 0 || Bangle.CLOCK) {
```
bobrippling: Could we convert this to an early return here?

```suggestion
    if (config.whenToShow === 1 && !Bangle.CLOCK)
        return;
```

apps/widalarmeta/settings.js
```diff
@@ -59,5 +60,14 @@
         writeSettings();
       }
     },
+    /*LANG*/'When To Show': {
```
thyttan, line 63: I'm inclined to wanting this renamed to "Only on clock" or similar and then have it as a regular boolean setting with the standard checkbox formatting. Something to consider for a future update maybe.

thyttan:
I left a comment on this settings entry - please consider 🙏

But I'll merge this as is now 🌞 thanks!

apps/widalarmeta/settings.js
kidneyhex: Will do. I wasn't sure if some future person might want a "Not On Clock", but didn't add that to avoid unused features.

bobrippling:
Nice!

PR #3954, opened 2025-07-31
[touchtimer] Enable swipe to show widget top bar

Rafaesp:
Enable swipe to show the widget top bar.

This is my first contribution but I have some other improvements in mind.


bobrippling:
LGTM, thanks!

PR #3955, opened 2025-07-31
[iOS Integration] Add BangleDumpLocation shortcut compatibility

RKBoss6:
Added a shortcut to update the bangle's location in `mylocation`, without needing to use the web IDE or watch GPS. Useful for auto updating the location on trips, so sunrise/sunset info stays accurate.

I also made some readme changes, and added a contributors section. I know @stweedo contributed to this app, so I added them under it. If anyone else contributed, and wants to see their name under there, just comment, and I can add it to this PR.

### DIFF
--- a/apps/ios/ChangeLog
+++ b/apps/ios/ChangeLog
@@ -19,3 +19,4 @@
 0.18: Fix UTF8 conversion (check for `font` library, not `fonts`)
 0.19: Convert numeric weather values to int from BangleDumpWeather shortcut
 0.20: Add feels-like temperature data field to weather parsing from BangleDumpWeather shortcut.
+0.21: Add BangleDumpLocation shortcut, to update location data on watch without needing to use the watch's GPS.
--- a/apps/ios/README.md
+++ b/apps/ios/README.md
@@ -35,20 +35,21 @@ for now. It's just a few bucks/pounds/euro's.
 If you like to try a free app first, you can always use NRF Toolbox or
 Bluetooth BLE Device Finder to find and connect your Bangle.
 
-### Weather and Calendar
+### Weather, Calendar, and Location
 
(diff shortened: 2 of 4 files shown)


bobrippling:
Tagging @gfwilliams for a central app change

gfwilliams:
This looks really cool, thanks!

Only thing I'd say is how often does the location get updated? I guess it depends on the shortcut? If it's more than a few times an hour we might want to have some kind of rate limiting so we're not constantly writing to storage

RKBoss6:
@gfwilliams It only updates when the shortcut is called, one time. Depending on how often you run the shortcut, the storage will be written to, but I'm not sure that it makes sense to put a limit on it, because I don't think anyone will run the shortcut constantly every minute...

RKBoss6:
@gfwilliams I can add something that blocks against the same location being re-written though... Good idea :)

RKBoss6:
@bobrippling @gfwilliams Added a check to save flash memory, wdyt?

gfwilliams:
That all looks great, thanks!

I wasn't sure how shortcuts work on iOS - I'd assumed it was something that worked in the background so might have generated quite a lot of info.

Just FYI Espruino is smart enough not to write the exact same contents to the same file twice, but checking the difference in location as you've done is good as the GPS coordinates will likely always change a bit.

RKBoss6:
@gfwilliams Thanks! I'm wondering whether or not I should make a setting for the distance before it will update, for more customizability, and preference on how specific you want the location to be... What do you think?

gfwilliams:
I reckon it's fine as-is. Looks like 0.0001 is about 10m at the equator (less elsewhere) and I really doubt anyone's going to complain about that accuracy. IMO it could be 10x at 0.001 

The nice thing is if someone really cares they can do a PR :)

RKBoss6:
@gfwilliams Alright then, if it's good with you, this is ready to merge!

gfwilliams:
Great, thanks! Just merged!

PR #3958, opened 2025-08-04
[Power Manager] Fix percentage not updating.

RKBoss6:
I noticed that the percentage when stable percentage was on did not update. This fixes it.

### DIFF
--- a/apps/powermanager/boot.js
+++ b/apps/powermanager/boot.js
@@ -1,9 +1,10 @@
+
 (function() {
   var settings = Object.assign(
     require('Storage').readJSON("powermanager.default.json", true) || {},
     require('Storage').readJSON("powermanager.json", true) || {}
   );
-  
+  var savedBatPercent=E.getBattery();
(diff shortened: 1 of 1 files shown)


sonicityV:
is the new power manager ready to test?

RKBoss6:
@RelapsingCertainly Yep!

RKBoss6:
My app loader does not have this, so just update the boot.js in the ide

sonicityV:
Installed it, will report back once I've tested the functionality

sonicityV:
Wow, uh finally can confirm, battery percent is working again. 

P.s. the bangle is by far the worse thing to test battery related since it last so long (i.e. unless i run gps and hrm at the same time, i get maybe 0.5%/hr or something) 😔.

sonicityV:
@gfwilliams pinging gordon since ideally this gets merged before the stable refresh

gfwilliams:
This looks good - thanks!

PR #3960, opened 2025-08-06
[Clock Info] Altitude clockInfo uses locale for distance units

RKBoss6:
Instead of always showing the altitude in meters, it now gets the correct unit set from locale, and can display in feet, meters, or kilometers/miles if the altitude is too great.

### DIFF
--- a/apps/clock_info/ChangeLog
+++ b/apps/clock_info/ChangeLog
@@ -19,3 +19,4 @@
 0.18: Tweak BLE icon to add gap and ensure middle of B isn't filled
 0.19: Fix Altitude ClockInfo after BLE added
       Tapping Altitude now updates the reading
+0.20: Altitude ClockInfo now uses the distance units set in locale.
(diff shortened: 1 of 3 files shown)


gfwilliams:
Hi - this looks good, but why is there also the change with stepUpdateInterval in here? I quite liked the way it updated for each step, and moving to 2.5 minutes per update feels a bit low to me?

RKBoss6:
@gfwilliams Oh, that was from another attempt at preserving battery life for my watch, as the updates to that clockInfo killed the battery quite fast for me... I'll remove it, as I wasn't intending on adding that.

gfwilliams:
Looks good to me - thanks!

> Oh, that was from another attempt at preserving battery life

Yes, I can see how it could hurt if it's updating once a second - it depends how good the clockinfos are at updating, but it  can be slow especially when the icon has a border applied on it.

I'd be up for maybe doing something like:

```
var lastStepTime = getTime();
function stepUpdateHandler() { 
  var t = getTime();
  if (t < lastStepTime+10) return;
  bangleItems.find(i=>i.name=="Steps").emit("redraw"); 
  lastStepTime = t;
}
```

So only updating the step count once every 10 seconds. 

... but I'll merge this now.

PR #3961, opened 2025-08-06
[iOS] Readme updates

RKBoss6:
I noticed that the shortcut for BangleDumpWeather was the outdated one that still uses the OpenWeather API. This PR changes it to the new shortcut created by me and @stweedo using Apple weather in the readme.


thyttan:
Thanks!

thyttan:
@RKBoss6 could you link to the thread where you worked on the updated link? I remember reading it but it would be nice to reference it on this PR as well. To make it easy to trace back.

RKBoss6:
@thyttan Of course! [Discussion](https://github.com/orgs/espruino/discussions/7791)

PR #3964, opened 2025-08-13
[Weather] Use local units for wind speed

RKBoss6:
Self explanatory, uses local units for wind speed, instead of adding kph to the end.
I kept the string splitting for more flexibility down the line


RKBoss6:
This can be merged after #3910. That way, @rengyr doesn't need to change their PR to fix merge issues, as this is simpler to fix than #3910

RKBoss6:
I also wanted to change the screenshots of the app to match the new updates, but as in #3965, my watch is currently down for the count, hopefully should be back in 3-6 days...

bobrippling:
I'm happy to merge (after the branch is updated) or can wait for the screenshots, just let me know :)

RKBoss6:
Yeah, I think I'll wait for the screenshots. No telling when that might be, as all sensors are off and the screen is no longer updating...

RKBoss6:
Alright, my bangle is back!
Just changed the screenshot, should be good to go. @bobrippling Would you prefer a screenshot without the widgets, or with? Just asking, it looks rather cluttered with the widgets...

bobrippling:
Great, no this is good - thanks :)

PR #3966, opened 2025-08-16
Deko: Add settings menu and make widgets optional

Trippnology:
As per the title, I've added a way to make widgets optional.

They can now be configured to be always on (the default, so existing users will see no change), always off, or only shown when unlocked, for a cleaner look.

screenshot-no-widgets

This is my very first attempt at any kind of Bangle programming, so please be gentle with me! 😄


thyttan:
Hi! Thanks!

Preferably you don't reformat the surrounding code (and even better stick to that style for your contribution as well - but I'm not really picky about that). If you really need to do it then do it in a separate commit so it's easy to verify it's just formatting. And then the functional code additions can be looked at by itself in its own commit(-s) as well.

Other than that this sounds like a nice addition - thanks again :)

thyttan:
Are you comfortable either dropping the formatting changes or splitting them out into a separate commit? :)

Trippnology:
Ah, apologies... that was automatically done by my editor on save, and not something I did manually. I would rather not try to "unpick" the changes now (if you don't mind) as that would probably be more work than the actual changes. I didn't notice at the time, but in the future I will be sure to disable Biome, or at least make a separate commit with any formatting changes first.

In terms of what are my changes to `app.js`, lines: 1, 16-23, and 74-82.

Thanks for the heads up!

thyttan:
Yeah, I'll let it go for this PR but please refrain from it in the future :)

The checker cought two things: https://github.com/espruino/BangleApps/actions/runs/17012215744/job/48230939987?pr=3966

Trippnology:
OK, cheers. I'm so used to just never thinking about formatting that I didn't notice. Perhaps it's worth adding a `biome.json` file to this repo or even incorporating [Biome](https://biomejs.dev/formatter/) directly into the web IDE to enforce consistent formatting rules for everyone, or maybe that would just cause chaos when existing apps are updated.

The error `App deko file deko.settings.js should be evaluated as a function but doesn't end in ')'` is again caused by my editor automatically adding a `;` so I'll fix that, and add the `data` entry to the metadata too.

apps/deko/app.js
```diff
@@ -49,16 +62,26 @@ g.clear();
...
+  if (settings.loadWidgets === 1 || (settings.loadWidgets === 2 && !locked)) {
+    widget_utils.show();
+  }
+});
 // Show launcher when middle button pressed
-Bangle.setUI("clock");
+Bangle.setUI('clock');
 // Load widgets
 Bangle.loadWidgets();
 Bangle.drawWidgets();
```
thyttan: The `drawWidgets` call should be dependent on the setting as well right?

So either call it or call `widgetUtils.hide()` instead - I think.

Trippnology: I did try that approach first, but it caused some weird behaviour, perhaps because of how the clock updates only once per minute?

I don't remember exactly what was happening, but it was kind of "remembering" previous settings and showing widgets when it shouldn't and/or not showing widgets when it should (until a redraw).

Once I put that line back to how it was and switched to only showing/hiding widgets when the lock state changed, it all worked as intended. I also wanted to minimise disruption to any existing users who update, but haven't set the option at all.

thyttan: Try this maybe then:

```
Bangle.drawWidgets();
if (settings.loadWidgets === 0 || (settings.loadWidgets === 2 && locked)) {
  widget_utils.hide();
  draw();
}
```

Trippnology: At that point in the code, we are definitely unlocked as it's either the first boot, or we are coming from the launcher. We don't have access to `locked` as that is from the `lock` event handler.

All working 100% now. 👍

thyttan: > At that point in the code, we are definitely unlocked as it's either the first boot, or we are coming from the launcher.

Not necessarily 100% true, eg if coming via an [autoreset](https://banglejs.com/apps/?q=autoreset) (or some other timeout implemented somewhere else) the watch will (probably) be locked, so I think `settings.loadWidgets === 0 || (settings.loadWidgets === 2 && Bangle.isLocked())` is a good idea ([reference](https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_isLocked)). Bit of an edge case I admit.

thyttan:
Nice - thanks again! 🚀

Trippnology:
Awesome! Thanks for your help getting this over the line.

PR #3967, opened 2025-08-17
chargerot: attempt `Bangle.uiRedraw()` if possible (optimisation)

bobrippling:
... so we don't require a full `load()`

### DIFF
--- a/apps/chargerot/boot.js
+++ b/apps/chargerot/boot.js
@@ -5,9 +5,14 @@
     Bangle.on('charging', (charging) => {
         if (charging) {
             g.setRotation(chargingRotation&3,chargingRotation>>2).clear();
-            Bangle.showClock();
         } else {
             g.setRotation(defaultRotation&3,defaultRotation>>2).clear();
+        }
+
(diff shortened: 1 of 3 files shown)


apps/chargerot/boot.js
```diff
@@ -5,9 +5,14 @@
...
         if (charging) {
             g.setRotation(chargingRotation&3,chargingRotation>>2).clear();
-            Bangle.showClock();
         } else {
             g.setRotation(defaultRotation&3,defaultRotation>>2).clear();
+        }
+
+        if (Bangle.uiRedraw) {
+            Bangle.uiRedraw();
+            Bangle.drawWidgets();
```
bobrippling, line 14: This is [safe to call unconditionally](https://github.com/espruino/Espruino/blob/43535a60ae681bb7394361897d5d8e86b19a2c17/libs/js/banglejs/Bangle_drawWidgets_Q3.js#L2-L2)

bobrippling:
Passed a few days of testing with #3968

PR #3968, opened 2025-08-17
anton/slope clocks: pass `redraw` to allow `Bangle.uiRedraw()`

bobrippling:
To allow other apps/bootcode to redraw without a reload

### DIFF
--- a/apps/antonclk/ChangeLog
+++ b/apps/antonclk/ChangeLog
@@ -14,3 +14,4 @@
 0.10: Use Bangle.setUI({remove:...}) to allow loading the launcher without a full reset on 2v16
 0.11: Moved enhanced Anton clock to 'Anton Clock Plus' and stripped this clock back down to make it faster for new users (270ms -> 170ms)
       Modified to avoid leaving functions defined when using setUI({remove:...})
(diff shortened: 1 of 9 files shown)


RKBoss6:
Hey! I was just wondering, what is the purpose of `Bangle.uiRedraw?` I noticed in the app that inside of `setUI`, you added `redraw:draw`. When is it called, and what benefits are there for using it? Just curious about this, as I've never heard of it before, thanks!

bobrippling:
Hello! So [it](https://www.espruino.com/ReferenceBANGLEJS2#l_Bangle_setUI)'s a way of letting other widgets or bootcode know how to redraw the clock without reloading ([`load()`](https://www.espruino.com/ReferenceBANGLEJS2#l__global_load)) - currently only [used by `voldisp`](https://github.com/espruino/BangleApps/blob/c8eec1a545eb2dd56e716f676d190e928999cbe4/apps/voldisp/boot.js#L26-L27), but I wanted to have a faster response when I plug my watch in to charge (#3967).

bobrippling:
Passed a few days of testing with #3967

PR #3972, opened 2025-08-22
messagegui: Fix bugs related to empty titles and bodies.

### LINKED ISSUE #3969: [messagegui] Blank on task items with short titles
### Affected hardware version

Bangle 2

### Your firmware version

2v27

### The bug

Certain notifications appear in the message UI with an almost-fully blank screen.  When this happens to me, two of my widgets appear (widbatpc and widclk), but absolutely nothing else (not even the back button).  After waiting for the message timeout, the display returns to the clock.  Here's a screenshot of the blank notification:
Image

This specifically happens for task notifications with short titles and no description.  I use [tasks.org](https://tasks.org) on my phone, and my quick tests have shown:
- task title "Test" and no description gives the blank screen.
- task title "Test with longer title" displays as it should (with title in both the top and bottom spots).
- task title "Test" and description "Desc" displays as it should (with title on top and description below).

This bug has been happening for a while, so I can't narrow down exactly which update caused it.  This bug did not occur back when I was on the previous firmware version (2v25?), but I think there was an update to messagegui around the same time, so I can't pinpoint which is the culprit.

### Installed apps

widlock (0.08), notify (0.14), widalarm (0.02), reply (0.03), widmessages (0.07), widbatpc (0.20), quicklaunch (0.16), widgps (0.09), widbt_notify (0.18), delaylock (0.01), widclk (0.08), kbtouch (0.04), clkinfocal (0.05), widclkinfo (0.04), dtlaunch (0.28), bwclk (0.35), android (0.44), sched (0.33), messages (0.67), messageicons (0.09), boot (0.66), messagegui (0.88), clock_info (0.19)

randyheydon:
Fixes #3969.  Notification events can now have title or body as empty strings or `undefined` without errors.  Tested using the below notification messages sent from console.  Prior to this commit, all but the two marked test messages created errors.  Also tested with actual task notifications from Android.

```
GB({"t":"notify", "id":987123, "title":"Title"})
GB({"t":"notify", "id":987124, "title":"Title", "body":""})
GB({"t":"notify", "id":987125, "body":"Body"})
GB({"t":"notify", "id":987126, "title":"", "body":"Body"}) // worked before this commit
GB({"t":"notify", "id":987127, "title":""})
GB({"t":"notify", "id":987128, "body":""})
GB({"t":"notify", "id":987129}) // does nothing
```

### DIFF
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -361,12 +361,12 @@ function showMessage(msgid, persist) {
   active = "message";
   // Normal text message display
   let src=msg.src||/*LANG*/"Message", srcFont = fontSmall;
-  let title=msg.title, titleFont = fontLarge, lines;
+  let title=msg.title||"", titleFont = fontLarge, lines=[];
   let body=msg.body, bodyFont = fontLarge;
   // If no body, use the title text instead...
-  if (body===undefined) {
+  if (!body) {
     body = title;
-    title = undefined;
+    title = "";
   }
   if (g.setFont(srcFont).stringWidth(src) > g.getWidth()-52)
     srcFont = "4x6";
--- a/apps/messagegui/ChangeLog
+++ b/apps/messagegui/ChangeLog
@@ -118,3 +118,4 @@
 0.86: Default to showing message scroller (with title, bigger icon)
 0.87: Make choosing of font size more repeatable
 0.88: Adjust padding calculation so messages are spaced out properly even when using international fonts
+0.89: Fix bugs related to empty titles and bodies
--- a/apps/messagegui/metadata.json
+++ b/apps/messagegui/metadata.json
@@ -2,7 +2,7 @@
   "id": "messagegui",
   "name": "Message UI",
   "shortName": "Messages",
-  "version": "0.88",
+  "version": "0.89",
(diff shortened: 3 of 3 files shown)


apps/messagegui/app.js
```diff
@@ -361,12 +361,12 @@ function showMessage(msgid, persist) {
...
   let body=msg.body, bodyFont = fontLarge;
   // If no body, use the title text instead...
-  if (body===undefined) {
+  if (!body) {
     body = title;
-    title = undefined;
+    title = "";
   }
   if (g.setFont(srcFont).stringWidth(src) > g.getWidth()-52)
     srcFont = "4x6";
```
bobrippling, line 372: Thanks, looks good! To avoid a future regression, perhaps we comment about the relation here?

```suggestion
    srcFont = "4x6";
    // one of `title` or `body` must be truthy at this point (#3969)
```

randyheydon: I just added another commit with a comment.  I wrote it differently, because the total set of issues ended up being not just about truthiness but variables being undefined yet referenced later.

bobrippling: Thanks!

PR #3974, opened 2025-08-24
[RAM Widget] Update colors for readability on dark themes

RKBoss6:
Just as the title says, on dark themes, the color is changed from blue to cyan.


bobrippling:
Nice, thanks!

PR #3975, opened 2025-08-26
messageicons: fix broken pngs

paul-arg:
I found out that the PNGs I uploaded last year had some sort of encoding problem that prevented the generation script from working. I ended up manually inserting the missing bits but it was tedious and obviously the problem happened again when someone else added new icons.

To address the root cause, I finally found out that I needed to open the culprit PNGs in Gimp and manually export them to 8-bit RBG.

Then everything worked!


gfwilliams:
Excellent, thank you! I just added a check to the generation script will now fail if any image is blank

PR #3976, opened 2025-08-26
[Bordle] Add all valid wordle words

RKBoss6:
Adds all valid wordle words (2309 words), which surprisingly takes less than 1% of storage space

Tested, and works well


bobrippling:
Nice :)

PR #3979, opened 2025-08-29
fix: Erroneous apostrophe preventing firmware updates

Maffsie:
Hi! I noticed this was preventing the fwupdate app from actually downloading and installing firmware updates on my watches, because the accidental dangling apostrophe in the URL for each firmware entry both would cause a 404, but also prevents the `downloadURL` function from passing because it expects the URL to end with `.zip` or `.hex`.


gfwilliams:
Thanks for looking into this and getting a fix in. This looks great, thanks! Merging now!

PR #3980, opened 2025-08-30
messagegui: cancel/resume unread timeout on lock event

### LINKED ISSUE #3941: [Messages/Messagesgui] unread timer
When I am reading unread message, say new notification, it will close as i am reading it/scrolling. This feels wrong, should the unreadTimer reset upon user interaction?

thyttan:
fixes #3941

Can be tried via https://thyttan.github.io/BangleApps/?q=messagegui

I think it should be fine, but I'll test it for a bit before marking as ready for review.


apps/messagegui/app.js
```diff
@@ -646,8 +646,18 @@ setTimeout(() => {
 }, 10); // if checkMessages wants to 'load', do that
 
 /* If the Bangle is unlocked by the user, treat that
-as a queue to stop repeated buzzing */
+as a queue to stop repeated buzzing.
```
bobrippling: Minor English thing while we're here :)

```suggestion
as a cue to stop repeated buzzing.
```

apps/messagegui/app.js
```diff
@@ -646,8 +646,18 @@ setTimeout(() => {
...
+let isTimeoutSuspended;
 Bangle.on('lock',locked => {
-  if (!locked)
+  if (!locked) {
     require("messages").stopBuzz();
+    isTimeoutSuspended = !!unreadTimeout;
+    cancelReloadTimeout();
+  }
+  if (locked)
+    if (isTimeoutSuspended) {
```
bobrippling: To avoid running more code than we need, we could do this - I'll leave it to you to decide though, perhaps the current way is more readable :)

```suggestion
  } else if (isTimeoutSuspended) {
```

bobrippling (commented):
Nice, looks good to me

thyttan:
Seems this may be preventing unread timeout at least sometimes for my setup. So not ready right now.

PR #3982, opened 2025-08-31
Fix mntclock for weather v2 data

Rengyr:
Because the Mountain Pass Clock app reacts to the weather events directly, it breaks when the Weather app requires weather data v2. The clock reads the fields that are not in v2 data, and this sets the temperature to "undefined".

To not add mandatory dependency of Weather app, I have added both options. React to Weather app update if it's installed and use old (but fixed) way if the app is not installed.


bobrippling:
LGTM, thanks!

PR #3984, opened 2025-08-31
Add typescript auto-update "bot"

bobrippling:
WIP

### DIFF
--- a/.github/workflows/typescript.yml
+++ b/.github/workflows/typescript.yml
@@ -1,0 +1,62 @@
+name: Update typescript bindings
+
+on:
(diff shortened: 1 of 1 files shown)


RKBoss6:
I was actually looking through some of the code that has a typescript version, and I had a question, why is typescript there? The watch uses javascript, and having a typescript version is not a requirement... Just a question, is it for user-readability, or does some part of the watch actually use typescript?

bobrippling:
No, it's all JS - typescript just provides more guarantees against bugs, and documentation about what shapes of objects are where

PR #3985, opened 2025-08-31
lcars: fix row setting's max

### LINKED ISSUE #3981: [LCARS] Bug in app settings code prevents last data option from being selectable
### Affected hardware version

Bangle 2

### Your firmware version

2v27

### The bug

The 'CoreT' data option for the (wonderful!) LCARS watchface isn't selectable from the app's settings.

In lcars.settings.js, the max value for selecting the Row 1-3 data options (on lines 52, 61, and 70) is hardcoded to 8 but should be 9 (or better yet, be set to "dataOptions.length - 1") to ensure that all available data options are selectable.

### Installed apps

_No response_

bobrippling:
Closes #3981


thyttan:
Thanks!

PR #3987, opened 2025-09-03
[widalarmeta] Add a couple font options; Change only on clock setting to a checkbox

kidneyhex:
Link: https://thinkpoop.github.io/BangleApps/?id=widalarmeta

1) Added 5x7 and 5x9 font options

2) Switched the whenToShow setting to be a checkbox per thyttan's request: [link](https://github.com/espruino/BangleApps/pull/3953#discussion_r2265236088)

I wasn't sure how to best handle switching whenToShow to Boolean after it'd already been merged... there's a good chance no one but me has used it, and I'm fine if you'd prefer a renaming the variable.

### DIFF
--- a/apps/widalarmeta/widget.js
+++ b/apps/widalarmeta/widget.js
@@ -47,7 +47,7 @@
   function draw(_w, fromInterval) {
 
     // If only show on clock and not on clock
-    if (!!config.whenToShow && !Bangle.CLOCK) {
+    if (config.whenToShow && !Bangle.CLOCK) {
       this.nextAlarm = undefined; // make sure to reload later
       return;
     }


apps/widalarmeta/widget.js
```diff
@@ -45,7 +47,7 @@
   function draw(_w, fromInterval) {
 
     // If only show on clock and not on clock
-    if (config.whenToShow === 1 && !Bangle.CLOCK) {
+    if (!!config.whenToShow && !Bangle.CLOCK) {
```
bobrippling: In fact we don't even need the `!!` here, the truthiness of `1` / falsiness of `0` already work

```suggestion
    if (config.whenToShow && !Bangle.CLOCK) {
```

bobrippling (commented):
Looks good - the 0/1 vs false/true will work ok, the only downside being the variable name. The alternative is to add a new setting name and handle both.

Or we put a change in the app loader so it can migrate settings when we go from `0.14` to `0.15`

kidneyhex:
Yeah, I considered the variable name options but wasn't sure if they were worth the effort compared to just having an oddly named setting.

Let me know if you want me to do the rename and checking both values. Or if you have any other relatively easy ideas

bobrippling:
Cool, yeah I think the app code is small enough that the odd variable name isn't an issue - let's bring the code in - and thanks!

PR #3990, opened 2025-09-07
[Smart Battery Module] Add dynamic set intervals

RKBoss6:
Added intervals that change depending on total cycles, to focus less on updating the prediction, and more on reliability


bobrippling:
LGTM, thanks!

PR #3993, opened 2025-09-09
[Smart Battery Module] API changes

RKBoss6:
This just returns more in the exports.get() function, and has updated documentation along with it. Because the most recent update (maybe a day ago) is not yet deployed to the app loader, should I change the version for this one as well?


bobrippling:
LGTM - yes let's bump the version too please

RKBoss6:
Alright, just changed version, should be good to go!

bobrippling:
Thanks!

PR #3994, opened 2025-09-10
Tag app authors in PRs to their apps

bobrippling:
As seen in #3932, it can take a bit of research to find the author of an app. This PR adds an author field so people can add themselves to be notified of PRs which might need their review.
I've also added a few people who I could programatically glean from the git history.

Doing this from a branch within the repo to avoid the permission problems in #3935. Going off #3996, it looks like this'll need to be merged to `master` before PRs will be able to work with it.

- [x] Improve the bot message (group by app or author)
- [x] Author field:
  - [x] Update [metadata README](https://github.com/espruino/BangleApps?tab=readme-ov-file#metadatajson-format)
  - [x] Update [tutorial](https://github.com/espruino/EspruinoDocs/blob/master/tutorials/Bangle.js%20App%20Loader.md#adding-your-app) - see espruino/EspruinoDocs#753
  - [x] Move author field in `metadata.json` - after version
  - [x] Allow arrays, i.e. `author: string | string[]`
- [x] Ensure the bot only runs once per PR?

### DIFF
--- a/apps/smartbatt/ChangeLog
+++ b/apps/smartbatt/ChangeLog
@@ -1,2 +1,3 @@
 v0.01: New app!
 v0.02: Add dynamic intervals, depending on total cycles recorded
+v0.03: Return more data inside of .get();
--- a/apps/smartbatt/README.md
+++ b/apps/smartbatt/README.md
@@ -34,15 +34,17 @@ From any app, you can call `require("smartbatt")` and then one of the functions 
 * `require("smartbatt").get()` - Returns an object that contains:
 
   
-  * `hrsRemaining` - Hours remaining
+  * `hrsLeft` - Hours remaining
   * `avgDrainage` - Learned battery drainage per hour
-  * `totalCycles` - Total times the battery has been recorded and averaged
+  * `cycles` - Total times the battery has been recorded and averaged
+  * `timeLastRecorded` - The time when the module last updated the average, in unix seconds. (1757421612)
+  * `battLastRecorded` - The battery percent when the module last updated the average.
   * `totalHours` - Total hours recorded
   * `batt` - Current battery level
 
-    
 * `require("smartbatt").deleteData()` - Deletes all learned data. (Automatically re-learns)
 ## Creator
 - RKBoss6
 ## Contributors
-- RelapsingCertainly
+- sonicityV
+
--- a/apps/smartbatt/metadata.json
+++ b/apps/smartbatt/metadata.json
@@ -2,7 +2,7 @@
   "id": "smartbatt",
   "name": "Smart Battery Module",
   "shortName": "Smart Battery",
-  "version": "0.02",
+  "version": "0.03",
   "description": "Provides a `smartbatt` module that returns the battery in days, and learns from daily usage over time for accurate predictions.",
   "icon": "app.png",
   "type": "module",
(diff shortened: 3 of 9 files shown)


RKBoss6:
Nice! This looks really good!

RKBoss6:
I'm wondering, does the field support multiple authors?

bobrippling:
> I'm wondering, does the field support multiple authors?

Knew there was something I'd forgot - yes, thanks!

gfwilliams:
Thanks! This looks good, and glad it's working now - as you say in the header it'd be nice if the author was higher up the JSON though.

I see it says '54 files changed' but there are 600-odd apps? Or is it only searching recent history?

>  I'm wondering, does the field support multiple authors?

You could always separate them with commas?

RKBoss6:
@gfwilliams as for the 54 files changed, @bobrippling manually added them, as far as I can tell. App authors will have to add their username to their apps in the future...

bobrippling:
> I see it says '54 files changed' but there are 600-odd apps? Or is it only searching recent history?
 
> @gfwilliams as for the 54 files changed, @bobrippling manually added them, as far as I can tell. App authors will have to add their username to their apps in the future...

Yeah they were manually added, the remaining apps would have to be either updated also in future or we could track down the first commit to that app an assume that's our author

gfwilliams:
Maybe I'm being a bit of a luddite here, but would it make sense for us to not have a GitHub action, but just a script that we ran manually every so often that set the author? It'd be easier to get it to set the `author` tag in the right place and to handle >1 author - it wouldn't be automatic but we could then just have a single commit that added the author info. 

If the GitHub action doesn't run on PRs but only branches, presumably it means we couldn't rely on it to always update the author (if we get a PR from someone who doesn't have GitHub Actions?)

I'm not entirely sure whether having something that modifies the contents or PRs for us is going to come back and bite us later? Like if I did a PR from my PC, then changed the metadata around where the author tag went in and went to push again, wouldn't I get a merge error?

RKBoss6:
Hmm, we could make a script that just scans for the first commit author and sets the metadata author tag to that person, asa way to update all apps, but I like the idea of having a GitHub action for tagging the authors in every PR

bobrippling:
> Maybe I'm being a bit of a luddite here, but would it make sense for us to not have a GitHub action, but just a script that we ran manually every so often that set the author? It'd be easier to get it to set the `author` tag in the right place and to handle >1 author - it wouldn't be automatic but we could then just have a single commit that added the author info.

> Hmm, we could make a script that just scans for the first commit author and sets the metadata author tag to that person, asa way to update all apps, but I like the idea of having a GitHub action for tagging the authors in every PR

Ah I've perhaps confused things here - the setting of authers I've left as entirely manual for now, I just used a hacky script initially to populate a few apps I could be certain about. But I don't plan on automating that hacky script any time soon

bobrippling:
Should now just get a tag for gfw 🤞 

(for the changes I made to example apps)

bobrippling:
I'm happy with this now, will run a test PR after it's merged to try it out

gfwilliams:
Ok, great - let's merge then!

gfwilliams:
Thanks for your work on this!

bobrippling:
Thanks for the reviews & merge!

PR #4002, opened 2025-09-16
Update de_DE.json

schneidr:
Corrected and added German translations

This is work in progress, I'll add some more translations before finishing this pull request.


thyttan:
Thanks!

PR #4005, opened 2025-09-17
[BackLite] Add button release check

RKBoss6:
Just checks if the button is released, then cancel the check for long press, to avoid users pressing once, then once more and the screen turns on.

### DIFF
--- a/apps/backlite/boot.js
+++ b/apps/backlite/boot.js
@@ -29,7 +29,7 @@
         } else {
           Bangle.setLCDBrightness(0);
         }
-        clearTimeout(longPressTimer)
+        longPressTimer = null;
       }, longPressTime); // Slight delay to allow unlock to settle
     }
   });
--- a/apps/backlite/boot.js
+++ b/apps/backlite/boot.js
@@ -11,26 +11,36 @@
   //Set LCD to zero every reboot
   let s = require("Storage").readJSON("setting.json", 1) || {};
   s.brightness = 0;
+  if (!("lcdTimeout" in s)) s.lcdTimeout = 5; // fallback so  logic doesn't break
   require("Storage").writeJSON("setting.json", s);
   //remove large settings object from memory
-  delete s;
+  s=null;
   const longPressTime=400; //(ms)
-
+  var longPressTimer;
   Bangle.on('lock', function(isLocked) {
     Bangle.setLCDBrightness(0);
 
     if (!isLocked) {
       // Just unlocked — give a short delay and check if BTN1 is still pressed
-      setTimeout(() => {
+      longPressTimer=setTimeout(() => {
         if (digitalRead(BTN1)) {
           //set brightness until. locked.
           Bangle.setLCDBrightness(getSettings().brightness);
         } else {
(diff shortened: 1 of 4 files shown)


apps/backlite/boot.js
```diff
@@ -11,26 +11,36 @@
   //Set LCD to zero every reboot
   let s = require("Storage").readJSON("setting.json", 1) || {};
   s.brightness = 0;
+  if (!("lcdTimeout" in s)) s.lcdTimeout = 5; // fallback so  logic doesn't break
   require("Storage").writeJSON("setting.json", s);
   //remove large settings object from memory
-  delete s;
+  s=null;
```
bobrippling: We're probably better keeping `delete s;` here, then the variable is gone entirely instead of one existing and holding onto `null`

apps/backlite/boot.js
```diff
@@ -11,26 +11,36 @@
...
       // Just unlocked — give a short delay and check if BTN1 is still pressed
-      setTimeout(() => {
+      longPressTimer=setTimeout(() => {
         if (digitalRead(BTN1)) {
           //set brightness until. locked.
           Bangle.setLCDBrightness(getSettings().brightness);
         } else {
           Bangle.setLCDBrightness(0);
         }
+        clearTimeout(longPressTimer)
```
bobrippling: Here there is no timeout to clear, since it's already fired. You could set it to `null` (not deleting here, since we want the variable to remain for future sets)

```suggestion
        longPressTimer = null;
```

bobrippling (commented):
Nice find, thanks! Just a couple of little things

RKBoss6:
This doesn't work, and if you press, release, and then press quickly, it still checks for the button and lights up the screen, even with the checks in place. @bobrippling do you have any idea why that could be?

apps/backlite/boot.js
```diff
@@ -11,26 +11,36 @@
...
+  var longPressTimer;
   Bangle.on('lock', function(isLocked) {
     Bangle.setLCDBrightness(0);
 
     if (!isLocked) {
       // Just unlocked — give a short delay and check if BTN1 is still pressed
-      setTimeout(() => {
+      longPressTimer=setTimeout(() => {
         if (digitalRead(BTN1)) {
           //set brightness until. locked.
```
bobrippling, line 27: > This doesn't work, and if you press, release, and then press quickly, it still checks for the button and lights up the screen, even with the checks in place. @bobrippling do you have any idea why that could be?

I don't think my below suggestion is right unless there's a race between 'clearTimeout()` and the invocation of the timeout func.

If you add debug logs to both the timeout function (each arm of the `if`) and the `setWatch` callback, what gets printed?


```suggestion
      longPressTimer=setTimeout(() => {
        if (longPressTimer == null) {
          // cancelled, ignore
          return;
        }
        if (digitalRead(BTN1)) {
          //set brightness until. locked.
```

RKBoss6:
I think I'll just close this for now, it doesn't look like it'll work either way, and it's not that big of a deal if this isn't implemented.

PR #4008, opened 2025-09-21
[schedclock] New app to schedule changing clock faces

kidneyhex:
Link to app: https://kidneyhex.github.io/BangleApps/?id=schedclock

New app to schedule changing clock faces.

There are [clockswitch](https://banglejs.com/apps/?id=clockswitch) and [rndmclk](https://banglejs.com/apps/?id=rndmclk) already, but neither selects specific faces or specific times.


thyttan:
Thanks! Seems to work as intended when I try it.

It seems it's not fully compatible with Fastload Utils - but that's no requirement. I scheduled two clock face changes a minute apart. The first one executed but not the second one.

apps/schedclock/metadata.json
```diff
@@ -0,0 +1,23 @@
...
+    {"url":"screenshot2.png"}
+  ],
+  "tags": "tool",
+  "supports" : ["BANGLEJS2"],
+  "readme": "README.md",
+  "dependencies": {"scheduler":"type"},
+  "storage": [
+    {"name":"schedclock.app.js","url":"app.js"},
+    {"name":"schedclock.settings.js","url":"settings.js"},
+    {"name":"schedclock.lib.js","url":"lib.js"},
```
thyttan, line 19:
```suggestion
    {"name":"schedclock","url":"lib.js"},
```

I'm more used to this suggested pattern for loading libraries to the watch. But I see there are also apps that do it like you have it here (`ha`, `timerclk`, `tinyheads`, `xxlmessage`).

Note: If you want to change there would need to be accommodating changes at other places in the app as well.

@gfwilliams do we have a preference for how to name libraries on the watch?

kidneyhex:
I'll just tested after installing Fastload Utils and both clock changes worked, but from the code it looks like it might cache `setting.json` and not do a full reload in some situations. 

It looks like `widdst` has a "if fastload is installed" test to force settings updates... It just tests for `global._load`, which doesn't seem safe. 

Rough brainstorming: maybe fastload needs a way to tell it that settings changed, or force a slow-load?

At the very least, I'll add a note to the schedclock readme that it's a known issue. Maybe test a bit more to figure out if the "detect settings changes" flag helps.

I'll submit that after the response from Gordon about library names -- I like the shorter names myself, but when I searched other apps I found the app.lib.js ones, so went with that.

apps/schedclock/metadata.json
kidneyhex: Side question: Is there a recommended way to migrate old settings?

I just tested adding a `{"url": "migrate_v01_v02.js", "evaluate":true}` to the `metadata.json`, and as long as I put a `console.log("OK\r");` at the end of the migrate.js, it worked. But I can't find anything in docs/discussions about it.

gfwilliams: > do we have a preference for how to name libraries on the watch?

If they're being exported from the app, definitely just `schedclock` - but in this case I think it's just being used internally, so naming it `schedclock.lib.js` seems fine - maybe even preferable?

PR #4010, opened 2025-09-25
[SleepLog] Fix thresholds not changing functionality

RKBoss6:
As spotted by @ondras12345, the threshold variable name was different, and wouldn't compare. This fixes that issue.


thyttan:
Looks alright to me - is it tested now?

ondras12345:
I have uploaded version 0.22 to my watch and tested it overnight, and it still shows 0h of sleep, even though I can see in the log that my heart rate was below the light sleep threshold. This PR was definitely a step in the right direction, but it still does not seem to work correctly for me.
I will try to investigate this further once I get the IDE working on my PC. For whatever reason, Chromium on Ubuntu 24.04 fails to connect to the watch, and I don't feel like debugging this on an Android phone.
It should be noted that I only installed the app after the HRM update was merged, and I have never had it work correctly. It is well possible that I am doing something wrong.

thyttan:
@RKBoss6 Does it work for you?

RKBoss6:
It was working perfectly fine for me yesterday, but today I checked and it also shows 0 hrs 0 min, and has a blank period... I'm wondering what's happening now, as before it was comparing it to undefined, and was returning all the sleep states, but now it's not working. @ondras12345 You may want to try reinstalling the app with the older version (0.21) as that was working for me. See if that works for you. That version game me accurate sleep, so you may have done something wrong. As for this one, we'll have to look into what's happening. In the meantime, should we make a PR reversing the changes of this one so as to not affect others?

ondras12345:
> @ondras12345 You may want to try reinstalling the app with the older version (0.21) as that was working for me.

I did have that installed, and it was showing 0h. That's what lead me to discovering the bad comparison in the first place.


> It should be noted that I only installed the app after the HRM update was merged

In case it wasn't clear, what I meant to write here is that 0.21 is the first version of the app that I tried. (I only bought the watch recently.) That version already has HRM features. I never tried 0.20 or other old versions that only used the accelerometer.

thyttan:
I think I managed to revert commits so you can try version 0.20 -  which should be the last one before HRM detection went in. Here: https://thyttan.github.io/BangleApps/?c=&q=sleeplog

Please report back if that works better.

ondras12345:
Thanks thyttan. I will try it overnight, but probably tomorrow. Tonight, I would like to test a custom patch on top of 0.22.
I noticed that this code
https://github.com/espruino/BangleApps/blob/f04b1d44b5aa3031c0cbc945f41933ff9d795e9c/apps/sleeplog/boot.js#L165-L170
uses `data.heartRate`, but I cannot see where that value is supposed to come from. The `health` event provides a field called `bpm`, so I will try using that instead and see if that helps. Perhaps both sides of the comparison were undefined in 0.21.

Edit: I tried printing out the entire `data` object, and neither `data.heartRate` nor `data.hrm` seem to exist.

If the app used movement data only, that could possibly explain the behavior I am seeing. The default thresholds for movement seem way too low for me, and I didn't bother setting them, because I wanted to use the heart rate.

RKBoss6:
Yes, I found that out as well! It seems I was hooked into the wrong data. try out this one from my app loader. Also, I noticed that the movement data was being compared to the wrong thresholds...

Give [this](https://rkboss6.github.io/BangleApps/?id=sleeplog) a shot

ondras12345:
Last night, I was running RKBoss6's version 0.23, and I finally got a nonzero reading:
screenshot_20251002-054036.bmp

However, there seems to be a huge gap in the data. Perhaps I have corrupted the log file when previously messing with the app myself. I think I will try deleting everything and reinstalling 0.23.

thyttan:
OK - please report back again 👍🙏

RKBoss6:
I tested that out for a while, and it seems as though it's using accelerometer data again, even though it should have changed to use BPM. It is varied, and often unreliable and unpredictable. Will work on figuring out what's happening here, just wanted to give a status update.

RKBoss6:
@thyttan Would you mind trying out [this](https://rkboss6.github.io/BangleApps/?id=sleeplog) as well, and see if you can spot something wrong? It'll just help us figure out what is going wrong. Thanks!

ondras12345:
I don't think 0.23 is using the accelerometer for me. It correctly detects deep sleep, which it previously wouldn't do due to the default accelerometer threshold being too low. I will try further reducing it to confirm.
I am currently trying to fine-tune my HRM thresholds, as I haven't seen the app report light sleep yet.

One more difference from previous tests: ever since reinstalling 0.23, I have kept debug mode off, just in case that affects things. I will try turning it back on now.

RKBoss6:
Let me also try fine-tuning it, perhaps it's just not properly tuned for me...
Debug mode shouldn't affect it, but you're right, it's good to test it out.

ondras12345:
0.23 is definitely using HRM for me, and it seems to be working correctly. Regardless of whether debug mode is on or off.
I did not test movement/accelerometer mode (HRM sensor disabled) though.

thyttan:
It seems @RKBoss6's `sleeplog` v. 0.23 is better than 0.22 that's currently on the development app loader then? If so I'd ask @RKBoss6 to do a PR with it? :)

RKBoss6:
Just created a PR!

PR #4021, opened 2025-10-06
i added the running app (id=rungps)

kasperrey:


bobrippling:
Looks good, has it had much testing? @gfwilliams what are your thoughts on apps that replicate core apps? Do they need any signposts saying there's the "official" run{,plus} apps for users to look at too?

gfwilliams:
Thanks! A screenshot might be really handy actually - it makes it much more likely folks will try it.

In general, I think stuff like GPX recording could be much better handled by using the `Record` app (like Run does) rather than reimplementing it yet again (a lot of work went in to Recorder to try and reduce all the duplication that was happening so it's a shame if we're going backwards. Save with `exstats` to get all the stats). But I think this is fine as-is.

@bobrippling we also need to try and figure out how to stop the linter running on JS files that are for the web browser?

I think in general I'd prefer stuff that doesn't deliberately seem confusing compared to the built-in ones, but I think this is ok? I mean, we already have `Run` (the standard one), and `Run+` which is the one with extra screens and stuff. The 'sort by favourites/installs' will still likely put Run at the top?

bobrippling:
Sounds good - I think short of getting the linter to inspect `metadata.json`, we just have an ignore comment.

But otherwise ok, let's merge ~- @kasperrey can you add `/* eslint-disable */` to `calculations.js` and `gpx_maker.js` ?~

bobrippling:
Made the change, squashed a few dupe commits while I was at it, will merge when the build passes

bobrippling:
Actually - @kasperrey would you mind adding a screenshot? Don't worry if github says about overwriting any changes, just force your changes and I'll sort it :)

kasperrey:
Sorry for the late response, I added the screenshot and a reference to the runplus app.

bobrippling:
No worries, thanks for the change!

Issue #4012, opened 2025-09-28
[SleepLog] Trigger returning undefined for prevStatus


RKBoss6:
### Affected hardware version

Bangle 1, Bangle 2

### Your firmware version

2v27

### The bug

I'm trying to create an app that needs the ability to see any sleep status changes, and there is a built in way to do this in `sleeplog` using triggers. However, with that implemented, and logging for it set up, it always returns undefined for the `prevStatus` object, which is quite frustrating for me. In the source code for sleeplog, I can see that it returns undefined if the status is the same as the last one, however, for me, the status changes but still returns undefined. Here's the code I use to set the trigger in a boot.js file, any help is appreciated!

Trigger code: 
```javascript
// first ensure that the sleeplog trigger object is available (sleeplog is enabled)
  if (typeof (global.sleeplog || {}).trigger === "object") {
    // then add your parameters with the function to call as object into the trigger object
    sleeplog.trigger["sleepsummary"] = {
      onChange: true,   // false as default, if true call fn only on a status change
      from: 0,           // 0 as default, in ms, first time fn will be called
      to: 24*60*60*1000, // 24h as default, in ms, last time fn will be called
        // reference time to from & to is rounded to full minutes
      fn: function(data, thisTriggerEntry) { 
      
       //Run code here

   logNow("checked, prev status: "+data.previousStatus+", current status: "+data.status+", promptLastShownDay: "+require("sleepsummary").getSummaryData().promptLastShownDay);
      
      let today = new Date().getDay();
      if(require("sleepsummary").getSummaryData().promptLastShownDay!=today){
        if (data.status==2&&(data.previousStatus==3||data.previousStatus==4)) {
          var settings=require("sleepsummary").getSettings();
  
          //woke up
          if(settings.showMessage) showSummary();
          logNow("shown")
          require("sleepsummary").recordData();
        }
        
      }
        
      
      
      } // function to be executed
    };
  }
``` 

### Installed apps

[...]

PR #4001, opened 2025-09-16
Create Sleep Summary Module

RKBoss6:
This creates a module that provides a sleep score out of 100 based on how good of a sleep you got at night. It also provides an app to view stats, and a good morning message when you wake up. It averages total sleep you get, and wake up times, and that is factored into the sleep score.

More documentation in README.md

This is still under testing... feel free to test as well from [my app loader](https://rkboss6.github.io/BangleApps/?id=sleepsummary)

PR #4023, opened 2025-10-07
[SleepLog] Fix important bugs with HRM being undefined

RKBoss6:
Fixed bugs relating to HRM data being incorrectly pulled in and comparing with undefined.
This comes from the discussion at #4010, and has been tested by me and @ondras12345

### DIFF
--- a/apps/sleeplog/boot.js
+++ b/apps/sleeplog/boot.js
@@ -161,17 +161,13 @@ if (global.sleeplog.conf.enabled) {
       // add preliminary status depending on charging and movement thresholds
       // 1 = not worn, 2 = awake, 3 = light sleep, 4 = deep sleep
       if(data.bpm){
-          if (!Bangle.isCharging()) {
-              if (data.bpm <= global.sleeplog.conf.hrmDeepTh) data.status = 4;
-              else if (data.bpm <= global.sleeplog.conf.hrmLightTh) data.status = 3;
-              else data.status = 2;
-          } else data.status = 1;
-      } else {
-          if (!Bangle.isCharging()) {
-              if (data.movement <= global.sleeplog.conf.deepTh) data.status = 4;
-              else if (data.movement <= global.sleeplog.conf.lightTh) data.status = 3;
-              else data.status = 2;
-          } else data.status = 1;
+        data.status = Bangle.isCharging() ? 1 :
+          data.bpm <= global.sleeplog.conf.hrmDeepTh ? 4 :
+          data.bpm <= global.sleeplog.conf.hrmLightTh ? 3 : 2;
+      }else{
+        data.status = Bangle.isCharging() ? 1 :
+          data.movement <= global.sleeplog.conf.deepTh ? 4 :
+          data.movement <= global.sleeplog.conf.lightTh ? 3 : 2;
       }
 
       // check if changing to deep sleep from non sleeping


apps/sleeplog/boot.js
```diff
@@ -151,38 +151,34 @@ if (global.sleeplog.conf.enabled) {
     // define health listener function
     // - called by event listener: "this"-reference points to global
     health: function(data) {
+      print("Health data acquired");
```
thyttan: I wonder if we should comment this out?

RKBoss6: ah yes, probably!

apps/sleeplog/boot.js
```diff
@@ -151,38 +151,34 @@ if (global.sleeplog.conf.enabled) {
...
       // check if global variable accessable
       if (!global.sleeplog) return new Error("sleeplog: Can't process health event, global object missing!");
-
+
```
thyttan: Please remove the added spaces here and on line 160.

apps/sleeplog/boot.js
RKBoss6: Although, for now, it might make sense to keep it in, just for testing in the long run
We can always remove in the next version, but for now, I would like to have it in, just to keep confirming it is working as expected

thyttan: OK - let's go with that :)

apps/sleeplog/boot.js
```diff
@@ -151,39 +151,29 @@ if (global.sleeplog.conf.enabled) {
...
-          data.movement <= global.sleeplog.conf.deepTh ? 4 :
-          data.movement <= global.sleeplog.conf.lightTh ? 3 : 2;
+      if(data.bpm){
+          if (!Bangle.isCharging()) {
+              if (data.bpm <= global.sleeplog.conf.hrmDeepTh) data.status = 4;
+              else if (data.bpm <= global.sleeplog.conf.hrmLightTh) data.status = 3;
+              else data.status = 2;
+          } else data.status = 1;
+      } else {
+          if (!Bangle.isCharging()) {
```
thyttan: Please don't change from the previous code style when using `movement` data in the `else`-statement (https://github.com/espruino/BangleApps/wiki/App-Contribution#for-modifications-to-existing-appsmodules). The original author for that code probably did it like that for brevity's sake.

Let's just focus on the hrm fixes for this PR - doing more changes makes it a bit harder for me to review 🙏

apps/sleeplog/boot.js
```diff
@@ -303,7 +293,7 @@ if (global.sleeplog.conf.enabled) {
               timestamp: new Date(data.timestamp),
               status: data.status,
               consecutive: data.consecutive,
-              prevStatus: data.status === this.status ? undefined : this.status,
+              prevStatus: this.status,
```
thyttan: Is this change needed? I wonder if it could have some unintended side effect?

apps/sleeplog/boot.js
```diff
@@ -151,39 +151,29 @@ if (global.sleeplog.conf.enabled) {
     // define health listener function
     // - called by event listener: "this"-reference points to global
     health: function(data) {
+      print("Sleep Log - Health Data Acquired");
       // check if global variable accessable
       if (!global.sleeplog) return new Error("sleeplog: Can't process health event, global object missing!");
-
       // check if movement is available
       if (!data.movement) return;
```
thyttan: Does this interfere if using `hrm` data?

apps/sleeplog/ChangeLog
```diff
@@ -18,3 +18,4 @@
 0.20: Increase default sleep thresholds, tweak settings to allow higher ones to be chosen
 0.21: Use HRM data is polling is enabled, and add appropriate thresholds in settings. Change settings to feel more intuitive, rather than copied settings in different pages.
 0.22: Fix bug with HRM threshold not updating
+0.23: Fix important bug with HRM data being undefined, fix bug with movement thresholds being compared against the HRM thresholds.
```
ondras12345: > fix bug with movement thresholds being compared against the HRM thresholds

What is this about? I don't see a corresponding change in the diff.

bobrippling:
Not for this PR, but would you find it useful to have typescript check this code, to stop similar errors in future? I wouldn't mind making the change if so

RKBoss6:
@bobrippling That would be good! Just a question: how does typescript checking differ from the checks that are already done?

bobrippling:
> @bobrippling That would be good! Just a question: how does typescript checking differ from the checks that are already done?

It'd be able to find typos such as `object.propertyy` that we might miss, or catch nulls and so on

apps/sleeplog/ChangeLog
RKBoss6: That was changed in version 0.22, but wasn't on the changelog, and with all the versions being released, I thought it would be good to stay as a record of what happened.

RKBoss6: I can change that to be in 0.22, which will be more accurate anyway

apps/sleeplog/boot.js
RKBoss6: It shouldn't, because movement data is always there.

apps/sleeplog/boot.js
RKBoss6: As I nested the 2 branches within the overarching if statement, it became very unreadable, and would have been much worse if I had gone with the code style, so I think we should keep it this way for now. I'm sorry about that, I'll note it for next time!

RKBoss6:
@ondras12345 Any updates from testing for you? For me it seems as though only deep sleep is being compared in consecutive sleep, and even when the light sleep Hrm threshold is 100, it does not register. Light sleep is still registered, but doesn't look like it is accurate. When the deep sleep threshold is 100 it works as expected and shows everything as sleep, as it should.

ondras12345:
Sorry for the late reply.

> For me it seems as though only deep sleep is being compared in consecutive sleep

That definitely isn't the case for me. E.g. today, it shows consecutive sleeping 9h 50min, true sleeping 9h 50min, deep sleep 5h 10min, light sleep 4h 40min. My HRM thresholds are 60 for deep sleep and 64 for light sleep.

RKBoss6:
@ondras12345  can you try something out? Try setting the light sleep threshold to 100. Theoretically, it should report that you sleep all the time, but when I try it does not register. Changing deep sleep as stated is what works for me. Test that out...

ondras12345:
You're right, it registers as awake when I set light sleep threshold to 100. Will experiment with this some more and let you know what I find.

ondras12345:
I think this behavior is by design:
https://github.com/RKBoss6/BangleApps/blob/06fc36115835384bb4ccd2ad964398e63197a5d4/apps/sleeplog/boot.js#L230-L234

Light sleep is corrected to awake if it did not start with a deep sleep.

RKBoss6:
Hmm, is that accurate though? From what I have researched, a daily sleep starts with light sleep, and then alternates between deep sleep and light sleep again. I'm tagging @storm64, the creator, to see what the decision behind this was.

storm64:
As the programming is some time ago I can't tell you what the decision process was, but I tried to document all behaviour and settings in the readme: [https://banglejs.com/apps/?id=sleeplog&readme](https://banglejs.com/apps/?id=sleeplog&readme)

ondras12345:
I think it is working well enough. We should probably get this PR through so that users who update don't get a completely broken app. A change to the light sleep behavior can be added later if deemed necessary.

RKBoss6:
True. I'll work on fixing checks so we can hopefully get this merged in.

apps/sleeplog/boot.js
RKBoss6: It doesn't seem to break anything for me, instead fixing the implementation of hooks into the system so far for other developers. If anyone has a problem, we can address this there, as this doesn't directly break anything I can see.

RKBoss6:
This looks good to go right now!

thyttan:
Two review comments waiting for answers in the files view :)

RKBoss6:
Should be resolved now

RKBoss6:
This is good to go, so if we can merge this in, that would be great for users to fix the currently broken app :)

bobrippling:
@thyttan I'll let you merge if you're happy, being the main reviewer here :)

thyttan:
I'll get back to this saturday I hope.

apps/sleeplog/boot.js
thyttan: I believe this is for #4001? Take it out of this PR and introduce it on that other PR please. 

I'm not comfortable making that change willy-nilly - the app is somewhat complex and some functionality could depend on the way that line is written. You may not be affected but others might. Why would it be written that way otherwise I wonder? 

Ideally we search through the code to try and figure that out before changing it - and let's do that on #4001.

apps/sleeplog/boot.js
thyttan: I will open a PR to your fork to get this to where I'm ok with merging.

Please don't mark comment as resolved if it isn't obvious we are on the same page with how to move forward 🙏

apps/sleeplog/boot.js
thyttan: Ok, I won't force a change here on this PR - it's not what it set out to address anyway. But it feels wrong to have that `return` trigger if the user chose hrm mode and that was available but movement wasn't for some reason.

apps/sleeplog/boot.js
RKBoss6: Alright, sorry about that!

apps/sleeplog/settings.js
```diff
@@ -13,9 +13,9 @@
     maxAwake: 36E5, //  [ms] maximal awake time to count for consecutive sleep
     minConsec: 18E5, // [ms] minimal time to count for consecutive sleep
     deepTh: 150, //     threshold for deep sleep
-    lightTh: 300,//    threshold for light sleep
-    hrmLightTh: 74,//    threshold for light sleep
-    hrmDeepTh:60,//     threshold for deep sleep
+    lightTh: 300,
+    hrmLightTh: 74,
+    hrmDeepTh:60,//    threshold for light sleep
```
thyttan: Why the change to comments on these lines?

apps/sleeplog/boot.js
thyttan: Made a PR to your branch now.

apps/sleeplog/boot.js
RKBoss6: That makes sense, I can modify it to return only if both movement data and hrm data are not there

apps/sleeplog/boot.js
RKBoss6: Alright, I've taken it out, and we can see to fixing #4012  in another PR. Thanks!

apps/sleeplog/settings.js
RKBoss6: Huh, that's funny. Fixing this now.

apps/sleeplog/boot.js
RKBoss6: Just merged it in, thanks so much for the help!

thyttan: Thanks!

thyttan:
I'll merge this now - thanks!

One thing I'm thinking is, should the user be able to prefer movement data over hrm data? The code now defaults to heart rate and falls back to movement (correct me if I'm wrong). For some people movement data may be more reliable and it has been the standard for a long time.

In my mind, the heart rate option should have been introduced as an option and not the new default.

On the other hand, if a user find heart rate readings unreliable they probably keep it disabled generally? thus getting the movement data version anyway.

@RKBoss6 @ondras12345 @bobrippling, thoughts?

RKBoss6:
I think it's fine as is, the code only uses hrm data if it's available, but yes, it does default to it if present. We can make a pr for this, I think it's a good setting to have

bobrippling:
Cool, sounds like a reasonable approach to me

PR #4024, opened 2025-10-08
Fix accelrec metadata: Remove BANGLEJS support

### LINKED ISSUE #4022: [Acceleration Recorder] accelrec metadata incorrectly supports BANGLEJS
### Affected hardware version

Bangle 1

### Your firmware version

2v27

### The bug

Hi there,

I installed the "accelrec" app on my Bangle.js 1 watch. The app starts the countdown (3, 2, 1) but crashes immediately after, showing no recording or error message. 

I think this is because [Bangle.accelWr()] is not available on Bangle.js 1 (only on Bangle.js 2), causing the recordStart() function to fail. The metadata.json incorrectly lists "supports": ["BANGLEJS","BANGLEJS2"], but it should be "BANGLEJS2" only. Please update the metadata. 

Thank you

### Installed apps

_No response_

saeidsobhani:
Fixes #4022

The accelrec app uses `Bangle.accelWr()` which is only available on Bangle.js 2.
Removed "BANGLEJS" from the supports array in metadata.json.

### DIFF
--- a/apps/accelrec/metadata.json
+++ b/apps/accelrec/metadata.json
@@ -6,7 +6,7 @@
   "description": "This app puts the Bangle's accelerometer into 100Hz mode and reads 2 seconds worth of data after movement starts. The data can then be exported back to the PC.",
   "icon": "app.png",
   "tags": "",
-  "supports": ["BANGLEJS","BANGLEJS2"],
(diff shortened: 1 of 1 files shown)


bobrippling:
Thanks!

PR #4025, opened 2025-10-09
New app: hidmouse

bobrippling:
Tinkering with BLE and mouse reports - made this basic app based on the discussion at https://github.com/orgs/espruino/discussions/7856

Deployed at [my gh-pages](https://bobrippling.github.io/BangleApps/?q=hidmouse)

### DIFF
--- a/apps/hidmouse/README.md
+++ b/apps/hidmouse/README.md
@@ -1,3 +1,5 @@
 # Mouse
 
 Connect to a computer, drag the screen to move your mouse. Tap or press the BangleJS2's button to left-click.
+
+Note that HID must be enabled in settings for bluetooth devices to recognise the bangle as a mouse.


pavelmachek:
Not tested, but looks good to me and I'm looking forward to testing it :-).

pavelmachek:
Actually, there's already "presentor" app. After clicking, it can be used as a  mouse.

thyttan:
> there's already \"presentor\" app. After clicking, it can be used as a mouse.

Ah! Maybe we should add some more tags to that one. E.g. `mouse`, `keyboard`, `hid`.

bobrippling:
Added those tags to the presentor app and a few to `hidmouse`, and given it a testing too :)

thyttan:
I'd try this out if it weren't for me having problems pairing with both my laptop and phone.

bobrippling:
No worries, I've been testing more - happy to merge if that's ok with you? Unless you want a go @pavelmachek?

thyttan:
Just saw the icon doesn't work well with dark theme. Both on launcher and in the app itself. Other then that OK to go ahead 👍 (still not able to test more than that)

pavelmachek:
I'd suggest merging it.

thyttan:
Does the Settings->Bluetooth->HID setting have to be set or not? I think that's worth mentioning in the readme.

I managed to test it now and it worked both on my phone and on my laptop (I'm unsure of if I had the aforementioned HID setting enabled or not). Nice!

So in my mind we just fix the icon and add a note regarding that setting in the readme and then merge.

bobrippling:
Thanks - all sorted

thyttan:
Thanks! :)

PR #4026, opened 2025-10-10
new app: bmoface

LucienStorm:
A watchface inspired by some popular cartoon characters.


apps/bmoface/ChangeLog
```diff
@@ -0,0 +1,11 @@
+0.01: Initial release with BMO character
+0.02: Added Finn and Jake characters
+0.03: Added settings menu for character selection
+0.04: Added temperature unit toggle (C/F)
+0.05: Fixed settings menu crash, added charging status indicators
+0.06: Fixed "Invalid Settings!" error with proper settings file handling
+0.07: Added character randomizer feature with multiple intervals
+0.08: Fixed lock screen character variable error, separated lock screen logic
+0.09: Improved lock screen character-specific drawing and positioning
+1.0.0: Major release with all core features complete
```
thyttan, line 10: We tend to keep it zero-ver, so please just increment to `0.10` here etc.

thyttan: Since it's the first time adding it to espruino repo - If you want you can just introduce it as version `0.01`.

LucienStorm: Ok, sorry about that. I'll update the versions and changelog.

PR #4028, opened 2025-10-14
widminbate: lower battery threshold to appear 30%->10%

thyttan:
The widget now shows only when battery level is 10% or lower, instead of 30%. In an effort to make it EXTRA extra minimal.


bobrippling:
Nice!

PR #4031, opened 2025-10-16
[dailycolorclk] New App

RKBoss6:
A clock with a color bar at the bottom that changes color every day. Forked from slopeclock++

### DIFF
--- a/apps/dailycolorclk/app.js
+++ b/apps/dailycolorclk/app.js
@@ -145,11 +145,13 @@ let clockInfoDraw = (itm, info, options) => {
 let clockInfoItems = require("clock_info").load();
 let clockInfoMenu = require("clock_info").addInteractive(clockInfoItems, {  
   app:"slopeclockpp",x:98, y:38, w:70, h:50,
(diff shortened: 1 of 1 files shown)


thyttan:
I'm thinking - would it make more sense to have this as a settings of slopeclock++?

RKBoss6:
I would still keep this app, as all the offsets have been modified, and there is no slope anymore, increasing battery life. I personally like the look of this clock face as well, but it can definitely be an addon to the slope clock app as well in the future.

thyttan:
Yes that makes a lot of sense then. Thanks for elaborating!

RKBoss6:
This looks good to go :)

RKBoss6:
This is good to go, @bobrippling if you want to give this a check...

bobrippling:
Thanks! @thyttan happy for me to merge?

thyttan:
@bobrippling Yes - it's a GO for me :)

bobrippling:
Great!

PR #4033, opened 2025-10-17
New app: Offset-based world clock clockinfo

chuckwagoncomputing:
A World Clock clockinfo already exists, but it has hard-coded places.
This one allows you to configure places in the app loader, giving them a name and an offset from UTC.

This is my first BangleJS app, so please critique my work.

### DIFF
--- a/apps/offsetclockinfo/clkinfo.js
+++ b/apps/offsetclockinfo/clkinfo.js
@@ -1,7 +1,7 @@
 (function() {
   let config = require("Storage").readJSON("offsetclockinfo.settings.json") || {};
 
-  var clocks = {
+  let clocks = {
     name: "Offset Clocks",
     items: []
   };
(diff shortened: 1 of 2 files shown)


bobrippling:
Sounds good - sufficiently different to [the existing app](https://github.com/espruino/BangleApps/tree/master/apps/worldclkinfo) so I'm happy to merge. @thyttan what do you think?

chuckwagoncomputing:
I did consider making this as a change to the preexisting app, but I think there are probably people who just want to see the times in famous cities around the world and would prefer to use it rather than manually configuring things themselves.

RKBoss6:
Hmm, it's fine as it is, but it would integrate pretty well with the existing one - just add the custom html, and preload it with the cities there. From there, the user can add or remove more locations. That way we don't have 2 apps that do almost the exact same thing... Wdyt?

chuckwagoncomputing:
Sounds good to me. I'll do that and PR.

PR #4034, opened 2025-10-18
recorder: add multi-select to interface.html

bobrippling:
This changes the batch operations to a more fine-grained selection, and allows users to multi-select delete too

Deployed on my [gh-pages](https://bobrippling.github.io/BangleApps/?id=recorder)


bobrippling:
Tested a few times now, no problems - merging

PR #4036, opened 2025-10-18
new app: bmoface

LucienStorm:
New commit with app.png and hopefully all errors cleared.


LucienStorm:
Ok, I think I finally figured out the issue with this one. I'm going to re-submit and also include two others that I've been working on.

PR #4037, opened 2025-10-19
Add config to worldclkinfo

chuckwagoncomputing:
Re: #4033
@RKBoss6
This should be feature-complete compared to the original.

### DIFF
--- a/apps/worldclkinfo/README.md
+++ b/apps/worldclkinfo/README.md
@@ -37,7 +37,11 @@ With Simple Mode disabled, there are 5 available display modes:
 - `London 5:45 Mon`
 
 
-## Creators
+## Creator
 
 [RKBoss6](https://github.com/RKBoss6)
-[ChuckWagon Computing](https://github.com/chuckwagoncomputing)
+
+## Contributors
+
+- [ChuckWagon Computing](https://github.com/chuckwagoncomputing)
+


bobrippling:
LGTM, @RKBoss6 would you fancy a bit of testing?

RKBoss6:
Of course! @chuckwagoncomputing Is it set up on your app loader? That way I can test the customizer as well. Thanks!

chuckwagoncomputing:
Yes it's currently up at https://chuckwagoncomputing.github.io/BangleApps/

It might not consistently stay up because I'm messing around with some other apps, but it is right now.

RKBoss6:
I gave it some testing, and the uploads all look good, but when you tap on the ClockInfo where it used to show the city name, now it shows something like `6pm sun` and only after tapping many times will it inconsistently show the city name.

Edit: It looks like there are many different modes it can render the string as, what do we need all of those for? Should we only keep the name and time?

apps/worldclkinfo/custom.html
```diff
@@ -0,0 +1,144 @@
...
+      document.getElementById("addrow").addEventListener("click", function() {
+        let tbl = document.getElementById("rows");
+        let tmp = document.getElementById("row-template");
+        let row = tmp.content.cloneNode(true);
+        tbl.append(row);
+      });
+
+      document.getElementById("delrow").addEventListener("click", function() {
+        let tbl = document.getElementById("rows");
+        tbl.deleteRow(tbl.rows.length -1);
```
RKBoss6: If you delete all the rows, it also deletes the labels, so add an if statement here to keep labels from being deleted.
```suggestion
        if(tbl.rows.length>1) tbl.deleteRow(tbl.rows.length -1);
```

RKBoss6:
Also, just bump the version in metadata up, and add an entry in changelog :)

apps/worldclkinfo/custom.html
```diff
@@ -0,0 +1,144 @@
...
+      <tr>
+        <th>Name</th>
+        <th>Short Name</th>
+        <th>UTC Offset</th>
+      </tr>
+    </table>
+
+    <button id="addrow" class="btn">+</button>
+    <button id="delrow" class="btn">-</button>
+    <br>
```
RKBoss6: I think we can add some more padding between the delete row and upload buttons, as it is quite close together, and resulted in me accidentally uploading it while I was testing...

chuckwagoncomputing:
Oh yeah, I forgot about the version number.

The different modes allow flexibilty for different personal prefernces and use cases. I like to see the day of the week.
Is there any downside to having so many modes?

chuckwagoncomputing:
I found and fixed a bug: The default settings values were different between clkinfo.js and settings.js, which resulted in "shorten cities" being enabled by default, but showing as not enabled in settings. Toggling in settings got things back in sync, but this should be fixed now.

RKBoss6:
>is there any downside to having so many modes?

For me, I would say most of the modes look the same, and having direct access to the city name is much more efficient for me. Perhaps we could make it a setting, for only city name and time, or the full 5 modes?

RKBoss6:
Also, another bug I found was if you have a clock with more than one clock info and both are set to the same city (eg. London), then both are updated when you tap on one, rather than only that one being rendered with the next mode

chuckwagoncomputing:
> > is there any downside to having so many modes?
> 
> For me, I would say most of the modes look the same, and having direct access to the city name is much more efficient for me. Perhaps we could make it a setting, for only city name and time, or the full 5 modes?

Hmmm, I think this could be done...

> Also, another bug I found was if you have a clock with more than one clock info and both are set to the same city (eg. London), then both are updated when you tap on one, rather than only that one being rendered with the next mode

This I already knew, it's a consequence of the mode being saved for persistence. I didn't see it as a bug because I couldn't think of a use case for having two of the same city on the same screen.

chuckwagoncomputing:
Simple mode now implemented. I don't know how to fix the other thing without removing persistence. 
I also added myself as a creator in the README :)

RKBoss6:
Hmm, in the olde version, it also had that error, and I never came across it, so that should actually be fine. I'll get to testing this out later today :)

apps/worldclkinfo/README.md
```diff
@@ -1,28 +1,43 @@
...
+- `London`
+- `5:45 Mon`
+- `London 5:45`
+- `London 5:45 Mon`
+
+
+## Creators
+
 [RKBoss6](https://github.com/RKBoss6)
+[ChuckWagon Computing](https://github.com/chuckwagoncomputing)
```
RKBoss6: As of right now, this is formatted so it looks like a continuous string: `RKBoss6 ChuckWagon Computing`
Bullet points will mitigate the issue.

Also, you might want to make a section called `contributors` so more people who change this app in the future can pop themselves in...

```suggestion
## Creator
[RKBoss6](https://github.com/RKBoss6)
## Contributors
- [ChuckWagon Computing](https://github.com/chuckwagoncomputing)
```
Otherwise, everything looks good! Nice job :)

RKBoss6:
Alright, tested and it all looks good! The only issue I see that's not in the older one is when you tapped to toggle between city name and time, if you swiped down, to change cities, that would stay persistent, showing the clock or city throughout changes. That goes away now, and some show city names, some show time...

chuckwagoncomputing:
That is intended. While I can't imagine a use case for having two of the same city in different modes, I can imagine a use case for having different cities in different modes, so I made the persistence city-specific.

RKBoss6:
Hmm, I feel like it would make selection process less efficient, as you need to swipe, tap a few times, swipe again, and repeat, just to get to the city you want. Could we bundle the persistence into having simple mode off?

chuckwagoncomputing:
I'm not sure if I completely understand what you're asking for, but give it a try now and let me know what you think.

RKBoss6:
That's much better! I'd say once you take a look at the reviews, this is good to go!!

RKBoss6:
Sorry if you couldn't understand what I was trying to day earlier, I think you got the gist of it, but effectively I preferred having the persistence global, so tapping once sets all of the clockinfos to city names, and tapping again sets all back to time regardless of scrolling through. Tying it into the simple mode is a good change, as users can now choose whether they like the old interface, or a newer one! Nice job, thank you!

bobrippling:
Sounds good - I'll merge unless there's anything outstanding?

chuckwagoncomputing:
OK to merge as far as I'm aware.

RKBoss6:
Can we add some padding and table bug fixes as per the reviews, for a more user friendly experience?

bobrippling:
> Can we add some padding and table bug fixes as per the reviews, for a more user friendly experience?

Ah, I suspect you might have your reviews in draft - if you go to [files changed](https://github.com/espruino/BangleApps/pull/4037/files), is there your review waiting to be submitted?

RKBoss6:
Ah, you were right! Thanks for the heads up, just published the review!

bobrippling:
Great, thanks you two :)

RKBoss6:
Thanks @chuckwagoncomputing! You've made this app much more customizable, nice job!

PR #4038, opened 2025-10-19
Check filename length in CI

chuckwagoncomputing:
The filename length limit tripped me up, so I thought it might be a good addition to the CI checks.
There was also one app with a too-long filename declared in metadata.json, but it wasn't actually using that name.


bobrippling:
Great!

PR #4042, opened 2025-10-21
[Clock Info] Add option for haptics when changing infos

RKBoss6:
I love haptics as a way to feel the change on an otherwise digital watch, so I made a setting to enable/disable this when changing clockInfos as a means of confirmation to those who have trouble seeing certain clockInfos on clocks.

### DIFF
--- a/apps/clock_info/ChangeLog
+++ b/apps/clock_info/ChangeLog
@@ -20,4 +20,4 @@
 0.19: Fix Altitude ClockInfo after BLE added
       Tapping Altitude now updates the reading
 0.20: Altitude ClockInfo now uses the distance units set in locale.
-0.21: Add option for haptics when changing infos (set to false by default)
+0.21: Add haptics when changing infos and a setting to toggle it on or off


bobrippling:
Great idea, LGTM - tagging @gfwilliams with clock_info being a core app

gfwilliams:
Thanks - looks good!

We do already buzz when tapping on a clockinfo (if it has something to run): https://github.com/espruino/BangleApps/pull/4042/files#diff-1239510a1796032e18d0af3c42256cd1ac83caedbf641bbe315baaad3350cebcR361

Should this be included by the Haptics option I wonder?

I guess my feeling is maybe we do include it, and then turn haptics on by default? This seems like quite a nice improvement to me and  I doubt too many people would complain esp as they can turn it off

RKBoss6:
That sounds good! Will do!

bobrippling:
LGTM, will merge soon (to give you a chance to check over if you wanted @gfwilliams)

gfwilliams:
Looks great to me, thanks! Sorry for not spotting your changes earlier!

PR #4043, opened 2025-10-21
New App: Air Quality Clockinfo

chuckwagoncomputing:
This clockinfo fetches air quality index and temperature from the IQAir API.
I put "Not affiliated with IQAir" in the readme, but I don't know if this is sufficient. I suppose that "IQAir" is probably a registered trademark, but the [IQAir Copyright and Trademark page](https://www.iqair.com/legal/trademark-list) just shows their privacy policy for some reason.

https://chuckwagoncomputing.github.io/BangleApps/

### DIFF
--- a/apps/airqualityci/clkinfo.js
+++ b/apps/airqualityci/clkinfo.js
@@ -60,16 +60,16 @@
           let aqi = data[config.rows[id].url]
           let txt = "";
 					let aqius = (aqi && typeof(aqi.aqius) == "number") ? aqi.aqius : "...";
-					let temp = (aqi && typeof(aqi.temp) == "number") ? aqi.temp : "...";
+					let temp = (aqi && typeof(aqi.temp) == "number") ? require("locale").temp(aqi.temp) : "...";
           switch ( config.rows[id].mode) {
             case 2:
-            txt = aqius + " " + temp + "°";
+            txt = aqius + " " + temp;
               break;
             case 3:
               txt = row.name + " " + aqius;
               break;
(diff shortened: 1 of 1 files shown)


RKBoss6:
Nice! For iOS, I was thinking of adding AQI to the weather data being pulled in, so later in the year, we can update this to use that data if available! Looks nice!

bobrippling:
Yeah looks good - @gfwilliams I guess the trademark mention is ok?

gfwilliams:
Looks great, thanks! I don't see a problem with using the IQAir name - it's fair use, you're not using it to sell anything, just to describe what it does.

chuckwagoncomputing:
I just realized there's an issue with this.
There's only one `aqiTime` slot, so if there are multiple instances, and one of them fails to fetch for some reason, it will be blocked from fetching for an hour.

bobrippling:
Cool!

bobrippling:
Was about to merge but I notice there's some new changes - @chuckwagoncomputing do you want me to let you test those for a little while first?

chuckwagoncomputing:
Let me test this one a little more. A few times I've looked at my watch and it's showing the AQI but not the temperature, which is a bit puzzling. I'm not quite sure how to debug that. I guess I'll create a version that detects that scenario and writes some debug info to a file.

chuckwagoncomputing:
I've been using this for a while now, and it seems solid.

bobrippling:
Cool, thanks for the app!

PR #4045, opened 2025-10-23
Add clockinfos to mtnclock

chuckwagoncomputing:
I'm creating this draft PR to get feedback before I polish things that I shouldn't do in the first place.

The first two commits are fixing the indentation and converting to posix line endings. If you would prefer not to change these things, let me know and I can rebase without them.

This adds up to 3 clockinfos below the date/weather text, shifting the time and date/weather text upwards.

I also moved one of the mountains a bit for background contrast when all 3 clockinfos are present.

Because the clockinfos are not very large, I created a second view for editing them to allow more room for swipes. This view is accessed by tapping, or by long-tapping on the Bangle.js 2 to reduce accidental activation. I guess this view could have been put in the settings screen, but I did it this way for two reasons which I am not certain about and am open to feedback.
1. Convenience - people might not change their clockinfos often, but it sure is handy for initial setup.
2. I'm not sure if a custom UI as the settings for an app is acceptable, as all the apps that I use just have menus as their settings.

App manager: https://chuckwagoncomputing.github.io/BangleApps/

TODO/Known issues:
- Should probably update all the screenshots
- A few times the edit screen has shown with the wrong clockinfos. I don't know how to reproduce this, and I've made quite a few changes since the last time I saw it, so it might be fixed.
- `clockinfosMain` seems like pretty dumb data structure design, but I haven't decided on a solution yet. I don't want to just tag `data.rows` because I don't want to persist the shown state. How do objects compare to arrays in Espruino in terms of performance?

### DIFF
--- a/apps/mtnclock/app.js
+++ b/apps/mtnclock/app.js
@@ -5,8 +5,6 @@ if (! Array.isArray(data.rows)) {
   require("Storage").writeJSON("mtnclock.json", data);
 }
 
-let showingSettings = false;
-let clockinfosSettings = [];
 let clockinfosMain = {};
 let clockinfos = require("clock_info").load();
 
@@ -236,7 +234,6 @@ function draw(color) {
 }
 
 function setWeather() {
-  if (showingSettings) return;
   var a = {};
   //clear day/night is default weather
   if ((data.code >= 800 && data.code <=802) || data.code == undefined) {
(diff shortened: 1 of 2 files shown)


chuckwagoncomputing:
> * `clockinfosMain` seems like pretty dumb data structure design, but I haven't decided on a solution yet. I don't want to just tag `data.rows` because I don't want to persist the shown state. How do objects compare to arrays in Espruino in terms of performance?

I did some benchmarking:
```
var foo = [[],[],[false, false,true]];
var bar = [[true]];
var baz = {2: {2: true}}
let t = Date.now(); for (let i = 0; i < 1000; i++) { let s = foo[2][2] }; console.log(Date.now() - t);
t = Date.now(); for (let i = 0; i < 1000; i++) { let s = bar[0][0] }; console.log(Date.now() - t);
t = Date.now(); for (let i = 0; i < 1000; i++) { let s = baz[2][2] }; console.log(Date.now() - t);
```
`foo` takes 11 bytes of memory and 511 ms to complete the loop
`bar` and `baz` take 5 bytes and 496 ms to complete the loop

So objects are the winner; change incoming

bobrippling:
> The first two commits are fixing the indentation and converting to posix line endings. If you would prefer not to change these things, let me know and I can rebase without them.

Sounds good, thanks for splitting them up like that :)
 
> Because the clockinfos are not very large, I created a second view for editing them to allow more room for swipes [...] I guess this view could have been put in the settings screen, but I did it this way for two reasons which I am not certain about and am open to feedback.

Yes - I think having it in settings would be better, as people are used to navigating to there to configure their clocks (etc), but I don't mind strongly so I'll let you decide, given you've been using the app for a while now. Another point is that the setting code loads every time the clock does, whereas if it were in settings, your clock would be faster to startup.


> A few times the edit screen has shown with the wrong clockinfos. I don't know how to reproduce this, and I've made quite a few changes since the last time I saw it, so it might be fixed.

I suspect it is fixed, I can't find any problems with that kind of stuff in the code

chuckwagoncomputing:
OK, I've moved the clockinfo config to settings. There's an odd bug though.
Occasionally when tapping the back button, it acts as though I tapped on the messages widget, which is hidden.

bobrippling:
Weird - if you edit the messages widget on your device (via the [IDE](https://www.espruino.com/ide/)), and change this code to print some logs about those dimensions it's looking at, what do you see?

https://github.com/espruino/BangleApps/blob/b9141d8986aea70836c7a9c6a71a70bc836fae61/apps/widmessages/widget.js#L50-L54

RKBoss6:
Hey - just wondering, have you considered making the existing weather display at the bottom a clockInfo? To me, it looks slightly crowded with all the extra data, and there's already a weather clockInfo if users still want to see the weather data. To me at least, that seems a more elegant solution for all users, leveraging the already existing space at the bottom. Wdyt?

thyttan:
Should we do like with "Pebble ++ Clock" and "Anton Clock Plus" et al, making this a new more feature rich version of the app? I think that's probably a good idea given this app is also compatible with the RAM limited Bangle.js 1.

chuckwagoncomputing:
> Hey - just wondering, have you considered making the existing weather display at the bottom a clockInfo? To me, it looks slightly crowded with all the extra data, and there's already a weather clockInfo if users still want to see the weather data. To me at least, that seems a more elegant solution for all users, leveraging the already existing space at the bottom. Wdyt?

It crossed my mind, but a good way of designing the configuration UI didn't occur to me, so I forgot about it. :)  
The difficulty (as I see it) is that we can't just assume that the first configured clockinfo should go after the date because some people might not want a clockinfo after the date (most likely because there's not much room there).

> Should we do like with "Pebble ++ Clock" and "Anton Clock Plus" et al, making this a new more feature rich version of the app? I think that's probably a good idea given this app is also compatible with the RAM limited Bangle.js 1.

This is what I did initially, but then I decided to see if the mtnclock maintainer wanted the changes. I'll gladly make it a separate app if that's preferable.

thyttan:
I think since `clock_info` module doesn't support Bangle.js 1 it really is the best idea to fork it. Or do like Desktop Launcher `dtlaunch` and have separate js code uploaded depending on the HW version.

PR #4049, opened 2025-10-30
recorder: load full map and add author info

bobrippling:
We now:
- No longer limit map drawing to tracks whose initial data contained a latitude (`if (trackData.Latitude) { ... }`)
- Return after `resolve`ing errors with track infos (this should be `reject` but I don't want to regress the slightly intertwined promise-heavy code)
- Have a slightly faster longitude/latitude lookup for previews
- Note BangleJS as the GPX author

I've also added a note about how we would specify the track type (run, ride, swim, etc) but not implemented this

### DIFF
--- a/apps/recorder/interface.html
+++ b/apps/recorder/interface.html
@@ -464,7 +464,7 @@ function trackLineToObject(headers, l) {
   return o;
 }
 
-async function downloadTrack(filename) {
+async function downloadTrack(filename) { // filename: string
   function parse(data) {
     const lines = data.trim().split("\n"), headers = lines.shift().split(",");
     return lines.map(l=>trackLineToObject(headers, l)).filter(t => t.Time);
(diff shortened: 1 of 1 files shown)


bobrippling:
@gfwilliams are you happy with GPX tracks exported from bangles being tagged as coming from here?

https://github.com/espruino/BangleApps/blob/6b1d0e1ff70b54cf4fe1009e92396524f6d2bf13/apps/recorder/interface.html#L418-L423

gfwilliams:
I think that's fine - it should probably say `Bangle.js` instead though.

The link could actually reference https://banglejs.com/apps/?id=recorder - it might be helpful to differentiate that as the actual source?

bobrippling:
Sounds good - [updated](https://github.com/espruino/BangleApps/pull/4049/commits/8ade24a2f16b4003cda7c40b684c795cc4f830c7). Will give it a little test

bobrippling:
Working fine, will merge shortly

PR #4050, opened 2025-10-31
[Smart Battery Widget] show alert when below 5%

RKBoss6:
This modifies the widget to display a '!' if the battery is 5% or less


bobrippling:
Thanks!

PR #4051, opened 2025-10-31
feat: add system Quiet/Silent mode integration to prevent buzzing

EmVee381:
I really like my grandfather clock on my wrist :) but I’ve been missing a feature that would prevent it from constantly vibrating while I’m asleep.

I added an option in the settings that, when set to true, follows the system’s silent mode.


bobrippling:
Nice!

PR #4054, opened 2025-11-03
Fix/grandfatherclock silent 

EmVee381:
This update moves quiet mode detection into the chime function. The previous implementation was based on an incomplete understanding of the widget’s behavior.

I also added the missing Changelog and meta updates that were omitted in the previous commit.


bobrippling:
Nice fix - I'll merge unless you've anything outstanding with the PR being in draft?

EmVee381:
> Nice fix - I'll merge unless you've anything outstanding with the PR being in draft?

There’s one more bug that I fixed this morning. I’m going to sleep now, and it should start running in the morning — that’ll serve as the test. I’ll make a new pull request; the fix is already in my branch, but I’ll squash it into another one so I don’t drag any mess into your repository, and then I’ll create a new pull from that.

PR #4056, opened 2025-11-06
fix: silent mode detection

EmVee381:
This update moves quiet mode detection into the chime function. The previous implementation was based on an incomplete understanding of the widget’s behavior.

I also added the missing Changelog and meta updates that were omitted in the previous commit.


bobrippling:
Great, thanks! I should've spotted that on the past review 🤦

PR #4059, opened 2025-11-11
runplus: show satelite strength bar (+setting)

bobrippling:
This adds an option to show the satelite strength as a bar, to allow people to gauge how accurate the pace (and related stats) are, while running.

- [ ] Test on a run

### DIFF
--- a/apps/runplus/ChangeLog
+++ b/apps/runplus/ChangeLog
@@ -33,3 +33,4 @@ Write to correct settings file, fixing settings not working.
 	track screen state
 0.30: Change the "time" stat to show active time (duration) rather than
 	elapsed time (fix #3802), and sync with recorder on load
+0.31: Add setting to show GPS strength bar on the left
--- a/apps/runplus/README.md
+++ b/apps/runplus/README.md
@@ -4,8 +4,6 @@ Displays distance, time, steps, cadence, pace and heart rate for runners. Based 
 It requires the input of your minimum and maximum heart rate in the settings for the app to work. You can come back to the initial run screen anytime by swiping left.
 To use it, start the app and press the middle button so that the red STOP in the bottom right turns to a green `RUN`.
 
-To focus on a single stat, tap on the stat and it will take up the full screen. Tap again to return to the main screen.
-
 ## Display 1st screen
 
 * `DIST` - the distance travelled based on the GPS (if you have a GPS lock).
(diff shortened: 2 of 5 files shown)


bobrippling:
@gfwilliams for some reason the main boxes are scooted over to the right (I would expect left), when I add a column to their right:

image

Compared to the setting being off, so no column added on the left:
image


The code that adds the bar is:

https://github.com/espruino/BangleApps/blob/4473072b0d89d4073ad2fcd93e15f4ed0cfdf5f0/apps/runplus/app.js#L226-L241

Is this something odd in [layout](https://github.com/espruino/BangleApps/blob/master/modules/Layout.js)?

gfwilliams:
It does seem to be something strange in Layout. I'm looking into it, but right now it seems to be caused by the button label that gets added. If you remove the `btns` argument it lays out properly.

gfwilliams:
Ok, I've just fixed it, but because we're building Layout in now, you'll need a whole new firmware for this to work :(

Short of hard-coding the widths of the columns I don't think there's a great solution for this.

But what about some other options:

* We already have 8 pixels over the right-hand side where the button's text is. Could you just draw in there? As long as you redraw after `layout.render();` it'd be ok
* Or what about adding a custom renderer for the `GPS` area? Then you could have your GPS accuracy bar in there, since we already change the background colour based on whether we have a fix or not we might as well change it to the colours you've used, and maybe add the percent bar in there in the background?
* Or you could add it as a widget in the widget bar (you can easily add one that only shows in runplus)?
* ...or what about a widget that can be installed and would display in any apps where GPS is enabled?

bobrippling:
Thanks for the fix!

> But what about some other options:
> 
> - We already have 8 pixels over the right-hand side where the button's text is. Could you just draw in there? As long as you redraw after `layout.render();` it'd be ok
> 
> - Or what about adding a custom renderer for the `GPS` area? Then you could have your GPS accuracy bar in there, since we already change the background colour based on whether we have a fix or not we might as well change it to the colours you've used, and maybe add the percent bar in there in the background?
> 
> - Or you could add it as a widget in the widget bar (you can easily add one that only shows in runplus)?
> 
> - ...or what about a widget that can be installed and would display in any apps where GPS is enabled?

Sure - I'll try those out and see how they look 🎨

pavelmachek:
In "sky spy", I was playing with different ideas to determine if GPS signal is reliable. Best way probably is to compare GPS and barometer heights, and if that difference is too big or changes rapidly, GPS is not reliable.

PR #4060, opened 2025-11-11
tally: fix duplicate menu entries

bobrippling:
e.g. if the user has the same tally type and time, we only show one

- [x] test


bobrippling:
Tested, works well - merging

PR #4063, opened 2025-11-16
recorder: show maps for tracks with late-GPS

bobrippling:
Previously if we didn't have GPS in the small snapshot we fetch:

https://github.com/espruino/BangleApps/blob/0bfa101e72ca70b14361337aaa0e3ee8a1c7df86/apps/recorder/interface.html#L720-L724

then we wouldn't create the map element:

https://github.com/espruino/BangleApps/blob/0bfa101e72ca70b14361337aaa0e3ee8a1c7df86/apps/recorder/interface.html#L834-L838

We now create the map element and populate it with the track coords, or if no coords are found in the entire track, we hide it.

- Best to look at the [diff ignoring the indent changes](https://github.com/espruino/BangleApps/pull/4063/files?diff=split&w=1).
- Tested with:
  - A normal GPS file
  - A late-GPS file
  - An empty GPS file
- Deployed at my [gh-pages](https://bobrippling.github.io/BangleApps/?id=recorder)

### DIFF
--- a/apps/recorder/interface.html
+++ b/apps/recorder/interface.html
@@ -829,17 +829,14 @@ return {headers:headers,l:data};
     trackContainer.innerHTML = '<div class="track-loading">Loading track data...</div>';
     const track = trackList[trackIndex];
     const trackData = trackLineToObject(track.info.headers, track.info.l);
-    let trackHtml = `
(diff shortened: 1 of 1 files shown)


bobrippling:
No problems with testing - merging

PR #4066, opened 2025-11-24
Recorder: Use average of HR readings

ruvilonix:
Currently, Recorder saves the last sent HR reading when saving to CSV for each recording interval. This updates the app to keep a running average of all HR readings during each interval (only readings with confidence > 80), and saves this instead. If no readings pass the confidence threshold, the reading with the highest confidence is saved.

See discussion: https://github.com/orgs/espruino/discussions/7888


gfwilliams:
Looks great - thank you!

PR #4073, opened 2025-11-28
[messages][messagegui] - fix music info handling

kidneyhex:
_**Closing this: The fix requires design choices, so I'll just log an issue and let someone else decide.**_ :P

### DIFF
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -67,39 +67,51 @@ if (Bangle.MESSAGES) {
   delete Bangle.MESSAGES;
 }
 
-// Ensure any music message is always kept at the front of the list
-function moveMusicToFront() {
-  for (var i=0;i<MESSAGES.length;i++){
-    if (MESSAGES[i] && MESSAGES[i].id=="music"){
-      if (i!==0) MESSAGES.unshift(MESSAGES.splice(i,1)[0]);
-      return;
-    }
+// Update or add music message to MESSAGES array at front
+function updateMusicMessage(musicMsg) {
+  var existingMusicIdx = MESSAGES.findIndex(m => m.id === "music");
+  if (existingMusicIdx >= 0) {
+    MESSAGES[existingMusicIdx] = musicMsg;
+    if (existingMusicIdx !== 0) MESSAGES.unshift(MESSAGES.splice(existingMusicIdx, 1)[0]);
+  } else {
+    MESSAGES.unshift(musicMsg);
   }
 }
-moveMusicToFront();
+
+// Load music message from storage if it exists
+var musicMsg = require("messages").getMusic();
+if (musicMsg && (musicMsg.track || musicMsg.artist)) {
+  updateMusicMessage(musicMsg);
+}
 
 var onMessagesModified = function(type,msg) {
   if (msg.handled) return;
   msg.handled = true;
-  require("messages").apply(msg, MESSAGES);
-  // Keep music message at the front so it remains the first item
-  moveMusicToFront();
+  
+  if (msg.id === "music") {
+    // For music messages, get the complete state from messages module
+    updateMusicMessage(require("messages").getMusic());
+    var musicMsg = MESSAGES[0]; // music is now at front
+    
+    if (musicMsg.state && musicMsg.state=="play") {
+      openMusic = true;
+    } else if (musicMsg.state && musicMsg.state!="play") {
+      openMusic = false; // no longer playing music to go back to
+    }
+    if ((active!=undefined) && (active!="list") && (active!="music")) return; // don't open music over other screens (but do if we're in the main menu)
+  } else {
+    require("messages").apply(msg, MESSAGES);
+    // Move music back to front since apply() may have pushed it down
+    var musicIdx = MESSAGES.findIndex(m => m.id === "music");
+    if (musicIdx > 0) {
+      MESSAGES.unshift(MESSAGES.splice(musicIdx, 1)[0]);
+    }
+  }
+  
   // TODO: if new, show this new one
   if (msg && msg.id!=="music" && msg.id!=="nav" && msg.new &&
       !((require('Storage').readJSON('setting.json', 1) || {}).quiet)) {
     require("messages").buzz(msg.src);
-  }
-  if (msg && msg.id=="music") {
-    // Track when music actually played so we can expire old music messages
-    if (msg.state && msg.state=="play") {
-      // Update the stored music message (if present) with a last-played timestamp
-      var mm = MESSAGES.find(m=>m && m.id=="music");
-      if (mm) mm._lastPlayed = Date.now();
-      openMusic = true;
-    } else if (msg.state && msg.state!="play") {
-      openMusic = false; // no longer playing music to go back to
-    }
-    if ((active!=undefined) && (active!="list") && (active!="music")) return; // don't open music over other screens (but do if we're in the main menu)
   }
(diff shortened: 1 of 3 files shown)


bobrippling:
I'll hold off reviewing / testing until this is out of draft, but with it being a core app I'll tag @gfwilliams in the meanwhile

kidneyhex:
The current code is working, but I ran into an app where GadgetBridge isn't sending musicinfo even on play/pause (libro.fm), so the music entry/controls aren't showing up.

If we could just request the info/state, it would simplify things and reduce the need for writing to flash. (I think?)

I see someone else proposed this too (https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/4886), and it wasn't outright rejected, so I might do that this weekend and see if it's accepted, then revise this to use it. 

--- I don't have access to an apple device though, so that does complicate things a little.

bobrippling:
Cool! For that last bit, you could ask in [discussions](https://github.com/orgs/espruino/discussions), I think there's a few iPhone users about

gfwilliams:
I'm really not convinced about requesting state (it feels like a hack to work around a bug somewhere) - I've put some more of my thoughts on this on https://codeberg.org/Freeyourgadget/Gadgetbridge/issues/4886#issuecomment-8622057

edit:

>  I ran into an app where GadgetBridge isn't sending musicinfo even on play/pause (libro.fm)

This is kind of what I mean - we should fix this in Gadgetbridge (or maybe it's a libro.fm issue?), and then it'll help everyone using it (even those without a Bangle.js).

kidneyhex:
@gfwilliams The issue I ran into is how to store the music info/state. We might get the music info, but we lose it unless we store it somewhere.

For Libro.fm the issue is audiobook chapters means no new musicinfo for a long time. But even if we had the musicinfo, we lose it as soon as we change apps unless we keep writing to flash. But if you listen to short things like songs or youtube shorts, you change musicinfo ever 30s-3min, so it adds up to a lot of writes to flash.

I wouldn't poll for music info/state, just once on opening messagegui... but then I ran into the "No messages" popup needing to be closed or delayed to show the new music info. Little design choices started adding up beyond what I was comfortable with.

Edit: oooh, wait... Just write to flash on kill events maybe? Set up the kill event handler in the listener and some flag that it was already set up. 🤦

kidneyhex:
I've updated this with logic to store the music in the `require('messages').music`, and then on kill event persist it into `messages.music.json`

Then `messagegui` uses `require('messages').getMusic()` to read the latest music info and there's a `_lastPlayed` flag to help expire old messages.

This still has a debug method in it to count number of times it writes to file--so not really ready for review/merge. Just here for info purposes for now.

gfwilliams:
> oooh, wait... Just write to flash on kill events maybe? Set up the kill event handler in the listener

Yes - to be honest I thought this is what happened already! The idea was always that `android/app.js` converted the music updates into a message type with `id:"music"` which are added with `require("messages").pushMessage` and should have got stored along with the other messages (where there's already code to handle saving to flash only when swapping apps).

But I've just looked into this, and as you say, that doesn't happen. I think it might be the return statement here? https://github.com/espruino/BangleApps/blob/master/apps/messagegui/lib.js#L41

So I'm not sure at what point we stopped storing those and why - it looks like it might have slipped in (perhaps even as a typo) as part of https://github.com/espruino/BangleApps/commit/2a506e7421f2320ba65afcadeb9d3458a8958dd1

So ideally, I'd really like to try and fix that rather than having code that deliberately extracts the music messages from the messages list and then writes it into a separate file.

gfwilliams:
I just tried it, and it does feel like literally just removing that one line (`else return;`) fixes it?

You can send music updates from the Web IDE by pasting in:

```
GB({"t":"musicstate","state":"play","position":1,"shuffle":1,"repeat":1})
GB({"t":"musicinfo","artist":"My Artist","album":"Album","track":"A song with a reasonably long name","dur":240,"c":-1,"n":-1})
```

And then you can use `require("messages").getMessages()` to get the messages (which should include music), and it persists even after restarting with `load()`

gfwilliams:
Just to add, the whole `exports.music` thing in `messages/lib.js` appears to be completely pointless. I can't actually see any code that's using it?

edit: Assuming that removing `else return;` works for you, I can push a change to remove it (as well as the `.music`) field.

kidneyhex:
> Just to add, the whole `exports.music` thing in `messages/lib.js` appears to be completely pointless. I can't actually see any code that's using it?

Yes, according to an old issue about it, exports.music was only still around to be what helps combine musicinfo/musicstate. I don't think it needs to be exposed, or even exist if that happens elsewhere.

> edit: Assuming that removing `else return;` works for you, I can push a change to remove it (as well as the `.music`) field.

I assumed the idea was to not write to flash too often. Just removing that return makes it write a ton. But I maybe overly cautious about how much wear that'd cause?

But I'm okay with anything, my caution was just trying to decern what the original intent was and not break anything. 😄

gfwilliams:
> Just removing that return makes it write a ton

Are you sure? because I thought the messages app was supposed to be smart enough to only write to flash when apps were switched (but maybe I'm mis-remembering)

kidneyhex:
I think the messagegui listener writes to flash for every message if you don't have auto-open enabled.

https://github.com/espruino/BangleApps/blob/702db69c399382c3463b8f1babbb12d97b73611d/apps/messagegui/lib.js#L48

It only does the write on kill in the app.js itself.

gfwilliams:
Ok, thanks - I'm looking into this.

Can you see any reason why making messages do a delayed write for every message type wouldn't work? I'm setting it up so `require("messages").getMessages()` would still return the updated messages even if they weren't saved right away.

kidneyhex:
Yep, I think that'd work fine.

gfwilliams:
Just to add I am still working on this - sorry for the delay.

kidneyhex:
No worries. I use my own music controls that I've been working on anyway. (I like to thumbs up/down stuff. I'm probably in the minority there.)

I was only attempting the fix cuz I thought I had a solution.

My latest attempt saving music msgs on "kill" caused a ton of trash file build up.

gfwilliams:
How much trash are we talking about here? I was working on the assumption that we actually stored the position in the track, but I just realised we ignore it (which kind of sucks as Gadgetbridge is sending loads of data when there's no need).

... so really if we save when swapping apps, we should only really be saving to storage when the song or play state changes and we swap apps - a lot of the time if you're just flicking between apps while on the same track it'll be ok, or if you're not doing anything with the watch it shouldn't be saving anything either.

> I like to thumbs up/down stuff

That's cool - how can you do that? Is there some event that can be sent via Gadgetbridge, or you have to do intents specific to one particular player?

kidneyhex:
I was getting about 60% full of trash per day. I didn't dig in to figure out for sure why it was happening, but stopped as soon as I reverted my changes.




On Android there's a "setRating" method on the same class as the play/pause methods and ratings are stored in the metadata with artist/album. I've only seen Music Boss on Pebble use it though.

At first, I modified Gadgetbridge with new media commands to do it. But I've started adding adapters for apps that don't follow the standard, so I'm working on a companion app that uses intents. Eventually I'll probably publish the app and source so anyone with tasker/whatever can also use it.

I keep hoping someone will beat me to it, but so far there's always a little bit missing. Music Boss Deluxe has intents for setting ratings but doesn't have a way to read the current rating. Tasker lets you read the rating but doesn't let you set it. etc. etc.

PR #4081, opened 2025-12-04
[messages/messagegui] More control over auto show new message

### LINKED ISSUE #4011: Feature request: Auto display messages on non-CLOCK apps
Hi,

Some apps I will leave running and by that messages will not auto display since `Bangle. CLOCK=1` is not set.

I want to list some apps that should let messages auto display. But I don't want clock face specific widgets to show as would be the case if I set `CLOCK=1`

Maybe it would also make sense to differentiate apps that should let messages auto-display always or just if the screen is locked.

So I have two suggestions:

1. Set `Bangle.CLOCK=2` to say this app should let messages auto-display but not show clock face widgets. Maybe set `=3` to say it should auto-display messages only if the screen is locked.

or

2. Add `Bangle.AUTO_SHOW_MESSAGES=0/1/2` or similar to allow messages to take over the app. (0 - no, 1 - yes, 2 - only when locked)

The next step, as I imagine, would be to create a bootloader type app for setting the value depending on which app is running and user preferences. 

Does this speak to anyone else? Is there a better solution?

thyttan:
The updated setting works like so:
* **Auto-open new msg** : Should we open the messages app for new messages?
  * **Never** : Never open for new messages
  * **On clock** : Open if a clock face is showing (default)
  * **If locked** : Open if the screen is locked
  * **Always** : Always open new messages

I debated if the `**If locked**` option should also open if unlocked but on the clock face - currently it will ~~not~~. **Please suggest how it that should behave!**

Try via my app loader:
- https://thyttan.github.io/BangleApps/?id=messages
- https://thyttan.github.io/BangleApps/?id=messagegui

Todo:
- [x] Check if other message gui's need to be tweaked to not break.
  - no problem

Closes #4011

### DIFF
--- a/apps/messages/settings.js
+++ b/apps/messages/settings.js
@@ -81,7 +81,7 @@
       onchange: v => updateSetting("quietNoAutOpn", v)
     },
     /*LANG*/'Auto-open new msg': {
-      value: settings.autoOpen,
+      value: settings.autoOpen&3,
       min:0,max:3,step:1,
       format: v=>["Never","On clock","If locked","Always"][v],
       onchange: v => updateSetting("autoOpen", v)


apps/messages/settings.js
```diff
@@ -79,9 +80,15 @@
...
       onchange: v => updateSetting("quietNoAutOpn", v)
     },
-    /*LANG*/'Disable auto-open': {
-      value: !!settings.noAutOpn,
-      onchange: v => updateSetting("noAutOpn", v)
+    /*LANG*/'Auto-open new msg': {
+      value: settings.autoOpen,
+      min:0,max:3,step:1,
+      format: v =>
+        v===0 ? "Never" :
```
gfwilliams: Can you just do ` v=>["Never","On clock","If locked","Always"][v]` here? That's what we normally do for options like this

apps/messages/settings.js
```diff
@@ -79,9 +80,15 @@
       value: !!settings.quietNoAutOpn,
       onchange: v => updateSetting("quietNoAutOpn", v)
     },
-    /*LANG*/'Disable auto-open': {
-      value: !!settings.noAutOpn,
-      onchange: v => updateSetting("noAutOpn", v)
+    /*LANG*/'Auto-open new msg': {
+      value: settings.autoOpen,
```
gfwilliams: maybe `settings.autoOpen&3` will convert it to an integer and keep it in range

apps/messagegui/lib.js
```diff
@@ -60,7 +64,7 @@ exports.listener = function(type, msg) {
   const quiet = (require("Storage").readJSON("setting.json", 1) || {}).quiet;
   const unlockWatch = appSettings.unlockWatch;
   // don't auto-open messages in quiet mode if quietNoAutOpn is true
-  if ((quiet && appSettings.quietNoAutOpn) || appSettings.noAutOpn)
+  if ((quiet && appSettings.quietNoAutOpn))
```
gfwilliams: Would be nice to ditch the double brackets?

gfwilliams:
Looks good to me, thanks! just made a few suggestions

> I debated if the **If locked** option should also open if unlocked but on the clock face - currently it will not

To me it feels like it should open if locked/unlocked on the clock face? Then the 4 options are in order of how likely a message is to open - each one increases the places a message would open?

apps/messages/settings.js
thyttan: I'm not knowledgeble with `&` operator but happy to go with it if you think it's right. Otherwise `settings.autoOpen ?? 1` ?

Also, line 15 sets the default:
```
    if (settings.autoOpen === undefined) settings.autoOpen = 1;
```

thyttan: > Otherwise settings.autoOpen ?? 1 ?

But maybe it's that doesn't save us from out of range?

gfwilliams: Yes - also not if for some reason it wasn't saved as an integer. `settings.autoOpen&3` is tidy and should work fine

thyttan: Ok - thanks!

thyttan:
Should be good to go then 👍

gfwilliams:
Ok, great! Looks good to me!

PR #4082, opened 2025-12-05
[multitimer] Fixes #3693; Adds support for 12h alarm display

### LINKED ISSUE #3693: [multitimer] Timer cannot be set over the day end
### Affected hardware version

Bangle 2

### Your firmware version

2v25

### The bug

When timer is being set to the point in time which must be in the next day, this timer is shown incorrectly (with negative hours and seconds) and fires at the wrong time (when the closest timer fires)

### Expected behaviour

I expect a timer to fire after the given amount of time, while showing correct left time.

### Steps to reproduce

Say when there is 23:40 of local time

- open the app
- set two timers: for 5 minutes and for 25 minutes with auto-removal after expiration
- look at the displayed time and see that the second timer is shown as `-24:41:-9` instead of `0:24:51`
- wait for the first timer to fire and see that the second (with indication `25 min`) also fired, and removed afterwards

### Installed apps

_No response_

kidneyhex:
Link: https://kidneyhex.github.io/BangleApps/?id=multitimer

`Multi Timer 0.12`

Changes:
* Fixes displaying timers that go past midnight for #3693
  * alarm.t = ms since midnight that the timer ends on; so `t - currentTime` goes negative if `t` is tomorrow morning (aka smaller than `currentTime`)
* Fixes `uiRemove` logic that's trying to `clearTimeout` for every `timerInt2` when it should `clearInterval`, instead using the common method `clearInt()` to do it.
  * Not sure this has any effects, just noticed it by chance while debugging
* Adds support for 12h format if locale is set to that

Triggering of timers past midnight was already fixed with this: https://github.com/espruino/BangleApps/commit/44e201326f78bb59ea90c018caf6ea6f89f10480

But timers still display as negative if they go past midnight. 
Since timers can't be greater than `23:59:59`, we don't need to worry about multiple days.

### DIFF
--- a/apps/multitimer/app.js
+++ b/apps/multitimer/app.js
@@ -658,7 +658,7 @@ function editAlarm(idx, a) {
       format: v => {
         // Show 12h or 24h based on locale settings
         const mer = require("locale").meridian(new Date(1999,1,1,v,0,0));
-        return (!mer) ? v : (v%12||12) + mer;
+        return mer ? (v%12||12) + mer : v;
       },
       onchange: v => {
         t.hrs = v;


kidneyhex:
Tagging original MultiTimer author: @frigis1 

Let me know if you have any concerns.

apps/multitimer/app.js
```diff
@@ -631,6 +655,11 @@ function editAlarm(idx, a) {
     },
     "Hours": {
       value: t.hrs, min: 0, max: 23, wrap: true,
+      format: v => {
+        // Show 12h or 24h based on locale settings
+        const mer = require("locale").meridian(new Date(1999,1,1,v,0,0));
+        return (!mer) ? v : (v%12||12) + mer;
```
bobrippling: This might be more easily read as:

```suggestion
        return mer ? (v%12||12) + mer : v;
```

bobrippling (commented):
Thanks for the fixes! LGTM, one small comment

Shall give @frigis1 a few days to take a look too

bobrippling:
Ok, merging - thanks!

PR #4084, opened 2025-12-08
Add GIF creation instructions to README

RKBoss6:
I noticed that some apps had an animated gif as a screenshot, which I thought was tremendously useful to provide a "video" of what the app does, instead of just pictures. I've added some instructions to the main repo readme for how to do that, so others can do it easier if they are just starting out.


gfwilliams:
Thanks! That sounds like a great idea

PR #4085, opened 2025-12-08
Deployment Pure Numerals Clock

peterbuehler:
Added ChangeLog

### DIFF
--- a/apps/purenumerals/ChangeLog
+++ b/apps/purenumerals/ChangeLog
@@ -1,3 +1,1 @@
-0.03:
-  - first released version.
- 
+0.03: first released version.


apps/purenumerals/ChangeLog
```diff
@@ -0,0 +1,3 @@
+0.03:
+  - first released version.
+
```
bobrippling: Let's match other changelogs:

```suggestion
0.01: First released version.
```

bobrippling:
And thanks - next time if you leave the pull request open (#4079), it'll automatically update when you make a change. Makes it easier to keep track of for maintainers too :)

apps/purenumerals/ChangeLog
peterbuehler: Well, Version 0.01 and Version 0.02 have been Beta-Versions, that have been tested by some people. They were not good enough to be deployed on espruino. This is the reason why the first version is is 0.03. (I would prefer Version 1.00 but that seems not to be common.)

peterbuehler:
Ok, sorry, I did not know, that I can do any commits and leave a request open.

apps/purenumerals/ChangeLog
bobrippling: Not a problem, shall we go with this, to match other formats?

```suggestion
0.03: first released version.
```

bobrippling:
No worries, glad to have shown you :)

bobrippling:
Thanks!

peterbuehler:
How long does it take until the Pure Numerals Clock is shown in the App Loader?

bobrippling:
Gorden tends to do weekly updates, but in the meanwhile you can check the [dev app loader](https://espruino.github.io/BangleApps/)

peterbuehler:
Ok, thank you!

PR #4087, opened 2025-12-09
Implement SOS mode with color selection

fparri:
Added SOS blinking functionality with user-selected color and touch interaction to set the main screen color.

Removed unused function draw2Pattern ( was previously defined but never called ), fixing lint error.


bobrippling:
Nice idea - can you bump the [changelog version](https://github.com/espruino/BangleApps/blob/master/apps/wristlight/ChangeLog) and update the version in `metadata.json`?

(you don't need to open a new pull request - any changes will be added to this one :) )

Tagging @pidajo as the original author too

fparri:
Thanks a lot, I've just done that! :)

bobrippling:
Thanks! We'll give @pidajo a few days and if there's no response by then I'll merge

bobrippling:
Ok, merging - thanks for the change :)

PR #4088, opened 2025-12-09
bigclkinfo: new app - one fullscreen clock info

thyttan:
Initial version of an app proposed at https://github.com/orgs/espruino/discussions/6957#discussioncomment-14091812

> - [ ] An app that is just a single clock info field, full screen. 
>  - I use a clock without clock infos, but having them accessible by a swipe on `quicklaunch` would be neat.

Try it out at https://thyttan.github.io/BangleApps/?id=bigclkinfo

### DIFF
--- a/apps/bigclkinfo/app.js
+++ b/apps/bigclkinfo/app.js
@@ -1,3 +1,7 @@
+// TODO:
+// Add setting to decide if the app shoud set Bangle.CLOCK=1 ?
+// Make an updating analog clock info entry to use as start card. (update clkinfoclk or make a new one?)
+
 let isClock = true; // TODO: make optional via setting.
 if (isClock) {
   Bangle.setUI("clock");


RKBoss6:
Tried it out, there are a couple of bugs: 
download
* Only clears half the text, leading to what you see above - The clockInfo dimensions might need to be tweaked
* You can't see it in the picture, but sometimes the widget bar will appear, and only when I switch a clockInfo does it clear.

Other than that, I would love to see where this is headed!

thyttan:
- [x] fix icon on watch - black on black with darl theme.

thyttan:
> - Only clears half the text, leading to what you see above - The clockInfo dimensions might need to be tweaked

I can't seem to reproduce this. Can you post a repro? 

> - You can't see it in the picture, but sometimes the widget bar will appear, and only when I switch a clockInfo does it clear.

The app doesn't load widgets so not sure why that is happening for you. Was thinking maybe fastload utils? But I don't think that should be the cause in this case...

RKBoss6:
I don't have widget utils or fastload utils installed... I think when a widget updates for me it triggers a redraw of the whole bar. I that normal behaviour? Otherwise it might just be a problem on my end.  That could also be what's causing the text to not be cleared. If it's shifting due to the widget bar, it could cause issues, right?

thyttan:
The apps are responsible for loading in widgets on launch then most is left up to the widgets themselves (barring widget_utils stuff).

Come to think of it, do you think it's a consequence of using the loading animation you're working on?

RKBoss6:
I tried removing`loadanim`, and it didn't affect anything. I think the widgets are being drawn when a widget updates (eg. bluetooth disconnects), and the clearing issue remains... Can you share a screenshot of what the app looks like on your side?

thyttan:
`bigclkinfo` doesn't load widgets. In the normal case when you launch it everything else is cleared and the watch is reset before the app is loaded. So no widgets will be present.

Is there a clock info that displays widgets? If you have such a clock info maybe that would be the culprit.

But other than that I read you have some memory leaking currently. I suspect when you work that out `bigclkinfo` will behave better.

RKBoss6:
Hmm, the leak seems to have been fixed now, and the problems still persist. Just to make sure, when you switch lists and items, no clearing issue happens? Also try disconnecting bluetooth or updating/redrawing another widget to see if that's shown. @gfwilliams or @bobrippling would you mind testing this out as well to see if it's an issue with my watch, or a global issue?

thyttan:
> Also try disconnecting bluetooth or updating/redrawing another widget to see if that's shown.

`bigclkinfo` doesn't (shouldn't) have any widgets present. I.e. there's NO logic from widgets loaded by the app and running when `bigclkinfo` is started. This is for the normal case even with fastload utils. 

Could it be that you have made customizations to either loading bahaviour or widgets behaviour with either your own code or some app installed from the app loader? I recommend you make a backup of your watch and then reinstall all apps (using the official app loader) and see if the issue persists for you.

thyttan:
> Just to make sure, when you switch lists and items, no clearing issue happens?

Not like the one you described.

I have noticed there is some problem with how I try to set the initial clock info and the first touch interactions. But draws to screen clears previous info just fine.

gfwilliams:
> bigclkinfo doesn't (shouldn't) have any widgets present

If it fast loads from the launcher, then the launcher may have loaded widgets already? It shouldn't though since your code doesn't mention `loadWidgets/drawWidgets` anywhere which I think the fast loading checks for.

Just looking at this, it'd be really cool if it used the clockinfo drawBorderedImage (https://github.com/espruino/BangleApps/blob/master/apps/clock_info/lib.js#L448) on top of a custom background: http://localhost/apps/?id=clockbg - Quite simple but it'd really brighten it up.

You could also add the 'clockinfo clock' that's in analogquadclk so you can be sure there's always the ability to show time: https://github.com/espruino/BangleApps/blob/master/apps/analogquadclk/app.js#L190-L212

I'd also use `Bangle.setUI("clock")` rather than setWatch/.CLOCK=1 you have right now. It was abstracted so that other apps could change the behaviour of the button if they needed to.

bobrippling:
> Hmm, the leak seems to have been fixed now, and the problems still persist. Just to make sure, when you switch lists and items, no clearing issue happens? Also try disconnecting bluetooth or updating/redrawing another widget to see if that's shown. `@gfwilliams` or `@bobrippling` would you mind testing this out as well to see if it's an issue with my watch, or a global issue?

Sure - no problems here either:

image

image


Perhaps try running the app from RAM, or give us a list of the boot code you have?

thyttan:
> > bigclkinfo doesn't (shouldn't) have any widgets present
> 
> If it fast loads from the launcher, then the launcher may have loaded widgets already? It shouldn't though since your code doesn't mention `loadWidgets/drawWidgets` anywhere which I think the fast loading checks for.

Yes fastload utils should not change anything because of the reason you say. I also use fastload utils and do not have the problem.

> Just looking at this, it'd be really cool if it used the clockinfo drawBorderedImage (https://github.com/espruino/BangleApps/blob/master/apps/clock_info/lib.js#L448) on top of a custom background: http://localhost/apps/?id=clockbg - Quite simple but it'd really brighten it up.

Yeah that could be cool! Potentially leaving it for a v0.02. We'll see.

> You could also add the 'clockinfo clock' that's in analogquadclk so you can be sure there's always the ability to show time: https://github.com/espruino/BangleApps/blob/master/apps/analogquadclk/app.js#L190-L212

I do basically this but by adding [`clkinfoclk`](https://espruino.github.io/BangleApps/?q=clkinfoclk) as a dependency and then try to display it directly on app launch. Am having some problem with that selection on first draw currently though.

> I'd also use `Bangle.setUI("clock")` rather than setWatch/.CLOCK=1 you have right now. It was abstracted so that other apps could change the behaviour of the button if they needed to.

Good point - I'll change it. It just felt a bit heavy to call out to setUI for just the things I needed. But it makes sense in light of your comment.

thyttan:
I believe this may be ready for merging now.

bobrippling:
I'll let Gordon have the final say, having reviewed previously :)

gfwilliams:
Thanks! Just made a few small tweaks.

I'll merge now, but a screenshot might convince more people to try it?

I also tried doing text by writing to a buffer and using filter on the buffer to get a nice background:

image

```
let isClock = true; // TODO: make optional via setting.
if (isClock) {
  Bangle.setUI("clock");
}

// once at the start
let background = require("clockbg");
let clock_info = require("clock_info");
const R = Bangle.appRect;
let txtg = Graphics.createArrayBuffer(R.w>>1, 32, 2);
txtg.transparent = 0;
txtg.palette = new Uint16Array([0,g.theme.bg,g.theme.bg,g.theme.fg]);

// Load the clock infos
let clockInfoItems = clock_info.load();
let clockInfoClockIdx = clockInfoItems[0].items.findIndex(c => c.name=="HRM");
if (clockInfoClockIdx>=0) { // pull the clock to the front
  let clockInfoClock = clockInfoItems[0].items[clockInfoClockIdx];
  clockInfoItems[0].items.splice(clockInfoClockIdx,1); // remove
  clockInfoItems[0].items.unshift(clockInfoClock); // add to front
}
// Add the
let clockInfoMenu = clock_info.addInteractive(clockInfoItems, {
  app: "bigclkinfo",
  // Add the dimensions we're rendering to here - these are used to detect taps on the clock info area
  x : 0, y: 0, w: R.w, h: R.h,  // You can add other information here you want to be passed into 'options' in 'draw'
  // This function draws the info
  draw : (itm, info, options) => {
    // itm: the item containing name/hasRange/etc
    // info: data returned from itm.get() containing text/img/etc
    // options: options passed into addInteractive
    // Clear the background
    background.fillRect(options.x, options.y, options.x+options.w-1, options.y+options.h-1);
    // indicate focus - we're using a border, but you could change color?
    if (options.focus) g.fillRect(options.x, options.y, options.x+options.w-1, options.y+options.h-1); // show if focused
    background.fillRect(options.x+3, options.y+3, options.x+options.w-5, options.y+options.h-5);
    // we're drawing center-aligned here
    var midx = options.x+options.w/2;
    let scale = 4;
    if (info.img) clock_info.drawBorderedImage(info.img, midx-12*scale,options.y+10, {scale:scale}); // draw the image
    let foundFontText = txtg.findFont(info.text, {
      w : txtg.getWidth()-4,    // optional: width available (default = screen width)
      h : txtg.getHeight()-4,    // optional: height available (default = screen height)
      min : 10,   // optional: min font height
      max : 24,   // optional: max font height
      wrap : true, // optional: allow word wrap?
      trim : true // optional: trim to the specified height, add '...'
    }); // TODO: g.findFont returns max size 28px. Would be nice with bigger font if there's room.
    //print(foundFontText);
    txtg.clear(1);
    txtg.setFont(foundFontText.font).setFontAlign(0,1);
    txtg.drawString(foundFontText.text, txtg.getWidth()/2, txtg.getHeight()-2); 
    txtg.filter([ // a gaussian filter
        0, 1, 1, 1, 0,
        1, 1, 1, 1, 1,
        1, 1, 1, 1, 1,
        1, 1, 1, 1, 1,
        0, 1, 1, 1, 0,
    ], { w:5, h:5, div:1, max:1, filter:"max" });
    g.drawImage(txtg, 0, R.h - txtg.getHeight()*2-4, {scale:2});
  }
});

Bangle.on("lock", function(locked) {
  // ensure that when unlocked, we automaticlly focus the clockinfo
  if (!locked && !clockInfoMenu.focus)
    Bangle.emit("touch",0,{x:100,y:100});
});
E.prependListener("kill", ()=>{clockInfoMenu.menuA = 0; clockInfoMenu.menuB = 0;}) // reset to initial menu state to prepare for next launch.
```

Do you think we should put that in?

thyttan:
That looks great - please do add that as well! Thanks for the tweaks :)

gfwilliams:
Thanks! Just done - and added screenshot

thyttan:
... but it pulls hrm to front instead of clock 🥲

also, do we have `findLastIndex`? could maybe be itsybitsy faster if so?

thyttan:
just looked - don't think we have findLastIndex 👌

gfwilliams:
Just fixed - I'd changed it to HRM to test and forgot to put it back!

thyttan:
Thanks!

Now I'm contemplating an app launcher clock info 😅

gfwilliams:
That's a really cool idea. With a settings page that lets you choose how many icons you want and what apps they should load? I'll take a quick look, as I reckon the `BTHome` app does pretty much everything we need for that already - it should be pretty easy to modify

thyttan:
Sounds like a good design choice 🤓

thyttan:
And listen to this: a clock info that holds four sub-clockinfos 🙃🫣😈

I'll stop now...

gfwilliams:
Just added: https://espruino.github.io/BangleApps/?id=clkinfolaunch

I did try resizing the app icon automatically, but it always looks rubbish so I didn't bother. It might be nice to add some extra icons to https://github.com/espruino/BangleApps/tree/master/apps/icons if it's being used more though.

I'm actually amazed nobody's done a googly eyes clockinfo yet, or a tamagotchi.

thyttan:
Nice! I tried it out and it works well. 👌 But I think it might be better to make the clkinfolaunch have its own entry instead of adding to the standard Bangle one?

gfwilliams:
> it might be better to make the clkinfolaunch have its own entry instead of adding to the standard Bangle one?

Can do - you want to do a PR?

It could be an option where it goes when you add the icon? Might make sense?

thyttan:
I'll look at making a PR 👍 no ETA though 👌

RKBoss6:
Just wanted to pop in here - the shortcuts clockinfo app does the exact same thing as the clock info launch

thyttan:
Aha... bummer I didn't look before thinking out loud... 🫢

thyttan:
But they do at least differ a tiny bit on the point that `clkinfolaunch` adds to the Bangle entry and `shortcuts` makes its own.

gfwilliams:
Argh, what a pain! Thanks for the heads up - I should have thought to look

What do you think - shall I just delete the launcher app then?

RKBoss6:
No worries! I think deleting it is the best course of action - Just so nobody gets confused by the two separate apps that do the same thing...

gfwilliams:
Just deleted - it's kind of a shame that the shortcuts app provides all its own icons instead of just using the `icons` lib though.

RKBoss6:
Yeah - perhaps we could migrate the icons not in the library to the lib, then use the library for the app, to centralize everything somewhat.

RKBoss6:
btw, I tried this app again with the background, and it looks fabulous! The widget and clearing issues seem to be gone for me, and the clockInfo with the background is looking amazing now!

thyttan:
@gfwilliams's final touches were chef's kiss! 👌

PR #4096, opened 2025-12-18
[schedclock] allow setting random faces per entry; fix 24h selection

kidneyhex:
Link: https://kidneyhex.github.io/BangleApps/?id=schedclock

## Overview

1. Add ability to select multiple clockfaces per schedule entry and have one set at random.

2. Fix bug making 24h always display as 12h when setting the hour in settings.

### readme.md

- Added note about `fastload utils` compatibility

### lib.js

- Removed passing clockface js file name into onAlarm, instead read clockface array from settings file

### settings.js

- Changed from selecting one clockface from a list, to using checkboxes to select multiple
- Add validation that at least face is selected before saving entry
- Fix spelling of `meridian` variable (from meridean)
- Remove passing "1" when getting meridian -- this was causing 24h to always display as 12h
- Rename `validationError` to `hasScheduleConflict` to be more specific

### DIFF
--- a/apps/schedclock/ChangeLog
+++ b/apps/schedclock/ChangeLog
@@ -1,1 +1,3 @@
 0.01: New App!
+0.02: Add multiple clock faces per entry and apply one at random
+      Fix bug handling 24h format in settings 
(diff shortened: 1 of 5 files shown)


bobrippling:
Nice feature, thanks!

PR #4100, opened 2025-12-20
stopwatch-touch: lap support

bobrippling:
This finishes some of the code to permit tracking laps

- [x] Test


bobrippling:
Tested, will merge after perhaps a week

PR #4104, opened 2025-12-22
clkinfo: add filtering

bobrippling:
This allows users to hide clkinfos they aren't using. No extra cost since we use the clkinfo cache

Closes espruino/EspruinoAppLoaderCore#83

- [x] Fix bootupdate: it isn't triggering when I change `clock_info.json`, need to force the boot code regeneration/`bootupdate.js` to run
- [X] Test

### DIFF
--- a/apps/clock_info/settings.js
+++ b/apps/clock_info/settings.js
@@ -7,6 +7,17 @@
     require('Storage').write("clock_info.json", settings);
   }
 
+  let forced = false;
+  function forceBootUpdate() {
+    if(forced) return;
+
+    const fname = "setting.json";
+    const j = require("Storage").readJSON(fname);
+    j.toggle = !j.toggle;
+    require("Storage").writeJSON(fname);
+    forced = true;
+  }
+
   let menu = {
     '': { 'title': 'Clock Info' },
     /*LANG*/'< Back': back,
@@ -55,6 +66,7 @@
                   delete settings.exclude[file];
               }
               save();
+              forceBootUpdate();
             },
           };
         });


RKBoss6:
Hey - quick question. If the user excludes a specific clock info, will that mean that apps specifically trying to load that clockInfo won't be able to? Eg. An app only uses one clock info like 'sunrise clockinfo' but everywhere else it's omitted?

bobrippling:
> Hey - quick question. If the user excludes a specific clock info, will that mean that apps specifically trying to load that clockInfo won't be able to? Eg. An app only uses one clock info like 'sunrise clockinfo' but everywhere else it's omitted?

It depends how the clkinfo is loaded. If `require('clkinfo').load()` is used, the excluded ones will be skipped:

https://github.com/espruino/BangleApps/blob/638d5cb2a5fa6ada754ef3c1bba9d54009387d9f/apps/clock_info/lib.js#L174-L190

How does the app in question use (for example) sunrise clkinfo?

bobrippling:
@gfwilliams how does this look?

apps/clock_info/settings.js
```diff
@@ -28,7 +29,47 @@
...
+            },
+          };
+        });
+
+      // clean up stale entries
+      Object
+        .keys(settings.exclude)
+        .filter(k => !(k.replace(re, "") in filterMenu))
+        .forEach(k => {
+          delete settings.exclude[k];
```
gfwilliams, line 80: One very tiny thing: suppose someone installs a clockinfo, excludes it, and then deletes it. You handle that here, but only save it if the user actually changes something in the menu. Maybe just add a `changed=true; ... if (changed) save()` so at least just looking at the menu would be enough to remove the entry?

gfwilliams:
That looks perfect to me - really nice and tidy.

I wouldn't worry about the 'what if an app used a clockinfo' thing - I'm not aware of any apps doing that, and even then someone would have to explicitly disable the clockinfo.

> bootupdate isn't triggering when I change clock_info.json

The easiest way to do that would be to just rewrite `'setting.json'`. Rewriting with the same data won't work as Espruino is smart enough to detect that, but I'm sure you could add something like `s.toggle = !s.toggle` that'd be enough. Or you could store the `exclude` list in `setting.json` instead?

I wouldn't add a check for the contents of the clockinfo JSON file as that's just another read and possible delay.

apps/clock_info/settings.js
bobrippling: Sounds good, done!

bobrippling:
> > bootupdate isn't triggering when I change clock_info.json
> 
> The easiest way to do that would be to just rewrite `'setting.json'`
> ...
> something like `s.toggle = !s.toggle` that'd be enough

Thanks, sorted!

gfwilliams:
Looks perfect - thank you! Shall we merge?

bobrippling:
Sounds good to me - thanks again!

thyttan:
Not sure it's from this. But clock infos misbehave on my watch, fw 2v28.70 iflash. Have you noticed anything? I'll try debugging some tomorrow. Haven't used clock infos in a while so not sure when it would have started.

RKBoss6:
How so? For me on regular cutting-edge fw, nothing seems out of the ordinary using this version

thyttan:
OK - thanks! Interesting and a little reassuring :)

bobrippling:
I've not noticed anything, but haven't changed the setup after I filtered down the clkinfos. What's the misbehaviour?

RKBoss6:
@thyttan What's the problem you're having? Recently, my ClockInfos started bugging out too, and wouldn't update the screen to show the new clockInfo, saying :
```javascript


Uncaught Error: Function code is undefined, not a string
    at menuHideItem (clock_info:293:23)
      itm.hide(options);
                      ^
    at clock_info:327:31
      menuHideItem(oldMenuItem);
```
Even though it was never a problem before, which now crashes my clock face each time. I had updated to this version prior, but the issue wasn't there, and now it just randomly is showing up...

thyttan:
Yes, I think we're seing the same thing now then 🙃

Would you have the time to file a bug report @RKBoss6?

RKBoss6:
Yeah, I can get on that in a bit!

gfwilliams:
Ah, interesting. Can you file it on the Espruino project?

The idea was to detect broken functions but I guess the interpreter decides not even to try allocating a code variable if the function is empty

RKBoss6:
Yeah, I can copy the issue over to the espruino repo

gfwilliams:
Just fixed that issue: https://github.com/espruino/Espruino/issues/2681

PR #4108, opened 2025-12-31
[iOS] Add automatic app name detection

RKBoss6:
This stems from [this discussion](https://github.com/orgs/espruino/discussions/7906), where a new way of determining app names was suggested by @gfwilliams. This PR implements that, with a setting (off by default) to enable that. This has been tested, and works well.

### DIFF
--- a/apps/ios/settings.js
+++ b/apps/ios/settings.js
@@ -21,7 +21,7 @@
       }
     },
     /*LANG*/"Auto-Detect App Names" : {
-      value : !! !settings.dontDetectNames,
+      value : !settings.dontDetectNames,
       onchange: v => {
         settings.dontDetectNames = !v;
         print(settings.dontDetectNames)


thyttan:
Leaving this to Gordon for when he's back. Core app and I don't have a iDevice.

RKBoss6:
Ah - one issue with marking the names with a `*` - It changes the src, so any app icons won't get shown. I don't feel comfortable with not showing a marker, in case it detects a name wrong and the user is confused, so should we make a modification to the app icons to ignore special signs like `*` when determining what icon to show?

Edit: just made PR #4110 to fix this

apps/ios/boot.js
```diff
@@ -272,11 +243,29 @@ E.on('notify',msg=>{
 
     return;
   }
-
+  let settings = require("Storage").readJSON("ios.settings.json",1)||{};
+  var name=""
+  if(settings.detectNames==true&&(!appNames[msg.appId]&&msg.appId)){
```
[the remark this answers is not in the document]
RKBoss6: Resolved

apps/ios/boot.js
```diff
@@ -272,11 +243,31 @@ E.on('notify',msg=>{
...
+    
+    let l = msg.appId.split(".");
+    // get the last part of the ID
+    name = l[l.length - 1]
+    // apply detection methods
+    name=name
+      .replace(/([a-z])([A-Z])/g, '$1 $2')  // Space between lower->upper (AppName → App Name)
+      .replace(/([A-Z]+)([A-Z][a-z])/g, '$1 $2') // Space between acronym->word (SMSMessage → SMS Message)
+      .replaceAll("  "," ");  // Correct duplicate spacing
+    //capitalize
```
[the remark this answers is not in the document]
RKBoss6: @gfwilliams is this OK to keep in, or should we change this as per copilot's suggestion? It works on bangle.js, and that would really be the only use case for it - I would like to keep it as readable for others as possible, instead of replacing with another comment and regex pattern...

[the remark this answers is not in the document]
gfwilliams: I'd keep it  - I like replaceAll and it's been in Bangle.js for long enough we can depend on it

gfwilliams:
Thanks, this looks great! I'm interested in what others think but my gut instinct is to not include the `*` and to turn it on by default.

This is only the app source, so it's not really the end of the world if it's wrong occasionally. Having the `*` on every detected app name is going to be annoying and will lead to people doing PRs to add translations for the app name just to get rid of it, which goes against the whole point of this PR.

... also it means we don't need #4110

RKBoss6:
Just removed the `*` from the end, I'll test this more though.

RKBoss6:
Why was this closed? I had changed it to remove the marker at the end, but it's still ongoing as of now...

gfwilliams:
Ahh - it's because I merged #4110 and GitHub saw it had the keyword `fix` and then a link to this issue - it thought merging that other one fixed this one

RKBoss6:
Oh, no worries, thanks!

RKBoss6:
I think this is tested well, and works like a charm! We can get this merged if you're good with it @gfwilliams

apps/ios/boot.js
```diff
@@ -65,92 +65,89 @@ E.on('notify',msg=>{
...
+    "com.google.earth": "Google Earth",
+    "com.google.keep": "Google Keep",
+    "com.google.translate": "Google Translate",
+    "com.kik.chat": "Kik",
+    "com.groupme.GroupMeApplication": "GroupMe",
+    "com.skype.SkypeForiPhone": "Skype",
+    "com.tencent.mobileqq": "QQ",
+    "com.google.ios.youtubemusic": "YouTube Music",
+    "com.paramountplus.app": "Paramount+",
+    "com.chimebank.Chime": "Chime",
```
gfwilliams: surely this special case isn't needed?

apps/ios/boot.js
```diff
@@ -65,92 +65,89 @@ E.on('notify',msg=>{
...
     "ph.telegra.Telegraph": "Telegram",
-    "tv.twitch": "Twitch"
+    "com.apple.garageband10": "GarageBand",
+    "com.google.authenticator": "Google Authenticator",
+    "com.google.earth": "Google Earth",
+    "com.google.keep": "Google Keep",
+    "com.google.translate": "Google Translate",
+    "com.kik.chat": "Kik",
+    "com.groupme.GroupMeApplication": "GroupMe",
+    "com.skype.SkypeForiPhone": "Skype",
```
gfwilliams: we can probably forget Skype now Microsoft killed it?

gfwilliams:
Thanks! Looks good! It might be worth removing those two lines I flagged.

How about turning this on by default? Doesn't feel like there's a downside to me.

apps/ios/boot.js
RKBoss6: Just committed!

apps/ios/boot.js
RKBoss6: Just committed!

RKBoss6:
I just added the default check to set it to true if it's undefined - is that the proper way to go about something like this, or is there a more efficient/recommended way?

gfwilliams:
> is there a more efficient/recommended way?

I'd do `settings = Object.assign({detectNames:true},require("Storage").readJSON("ios.settings.json",1)||{})` but then you'd want to do that for the settings page too.

Often in a case like this I'd just rename the setting so that the default state is 0 - so rename `detectNames` to `noDetectNames` and now you just check `if (!noDetectNames) do name detection`

RKBoss6:
Thanks, just added that!

gfwilliams:
Looks great - thanks!

PR #4110, opened 2025-12-31
[Message Icons] Add special character screening for 'marked' icons

RKBoss6:
This just increases match compatibility even if an app sets the src to be 'Slack*'. This is made to fix [the issue](https://github.com/espruino/BangleApps/pull/4108#issuecomment-3702536561)  at PR #4108

Also updated the generate.js file, as I only added colors for the previous icons I added to the lib.js in #4105 
I also added more documentation and warning to help others not make this mistake

### DIFF
--- a/apps/messageicons/README.md
+++ b/apps/messageicons/README.md
@@ -10,5 +10,5 @@ Provides icons for phone apps that are utilized by `message gui` and any other m
 - Run the `generate.js` script. Be sure you have the `png-js` dependency needed. (you can install it by running `npm install png-js` in the terminal)
 
   
-<i><b>Don't modify the Lib.js file at all. When someone adds new icons, your old ones will be wiped</b></i>
+<i><b>Don't modify the `lib.js` file at all. When someone adds new icons, your old ones will be wiped</b></i>
 


apps/messageicons/icons/generate.js
```diff
@@ -83,12 +86,14 @@ Promise.all(promises).then(function() {
...
-  require("fs").writeFileSync(__dirname+"/../lib.js", `// This file is auto-generated - DO NOT MODIFY
-// If you want to add icons, change icons/icon_names.json and re-run icons/generate.js
+  require("fs").writeFileSync(__dirname+"/../lib.js", `// This file is auto-generated. --- DO NOT MODIFY AT ALL ---
+// If you want to add icons, import your icon as a 24x24 png, change icons/icon_names.json and re-run icons/generate.js
 exports.getImage = function(msg) {
   if (msg.img) return atob(msg.img);
   let s = (("string"=== typeof msg) ? msg : (msg.src || "")).toLowerCase();
   if (msg.id=="music") s="music";
+  // trim special characters at end for better matching. Trims: *!?.-_ and whitespaces
+  s = s.replace(/[*!?.\-_\s]+$/i, '');
```
bobrippling: To avoid the hyphen being interpreted as a range (e.g. `a-z`), let's place it at the end

```suggestion
  s = s.replace(/[*!?._\s-]+$/i, '');
```

apps/messageicons/icons/generate.js
```diff
@@ -83,12 +86,14 @@ Promise.all(promises).then(function() {
   require("fs").writeFileSync(__dirname+"/../icons.img", Buffer.from(iconData,"binary"));
 
   console.log("Saving library");
-  require("fs").writeFileSync(__dirname+"/../lib.js", `// This file is auto-generated - DO NOT MODIFY
-// If you want to add icons, change icons/icon_names.json and re-run icons/generate.js
+  require("fs").writeFileSync(__dirname+"/../lib.js", `// This file is auto-generated. --- DO NOT MODIFY AT ALL ---
```
bobrippling, line 89: Probably don't need this, I think the existing warning is clear enough :)

(same for the others)

RKBoss6: I don't know... I made that mistake, so I'm hoping this makes it a bit more clear not to modify it at all, but just a part of it which is what I thought the warning meant.

gfwilliams:
Thanks, although IMO I don't think we should be trying to filter out special chars. For instance by filtering out `-` you've now stopped `corona-warn` and `f-droid` from working, and by stripping whitespace you've stopped all the apps with spaces in the description from working.

... given the only reason for this is for filtering out the`*` which #4108 adds, and I'm not too fond of that, I think we should just not put the star in.

If you wanted to strip out that code and keep the colour changes that'd be good though.

RKBoss6:
Could we only strip end characters that are special instead?

gfwilliams:
I just think it's a slippery slope - the messageicons lib should really just take app name (ideally more of an ID) and return an icon, not try and perform other functions.

I don't really get why you wanted the `*` shown anyway - most apps will be fine, and those that aren't, someone will complain and we'll have to add an exception - but the * won't really make a difference to them apart from being a slight annoyance in the case that it had guessed correctly

RKBoss6:
Ok, just updated this to only change the icon color bug fixes, and added a new health icon. I removed the character filtering. I think this can be merged, although I still want to do some testing to #4108 before we merge that too.

gfwilliams:
Thanks! Looks great!

Issue #4099, opened 2025-12-19
Fastload Utils: Implement and strictly enforce whitelist 


thyttan:
> Thanks! Looks like it's because the messagegui's music page uses `layout` with a `remove` handler - which makes fastload think it can remove it and have everything unloaded. That's going to happen for any app that uses `remove` in a setUI/layout/etc and doesn't actually fully remove itself.
> 
> ... Perhaps what we need is actually a whitelist of apps that support fastload - because I feel like for pretty much every casual user, it's just a fantastic way to make your Bangle really unreliable.

_Originally posted by @gfwilliams in https://github.com/espruino/BangleApps/issues/4077#issuecomment-3675346908_

I imagine the whitelist to be hardcoded without being configurable by the user.

Ping @halemmerich as original author.

gfwilliams:
Yes - I guess either we just make an app set `Bangle.fastload=1` or we add something to the `app.info` file from the metadata?

Another thing I think we should check for is since we're loading the app file and checking for loadWidgets in it, we should also check if the file begins with `{` or `(function() {` since without it, it's pretty much guaranteed that the app will leak memory as it defines vars in the global scope. 

It wouldn't work well for non-pretokenised code, but if you're using fastload you probably care about speed enough to want pretokenising on anyway

gfwilliams:
Ok, so I did a bit of searching with a script to see how many apps you could use fastload without leaking memory (checking for loadWidgets, remove/uiRemove, and then manually checking correct scoping). So far I have:

* `apps/calendar/calendar.js` - maybe, but it doesn't unload the font
* `apps/gbdiscon/app.js` - fast loads (works without fast load utils)
* `apps/gpstrek/app.js` seems close but it never sets `uiRemove` so never fast loads 
* `apps/grocery/app.js` - fast loads
* `apps/heatsuite/heatsuite.app.js` - fast loads, but uses `function` so leaks memory
* `apps/messagelist/app.js` - fast loads I think (but has comments about not using Bangle.load because of debugging)
* `apps/multitimer/app.js` - fast loads, but uses `function` so leaks memory
* `apps/podadrem/app.js` - fast loads,
* `apps/qcenter/app.js` - fast loads,
* `apps/rebbleagenda/app.js` - fast loads (works without fast load utils)
* `apps/spotrem/app.js` - fast loads,
* `apps/stopwatch/stopwatch.app.js` - fast loads,
* `apps/torch/app.js` - fast loads (works without fast load utils)

[...]

PR #4114, opened 2026-01-05
Enable fast loading on apps that appear to support it

gfwilliams:
effectively whitelisting apps that can unload quickly - see #4099

@thyttan has checked these and there's one minor memory leak that's a firmware issue https://github.com/espruino/Espruino/issues/2674 so we'll ignore that for now


kidneyhex:
I think this breaks `qcenter` if there's a timeout set. I have no idea why.

1) Reinstall just the default apps 
2) Add the 0.06 qcenter (https://espruino.github.io/BangleApps/?id=qcenter)
3) Test qcenter and see it works
4) Go to settings and set the Timeout in qcenter to 30s
5) Test qcenter and it does not work

In the button event, it isn't doing anything. No exception, but also not loading the app. 

Broken: `cb: (l) => Bangle.load(app.src),`
Works: `cb: (l) => load(app.src),`
Works: `cb: (l) => setTimeout(() => Bangle.load(app.src), 0),`

https://github.com/espruino/BangleApps/pull/4114/changes#diff-ffdc5bdd15075e5282f6b67931a08ab4a47ed47002c93118aa49bfc0e2bf30b6R75

thyttan:
Hm, it seems to work for me when trying that.

What fw do you have on? I have fw 2v28.37 iflash build.

kidneyhex:
I was on the normal build of 2v28.37 when I first noticed, but updated to 2v28.38

gfwilliams:
Ok, thanks - well, I guess it doesn't hurt to add the setTimeout, so I've just done that and pushed it in.

I wonder whether it depends what app you're trying to fast load into? What were you trying @kidneyhex ?

kidneyhex:
For my normal setup: weather, multitimer, and gbdiscon 
And normally I have timeout at 30s.

When I tried with the simplified reinstall: it has settings and about as its hardcoded fallbacks and timeout off, so it works at first.

Eventually I found if I had set timeout to off it works, but any other value it just silently doesn't. 

Console.log still triggered if I added that to make sure it was getting touch events, but the Bangle.load just silently did nothing. But then plain load() or the setTimeout versions work for some reason... 🤷 

Just reinstalled after the latest commit and it's working.

kidneyhex:
@gfwilliams I forgot to add: I have not been using fastload, my attempts are all without fastload installed.

kidneyhex:
@gfwilliams @thyttan 

I can't get this to re-create in the emulator, but I can get it to happen on the watch with 2v27, 2v28, and 2v28.38. 

I put more `console.log`s in there and have a rough theory...

The timeout for leaving `qcenter` is cleared in two spots: 
A) in the Layout's `remove`
B) in a `Bangle.on("touch",updateTimeout)`. 

Neither one sets the variable `timeout = undefined` after `clearTimeout` though, so it could end up clearing the wrong timeout?

1) `qcenter` button event calls `Bangle.load(app.src)`
2) `Bangle.load` follows the `uiRemove` code path and calls `Bangle.setUI`
3) `Bangle.setUI` calls the Layout's `remove`
4) `remove` does a `if (timeout) clearTimeout(timeout);`
5) `Bangle.load` calls `setTimeout(eval,0,...)`
6) `Bangle.on("touch",updateTimeout)` fires
7) `updateTimeout` calls `if (timeout) clearTimeout(timeout);` -- this ends up being the wrong timeout to clear, so we never move to the next app
8) `setUI` has already removed the UI, so the app is now frozen


It would explain why my blindly doing a `setTimeout(()=>Bangle.load(app.src),0)` fixed it: with an extra setTimeout it gets around the timeout ID issue.


~~There also seems to be an existing issue where if the `updateTimeout` touch handler runs first, it isn't getting to the Layout's touch handler... or maybe~~ I am just terrible at hitting the buttons correctly.

thyttan:
>  I am just terrible at hitting the buttons correctly.

Have you calibrated the touch-screen?

kidneyhex:
Taps started acting normal again after I took my `console.log` statements out, so I don't know what was happening. 

I did make a version where I set `timeout=undefined;` and it's working, so I think that was the fastload problem.

thyttan:
Just a note on this I came to think about. Are we not a bit naive when calling `Bangle.load` to exit apps this way? I. e. we're not sure the next app (mostly clocks I assume) uses widgets.

So if we want to be safe we can do either of:
- only do this if widgets are not active
- use the logic in `launch_utils.loadApp`

Right?

gfwilliams:
Hmm - well, if we're calling `Bangle.load()` without arguments it's always going to go to `.bootcde`, which should check: https://github.com/espruino/BangleApps/blob/master/apps/boot/bootloader.js#L6

So I think we're ok

thyttan:
Ah - nice. Sorry for the noise!

thyttan:
Looking at that line, do we want to add to it like this?:

```
if (global.__FILE__ && (!!global.WIDGETS && !s.clockHasWidgets)) {load();throw "Widgets loaded and clock has no widgets, can't fast load";}
```

gfwilliams:
good idea! I realised you probably want `global.WIDGETS!==undefined` though (I chaged my last commit) as I think having WIDGETS defined is enough to reserve a 24px bar at the top which would upset any clocks that don't expect it.

apps/qcenter/app.js
```diff
@@ -72,7 +72,7 @@ let appButtons = groupBy3(pinnedApps).map((appGroup, i) => {
       pad: 5,
       src: require("Storage").read(app.icon),
       scale: 0.75,
-      cb: (l) => load(app.src),
+      cb: (l) => Bangle.load(app.src),
```
thyttan, line 75: I guess this case may not be entirely safe wrt widgets on next app though (the only unsafe use I see). So maybe we just change that back to `global.load` for now? Or we could pull in the `launch_utils` module.

gfwilliams: Yes, true - thanks! Using launch_utils would be an option but then we have to load the cache, which will take a while anyway. As you say using `load` would be more sensible

gfwilliams: Just done

PR #4115, opened 2026-01-05
Update txtreader to 0.03 - Ability to save pages

tonykakuuu:
The program now creates a new line in `txtreader.config.json` for every txt file it reads and saves the last viewed page there. Every flip to a new page updates this file. Pressing the top right corner of the screen now takes the user back to page one. 

Test & working, currently only one drawback: The user can't turn back pages beyond the page they continue on. Implementing this is not trivial because the application doesn't save every pages offset but I might implement it on a snowy weekend. 

For now, this lets users read longer text files without having to go through a gazillion pages every time.

### DIFF
--- a/apps/txtreader/README.md
+++ b/apps/txtreader/README.md
@@ -17,6 +17,10 @@ Bangle.js 2
 - tap the top right corner of the screen to go back to the first page
 - exit by pressing the physical button 
 
+## TODO
+
+- It is currently not possible to go backwards after continuing from a saved page. Instead, press the top right to go back to page one. 
+
 ## Creator
 
 <https://topkekker.rip/>


thyttan:
That sounds like a good feature for sure!

Am I right in seeing that it will interact with storage everytime a page is turned? I think it would be sensible to refactor to just do that once when the app is started and again when the app is closed (by acting on the `kill` event). That would lessen the wear on storage and be better for battery drain I believe (don't know to what extent). Would you like to do that before I look at merging? Otherwise it's fine to leave to a future version.

tonykakuuu:
Good call, I didn't know about the `E.on('kill', ...)` event but it works flawlessly, is more efficient, easier on the memory wear and needs fewer lines of code. So thanks for the hint!

thyttan:
> currently only one drawback: The user can't turn back pages beyond the page they continue on. 

Please just add a note about this in the readme and I'll merge. Thanks! :)

tonykakuuu:
Done!

thyttan:
Thanks for contributing! 🥳

PR #4122, opened 2026-01-11
btadv: permit more apps (HRM) to connect

bobrippling:
Some apps won't recognise a HRM advert if there's data attached - this "detaches" data from the advert


thyttan:
So the advert says "you can attach to me if you want data" and once attached it will start sending bpm's?

bobrippling:
Yes - a HRM advert can optionally advertise the HRM payload but this appears to cause Strava to ignore the device (and so it can't be paired with).

I'm going to have a little tinker before merging but I think this is the best way to have the app be compatible

bobrippling:
Yeah - gonna keep it as `undefined`, will merge unless you want to do the honours?

thyttan:
Feel free to go ahead! 🤓

PR #4124, opened 2026-01-12
New widget: hrmadv

bobrippling:
This lets users turn on HRM adverts from the widget (e.g. you're in runplus and want to also stream the HRM to your phone)

Also brought in `require("ble_advert").remove()` too

### DIFF
--- a/apps/widhrmadv/widget.ts
+++ b/apps/widhrmadv/widget.ts
@@ -2,6 +2,10 @@
 type WidHrmAdv = Widget & { userReq?: boolean };
 let serviceRegistered = false;
 
+const enum Consts {
+	WIDTH = 14,
+}
+
 Bangle.on('touch', (_btn, xy) => {
 	const oversize = 5;
 
@@ -11,7 +15,7 @@ Bangle.on('touch', (_btn, xy) => {
 	const x = xy!.x;
 	const y = xy!.y;
 
-	if(w.x! - oversize <= x && x < w.x! + 14 + oversize
+	if(w.x! - oversize <= x && x < w.x! + Consts.WIDTH + oversize
 	&& w.y! - oversize <= y && y < w.y! + 24 + oversize)
 	{
 		E.stopEventPropagation();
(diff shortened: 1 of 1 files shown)


bobrippling:
@gfwilliams small change to `ble_advert` if that's ok with you (tested - works for dropping the given service from adverts)

modules/ble_advert.js
```diff
@@ -86,3 +86,31 @@ exports.push = function (adv, options) {
...
+    if (manyAdv(bangle.bleAdvert)) {
+        var i = 0;
+        for (var _i = 0, _a = bangle.bleAdvert; _i < _a.length; _i++) {
+            var ad = _a[_i];
+            if (Array.isArray(ad))
+                continue;
+            if (ad[id]) {
+                delete ad[id];
+                var empty = true;
+                for (var _ in ad) {
```
gfwilliams: Is there not a tidier way of checking if it's empty? like `Object.keys(ad).length`?

edit: sorry - meant to post on the typescript, but it's the same there I think

bobrippling: Yes, sounds good - I was thinking the loop would be faster than a function call + object lookups, but that was just an old guess

gfwilliams:
Thanks - that looks great - good to merge if you're ready?

modules/ble_advert.js
gfwilliams: Ahh - I'm not sure actually. Although I guess this code isn't really on the 'critical path' - I think for most things, tidy, short and understandable wins (the less vars the interpreter has to look up and less code it has to parse the fastest it tends to be) and then if we find performance is bad we can tweak it later.

Honestly most of the performance issues with Bangle.js I've found tend to be the more coarse-grained stuff - loading files when we don't need to and stuff like that.

bobrippling: Makes sense, I made the mistake of making an "optimisation" without profiling first too, thanks for keeping me honest! And yes, this code is much nicer to read

bobrippling:
Thank you - I was giving the widget a try last night in an actual run and it did disappear off the widget list (I have `widclk` and the recorder widget there). So I might tweak the press detection so I can still toggle the BLE adverts even if the widget is pushed off screen, if that's ok

bobrippling:
Ok - all sorted. The reason it wasn't working wasn't because of a touch problem, it was because I'd not re-uploaded `ble_advert` to my watch. Doh. Good to merge

PR #4128, opened 2026-01-17
smpltmr: Fix trigger happy buzzing

thyttan:
Try the change on my app loader: https://thyttan.github.io/BangleApps/?id=smpltmr

Tagging @peerdavid and @sir-indy as previous authors for feedback.

### DIFF
--- a/apps/smpltmr/ChangeLog
+++ b/apps/smpltmr/ChangeLog
@@ -12,4 +12,5 @@
 0.12: Fix ClockInfo so it updates after leaving +5min/etc by blurring
       ClockInfo now shows time left in icon, updates more often the closer to the time
       Tapping clockinfo adds 1 min, added +30min option too
-0.13: Fix trigger happy buzzing.
+0.13: Fix trigger happy buzzing, from overlapping drag and touch events. Also
+      if a touch didn't actually cause action.


apps/smpltmr/ChangeLog
```diff
@@ -11,4 +11,5 @@
 0.11: Fix to handle updated firmware on fw 2v25 (or cutting edge >2v24.164), btn -> btnRelease. The timer would fire on long press before the watch reset.
 0.12: Fix ClockInfo so it updates after leaving +5min/etc by blurring
       ClockInfo now shows time left in icon, updates more often the closer to the time
-      Tapping clockinfo adds 1 min, added +30min option too
\ No newline at end of file
+      Tapping clockinfo adds 1 min, added +30min option too
+0.13: Fix trigger happy buzzing.
```
bobrippling:
```suggestion
0.13: Fix trigger happy buzzing from repeat touch events yielding the same point
```

bobrippling (commented):
LGTM! Will wait for the others you've tagged, but perhaps a small suggestion to explain what's causing the extra buzzing?

apps/smpltmr/ChangeLog
thyttan: Added a little to it :)

bobrippling: That's even better, great stuff!

peerdavid:
Just a quick comment - I had a really really great time developing for the BangleJs but I recently moved to Pebble (something I always wanted to try as well) Therefore, I'm no more developing for the BangleJs. So for me any improvements and fixes are fine but I can no more contribute :)

RKBoss6:
Tried it out and it looks good! One suggestion - currently when you drag, it increments in large values (drag once and it goes from 3--> 17). Is it possible for you fix this to be more like the show menu pickers where they increment predictably, as it'll tie into removing extra buzzes?

thyttan:
> One suggestion - currently when you drag, it increments in large values (drag once and it goes from 3--> 17). Is it possible for you fix this to be more like the show menu pickers where they increment predictably, as it'll tie into removing extra buzzes?

Yes, I added a comment to this commit https://github.com/espruino/BangleApps/pull/4128/changes/66e5e21be4a4e9d884804cf75f46a74ac010ea08

>I didn't manage to make an ideal solution. In my mind one step of either
hour, minute or second indicator should correspond to one buzz. One can
look to e.g. the dial module for inspiration. But I didn't want to
change the logic too much for now at least.

So while it shouldn't be too hard it wasn't trivial and requires a slightly bigger rewrite. Which I didn't want to do now in this PR. :) 

But I agree it would be nice 👍

PR #4129, opened 2026-01-17
Recorder: don't emit `(0, 0)` coordinates

bobrippling:
... as these make it look like I ran to the equator and back.

The crucial commit is the latest one, 3a5a1ada45c. Rest is scaffolding

- [x] Deploy and test


bobrippling:
Working well, merging

PR #4130, opened 2026-01-18
widhrmadv: wake NRF for adverts

bobrippling:
Otherwise the user has to have BLE turned on in settings in order for a device to then connect (once it notices the adverts)


bobrippling:
Tested on a run, worked a treat

PR #4131, opened 2026-01-18
runplus: don't infinitely resume a track during a day

bobrippling:
The "always resume" setting is intended to always resume during a single duration of the app, not across separate launches


bobrippling:
Tested on several runs, working great

PR #4135, opened 2026-01-20
[Control Pad] Refactor app for consistency and reliability, as well as UI updates

RKBoss6:
Refactors for better responsiveness, reliability, toggle behaviour, ui look and feel, etc. Look at the changelog for a full description. Also, tagging @bobrippling as the author of this app.

### DIFF
--- a/apps/ctrlpad/main.js
+++ b/apps/ctrlpad/main.js
@@ -325,7 +325,7 @@ var controls = [
       ui.ctrls.draw(ui.overlay.g2);
       
       // Force an update through the overlay
-      var y = g.getHeight() - ui.overlay.height; 
+      var y = g.getHeight() - ui.overlay.height;
       Bangle.setLCDOverlay(ui.overlay.g2, 2, y - 10);
       Bangle.buzz(10);
       


apps/ctrlpad/main.js
```diff
@@ -80,7 +106,7 @@
                     .setColor(ctrl.bg)
                     .fillCircle(ctrl.x, ctrl.y, 23)
                     .setColor(ctrl.fg)
-                    .drawString(ctrl.text, ctrl.x, ctrl.y);
+                    .drawImage(ctrl.img,ctrl.x-12,ctrl.y-12)
```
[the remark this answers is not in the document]
RKBoss6: I'd like to keep the text as an id for the object, but drawing it results in less space and a cramped feeling...

RKBoss6:
I want to add a screenshot, but I don't think `g.dump()` supports lcd overlays - any alternatives?

bobrippling:
Hey, this looks great! One problem however - the app is typescript. Before you go any further, would you be able to migrate your changes to `main.ts` ?

For the screenshot, you'll want to call `ui.overlay.g2.dump()` - here after `ui` is assigned, you can copy it to `global.ui` or similar:

https://github.com/espruino/BangleApps/blob/45936302cbb875139ed4215c0a5a37eae401fa27/apps/ctrlpad/main.ts#L247-L250

RKBoss6:
Thanks! Is there a way to generate typescript from the js code, or do I have to manually reimplement everything?

bobrippling:
> Thanks! Is there a way to generate typescript from the js code, or do I have to manually reimplement everything?

It's the other way round in fact, JS is generated from TS. You could try your hand at patching across but that's not likely to help much (`git diff | sed 's/\.js/\.ts/' | patch -p1`)

RKBoss6:
Hmm, I don't know typescript (at all), so I'm afraid I won't be of much use. Do you think it's possible for you to make those changes? The other option is I can ask an AI to see if that'll work, but I'm not sure about that either. Perhaps I'm misunderstanding something fundamental with TypeScript though, but I'm just not sure how to go about this...

(the `git diff | sed 's/\.js/\.ts/' | patch -p1` didn't work...)

apps/ctrlpad/metadata.json
bobrippling, line 1: Love the new look!

apps/ctrlpad/README.md
```diff
@@ -23,3 +22,8 @@ The control pad disables drag and touch event handlers while active, preventing
 
 - Handle rotated screen (`g.setRotation(...)`)
 - Handle notifications (sharing of `setLCDOverlay`)
+
+## Creator
+Bobrippling
+## Contributors
+RKBoss6
```
bobrippling: If you wouldn't mind:

```suggestion
[bobrippling](https://github.com/bobrippling)

## Contributors
[RKBoss6](https://github.com/rkboss6)
```

apps/ctrlpad/main.js
```diff
@@ -15,28 +15,45 @@
...
-                .setColor(g.theme.bg)
-                .fillRect(0, 0, this.width, this.height)
-                .setColor(colour.on.bg)
-                .drawRect(0, 0, this.width - 1, this.height - 1)
-                .drawRect(1, 1, this.width - 2, this.height - 2);
-        };
+          var border = 3;
+          this.g2
+              .reset()
+              .setColor(13)
```
bobrippling, line 36: What's the `13` all about here?

apps/ctrlpad/main.js
```diff
@@ -109,79 +135,71 @@
...
+  {
+    text: "DnD",
+    img:atob("GBiBAAAAAAAAAAA8AAAYAAAYAAD/AAH/gAH/gAP/wAP/wAP/wAP/wAP/wAP/wAP/wAf/4Af/4A//8B//+A//8AAAAAA8AAAAAAAAAA=="),
+    get: () => {
+      return require("Storage").readJSON("setting.json", 1).quiet === 1;
+    },
+    toggle: () => {
+      let s = require("Storage").readJSON("setting.json", 1);
+      s.quiet = s.quiet ? 0 : 1;
+      require("Storage").writeJSON("setting.json", s);
```
bobrippling, line 160: I think we should make this optional, I want the quiet mode to be temporary and also avoid the flash writes on each tap

apps/ctrlpad/main.js
```diff
@@ -109,79 +135,71 @@
...
+      //if qm widget is present, update that
+      if (typeof WIDGETS === "object" && "qmsched" in WIDGETS) WIDGETS["qmsched"].draw();
+      if (global.setAppQuietMode) setAppQuietMode(s.quiet);
+    }
+  },
+  {
+    text: "HRM",
+    img:atob("GBiBAAAAAA+B8B/D+D/n/H///v/////////P///P///H/3+WQAC2Xj82HB86uA/48Af54AP9wAH9gAD/AAB+AAA8AAAYAAAAAAAAAA=="),
+    get: () => Bangle.isHRMOn(),
```
bobrippling, line 169: This will mislead users - tapping the button will have no visible effect if another app/widget has enabled it (it'll remain on)

apps/ctrlpad/main.js
```diff
@@ -289,6 +307,30 @@
...
             touchDown = true;
     });
+    var onCtrlTap = function(ctrl) {
+      Bangle.buzz(20);
+
+      var result = ctrl.cb(true); 
+      if (result === "close") {
+        terminateUI();
+        return;
+      }
```
bobrippling, line 317: I think it's ok to have the callback terminate the UI itself, rather than spread the responsibility and allocate strings in the code

bobrippling (commented):
Thanks - I've moved the changes into `main.ts`, it doesn't compile yet but thought I'd comment a few thoughts before going further. Let me know what you think!

apps/ctrlpad/main.js
RKBoss6: ah- it was just a number in the palette that's used as a transparent background color for the corners. It's the best solution I was able to come up with, but if you have another way, I would prefer having a more elegant solution over this.

apps/ctrlpad/main.js
RKBoss6: My only gripe with that is when I first used the control panel, bluetooth wouldn't connect, and it took me a few days before I figured out that it was locally turning bluetooth off, and just wouldn't display that. I would prefer to have it global so all apps are affected but haptics still work, but possibly avoiding flash writes if we can.

apps/ctrlpad/main.js
RKBoss6: This was due to ui being undefined when it's called, because it's defined later in scope, so this just passes it down to a place where it can actually terminate the UI first.

apps/ctrlpad/main.js
bobrippling: I think there's a mix-up here? This is quiet mode, not bluetooth?

apps/ctrlpad/main.js
bobrippling: Gotya - I don't know of any way to set transparent otherwise. Do you have the source palette for us to see how likely it is that this will change?

apps/ctrlpad/main.js
bobrippling: Can you show me the code where that's the problem and the error output? These functions are all executed after they're all defined, so I don't see how it can be a scoping problem.

For it being a problem that `ui` isn't set (not a scope problem, just logic), then:

```javascript
var result = ctrl.cb(...); // if ui was `undefined` here then

if(result == "close"){
  terminateUI(); // there's no code between there and here which will make it defined
  return;
}
```

apps/ctrlpad/main.js
RKBoss6: Sorry, I was using it as an example. The point was I didn't want any local change to quiet mode when there's already a global implementation. Could we have a global variable used in messages instead that is saved on e. Kill? I think that would be a better solution for system-level quiet mode, as it avoids writes unnecessarily until E.kill is triggered, and can just be appended in bootloader for simplicity.

apps/ctrlpad/main.js
RKBoss6: I'm not sure what you mean by source palette :( If it helps, in the screenshot before I edited the layers on, the transparent color was shown as an orange..

apps/ctrlpad/main.js
RKBoss6: Hmm, how do you propose we go about fixing it? It has tripped me up when I used it, so is there a way to distinguish whether we triggered it, or if it's on in general? I suppose if we really want the 'control center feel' that we get on phones, it might be best to have it as it is now, as it just shows the state of the HRM and allows you to toggle...

apps/ctrlpad/main.js
bobrippling: Ah, I mean just where the 13 comes from, like where you found it - the image bytes themselves?

apps/ctrlpad/main.js
bobrippling: Ah I see - my use case is different - I want a global change to just buzz but I don't want quiet mode installed or to have it as a hard dependency of the app. Perhaps have the code check if quiet mode scheduler or similar is installed and enable the functionality in that case?

And yes, agreed - saving quiet mode setting in the `kill` event handler sounds good

apps/ctrlpad/main.js
bobrippling: There is - we can either have a boolean to keep track if we triggered it, or we delve into Bangle internals to find out, which is what the code does atm

apps/ctrlpad/main.js
RKBoss6: Sounds good! For the e.kill handler, we'll need a global variable to act in the file's place, which we have to create. Should we do that, and then update all the apps that use quiet mode? Alternatively, we could have a more system level one, and create a `Bangle.quietMode` variable for the system instead that achieves the same effect. @gfwilliams wdyt?

apps/ctrlpad/main.js
RKBoss6: No, like I said before, it's just a random number I chose, with no significance at all. That probably isn't the best way to go about something like that, but I didn't find another way :(
Do we need to change it/actually source the value from somewhere? It has worked for all I've been using it, you can upload the boot.js to your watch as well if you need to test it.
(I could still be misunderstanding what you're trying to ask, I'm sorry for the confusion)

apps/ctrlpad/main.js
RKBoss6: Gotya, let me look and see how it's done now to see if we can add that functionality in!

apps/ctrlpad/main.js
gfwilliams: I really don't like the idea of starting to mess with other apps and adding global variables just to deal with this case. So the reason you want to save is because the messages lib looks at `quiet` in `setting.json`?

It looks like you and @bobrippling want different things, and as he says maybe you need it to be an option.

Either:

* You have a temporary change which you can do by just overwriting the `.buzz` function (**although if `quiet` was set already in settings, you won't ever be able to unset it by changing .buzz**)
* You change the global setting by writing to `setting.json` which I think is the best bet - I wouldn't worry about the `kill` handler (as then you've got these two ways of disabling buzzing which are overlapping). You're only writing in response to a user input which you'll do at most a few times a day, so I don't think this is a big deal.

I think probably the best bet it to switch to just changing `.quiet`, and then if it's temporary (as bob would like) in a `kill` handler you set the setting back to what it was before.

Interestingly internally Bangle.js actually reads the `settings` file each time an app is loaded and looks at the value of `vibrate`, and if so disables `JSBF_ENABLE_BUZZ` which stops vibration regardless of what you do to `.buzz` or what `quiet` is set to. Really, longer term we should probably just expose the ability to set that flag and delete all the checks for `.quiet`.

I guess `quiet` was added rather than changing `vibrate` because someone still wanted the haptics even when quiet, but maybe we could have a `Bangle.haptic` that the built-in menus use, and then that gives us a bit more flexibility (to disable haptics as well as making them stronger/weaker).

RKBoss6: That sounds good! As for this however, I can make it a setting and use that to decide whether we want buzzes at all, or only quiet mode for messages, etc. It might take some time though, as I'm a bit busy now, but I'll try to get it done soon.

apps/ctrlpad/main.js
bobrippling: Sounds good - the code had the checks in, you could pinch that:

https://github.com/espruino/BangleApps/blob/1782cedaaffb04118fa983e1f024ae9974e88621/apps/ctrlpad/main.ts#L220-L221

apps/ctrlpad/main.js
bobrippling: Ah, I see! I think 13 is ok then, let's make it clear in a comment that it's a somewhat arbitrary variable chosen out of nowhere in particular :)

PR #4137, opened 2026-01-21
boot: allow fast load to clock if WIDGETS is not defined.

thyttan:
As discussed in https://github.com/espruino/BangleApps/pull/4114#issuecomment-3778294375.


gfwilliams:
Great - thanks!

PR #4141, opened 2026-01-25
widbtstates: sync on `draw()`

bobrippling:
The widget appears even when bluetooth is off sometimes, this syncs when we draw

### DIFF
--- a/apps/widbtstates/ChangeLog
+++ b/apps/widbtstates/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New App!
 0.02: Improve generated JS
+0.03: Recheck state on draw(), to get in-sync
--- a/apps/widbtstates/metadata.json
+++ b/apps/widbtstates/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "widbtstates",
   "name": "Bluetooth States",
-  "version": "0.02",
+  "version": "0.03",
   "author": "bobrippling",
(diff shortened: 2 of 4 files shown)


thyttan:
I have a similar thing happen with `widbt_notify`. I opened this PR but closed it as I thought the logic already looked good. But the delayed update of widget showing new bt state persists.

https://github.com/espruino/BangleApps/pull/3735

bobrippling:
> I have a similar thing happen with `widbt_notify`. I opened this PR but closed it as I thought the logic already looked good. But the delayed update of widget showing new bt state persists.
> 
> #3735

Interesting, thanks for the reference. I'm currently testing this MR, so I'll see if the bug does reappear. If not, we know that the problem is something todo with either the listeners on `NRF` or that `wake()` / `sleep()` aren't the only ways that BLE advertising is enabled/disabled

(I suspect that is the case - likely we may need to hook `setAdvertising` or similar)

bobrippling:
Seems to have sorted - we could merge this now and refine further, wdyt?

thyttan:
I'm fine with merging if it improves the app :)

I don't really understand what the changed code does differently? Could you help me understand so I can fix the other app as well?

Is this more a workaround or actually fixing the problem? I.e. should we really do something differently somewhere else?

thyttan:
Ah, sorry. I see. It makes sure to check the state before each draw.

bobrippling:
Yeah exactly - I'll merge and see if I can think of anything better. Thanks!

PR #4144, opened 2026-01-27
[iOS] Fix bug

RKBoss6:
Fixes a bug where a space is added on to the auto-detected name, causing the icons to not be shown accordingly


bobrippling:
LGTM! But tagging @gfwilliams with this being a core app

gfwilliams:
Looks good - thanks!

PR #4145, opened 2026-01-28
adaptive heart rate monitor app 

antimo0f:

### DIFF
--- a/apps/ahrm/adaptivemode.js
+++ b/apps/ahrm/adaptivemode.js
@@ -1,96 +1,114 @@
-class UintRecorder{
-    constructor(dim,type){
-      if(type==8){
-        this.array= new Uint8Array(dim);
-      }else if(type==32){
-        this.array=new Uint32Array(dim);
-      }
-      this.dim=dim;
+class UintRecorder {
+  constructor(dim, type) {
+    if (type == 8) {
+      this.array = new Uint8Array(dim);
+    } else if (type == 32) {
+      this.array = new Uint32Array(dim);
     }
-    shift(){
+    this.dim = dim;
+  }
+  shift() {
     //Shifts all elements one position to the left, freeing up the last spot when the array is full.
-      for(let i=1; i<this.dim; i++){
-        this.array[i-1]=this.array[i];
-      }
-    } 
-    push(value){
+    for (let i = 1; i < this.dim; i++) {
+      this.array[i - 1] = this.array[i];
+    }
+  }
+  push(value) {
     //If the array is full, it shifts and then inserts the value (losing the first element).
-        this.shift();
-        this.array[this.dim-1]=value;
-    } 
-    print(){
-     //Prints the array for debugging purposes.
-      for(let i=0;i<this.dim;i++){
-        console.log(this.array[i]);
-       }
+    this.shift();
+    this.array[this.dim - 1] = value;
+  }
+  print() {
+    //Prints the array for debugging purposes.
+    for (let i = 0; i < this.dim; i++) {
+      console.log(this.array[i]);
     }
-    //Returns the average of the elements, excluding zero values.
-    avg(){
-      var sum=0;
-      var n=0;
-      this.array.forEach((element)=> {sum=sum+element;
-                                        if(element!=0){n=n+1;}
-                                     });
-      return sum/n;
-    }
-    //Returns the standard deviation of the elements, excluding zero values.
-    sd(){
-      var avg= this.avg();
-      var sd=0;
-      var n=0;
-      this.array.forEach((element)=>{if(element!=0){
-                                     sd=sd+Math.pow(element-avg,2);
-                                     n=n+1;
-                                     }
-                                      });
-        return Math.sqrt(sd/n);
-    }
+  }
+  //Returns the average of the elements, excluding zero values.
+  avg() {
+    var sum = 0;
+    var n = 0;
+    this.array.forEach((element) => {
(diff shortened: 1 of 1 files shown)


RKBoss6:
This looks really cool! One question - does this modify the existing heart rate function, so if another app calls `Bangle.setHRMPower`, will it use this, and return the value in firmware-level events, like the `health` event?

antimo0f:
> This looks really cool! One question - does this modify the existing heart rate function, so if another app calls `Bangle.setHRMPower`, will it use this, and return the value in firmware-level events, like the `health` event?

Thank you for your answer, this code does not modify or overwrite the native Bangle.setHRMPower function.
It does not affect the core Espruino functionalities. 
The firmware will continue to handle its standard events exactly as it did before. Consequently, the values returned in firmware-level events (like the health event) will not be altered by the script.

I’m very open to receiving other comments or suggestions! 
I plan to improve this over time and add new features, including a dedicated UI.

RKBoss6:
Is it possible to modify the hrm to use this? From what I can tell, this looks like it's another (maybe better) hrm algorithm, so it would make sense to have this modify the existing hrm... but correct me if I'm wrong about anything!

Also, you don't need to add more changelog entries if it's all in the first pr. Since this will be the first pr update on this app, all the changes can be bundled into 0.01 :)

apps/ahrm/adaptivemode.js
```diff
@@ -0,0 +1,233 @@
...
+    mode_prec=0;
+    mode=0;
+    changed=false;
+}
+
+function HRMPOWER(){
+    Bangle.setHRMPower(1,"ahrm");
+    start_time=new Date();
+    setTimeout(function(){
+        if(Bangle.isHRMOn&&Math.abs(new Date()-start_time)>=69*1000){ //turns off after one minute from power on
```
bobrippling: Perhaps you meant to call this function?

```suggestion
        if(Bangle.isHRMOn()&&Math.abs(new Date()-start_time)>=69*1000){ //turns off after one minute from power on
```

antimo0f:
Yes! the algorithm surely decides to switch on and off the hrm  in a smarter way. I originally designed this app to function primarily as data logger (saving to CSV). The main goal was to move away from fixed-interval measurements and instead use a 'smart' approach—triggering measurements only when it is actually useful to have the heart rate data, i tested it for a long time while i was writing my thesis at university, and it worked !

- I thought it would be better to keep the algorithm logic separate and use the standard APIs, rather than doing a direct override. I want to avoid forcing other apps that rely on the HRM API to adopt my specific sensor switching logic, as that might interfere with their functionality.

- If setHRMPower(0) and setHRMPower(1) are called too frequently from outside, the logic technically still functions, but the timing at which measurements arrive in the data structures (used to decide when to switch the sensor) will no longer match the intended period. This would break expectations for other apps. For example, if an app calls setHRMPower(1), it expects to receive an HR event immediately. However, that event might never arrive because in that case my algorithm has decided to only trigger a measurement every 10 minutes!

The answer is, Yes  it is possible to use it to override the original HRM, but  after you should consider the differences from the default one. The logic is explained in the Readme file.

Let me know your thoughts on this!

Also, I’ve adjusted the changelogs as requested. I'll bundle all the changes into the initial 0.01 release." ;)

apps/ahrm/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+### How to Access Data
+
+Since the app has no graphical interface, to view the data you need to connect the Bangle.js to your computer via the **Web IDE** or **Bangle.js App Loader** and download the `.csv` files from the device's storage (in the "Storage" section).
+
+## Settings
+
+The only available setting allows you to enable or disable data recording.
+
+*   **Path**: It is located in the main Bangle.js menu, under `Settings -> App Settings -> Health`.
```
thyttan: 'Health' is a typo - should be your app name from looking at the files here, right?

antimo0f: it was the old name of the app, i updated it now !

bobrippling (commented):
Thanks for the app! A few comments

apps/ahrm/adaptivemode.js
antimo0f: Yes, that is correct. I intended to call the function.

bobrippling:
Thanks for the changes, LGTM - @thyttan ok to merge if you're happy?

thyttan:
I'd suggest maybe using some auto-formatter to make the code more in line with https://www.espruino.com/Code+Style ?  (mostly wrt white space) 

Other than that, if it's a go from you it's fine with me :)

I only did the odd comment some weeks ago. Didn't look closer, sorry. 🥲🙂

antimo0f:
Thanks for the review. I've just run the code through Prettier to match the Espruino Code Style (2 spaces, 1TBS, etc.) and pushed the updated formatting. Let me know if everything looks good now!

bobrippling:
Great stuff, thanks! One thing we could check/lean on in future, if you were interested, is that this works with the [recorder](https://banglejs.com/apps/?id=recorder) and we could defer to its writing of csvs (so we can use all the charts etc that come with that)

antimo0f:
Thanks! That's a really interesting suggestion. I'll take a look at the recorder app and see how i can do that for a future update


> Great stuff, thanks! One thing we could check/lean on in future, if you were interested, is that this works with the [recorder](https://banglejs.com/apps/?id=recorder) and we could defer to its writing of csvs (so we can use all the charts etc that come with that)

PR #4147, opened 2026-01-31
patch health date parsing in <utc-0

KTibow:
in all timezones with less than a 0 shift from utc, the current health data downloader displays september twice and december never. this is because of how Date works:
```ts
> new Date("2025-9")
< Mon Sep 01 2025 00:00:00 GMT-0700 (Pacific Daylight Time)
> new Date("2025-10")
< Tue Sep 30 2025 17:00:00 GMT-0700 (Pacific Daylight Time)
```
this fixes it by explicitly constructing a date


gfwilliams:
Thanks - great spot! I've just changed the variable names to make it a bit clearer, but merging now!

PR #4142, opened 2026-01-25
Create Calorie Tracking app w/ health integration

RKBoss6:
Creates an app for calorie tracking that resolves #4092, that counts calories burned in a day and splits it by BMR, Active, and total calories. Uses MyProfile to get user health data, and hooks onto health events for new data. Also modifies myprofile to use RHR readings as an additional piece of data, with its own measurement UI.

Check it out at [my app loader](https://rkboss6.github.io/BangleApps/?id=calories)

Issue #4092, opened 2025-12-14
Burnt calories support?


GiornoGiovanna2605:
Hi y'all, have just a small question: is there any app that can stimate the amount of calories burnt based from the amount of steps in the bangle js 2? It's not a big deal for me, though it would be a good feature for runners maybe.
Thanks in advance for any answers! ^^

bobrippling:
I expect we could get a rough estimate if a user were to enter their weight and perhaps other details (maybe via [myprofile](https://github.com/espruino/BangleApps/blob/e517b27a107d1ad2e3db6df20ad43c831d357336/apps/myprofile/README.md)). @gfwilliams is this something that's come up before?

thyttan:
There's a previous discussion here: https://github.com/orgs/espruino/discussions/6359

Calories mentioned on the forum: https://github.com/orgs/espruino/discussions/categories/bangle-js?discussions_q=is%3Aopen+category%3ABangle.js+calories

gfwilliams:
Thanks @thyttan! Looks like there is already weight in `myprofile` so it could be pretty easy to add.

I wonder where it would make sense to do it? As its own app? Or I guess the `health` app could have an option to show it?

GiornoGiovanna2605:
Yea maybe the health app could be a good option

RKBoss6:
I would also like to see this happen, maybe by coupling steps, hrm and movement?

This link came up in a search that has a formula for calories: [here](https://www.nasm.org/resources/calorie-calculator?srsltid=AfmBOooJ4NwSXvNvoVJ5QqQ5w9PSiAb_JhE8JBdG-jIcd2ET5DCDZJRb).
It looks like you first need to calculate basal metabolic rate from height, weight, age, and gender, then determine the activity per day/interval, and update it accordingly

RKBoss6:
>I wonder where it would make sense to do it? As its own app? Or I guess the health app could have an option to show it?

I think having it as a menu option in health will help, that points to the myprofile page.


I've researched, and have narrowed it down to a set of calculations for bangle.js:
1. get the user's resting heart rate (this can be done by the user manually with the health app getting a user-triggered reading with the user being still)
2. get user's weight
3. every health interval, calculate the HRR ( (current HR-rest HR)/(max HR - rest HR) )
4. get the coefficient of energy from that reading ( 10*HRR+1)
5. Use the HRR coefficient and average that with the calculated coefficient of energy from steps (0.04*steps/min)+1.25
[...]

PR #4148, opened 2026-02-02
[Clock Info] Add health category setting

RKBoss6:
Adds a setting for whether or not health clockInfos (steps, hrm) should be moved to a `Health` list. This is from the discussion at #4092 and is related to PR #4142. The default is false.


gfwilliams:
Thanks! Looks good to me!

I just changed the `Default:` text back to `default=` in the Readme as you'd left at least one other line in there with `default=` - better to keep them all in the same style.

Issue #88, opened 2020-01-31
Don't enter menu if we've just powered on


gfwilliams:
We changed to starting the menu on button release (not press) - but that means that now the press of the button turns the LCD on, and the release opens the menu.

Previously it worked so that when in clock you could press any button to wake Bangle.js up (including the middle one) but it wouldn't go direct to settings. Now it's broken and you usually end up entering settings pretty often by accident

gfwilliams:
fixed with https://github.com/espruino/Espruino/commit/9105cd73a5d4d09a28e9a6974e999bd27dc2273b

PR #4150, opened 2026-02-03
App Loader: Add dropdowns for sorting and filtering

RKBoss6:
This stems from the discussion at [#7920](https://github.com/orgs/espruino/discussions/7920), where I had asked about having dropdowns for sorting and filtering, rather than a bunch of chips before. They got pretty messy with such a huge load of them on the screen, particularly for phones, so I made dropdowns instead. It's up on [my app loader](https://rkboss6.github.io/BangleApps/), and I have screenshots below. Any suggestions/feedback is very appreciated, check it out!
There is a JS component in appLoaderCore, which I'll try to get a PR for soon! Until then, I'm hoping we can discuss to make something everyone enjoys using.

### Before:
Screenshot 2026-01-22 at 5 54 02 PM
 

Screenshot 2026-01-22 at 5 54 40 PM

### After:

Screenshot 2026-02-02 at 9 35 37 PM

Screenshot 2026-02-02 at 9 36 12 PM


bobrippling:
Very nice!

css/main.css
```diff
@@ -330,6 +353,22 @@
...
+      gap: 0.75rem;
+      flex-wrap: wrap;
+      justify-content: center;
+      margin-left:0.7em;
+      margin-right:0.7em;
+    }
+
+    #searchform {
+      flex: 1;
+      min-width: 450px !important;
```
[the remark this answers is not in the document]
RKBoss6: Tested already, works well

gfwilliams:
I think we need to wait for https://github.com/espruino/EspruinoAppLoaderCore/pull/87 to go in before we merge this? As we're just putting this in, please could you change `<a sortid="explore">Random</a>` to `<a sortid="random">Random</a>` (and the relevant bits in the EspruinoAppLoaderCore pr?)

... also the EspruinoAppLoaderCore pr still needs the indentation changing and that `'chips' is not defined` error fixing

index.html
```diff
@@ -96,17 +145,9 @@
           <label class="chip hidden tooltip" sortid="installs" data-tooltip="Most installed by users">Installed</label>
```
gfwilliams: We should really remove this rather than just commenting it out

index.html
```diff
@@ -61,17 +61,66 @@
...
+            </ul>
+          </div>
+          <div class="dropdown sort-nav">
+            <a href="#" class="btn btn-link dropdown-toggle" tabindex="0">
+              <span>Sort by:</span><i class="icon icon-caret"></i>
+            </a>
+            <!-- menu component -->
+            <ul class="menu">
+              <li class="menu-item"><a sortid="">None</a></li>
+              <li class="menu-item"><a sortid="created">New</a></li>
```
gfwilliams: The idea of these is that `New/Updated` are hidden until we actually have the data to make them work - in the one you commented out `<label class="chip hidden tooltip" sortid="created"  data-tooltip="Most recent apps">New</label>`you can see it says `hidden`

But at least in your app loader (https://rkboss6.github.io/BangleApps) this is now broken - `appdates.csv` can't be found but the menu items are still shown anyway, they just don't work.

I'm also a bit concerned that we had tooltips on all the items before so that people could hover and see what they actually meant - but you've deleted those?

RKBoss6:
Finally gave a chunk of my time to polish this up, I've added back all the tooltips and hiding of the sorts that have no data (modified/created), and I'd say this is good to go. I know you're on vacation Gordon, so feel free to look at this whenever you have time. I would like to get in [#88](https://github.com/espruino/EspruinoAppLoaderCore/pull/88) for AppLoaderCore first though, and then we can sort this out and get it merged in nicely ;)

gfwilliams:
Thanks! Just checking - this isn't up on https://rkboss6.github.io/BangleApps/ right now, because it's instead got https://github.com/espruino/EspruinoAppLoaderCore/pull/88?

RKBoss6:
Yup! I can put this up, as I have to tweak that PR

gfwilliams:
No rush - as you said, let's sort out #88 first

RKBoss6:
Alright, I've put it up now if you want to take a look!

index.html
```diff
@@ -234,7 +232,7 @@ <h3>Device info</h3>
     <script src="core/lib/marked.min.js"></script>
     <script src="core/lib/espruinotools.js"></script>
     <script src="loader.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/jszip/3.7.1/jszip.min.js"></script> <!-- for backup.js -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jszip/3.7.1/jszip.min.js"></script>
```
gfwilliams: it's helpful having comments like `for backup.js` in here - otherwise we have no idea why we're including this script

index.html
```diff
@@ -210,20 +212,16 @@ <h3>Device info</h3>
...
       </div>
     </footer>
 
     <script src="webtools/uart.js"></script>
     <script src="core/js/utils.js"></script>
     <script>
-      UART.ports = UART.ports.filter(e => e.includes("Bluetooth")); // all watches are Bluetooth
-      Const.CONNECTION_DEVICE = "Bluetooth"; // force Bluetooth because we know (don't look it up)#
-      // If "?dev=Bangle.js abcd" specified in URL, filter by that name
+      UART.ports = UART.ports.filter(e => e.includes("Bluetooth"));
```
gfwilliams: Same here - I don't think we should be removing comments?

index.html
```diff
@@ -56,53 +48,63 @@
...
-          <label class="chip tooltip" filterid="fonts" data-tooltip="Extra fonts for non-latin languages">Fonts</label>
-          <label class="chip tooltip" filterid="favourites" data-tooltip="Apps that you've liked ❤️">Favourites</label>
+        <div class="dropdowns-pill">
+          <div class="dropdown devicetype-nav">
+            <a href="#" class="btn btn-link dropdown-toggle" tabindex="0">
+              <span>All apps</span><i class="icon icon-caret"></i>
+            </a>
+            <ul class="menu">
+              <li class="menu-item"><a>All apps</a></li>
+              <li class="menu-item"><a dt="BANGLEJS">Bangle.js 1</a></li>
```
gfwilliams: I actually made a change since your initial PR where the contents of this are pulled from the JS: https://github.com/espruino/BangleApps/blob/master/loader.js#L161-L164

It allows me to add Bangle.js 3 support via a config option so I can test stuff without confusing everyone before the release.

... so we'd need to slightly tweak those lines linked above, and remove the explicit Bangle.j2 lines here.

gfwilliams:
This looks great! Just added some comments above, but it's very minor.

and for my reference: this also needs https://github.com/espruino/EspruinoAppLoaderCore/pull/87

thyttan:
I also really like this!

Two things I'd prefer to be tweaked (but no big deal if not):
- when one of the three menus are open and I touch it's header again I'd like for it to close the menu.
- I think it would look nicer if the categories menu did not display the pop-up tip beside the selected entry (also not all of them fit on my phone's screen). Although they are fun and a little quirky!

This is from testing on an Android phone.

What do you two think?

RKBoss6:
Alright, I've made it so menus close when you tap the header, and put it up on my app loader - I'll still work on this though, as I know there's some reviews left.

thyttan:
> I've made it so menus close when you tap the header,

Cool! But at least on android I need to tap twice on the header to dismiss where I'd expected once to be enough. Tried this in Brave browser and Firefox mobile.

thyttan:
I noticed the changes does not seem to carry over fully to the android.html version yet.

See https://rkboss6.github.io/BangleApps/android.html

Compare:

gfwilliams:
> I noticed the changes does not seem to carry over to the android.html version yet.

Good point! Yes, `android.html` needs changing too - you'll just have to diff the two and copy the changes.

Also something looks a bit screwy on desktop now - pretty sure it didn't look like that before!

image
 
>  I've made it so menus close when you tap the header

The dropdown menus? Seems like a good idea!

RKBoss6:
> Also something looks a bit screwy on desktop now - pretty sure it didn't look like that before!

Should be fixed now! I'll get android.html changed as well. Just out of curiosity, what does android.html do? Is it a gadgetbridge override for connection on load or something similar?

index.html
RKBoss6: Fixed!

gfwilliams:
Thanks!  Yes, it's what Gadgetbridge loads for its embedded app loader. Some options are removed/added, but also we change the scaling and have some extra code in there which handles sending all the IO to the Bangle through the Android app rather than using Web Bluetooth

thyttan:
The 'My apps' and 'More' screens seem broken:

RKBoss6:
Should be fixed now! 😅

thyttan:
Yes - but not on the /android.html version 😉

RKBoss6:
Ah yeah - fixed it as well - and I pushed the overall android changes from my demo branch to this branch as well, I had forgotten to do that originally

thyttan:
Those pages look good on my end now for both versions :)

thyttan:
> > I've made it so menus close when you tap the header,
> 
> Cool! But at least on android I need to tap twice on the header to dismiss where I'd expected once to be enough. Tried this in Brave browser and Firefox mobile.

This is still the case on Android for me. I see on my laptop it works as I expect - one click on header to open the menu and then the next click on header closes it.

But on Android it's currently (both the android.html and standard version): 
- one touch on the header to open the menu,
- a second touch of the header does nothing,
-  and then a third touch of the header closes the menu.

RKBoss6:
Hmm, if it works fine on desktop, it might be a quirk with how phones handles interaction from touch, as it just uses the default button characteristics in the code. It looks like it taps once to open dropdown, second time deselects what was pressed in dropdown, and the third time closes it - I'm not sure if we really can do anything about that if that is the case, as it's a phone behavior to make websites compatible with hover/touch features. But you can also tap outside the dropdown to close it, and as you say, tapping twice still gets the expected result so it should be fine... 
It works fine on desktop for me, but doesn't work on phone as well. Perhaps we leave this for a separate PR?

thyttan:
Yes, I won't say it should be a blocker against merging. It would have been nice if it was achievable though :) Thanks for taking it this far!

thyttan:
> But you can also tap outside the dropdown to close it

The UI problem with this is one may happen to click a hyperlink (or something else) that was sitting in the background and opening it without meaning to. It's not a super big problem but annoying if it happens.

RKBoss6:
Yeah, that's a thought I had as well. I'd say we try to get this PR merged, and then we can work on a dedicated PR to help out here :)

android.html
```diff
@@ -128,27 +128,28 @@ <h2>App Loader</h2>
...
+          <button class="btn tooltip" id="installfavourite" data-tooltip="Delete everything, install your favourites">Install favourite apps</button>
+          <button class="btn tooltip" id="defaultbanglesettings" data-tooltip="Reset your Bangle's settings to the defaults">Reset Settings</button>
         </p><p>
-           <button class="btn tooltip" id="removeall" data-tooltip="Delete everything, leave it blank">Remove all Apps</button>
-           <button class="btn tooltip" id="reinstallall" data-tooltip="Re-install every app, leave all data">Reinstall apps</button>
-           <button class="btn tooltip" id="installdefault" data-tooltip="Delete everything, install default apps">Install default apps</button>
-           <button class="btn tooltip" id="installfavourite" data-tooltip="Delete everything, install your favourites">Install favourite apps</button>
-           <button class="btn tooltip" id="defaultbanglesettings" data-tooltip="Reset your Bangle's settings to the defaults">Reset Settings</button>
+          <button class="btn tooltip" id="installappfromfiles" data-tooltip="Install an app by selecting its files">Install App from Files</button>
```
gfwilliams: I'm pretty sure this one (installappfromfiles) shouldn't be in the android html as it doesn't work.

Also `webideremote` **should** be in there and has been deleted.

android.html
```diff
@@ -160,6 +161,10 @@ <h3>Settings</h3>
             <input type="checkbox" id="settings-settime">
             <i class="form-icon"></i> Always update time when we connect
           </label>
+          <label class="form-switch">
```
gfwilliams: Bluetooth Compatibility mode shouldn't be in here as it's implemented in Gadgetbridge

android.html
```diff
@@ -184,6 +189,10 @@ <h3>Settings</h3>
               <input type="checkbox" id="settings-alwaysAllowUpdate">
               <i class="form-icon"></i> Always show "reinstall app" button <i class="icon icon-refresh"></i> regardless of the version
             </label>
+            <label class="form-switch">
```
gfwilliams: This shouldn't be in here either as we don't support the emulator in Gadgetbridge as far as I know

gfwilliams:
Thanks - it's looking good! Please can you sort out the comments on https://github.com/espruino/BangleApps/pull/4150/changes#r3379415116 though

Mainly as I'd said in https://github.com/espruino/BangleApps/pull/4150#issuecomment-4632910665 some features are added/removed when running in Gadgetbridge, so we need to make sure those stay the same as they were.

index.html
RKBoss6: Good to go!

RKBoss6:
All the buttons and discrepancies with `android.html` should be fixed now - I also added back those comments that I think were removed accidentally during a merge. I also changed the `android.html` button code's indents, as the old version had one too many indents.

gfwilliams:
Great, thanks! Merging now, and I'll pull in the new core

thyttan:
EDIT 2: I spoke to soon - I hadn't updated submodules on my computer. Sorry for the noise. It seems to work as expected.

~~I think we broke the app loader if served from localhost, in that the `Library` page no longer shows any apps.~~

~~If I revert to 9ad0c90cd1f7f7133e7b9d060527c4563d89d2ed then serving on localhost seems to work fine. But current master at f37d8301471f46c50c215d63c00ac43aca735ada shows no apps under `Library`. Also if I checkout RKBoss6/dropdown-chips-final I get the same thing.~~

~~The dev app loader at espruino.github.io/BangleApps seems to work as it should though.~~

EDIT: Or, and I will go look now, I may need to update submodules on my maching. Will report back. (Yes, that was it)

gfwilliams:
Yes, confirmed here - all seems to work fine, although it looked a bit odd before I hit the 'refresh' button

PR #4152, opened 2026-02-03
Fix checkMemoryUsage to detect leaks only (allow lower memory)

elijahr:
Changed from exact equality to <= comparison. The test's purpose is to detect memory leaks (memory going up), not to enforce exact memory usage. Lower memory on subsequent loads indicates no leak and should pass.


gfwilliams:
Thanks!

PR #4153, opened 2026-02-04
[Message GUI] Add display for time if message has date value

RKBoss6:
Adds a display for the time ago from a notification (15m ago, 3h 24m ago), using the `date` parameter when pushing a message. If there is no date, the view is the same as the past version.
I updated the screenshot accordingly to match the changes.

### DIFF
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -439,11 +439,12 @@ function showMessage(msgid, persist) {
   let drawDate=function(r,rd,ld){
    // draws time ago. If there's nothing in one of the columns, take that space.
     g.setColor(g.theme.fg).setBgColor(g.theme.bg).setFont(srcFont);
-    if ((rd&&ld)||(!rd&&!ld))
+    if ((rd&&ld)||(!rd&&!ld)) {
+      let m = g.stringMetrics("--"+date), x = r.x+(r.w/2);
       g.setFontAlign(0,-1)
-       .clearRect(r.x+(r.w/2)-(g.stringWidth("--"+date)/2),r.y+2,r.x+(r.w/2)+(g.stringWidth("--"+date)/2),r.y+g.stringMetrics(" "+date).height)
-       .drawString(date,r.x+(r.w/2),r.y+5,1);
-    else if (!rd)
+       .clearRect(x-(m.width/2),r.y+2,x+(m.width/2),r.y+m.height)
+       .drawString(date,x,r.y+5,1);
+    } else if (!rd)
       g.setFontAlign(1,-1).drawString(date,r.x+r.w-6,r.y+5);
     else if (!ld)
       g.setFontAlign(-1,-1).drawString(date,r.x+6,r.y+5);


apps/messagegui/app.js
```diff
@@ -120,6 +120,28 @@ function addDragHandlerToChangeMessage(idx, scroller) {
...
+    var h = Math.floor(totalMin / 60);
+    var m = totalMin % 60;
+    return h + "h " + m + "m ago";
+  }
+
+  // Days and hours
+  var d = Math.floor(totalMin / 1440);
+  var h = Math.floor((totalMin % 1440) / 60);
+  return d + "d " + h + "h ago";
+}
```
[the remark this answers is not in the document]
RKBoss6: The time<1 will handle this.

apps/messagegui/app.js
```diff
@@ -404,6 +432,9 @@ function showMessage(msgid, persist) {
     };
     rowLeftDraw = function(r) {g.setColor("#f00").drawImage(atob("PhAB4A8AAAAAAAPAfAMAAAAAD4PwHAAAAAA/H4DwAAAAAH78B8AAAAAA/+A/AAAAAAH/Af//////w/gP//////8P4D///////H/Af//////z/4D8AAAAAB+/AfAAAAAA/H4DwAAAAAPg/AcAAAAADwHwDAAAAAA4A8AAAAAAAA=="),r.x+2,r.y+2);};
   }
+  let drawDate=function(r){
+  g.setColor(g.theme.fg).setBgColor(g.theme.bg).setFont(srcFont).setFontAlign(0,-1).clearRect(r.x+(r.w/2)-(g.stringWidth("--"+date)/2),r.y+2,r.x+(r.w/2)+(g.stringWidth("--"+date)/2),r.y+g.stringMetrics(" "+date).height).drawString(date,r.x+(r.w/2),r.y+3,1);
```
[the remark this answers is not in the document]
RKBoss6: I'm going to keep this in for brevity, and to avoid using more variables than we need to for an already complex app.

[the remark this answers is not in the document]
RKBoss6: works well, has been tested without any issues.

apps/messagegui/app.js
```diff
@@ -346,7 +370,12 @@ function showMessage(msgid, persist) {
   if(!persist) resetReloadTimeout();
   let idx = MESSAGES.findIndex(m=>m.id==msgid);
   let msg = MESSAGES[idx];
+  if(redrawMsgInterval)clearInterval(redrawMsgInterval);
+  redrawMsgInterval=null;
```
bobrippling: Perhaps to make it clear this is intentionally for when we find the interval:


```suggestion
  if(redrawMsgInterval){
    clearInterval(redrawMsgInterval);
    redrawMsgInterval=null;
}
```

apps/messagegui/app.js
```diff
@@ -457,6 +491,7 @@ function showMessage(msgid, persist) {
         if (idx!=1) return;
         if (rowLeftDraw) rowLeftDraw(r);
         if (rowRightDraw) rowRightDraw(r);
+        if(date!="") drawDate(r);
```
bobrippling: We can shorten this :)


```suggestion
        if(date) drawDate(r);
```

apps/messagegui/app.js
```diff
@@ -477,6 +512,8 @@ function showMessage(msgid, persist) {
       Bangle.removeListener("drag", dragHandler);
       Bangle.removeListener("swipe", swipeHandler);
       Bangle.removeListener("lock", lockHandler);
+      if(redrawMsgInterval)clearInterval(redrawMsgInterval);
+      redrawMsgInterval=null;
```
bobrippling: Same here as above

RKBoss6: Fixed!

apps/messagegui/app.js
```diff
@@ -346,7 +370,14 @@ function showMessage(msgid, persist) {
   if(!persist) resetReloadTimeout();
   let idx = MESSAGES.findIndex(m=>m.id==msgid);
   let msg = MESSAGES[idx];
+  if(redrawMsgInterval){
+    clearInterval(redrawMsgInterval);
+    redrawMsgInterval=null;
+}
   if (!msg) return returnToClockIfEmpty(); // go home if no message found
+  redrawMsgInterval=setInterval(function(){
+    showMessage(msgid, persist);
```
gfwilliams: I'm not sure I like this in here before the checks for music/nav - if `date` was set it'll them start trying to redraw those every minute even though they don't display anything *or* remove the redraw in the remove handler.

Maybe move this ~8 lines down, but keep the clearInterval where it is.

apps/messagegui/app.js
```diff
@@ -457,6 +493,7 @@ function showMessage(msgid, persist) {
         if (idx!=1) return;
         if (rowLeftDraw) rowLeftDraw(r);
         if (rowRightDraw) rowRightDraw(r);
+        if(date) drawDate(r);
```
gfwilliams: Just in general for changes you make, please can you keep with the formatting we use for the rest of Espruino/Bangle.js? eg. `if (date) drawDate(r);` not `if(date) drawDate(r);`

Not a big deal for this one, but for other stuff you make it'd be nice to keep all the code looking the same.

gfwilliams:
Thanks - your 'time ago' code is fine, but we do have `require("time_utils").formatDuration` which does pretty much the same thing already for time - it just includes too much accuracy: https://github.com/espruino/BangleApps/blob/master/modules/time_utils.js#L64-L74

I wonder whether we shoud look at tweaking that slightly.

RKBoss6:
Huh, I didn't see that at all! I think we could tweak it by asking for a `limit` parameter that limits the output levels of precision: eg 2--> 1h56m, or 4-->0d 1h 56m 32s

gfwilliams:
Yes - just pushed a change which adds this - also spotted a bug in it as well: https://github.com/espruino/BangleApps/blob/master/modules/time_utils.js#L64-L74

RKBoss6:
Nice, thanks! For testing this out using the module, are modules auto-uploaded to the latest version if I just upload something to the IDE, or is it built-in to the firmware now?

gfwilliams:
It depends how the IDE is set up - check out https://github.com/espruino/BangleApps/blob/master/modules/README.md?plain=1#L35-L36

apps/messagegui/app.js
RKBoss6: fixed now

RKBoss6:
Thanks! One other question - when users update the app, will it update the module as well for proper integration?

RKBoss6:
Also, I have no experience in Gadgetbridge integration, so if we want this there as well, someone who uses/knows how to update the app to push the date would be lovely.

gfwilliams:
> when users update the app, will it update the module as well for proper integration?

Yes - as long as the user re-installs the app (eg the version changes) you'll get the new module

gfwilliams:
Thanks! I do feel like 20+ lines of code just to draw one line for `Date` is a bit extreme though? https://github.com/espruino/BangleApps/pull/4153/changes/6dc15ac461b2a5df4be0c7b9181aa2f9fbe64bc4

I've just made some changes to try and tidy this up. But please can you try to keep an eye on code duplication? It's not just that it affects execution speed or storage, but it's a pain to maintain. For instance if someone wanted to change the font on the date message after your change, they have to change it in 3 places and find a way to test it with messages with/without negaction.

But this also makes me a bit concerned: The vertical position of the date text depends on whether one or both actions are visible. That can't be right surely. When one action is shown, you draw at `r.y+5`, if both you also clear the rectangle and then use `r.y+g.stringMetrics(" "+date).height`

If `r.y+5` works, please can you just use that always and remove the clearRect?

RKBoss6:
Thanks for telling me, sorry about that! The clearRect is mainly used out clear the space behind the text, so it doesn't overlap the posaction and negaction icons, so think  it's best if we keep it for now? I can get to work making it more consistent for all the positions though, thanks.

gfwilliams:
Ok, thanks - that makes sense then! I've just pushed a change to avoid using stringWidth/Metrics so much, because you only really needed to call it once. Hopefully it's a bit more readable now too.

PR #4154, opened 2026-02-04
[iOS] Pass ANCS notification datetime to message GUI

RKBoss6:
This is in tandem to #4153, where you can display the time of a message like "14m ago". This adds that functionality to iOS notifications, so phone notifications can be sent with a date parameter that will be displayed.

### DIFF
--- a/apps/ios/boot.js
+++ b/apps/ios/boot.js
@@ -308,7 +308,6 @@ E.on('notify',msg=>{
     // use exception or app id itself
     name = appNames[msg.appId]||msg.appId;
   }
- 
   require("messages").pushMessage({
     t : msg.event,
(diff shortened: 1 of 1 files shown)


apps/ios/boot.js
```diff
@@ -23,6 +44,8 @@ E.on('ANCS',msg=>{
   // don't need info for remove events - pass these on
   if (msg.event=="remove")
     return E.emit("notify", msg);
+  
+
```
bobrippling: Would you mind backing out the whitespace changes?

RKBoss6: Fixed!

bobrippling: Thanks!

PR #4157, opened 2026-02-06
backswipe_edge: swipe from the right edge of the screen to go back

ssievert42:
This allows going back by swiping from the right edge of the screen, Android style.

I hacked this together to prevent my button from dying (again).

Not sure if this needs some sort of calibration / settings, but loading a config file at boot seemed like it would slow down load times a bit.

Has been working fine as far as I can tell :)


apps/backswipe_edge/boot.js
```diff
@@ -0,0 +1,71 @@
...
+  Bangle.on("drag", (e) => {
+    // do nothing when showing the clock
+    if (Bangle.CLOCK == 1) {
+      drag = false;
+      trig = false;
+      return;
+    }
+    // where did the drag event start?
+    let xBefore = e.x + e.dx;
+    if (xBefore >= 191) {
```
thyttan: 191 looks high to me given the lcd has 176px width. But maybe it makes sense with the 'touch pixels'. 

Out of curiosity, have you done the touch screen calibration under LCD settings?

thyttan:
Didn't test but looks good I think. One question re pixel counting in review.

apps/backswipe_edge/boot.js
```diff
@@ -0,0 +1,71 @@
+{
+  let sui = Bangle.setUI;
+  let back = null;
+  Bangle.setUI = function(mode, cb) {
+    if (typeof mode == "object" && mode.mode == "custom" && typeof mode.back == "function") {
```
bobrippling: Perhaps we capture `back` regardless of the `mode`, to cater for potentially "out there" usage

```suggestion
    if (typeof mode == "object" && typeof mode.back == "function") {
```

apps/backswipe_edge/boot.js
```diff
@@ -0,0 +1,71 @@
...
+  };
+  
+  // drag started
+  let drag = false;
+  // backswipe triggered
+  let trig = false;
+
+  Bangle.on("drag", (e) => {
+    // do nothing when showing the clock
+    if (Bangle.CLOCK == 1) {
```
bobrippling: Since this'll be called a lot, we can make this code run a bit faster

```suggestion
    if (Bangle.CLOCK === 1) {
```

apps/backswipe_edge/boot.js
```diff
@@ -0,0 +1,71 @@
...
+      trig = false;
+      return;
+    }
+    // where did the drag event start?
+    let xBefore = e.x + e.dx;
+    if (xBefore >= 191) {
+      // drag from the right side of the screen started
+      drag = true;
+    }
+    if (drag) {
```
bobrippling: Let's invert this to simplify the code further down?

```javascript
if (!drag) return;
```

then we can avoid the double checks below

apps/backswipeedge/metadata.json
```diff
@@ -0,0 +1,21 @@
+{
+    "id": "backswipe_edge",
+    "name": "Backswipe from edge",
+    "shortName": "Backswipe edge",
+    "version": "0.01",
+    "description": "Go back by swiping from the right edge of the screen, Android style.",
```
bobrippling, line 6: We may as well add you in :)

```suggestion
    "description": "Go back by swiping from the right edge of the screen, Android style.",
    "author": "ssievert42",
```

thyttan:
Nice idea with the edge, so you don't need to count active swipe and drag handlers like we do in `backswipe` 👍

apps/backswipe_edge/boot.js
ssievert42: Thanks for taking a look at this :)

191 was the highest I got from uploading this:
```js
Bangle.on("drag", (e) => {
  // where did the drag event start?
  let xBefore = e.x - e.dx;
  print(xBefore);
});
```
(Both on my old and shiny new screen that is still waiting to be glued down.)

I didn't do the touch screen calibration.  
(Can't find `touchX1` and friends in my `setting.json` and `Bangle.getOptions()` reports `touchX1: 0, touchY1: 0, touchX2: 160, touchY2: 160`.)

Doing
```js
Bangle.setOptions({touchX1: 10, touchY1: 0, touchX2: 180, touchY2: 160});
```
causes this app to stop working entirely...

Not sure how I would go about accounting for touchscreen calibration though.

ssievert42:
> Nice idea with the edge

Thanks :)  
Although, as the description says, that idea is borrowed from Androids gesture navigation mode.

I guess it would also be possible to draw a little back arrow with an overlay, but you'd probably not see it below your finger :see_no_evil:

RKBoss6:
There's already a `backswipe` app, so would it make more sense to just modify that to have a setting for direction, rather than making a new app?

thyttan:
> There's already a `backswipe` app

I think the logic and design choices are different enough for it to make sense as a new app :)

apps/backswipe_edge/boot.js
thyttan: I don't know what's best:
- Design without touch screen calibration in mind
- Design only with expecting touch screen calibration to be active
- handle both cases.

I guess for the first version we could just put a disclaimer very visibly in the readme (maybe the short description as well?) that it doesn't work with touch calibration. I'm not too keen on that personally though since it's often recommended to do the calibration.

bobrippling:
> I guess it would also be possible to draw a little back arrow with an overlay, but you'd probably not see it below your finger

Yeah I was wondering about this - we could offset it into the middle of the screen but we have the problem of redrawing the original UI when the arrow's gone. Or we could draw the arrow using the overlay

thyttan:
> > I guess it would also be possible to draw a little back arrow with an overlay, but you'd probably not see it below your finger
> 
> Yeah I was wondering about this - we could offset it into the middle of the screen but we have the problem of redrawing the original UI when the arrow's gone. Or we could draw the arrow using the overlay

Maybe we can do paint over the back button so it "flashes"? Then we wouldn't have to worry about the rest of the screen. 

Either way I suggest leaving that for a later version :)

bobrippling:
> Maybe we can do paint over the back button so it "flashes"? Then we wouldn't have to worry about the rest of the screen.

Yeah that could be cool - I do like the vibrate as feedback too, tried it out and it feels just right

apps/backswipe_edge/boot.js
ssievert42: Yeah, ideally this should work whether someone did the calibration or not.

From looking through the code in Espruino:
https://github.com/espruino/Espruino/blob/654b3a3847392741bffd6c5b149a42f82844f692/libs/banglejs/jswrap_bangle.c#L2066-L2068
I landed on something like this to undo the calibration:
```js
let calcTouchX = function(touchX) {
  let opts = Bangle.getOptions();
  let x1 = opts.touchX1;
  let x2 = opts.touchX2;
  return x1 + ((touchX * (x2 - x1)) / g.getWidth());
};
```
Which seems to work alright; will push in a moment.

ssievert42:
Thanks for the review and feedback :)

I think I should've addressed all suggestions up until now; no idea how I managed to forget to fill out the "author" field :see_no_evil:

Painting over the back button sounds like a neat idea :+1:  
But yea, probably best left as a feature for a follow up to this.

Accounting for touchscreen calibration seems to work with
```js
Bangle.setOptions({touchX1: 0, touchY1: 0, touchX2: 160, touchY2: 160});
Bangle.setOptions({touchX1: 10, touchY1: 0, touchX2: 180, touchY2: 160});
Bangle.setOptions({touchX1: 0, touchY1: 0, touchX2: 140, touchY2: 160});
```
which should ideally sort of somewhat cover real world touchscreen calibrations.

thyttan:
Would you mind if we changed the id to `backswipeedge` without the underscore? Shorter names are a little bit faster and most apps don't use underscores in the id. I would have probably even done `backswedge` but that's probably a step too far :P

bobrippling:
Changes look great to me, thanks for the nice app idea btw! And good thinking both of you with the touchscreen calibration

ssievert42:
> Would you mind if we changed the id to backswipeedge without the underscore?

Not really set on the name, so can do :+1:

But would the name really make much of a difference if everything is copied to `.boot0` anyways?

On the topic of shorter names being faster: I changed `calib` to `cal`, which I think is still readable enough.

Maybe for hooking up the back button, the `backswipe` event could get a type parameter?  
Something along the lines of:
- `1`: started (will go back on finger up)
- `2`: cancelled
- `3`: go back

So we could do:
```js
Bangle.on("backswipe", (type) => {
  switch (type) {
    case 1:
        // make back button flash / change color
        break;
    case 2: // fallthrough
    case 3:
        // return back button back to normal
        break;
  }
});
```

apps/backswipeedge/boot.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if (e.b == 0) {
+      if (trig) {
+        Bangle.emit("backswipe");
+      }
+      drag = false;
+      trig = false;
+    }
+  });
+  Bangle.on("swipe", (lr, ud) => {
```
thyttan: I think you'll want  to do 

```suggestion
  Bangle.prependListener("swipe", (lr, ud) => {
```

to be at/near the start of the listener list.

apps/backswipeedge/boot.js
```diff
@@ -0,0 +1,86 @@
...
+    cal.add = x1;
+    cal.mult = (x2 - x1) / g.getWidth();
+  }
+  
+  // drag started
+  let drag = false;
+  // backswipe triggered
+  let trig = false;
+
+  Bangle.on("drag", (e) => {
```
thyttan: To have precedence over other drag listeners:
```suggestion
  Bangle.prependListener("drag", (e) => {
```

apps/backswipeedge/boot.js
ssievert42: Makes sense :+1:

thyttan:
I think it's time to merge :) what do you say?

ssievert42:
> the backswipe event could get a type parameter

Please disregard that last bit; I just assumed that the back button widget was an app and not part of the firmware https://github.com/espruino/Espruino/blob/654b3a3847392741bffd6c5b149a42f82844f692/libs/js/banglejs/Bangle_setUI_Q3.js#L143-L157

Modifying the back button sources to respect this app also seems like a bad idea; we could patch the draw function of `WIDGETS.back` instead though?  
That way people who haven't installed this app are not running any extra code to accommodate this app at all times.

Also, the `drag` handler seems like a prime example of something that can be thrown at the jit for a bit of extra perf, or could that cause issues?

> I think it's time to merge

If there aren't any objections to using the jit I'd like to do that, but the back button flashing thing is probably really best left for the future™ so :+1:

ssievert42:
JITing the `drag` handler seems to work alright; I've just pushed that change.

This should be ready now :)

ssievert42:
On the topic of making the back button "flash":  
Behold my unholy patching stuff to make the back button turn green https://github.com/ssievert42/BangleApps/tree/backswipeedge_change_back_button (It actually works :see_no_evil:)

thyttan:
If jit works that's great! Have you looked through https://www.espruino.com/JIT and https://github.com/espruino/Espruino/blob/master/README_JIT.md ?

I usually do "ram" keyword for perfmance but have not learned when to use "jit". I wonder if they would work together as well 🤔

apps/backswipeedge/boot.js
```diff
@@ -0,0 +1,83 @@
...
+        Bangle.buzz(50, 0.6);
+      }
+    }
+    if (e.b == 0) {
+      if (trig) {
+        if (back != null) {
+          back();
+        } else {
+          Bangle.load();
+        }
```
bobrippling, line 70: Not sure if it was intentional, but looks like the `backswipe` event got reverted. I think it's handy to have but if intentional then not a problem

bobrippling (commented):
> JITing the `drag` handler seems to work alright; I've just pushed that change.

Nice! Like @thyttan says, jit is buggy in places but if it works well under test then the language features being used are likely fine here

> On the topic of making the back button "flash": Behold my unholy patching stuff to make the back button turn green https://github.com/ssievert42/BangleApps/tree/backswipeedge_change_back_button (It actually works 🙈)

That is rather unholy! The pattern a lot of apps use for overwriting is to copy [the code](https://github.com/espruino/Espruino/blob/654b3a3847392741bffd6c5b149a42f82844f692/libs/js/banglejs/Bangle_setUI_Q3.js#L146) and tweak it, which in this case is probably less of a diff than modifying the function code at runtime. But we can discuss in a separate PR :)

thyttan:
For flashing the button I had a naive solution in mind just painting at the same position of the widget but not really interacting with it. Patching the widget may be better though? Without working it through it feels like it takes a bit more logic, but shouldn't matter.

apps/backswipeedge/boot.js
ssievert42: My plan was originally to have that backswipe event to decouple gesture detection and what happens when a backswipe is detected, as well as having some way to prevent backswipes from having any effect by removing that default handler.

But emitting that event didn't work with the jit.

thyttan: Just adding that I also found the idea of emitting the event was interesting for interoperability with other apps. But let's go with whatever you want 🙂 (as long as you're sure about jit)

ssievert42:
> Have you looked through https://www.espruino.com/JIT and https://github.com/espruino/Espruino/blob/master/README_JIT.md ?

:+1:

Everything we're doing here should ideally fall under the "works" category.

> That is rather unholy!

I know, right? It's probably only a matter of time before that breaks, so copying and then modifying the code sounds like a better idea (for a future PR).  
I'd prefer adjusting the code to drawing over the widget; that way the widget redrawing itself ideally has less of a chance of breaking things.

thyttan:
So we go ahead? 🙂

ssievert42:
Yes, please :+1:

Also, thanks again for the review, feedback and stuff :)

thyttan:
Thanks 😁

apps/backswipeedge/boot.js
bobrippling: No worries - @gfwilliams do you know what the particular missing JIT feature is that prevents emitting an event? I'd be interested in having a go at it

bobrippling:
> I'd prefer adjusting the code to drawing over the widget; that way the widget redrawing itself ideally has less of a chance of breaking things.

Sounds good - you may find [`widget_utils`'s hiding](https://github.com/bobrippling/BangleApps/blob/7e6c713389902e709bb3cab4bbd6193c0d472da9/modules/widget_utils.js#L7-L9=) interesting inspiration

And thanks for the PR!

apps/backswipeedge/boot.js
gfwilliams: I have no idea - it should work fine. If someone can narrow it down to a really basic function that works normally and not in JIT I can take a look?

But actually I'm not sure about putting JIT on this function - have you done any tests to check what the speed improvement really is?

I feel like for a drag handler it's going to be called maybe 20 times a second - the small improvement in speed may not be worth the extra RAM it's going to use up?

ssievert42: > But actually I'm not sure about putting JIT on this function

You're probably right; I just did some measurements and without JIT the time that drag handler takes to execute is about 2.4 ms on average, while with the JIT that shrinks to about 1.5 ms on average.  
So only a speedup of 1.6. (Which is already impressive in its own right.)  
OpenStreetMap, where I was worried that a slow additional drag handler may make moving the map around laggy, doesn't show any noticeable difference.

So yeah, the `"jit";` at the start of that drag handler can probably go :+1:    
Don't want to snatch away ram from other applications that may need it more :)

> If someone can narrow it down to a really basic function that works normally and not in JIT I can take a look?

From my brief testing, emitting events on builtin objects from the JIT doesn't work, so:
```js
let obj = {};

NRF.on("someevent", (p) => {
  print("NRF event: " + p);
});

obj.on("someevent", (p) => {
  print("obj event: " + p);
});

let emitEvent = function(name, p) {
  NRF.emit(name, p);
  obj.emit(name, p);
};

let main = function() {
  "jit";
  NRF.emit("someevent", "from jit");
  obj.emit("someevent", "from jit");
  emitEvent("someevent", "from js");
};

main();
```
prints:
```
obj event: from jit
NRF event: from js
obj event: from js
```
instead of (what I would expect):
```
NRF event: from jit
obj event: from jit
NRF event: from js
obj event: from js
```

RKBoss6: Just as a question - What is jit? I've seen it before in the main espruino repo, and don't yet know what it does. It it a way of making events prioritized?

thyttan: jit stands for 'just in time' and is a concept in programming and broader in industry. I think Toyota pioneered the concept but not sure. You'll find out more by a search on the web, e.g. 'jit programming'. 

https://www.espruino.com/JIT

gfwilliams: @ssievert42 thanks, that's great! Yes, I think for now removing JIT is more sensible,

I've just [filed an issue on the Espruino repo about that JIT issue](https://github.com/espruino/Espruino/issues/2690), the minimal function is perfect. I added some optimisations to JIT recently that try and build in built-in functions rather than searching for them at runtime, and I bet something's gone wrong there.

PR #4158, opened 2026-02-07
backswipeedge: fixup readme

ssievert42:
Forgot to remove the `Bangle.backswipe` event from the readme :see_no_evil:  
(That ended up being removed, mainly because it didn't want to work from jitd code)


thyttan:
Ah, nice! 🙂

PR #4160, opened 2026-02-08
Fix emulator race condition in tests

elijahr:
## Problem

Tests occasionally fail with:

```
Uncaught SyntaxError: Got [ERASED] expected EOF
```

Reported in https://github.com/espruino/BangleApps/pull/4132 by @thyttan: tests fail initially but succeed on re-run.
Example: https://github.com/espruino/BangleApps/actions/runs/21783161416/attempts/1

## Cause (suspected)

The test runner sends commands without waiting for the emulator to finish processing. When one test ends and the next begins, `factoryReset()` can be called while the previous test's commands are still being processed, corrupting the input stream.

## Solution

- Track emulator ready state (has it output a response?)
- Add `waitForResponse()` to poll until command completes
- Add `safeFactoryReset()` that ensures previous commands finished
- Wait after each test step (`cmd`, `emit`, `load`, etc.)
- Wait after app upload and reset in test setup

## Testing

- All functional tests pass locally and in CI
- Full test suite (all 4 apps): https://github.com/elijahr/BangleApps/actions/runs/21789313739
- No significant increase in test duration
- Also detects test infrastructure changes as a condition for running all app tests in CI (`bin/runapptests.js` `.github/workflows/nodejs.yml` and `core` submodule).

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -52,18 +52,43 @@ jobs:
         id: detect-apps
         run: |
           set -e  # Exit on error, but handle grep specially
+          
+          # Files that should trigger running all tests
+          # Note: "core" is the submodule - when its commit hash changes, "core" appears in the diff
+          TEST_INFRA_FILES="bin/runapptests.js .github/workflows/nodejs.yml core"
+          
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            # PR: test only changed apps
             CHANGED=$(git diff --name-only ${{ github.event.pull_request.base.sha }}..HEAD)
-            APPS=$(echo "$CHANGED" | grep '^apps/' | cut -d'/' -f2 | sort -u) || APPS=""
           elif [ "${{ github.ref }}" = "refs/heads/master" ]; then
-            # Push to master: test all apps
-            APPS=$(ls apps/*/test.json 2>/dev/null | cut -d'/' -f2) || APPS=""
+            # Push to master: always test all apps
+            APPS=$(ls apps/*/test.json 2>/dev/null | cut -d'/' -f2 | tr '\n' ' ') || APPS=""
+            echo "apps=$APPS" >> $GITHUB_OUTPUT
+            echo "Push to master - testing all apps:$APPS"
+            exit 0
           else
-            # Push to other branches: test only changed apps since master
             CHANGED=$(git diff --name-only origin/master...HEAD)
-            APPS=$(echo "$CHANGED" | grep '^apps/' | cut -d'/' -f2 | sort -u) || APPS=""
           fi
+          
+          # Check if test infrastructure changed - if so, run all tests
+          RUN_ALL=false
+          for f in $TEST_INFRA_FILES; do
+            if echo "$CHANGED" | grep -q "^$f$"; then
+              RUN_ALL=true
+              echo "Test infrastructure changed: $f"
+              break
+            fi
+          done
+          
+          if [ "$RUN_ALL" = "true" ]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/nodejs.yml
```diff
@@ -52,18 +52,43 @@ jobs:
...
-            # Push to master: test all apps
-            APPS=$(ls apps/*/test.json 2>/dev/null | cut -d'/' -f2) || APPS=""
+            # Push to master: always test all apps
+            APPS=$(ls apps/*/test.json 2>/dev/null | cut -d'/' -f2 | tr '\n' ' ') || APPS=""
+            echo "apps=$APPS" >> $GITHUB_OUTPUT
+            echo "Push to master - testing all apps:$APPS"
+            exit 0
           else
-            # Push to other branches: test only changed apps since master
             CHANGED=$(git diff --name-only origin/master...HEAD)
```
bobrippling, line 69: Let's keep this comment in, as it's not clear what case the `else` handles


```suggestion
            # Push to other branches: test only changed apps since master
            CHANGED=$(git diff --name-only origin/master...HEAD)
```

.github/workflows/nodejs.yml
```diff
@@ -52,18 +52,43 @@ jobs:
...
-            APPS=$(echo "$CHANGED" | grep '^apps/' | cut -d'/' -f2 | sort -u) || APPS=""
           fi
+          
+          # Check if test infrastructure changed - if so, run all tests
+          RUN_ALL=false
+          for f in $TEST_INFRA_FILES; do
+            if echo "$CHANGED" | grep -q "^$f$"; then
```
bobrippling, line 75: It's minor, but `$f` is treated as a regex when it isn't. We could:


```suggestion
            if echo "$CHANGED" | grep -qx "$f"; then
```

Or we pull out the refspec above and one-shot this without needing to loop:

```sh
# e.g. refspec is origin/master...HEAD

if git diff --quiet "$refspec" -- $TEST_INFRA_FILES; then
...
```

.github/workflows/nodejs.yml
```diff
@@ -52,18 +52,43 @@ jobs:
...
+              RUN_ALL=true
+              echo "Test infrastructure changed: $f"
+              break
+            fi
+          done
+          
+          if [ "$RUN_ALL" = "true" ]; then
```
bobrippling, line 82: We can shorten this:

```suggestion
          if "$RUN_ALL"; then
```

bin/runapptests.js
```diff
@@ -114,10 +114,69 @@ var emu = require(BASE_DIR+"/core/lib/emulator.js");
...
+}
+
+// Factory reset with safety checks
+function safeFactoryReset() {
+  // Ensure emulator is ready before resetting (previous commands complete)
+  if (!emuReady) {
+    waitForResponse();
+  }
+  if (!emuReady) {
+    throw new Error(
```
bobrippling, line 169: I suppose this is unreachable code as we can't leave `waitForResponse()` with `emuReady === false`

Perhaps we need a way of detecting if `maxIterations` has been reached?

bin/runapptests.js
```diff
@@ -187,6 +246,7 @@ function wrap(func, id){
   }(${func}));\n`;
 
   emu.tx(wrappingCode);
+  waitForResponse();
```
bobrippling, line 249: Would using `txAndWait()` for this and other `emu.tx(...); waitForResponse()` be an approach to avoiding the bug sneaking in, in future?

bobrippling (commented):
Thanks for the rapid fix, this is great - I have some small questions, but nothing major about the design etc :)

gfwilliams:
Fixed with https://github.com/espruino/BangleApps/pull/4252

PR #4161, opened 2026-02-09
lcdclock/lcdclockplus: fix/extend regexp for clockinfo menus

Freubert:
When syncing e.g. weather info on the watch, the display/layout for temperature is broken/misplaced/invisible on lcdclock and lcdclockplus inside the clockinfo menus.
Download
This PR adds an escape character to the hyphen, which broke the weather clockinfo display. It also adds the colon to the regexp to allow this character in the clockinfo.
Download (1)

https://freubert.github.io/BangleApps/

### DIFF
--- a/apps/lcdclockplus/app.js
+++ b/apps/lcdclockplus/app.js
@@ -67,7 +67,7 @@ let clockInfoDraw = (itm, info, options) => {
   g.clearRect({x:options.x,y:options.y,w:options.w,h:options.h,r:8});
 
   if (info.img) g.drawImage(info.img, options.x+1,options.y+2);
-  var text = info.text.toString().toUpperCase().replace(/[^A-Z0-9.:\-]/g, "");
+  var text = info.text.toString().toUpperCase().replace(/[^A-Z0-9.:-]/g, "");
   if (g.setFont("7Seg:2").stringWidth(text)+24-2>options.w) g.setFont("7Seg");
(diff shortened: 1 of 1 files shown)


apps/lcdclock/app.js
```diff
@@ -123,7 +122,7 @@ let clockInfoDraw = (itm, info, options) => {
 
   if (info.img) g.drawImage(info.img, options.x+2, options.y+2);
   var title = clockInfoItems[options.menuA].name;
-  var text = info.text.toString().toUpperCase().replace(/[^A-Z0-9-.]/g, "");
+  var text = info.text.toString().toUpperCase().replace(/[^A-Z0-9.:\-]/g, "");
```
bobrippling: Would you mind checking this without the espace character? It won't have any effect since `-` isn't a special character:

```javascript
/[-]/ .test("hi - there") // true
/[\-]/.test("hi - there") // true
/[-]/ .test("hi there") // false
/[\-]/.test("hi there") // false
```

Freubert: Well, it's working without escaping. But interestingly only when the hyphen is placed on last position in the RegEx (placing it not there seems to have been to original bug).

bobrippling: Yes - a hyphen in a regex denotes a range, unless it's at the end

However, `0-9-.` shouldn't be two ranges, just `0-9` and the two characters `-` and `.`. This is perhaps a differing behaviour in espruino JS

bobrippling:
Thanks for the fix!

PR #4163, opened 2026-02-10
backswipeedge: don't use JIT for drag handler

ssievert42:
As discussed in #4157

While the JIT provides a speedup of around 1.6 for the drag handler, we're talking about 2.4 ms without the JIT, on a function that is called maybe 20 times a second.

Plus using the JIT uses up some precious RAM.


thyttan:
🙏

thyttan:
🙏

PR #4165, opened 2026-02-11
[ClockInfo] Refactor BLE clock_info 

RKBoss6:
- Changes the ble clockInfo to use event based updating, (`NRF.on(...)`)
- Adds a connected message if it is connected, instead of showing on no matter what.
- Removed redraw after timeout since events handle it
- Overall, makes the updates snappier due to relying on state changes

### DIFF
--- a/apps/clock_info/lib.js
+++ b/apps/clock_info/lib.js
@@ -87,15 +87,14 @@ exports.load = function() {
       hide : function() { clearInterval(this.interval); delete this.interval; Bangle.removeListener("charging", batteryUpdateHandler); },
     },
     { name: "BLE",
-      getText: () => {
-        const s = NRF.getSecurityStatus();
-        if (s.connected) return "Conn.";
-        else if (s.advertising) return "On";
-        else return "Off";
-      },
       get: function() {
+        const s = NRF.getSecurityStatus();
+        let text;
+        if (s.connected) text="Conn";
+        else if (s.advertising) text="On";
+        else text="Off";
         return {
-          text: this.getText(),
+          text: text,
           img: atob("GBiBAAAAAAAAAAAYAAAcAAAWAAATAAARgAMRgAGTAADGAAB8AAA4AAA4AAB8AADGAAGTAAMRgAARgAATAAAWAAAcAAAYAAAAAAAAAA==")
           // small gaps added to BLE icon to ensure middle of B isn't filled
         };


apps/clock_info/lib.js
```diff
@@ -83,28 +87,38 @@ exports.load = function() {
...
       run: function() {
-        if (this.isOn()) {
+        if (this.state()=="Conn."||this.state()=="On") {
           NRF.sleep();
         } else {
           NRF.wake();
           Bluetooth.setConsole(1);
         }
-        setTimeout(() => this.emit("redraw"), 250);
       },
```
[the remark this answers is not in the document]
RKBoss6: fixed

apps/clock_info/lib.js
```diff
@@ -83,28 +87,39 @@ exports.load = function() {
...
         return {
-          text: this.isOn() ? "On" : "Off",
+          text: this.getText(),
           img: atob("GBiBAAAAAAAAAAAYAAAcAAAWAAATAAARgAMRgAGTAADGAAB8AAA4AAA4AAB8AADGAAGTAAMRgAARgAATAAAWAAAcAAAYAAAAAAAAAA==")
           // small gaps added to BLE icon to ensure middle of B isn't filled
         };
       },
       run: function() {
-        if (this.isOn()) {
+        var state=NRF.getSecurityStatus();
```
bobrippling: Let's keep with the theme :)


```suggestion
        const state = NRF.getSecurityStatus();
```

apps/clock_info/lib.js
```diff
@@ -83,28 +87,39 @@ exports.load = function() {
...
+        if (state.connected||state.advertising) {
```
bobrippling: And as above
```suggestion
        if (state.connected || state.advertising) {
```

apps/clock_info/lib.js
```diff
@@ -83,28 +87,39 @@ exports.load = function() {
       hide : function() { clearInterval(this.interval); delete this.interval; Bangle.removeListener("charging", batteryUpdateHandler); },
     },
     { name: "BLE",
-      isOn: () => {
+      getText: () => {
```
gfwilliams: Please can we move the code from this into the `get` function and delete this function? It's not used by anything else.

gfwilliams:
Thanks - looks good - please can we remove `getText`? `isOn` was only there initially to reduce code duplication originally.

Quick question though: Is `Conn.` a good message? `Conn./On/Off` feels a bit confusing to me - if I saw it I'm not sure I'd really know what it meant? At minimum I'd remove the dot.

RKBoss6:
The reason I didn't put the full "connected" was because I thought it wouldn't fit well on existing clock infos, so I put an abbreviation instead. Wdyt we should do?Just remove the '.M Or do something else instead??

gfwilliams:
Yes, I can see why it's done - and I can't think of better abbreviations. Not sure what others think? 

Maybe just remove the `.` for now. I think if we're trying to save characters (which we are) it makes sense to leave it off

apps/clock_info/lib.js
RKBoss6: Just committed!

gfwilliams:
Great - thanks!

PR #4166, opened 2026-02-12
Add authors to apps

RKBoss6:
This adds many authors to apps' metadatas, based on the app's readmes, in which all of them had a creator or author listed. I just took that, and added an author field to each app based on that. All of the authors are verified in the README, so I'm not assuming anything, instead taking what is already available and adding it to the metadata.


bobrippling:
Nice!

PR #4172, opened 2026-02-14
accelsender: add stop on disconnect option; fix saving config

kidneyhex:
link: https://kidneyhex.github.io/BangleApps/?id=accelsender

Tagging: @AnotherStranger , @bobrippling 

1) I had `Bangle.on("kill"...)` instead of `E.on("kill"...)` so config wasn't saving and I didn't notice because it kept loading the defaults.
2) Added option to stop/disable `accelsender` when bluetooth is disconnected
3) Moved defaults to happen with `Object.assign` instead of in the getters.

In the future someone might want to add restarting `accelsender` on reconnect if it's been some short time since disconnect (15-30min). But for now, it seems safe and simple to just stay stopped. 

Reference suggestion: https://github.com/espruino/BangleApps/pull/4159#issuecomment-3870143635


bobrippling:
Nice spot with the `E.on()`, the disconnect changes LGTM too, thanks. Shall give @AnotherStranger a few days to see then I'll merge

bobrippling:
Thanks for the PR!

PR #4173, opened 2026-02-14
exactwords: handle widgets; add author tag

kidneyhex:
link: https://kidneyhex.github.io/BangleApps/?id=exactwords

Tagging author: @bmsleight

The original `exactwords` will clear out widgets, so they disappear at times. This just adds using `appRect` to avoid that.

## app.js
* Widgets get cleared by the `g.clear()`, so changing to `g.clearRect(Bangle.appRect)`
* Moved first `draw()` to happen after `loadWidgets` so `appRect` is correct.
* Use `appRect` for some of the calculations (doesn't really affect anything)
* Changed bottom text to align bottom and draw at the bottom of the `appRect`.

### DIFF
--- a/apps/exactwords/ChangeLog
+++ b/apps/exactwords/ChangeLog
@@ -1,1 +1,2 @@
 0.1: New App! Need to work out locale settings
+0.2: Fixes to not draw over widgets
--- a/apps/exactwords/app.js
+++ b/apps/exactwords/app.js
@@ -176,8 +176,10 @@ function wordsFromDayMonth(day, date, month)
 }
 
 function draw() {
-  var x = g.getWidth()/2;
-  var y = g.getHeight()/2;
+  const R = Bangle.appRect;
+  const midX = (R.x + R.x2) / 2;
+  const midY = (R.y + R.y2) / 2;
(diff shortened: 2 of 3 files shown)


bmsleight:
Thanks. Not tested.

Does - "Changed bottom text to align bottom and draw at the bottom of the appRect." still work when month has a very long text (today for example is Saturday, Fourteenth of February  ?

kidneyhex:
>Does - "Changed bottom text to align bottom and draw at the bottom of the appRect." still work when month has a very long text (today for example is Saturday, Fourteenth of February ?

Yep, the `g.setFontAlign(0,1)` has it align based on the bottom of the string, so it works its way up from there. 

I just enabled allowing emulator too, since there wasn't anything stopping that. 

I also reverted my change centering based on appRect, and went back to centering based on the screen height/width. I'd been testing with top+bottom widgets, and then went back to just top and realized if it centered based on appRect, it moved the text down and looked weird.

bobrippling:
LGTM, will give @bmsleight a day or two then I'll merge

bobrippling:
Thanks again!

PR #4177, opened 2026-02-21
Bump accepted JS version in sanity checks, to allow static properties in apps

mdoggydog:
Espruino supports static properties in classes, introduced by Ecmascript 2022, e.g.:
```
class X {
  static y = "z";
}
```
However, sanitycheck.js and the eslint configuration do not currently understand or permit that syntax (i.e., failing with parse errors).

So, tell the sanity checkers to accept version 2022 of Ecmascript.

This may cause the checkers to allow additional 2022 features which are not supported by Espruino.  However, since this is merely sanity checking, it is probably better to be lenient, and allow Bangle apps to make complete use of those features that *are* supported by Espruino.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -192,7 +192,7 @@ module.exports = {
         "bleServiceOptions": "writable", // available in boot.js code that's called ad part of bootupdate
     },
     "parserOptions": {
-        "ecmaVersion": 11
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks good to me - thanks!

PR #4188, opened 2026-03-10
drained: handle missed "changing" events

bobrippling:
Sometimes my watch will change (with `drained` active) past the threshold and not deactivate `drained`. On debugging, the interval isn't set, so we set it unconditionally to ensure we have a charge "watcher".

### DIFF
--- a/apps/drained/app.js
+++ b/apps/drained/app.js
@@ -120,12 +120,13 @@ Bangle.on("charging", function (charging) {
         drainedInterval = setInterval(checkCharge, interval * 60 * 1000);
     draw();
 });
+drainedInterval = setInterval(checkCharge, interval * 60 * 1000);
 if (!keepStartup) {
-    var storage = require("Storage");
+    var storage_1 = require("Storage");
     for (var _i = 0, exceptions_1 = exceptions; _i < exceptions_1.length; _i++) {
         var boot = exceptions_1[_i];
         try {
-            var js = storage.read("".concat(boot, ".boot.js"));
+            var js = storage_1.read("".concat(boot, ".boot.js"));
             if (js)
                 eval(js);
         }
--- a/apps/drained/app.js
+++ b/apps/drained/app.js
@@ -111,16 +111,15 @@ var checkCharge = function () {
     }
     drainedRestore();
 };
-if (Bangle.isCharging())
-    checkCharge();
-Bangle.on("charging", function (charging) {
+var onChargeChange = function (charging) {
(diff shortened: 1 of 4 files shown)


thyttan:
Does this mean something is broken in the event handling somewhere?

I mean, why would we be missing to set the interval in the 'charging'-listener?

thyttan:
We also discussed some weird event listener behavior on https://github.com/espruino/BangleApps/pull/4141.

Could this be something similar?

bobrippling:
Good memory! I would initially say no, and that drained just starts too late but that can't be the case - drained is running when I plug my watch in to charge 🤔 

It could well be something with the firmware, like #4141. I'll see if I can get more details

thyttan:
@bobrippling I'm thinking of opening an issue re event handling on espruino/Espruino. Have you managed to understand the problem any more?

bobrippling:
I'm afraid not - still at square one!

stweedo:
There is one specific case i found with version 0.08 of drained where if it's started while the watch is charging, it will not deactivate when crossing the threshold until a new charging event happens by reconnecting the charger.

gfwilliams:
Just mentioned in https://github.com/espruino/Espruino/issues/2702#issuecomment-5608043490 but:

I don't think `drained` uses `notify`? But there was an issue with notify calling stopEventPropagation on hide, so if any other app did show a notification with `notify` (eg 'low battery') and then hide it on a `charging` event then any other 'charging' event handlers would not be called and it'd cause this exact issue

bobrippling:
Yeah, no use of `notify` - I haven't gotten to the bottom of this so maybe we merge this PR to handle the problem until I do suss it out?

thyttan:
It feels a bit antithetical to start running an interval while we try to preserve battery as much as possible. But maybe it's negligible.

apps/drained/app.js
```diff
@@ -120,12 +120,13 @@ Bangle.on("charging", function (charging) {
         drainedInterval = setInterval(checkCharge, interval * 60 * 1000);
```
thyttan, line 118: What if we put this inside a timeout? If the problem is the watch is too occupied to register the interval at the moment the charger is connected then maybe that can help. But also, if that's the problem the setTimeout would maybe suffer the same fate...

bobrippling:
What about this change I've just pushed?

thyttan:
Looks good to me!

gfwilliams:
Looks great to me!

PR #4197, opened 2026-03-23
Create colorpicker module

RKBoss6:
Creates colorpicker module as an easy way for apps to present users with a color selection menu. 
I've been wanting a menu that could do this for a while, as the alternative before was to make a scroll menu with all the colors and have the users scroll through all of them, instead of being able to see them at a glance. Currently this only works on bangle.js 2, I'm not exactly sure how to convert it to bangle.js 1 as well. This can be used in places like clock_bg, where users need to select multiple colors, or even in the theme menu and other system-level places if the need arises.

This is my first module, so let me know if I need to do anything else or did something wrong :)

Single color mode:
d

Multi-select mode:
download (7)

You can look at an example implementation in the [Clock Background app on my app loader](https://rkboss6.github.io/BangleApps/?id=clockbg), and just upload the module to your watch as `colorpicker` via web ide.

For bangle.js 1 compatibility, would the best course of action be to make the up/down buttons change the highlighted color, then a select is used to select it?

### DIFF
--- a/modules/colorpicker.js
+++ b/modules/colorpicker.js
@@ -15,6 +15,7 @@ exports.show = function(options) {
   if(colors.length>36) throw new Error("More than 36 colors provided, cannot display");
   
   if(!options.onSelect) throw new Error("No onSelect function provided");
+  if(!options.back) throw new Error("No back function provided");
   
   var rect = Bangle.appRect;
   var W = rect.w;
--- a/modules/colorpicker.js
+++ b/modules/colorpicker.js
@@ -4,7 +4,7 @@ exports.show = function(options) {
   var previewTimeout;
   if (!options.colors||options.colors.length==0) {
     colors = [
-      "#000000", "#555555", "#AAAAAA", "#FFFFFF",
+      "#000000", "#808080", "#AAAAAA", "#FFFFFF",
       "#FF9999", "#FFCC99", "#FFFF99", "#99FF99", "#99FFFF", "#9999FF", "#FF99FF",
       "#FF0000", "#FF8800", "#FFFF00", "#00FF00", "#00FFFF", "#0000FF", "#FF00FF",
       "#880000", "#884400", "#888800", "#008800", "#008888", "#000088", "#880088"
--- a/modules/colorpicker.js
+++ b/modules/colorpicker.js
@@ -65,12 +65,11 @@ exports.show = function(options) {
   }
 
   function remove() {
-    if(previewTimeout)clearTimeout(previewTimeout);
-    if(options.back){
-      options.back();
-    }else{
-      throw new Error("No back function provided");
+    if(previewTimeout){
+      clearTimeout(previewTimeout);
+      previewTimeout=null;
     }
+    options.back();
   }
 
   function onTouch(btn, xy) {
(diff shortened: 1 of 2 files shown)


RKBoss6:
Alright, this is good to go now :)

modules/colorpicker.js
```diff
@@ -0,0 +1,115 @@
...
+    if(y<rect.y)return null;
+    var col = ((x - rect.x) / CW) | 0;
+    var row = ((y - rect.y) / CH) | 0;
+    var i = row * COLS + col;
+    if (i < 0 || i >= n) return null;
+    return colors[i];
+  }
+
+  function remove() {
+    if(previewTimeout)clearTimeout(previewTimeout);
```
bobrippling: Do we want:


```suggestion
    if(previewTimeout)previewTimeout=clearTimeout(previewTimeout);
```

modules/colorpicker.js
```diff
@@ -0,0 +1,115 @@
...
+    if(options.back){
+      options.back();
+    }else{
+      throw new Error("No back function provided");
```
bobrippling: Is this not already covered?

modules/colorpicker.js
```diff
@@ -0,0 +1,115 @@
...
+        options.onSelect(col);
+        if (options.showPreview === undefined || options.showPreview) {
+          g.setColor(col);
+          g.fillRect(rect);
+          previewTimeout=setTimeout(remove, 0.7 * 1000);
+        } else {
+          remove();
+        }
+      }else{
+        Bangle.haptic()
```
bobrippling: We test `Bangle.haptic` above but not here?

modules/colorpicker.js
```diff
@@ -0,0 +1,115 @@
...
+      var col = colorAt(xy.x, xy.y);
+      if (!col) return;
+      if(!options.multiSelect){
+        isPicking=false;
+        if(Bangle.haptic) Bangle.haptic();
+        options.onSelect(col);
+        if (options.showPreview === undefined || options.showPreview) {
+          g.setColor(col);
+          g.fillRect(rect);
+          previewTimeout=setTimeout(remove, 0.7 * 1000);
```
bobrippling, line 90: Do we want to clear the timeout first?

bobrippling (commented):
Nice idea, seems useful

There's a few mistakes that feel like they're a bit sloppy, is it from an AI or just some changes later on made only in one place?

RKBoss6:
Ah - I didn't use ai, that was just me making silly mistakes :) I'll get to fixing these, but no promises when as I'll be on vacation this week

bobrippling:
No worries, thanks for your honesty - let me know if anything isn't clear / if my comments could be more helpful :) And enjoy your vacation!

modules/colorpicker.js
RKBoss6: Fixed!

modules/colorpicker.js
RKBoss6: Just fixed!

RKBoss6:
I think all your suggestions are added now. The only thing left is bangle.js 1 support. Also, I don't need to add the module to a list or anything like that for it to be pulled in through the app loader, right?

modules/colorpicker.js
RKBoss6: Not sure if this would be useful, as we can just set it to null...

thyttan:
Maybe a possible Bangle.js 1 implementation/placeholder could be a showmenu menu with all options?

No opinion on if you need to make it from me though 😺

bobrippling:
> I think all your suggestions are added now. The only thing left is bangle.js 1 support.

Thanks, LGTM! I think we leave BJS1 support for another PR :)

> Also, I don't need to add the module to a list or anything like that for it to be pulled in through the app loader, right?

Not to be pulled through - what you have will work well. I can't find it so I think I might have made it up, I don't think there's anything on banglejs.com that would need updating either. Will merge if you're happy @thyttan?

modules/colorpicker.js
bobrippling: Yes, you're right, same thing - I was keeping it to one line as a more "terse" style of JS I see from time to time

RKBoss6:
I think this is good to merge then!

bobrippling:
Sweet, thanks!

PR #4201, opened 2026-03-26
[qmsched] Add day-specific scheduling

RKBoss6:
Adds day specific scheduling using a bitmask similar to `Alarms`, and added a clockInfo for fast switching directly from the clock. Also created a method `getQuietMode()` that gets the mode active instead of clockinfo, and widgets doing it separately


bobrippling:
Drop me an @ when you're happy :)

RKBoss6:
@bobrippling Not sure about these lint errors - from what I can tell, they had long exemptions and are now gone due to edits. Would I re-add those lint exemptions, or try to fix it?

bobrippling:
Ah yes - because those files have been edited it's removed the exemptions. If you fancy, have a go at fixing them or if not let me know and I'll sort them

RKBoss6:
Hmm, I've no idea how to add back the lint exemptions, and don't want to mess something up :(
 If possible, could you re-add them? Thanks!

bobrippling:
Sorted

PR #4204, opened 2026-04-08
feat(health): add temperature-based wear check and configurable HRM settling timeout

FabianK-Dev:
### Description
This PR introduces two features to improve HRM data quality and battery efficiency by preventing the sensor from firing when off-wrist, and allowing the optical sensor hardware to settle for improved accuracy.

### Technical Implementation

**1. Temperature-Based Wear Check (`boot.js`):**
The legacy wear check relied solely on the Z-axis orientation (`Math.abs(Bangle.getAccel().z)>0.99`) and low movement. This caused false negatives (disabling HRM) when users were sleeping with their arm resting flat.
Introduced `wearCheckTemp`. If configured (`> 0`), the logic bypasses the accelerometer checks and evaluates `E.getTemperature() < wearCheckTemp`. This guarantees the HRM fires reliably during sleep tracking while still correctly identifying a cold nightstand.

**2. HRM Stop Delay (`boot.js`):**
The optical sensor often reports an inaccurate BPM spike on its very first reading at >90% confidence. Modified the early-return logic in `Bangle.on("HRM")`. Instead of powering off immediately, if `hrmTimeout` is set, it flags `accTimeout = true` to ignore subsequent confidence events and keeps the sensor powered for `hrmTimeout * 1000` ms via `setTimeout`. This allows the HRM recordings to stabilize, pushing a much more accurate average into the 10-minute database block.

**3. Documentation (`README.md`):**
- Documented the new Wear Check and HRM Stop Delay parameters.
- Added a to-do point regarding REM sleep approximation possibilities using the existing data structure (high BPM + low movement).

**Checklist:**
- [x] Tested locally on Bangle.js 2.
- [x] Version bumped in `metadata.json`.
- [x] `boot.min.js` successfully regenerated from `boot.js`.
- [x] No changes to other apps included.

**App Loader Link:**
https://fabiank-dev.github.io/BangleApps/


gfwilliams:
Thanks! This looks good!

However:

>  This caused false negatives (disabling HRM) when users were sleeping with their arm resting flat.

Has anyone actually experienced this? the check is intentionally pretty tight so you'd have to have your arm astonishingly flat for it to be an issue.

FabianK-Dev:
> Has anyone actually experienced this? the check is intentionally pretty tight so you'd have to have your arm astonishingly flat for it to be an issue.

I haven't experienced this live while asleep (obviously :-)) but I once noticed that sometimes HRM values in the activity page in Gadgetbridge aren't shown and that there are gaps inbetween (please see screenshot):
image

The following line will not activate the HRM if the condition is met:
```js
if (Bangle.getHealthStatus("last").movement<100 && Math.abs(Bangle.getAccel().z)>0.99) return;
```
So, I suspected this could be due to the hand lying "flat" during sleep so I tested the following 1-liner in the IDE:
```js
setInterval(() => print("Z: " + Bangle.getAccel().z), 1000);
```
To my surprise:
- 1. `Bangle.getAccel().z` gave values > 1.0 (I thought `Bangle.getAccel().z` will only give values between 0 and 1)
- 2. After placing my hands in a typical "sleep position" (i.e. a position that I would naturally keep it during sleep) I could easily get values at around 1 to 1.1

So I assumed this was the cause of the gaps in Gadgetbridge, thus I implemented the temperature-wear-check feature.

Just rechecked it: Laying my hand flat like in the photo and running the following code gives the following output:
PXL_20260418_143655068 MP
Screenshot_20260418-163721

Later on I figured out the gaps are probably a different cause:
https://github.com/espruino/Espruino/blob/master/libs/banglejs/jswrap_bangle.c#L1743
This line defines the current activity as `HSA_NOT_WORN` when the movement is < 120.

After logging the recorded activity of the night (`require("health").readDay(new Date(), print);`), I saw the following:
```js
{ "steps": 12, "bpmMin": 60, "bpmMax": 60, "movement": 400,
  "battery": 53, "isCharging": false, "temperature": 33, "altitude": 536,
  "activity": "UNKNOWN",
  "bpm": 60, "hr": 2, "min": 50 }
{ "steps": 0, "bpmMin": 55, "bpmMax": 55, "movement": 264,
  "battery": 55, "isCharging": false, "temperature": 33, "altitude": 436,
  "activity": "UNKNOWN",
  "bpm": 55, "hr": 3, "min": 0 }
{ "steps": 0, "bpmMin": 107, "bpmMax": 107, "movement": 96,
  "battery": 52, "isCharging": false, "temperature": 35, "altitude": 416,
  "activity": "NOT_WORN",  // <-- Here!
  "bpm": 107, "hr": 3, "min": 10 }
```
At 03:10 at night the activity was considered "NOT_WORN" even though the temperature was >= 35° (so clearly worn). I suspect that Gadgetbridge will simply not show any HRM bar when the device is not worn (which makes sense) but the Bangle.js 2 internally categorizes the activity as not worn due to the activity being to low.

Hope this helps!

P.S.: I love the Bangle.js 2 and it is a VERY GREAT smartwatch!!!

gfwilliams:
Great, thanks for all that info! There will be some noise in the accelerometer but I'm surprised yours can get to 1.1 - most I see here is 1.03.

If with your code you're still seeing issues with recording, it's still possible to override the activity that's sent to Gadgetbridge in JS, so that could help. But maybe in Espruino I could add a check that says that if the watch temperature is over 30 degrees we assume it's worn even if it's not moving?

FabianK-Dev:
Sorry for the late response!

> But maybe in Espruino I could add a check that says that if the watch temperature is over 30 degrees we assume it's worn even if it's not moving?

Yes, I think that would be a very smart approach and is better than "blindly" checking if the user's movement is less than a given threshold.

Thank you very much!

gfwilliams:
Just done! https://github.com/espruino/Espruino/commit/0bf4b722e7bff9fe1bfc4afd43c2f35bf59df863

I didn't quite realise that we were just looking at movement without checking the angle too. Hopefully this recent fix will help but I guess we could start checking if the watch was lying flat as well - having all 3 checks would make it *very* unlikely we got false positives.

FabianK-Dev:
Awesome and thanks! Yes, you're right. All 3 checks at once would make it incredibly unlikely for false positives and very accurate.

PR #4205, opened 2026-04-08
feat(sleeplog): introduce multi-sensor 'Sleep Mode' and refine thresholds

FabianK-Dev:
### Description
This PR addresses false sleep detections (e.g., classifying stationary desk work as sleep) by allowing users to mandate both Movement and HRM thresholds simultaneously. It also adjusts settings bounds to allow finer tuning for users with higher resting movement baselines.

### Technical Implementation

**1. Multi-Sensor Logic (`boot.js`):**
Replaced the rigid `preferHRM` boolean tree with a calculated state approach. We calculate theoretical sleep statuses (`hStatus` and `mStatus`) for both sensors independently.
If `sleepMode === 2` (Both), the logic uses `Math.min(hStatus, mStatus)`. Since lower integers represent "more awake" states (2 = awake, 4 = deep sleep), the sensor that detects the most activity acts as an overriding failsafe, effectively filtering out false positives. If `data.bpm` is falsy, it automatically falls back to `mStatus`.

**2. Graceful Settings Migration (`settings.js`):**
To maintain backward compatibility with existing `sleeplog.json` configurations, a silent migration block was added. Upon opening the settings, if the legacy `preferHRM` key exists, it translates the boolean into the new `sleepMode` integer format (0 or 1) and cleans up the old key.

**3. UI/Thresholds:**
- Replaced the `preferHRM` toggle with a 3-state Choice Menu (Movement / HRM / Both).
- Increased `max` bounds for deep/light HRM thresholds up to 200 and adjusted `step` to 1 to allow high-granularity tuning.

**Checklist:**
- [x] Tested locally on Bangle.js 2.
- [x] Version bumped in `metadata.json`.
- [x] No changes to other apps included.

**App Loader Link:**
https://fabiank-dev.github.io/BangleApps/

### DIFF
--- a/apps/sleeplog/settings.js
+++ b/apps/sleeplog/settings.js
@@ -321,7 +321,7 @@
         value: settings.hrmDeepTh,
         step: 1,
         min: 30,
-        max: 100,
+        max: 200,
         wrap: true,
         noList: true,
         onchange: v => {
@@ -333,7 +333,7 @@
         value: settings.hrmLightTh,
         step: 1,
         min: 30,
-        max: 100,
+        max: 200,
         wrap: true,
         noList: true,
         onchange: v => {
@@ -373,7 +373,7 @@
       
       /*LANG*/"Deep Sleep": {
         value: settings.deepTh,
-        step: 10,
+        step: 1,
         min: 30,
         max: 500,
         wrap: true,
@@ -385,7 +385,7 @@
       },
       /*LANG*/"Light Sleep": {
         value: settings.lightTh,
-        step: 10,
+        step: 1,
         min: 100,
         max: 800,
         wrap: true,
--- a/apps/sleeplog/settings.js
+++ b/apps/sleeplog/settings.js
@@ -16,7 +16,7 @@
     lightTh: 300,//    threshold for light sleep
     hrmLightTh: 74,//    threshold for light sleep with HRM
     hrmDeepTh:60,//     threshold for deep sleep with HRM
-    preferHRM: false, // prefer hrm based sleep state determination
+    sleepMode: 0, // 0=Movement, 1=HRM, 2=Both
     wearTemp: 19.5, //    temperature threshold to count as worn
     // app settings
     breakToD: 12, //    [h] time of day when to start/end graphs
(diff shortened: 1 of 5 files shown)


apps/sleeplog/settings.js
```diff
@@ -26,6 +26,22 @@
...
+  // --- MIGRATION LOGIC (Added in v0.26) ---
+  // WHY: In v0.25 and earlier, the HRM setting was a simple boolean called 'preferHRM'.
+  // In v0.26, this was replaced by a 3-state 'sleepMode' (0=Movement, 1=HRM, 2=Both).
+  // WHAT: This block silently migrates existing users to the new format upon opening
+  // the settings, ensuring they don't lose their preference and the app doesn't crash.
+  // REMOVAL: This block can be safely removed in a future major update (e.g., v1.0 or 
+  // after ~1 year), once we can assume all active users have updated past v0.25.
+  // CONSEQUENCE: If removed, users updating directly from <=v0.25 to that future version 
+  // will simply lose their old 'preferHRM' preference and default to sleepMode 0.
+  if ("preferHRM" in settings) {
```
bobrippling, line 38: What if the user never opens settings? Do we need to migrate in `boot.js` ?

bobrippling (commented):
Thanks for the idea! One small question

apps/sleeplog/settings.js
FabianK-Dev: Thanks a lot. Great catch! If they just update via the App Loader and go to sleep, boot.js would fall back to sleepMode: 0 and they'd silently lose their HRM preference. I will push a commit to handle the runtime migration in memory inside boot.js asap.

FabianK-Dev: @bobrippling I added the following migration logic code and comments (https://github.com/espruino/BangleApps/pull/4205/commits/e4db114cb783b9935edefbd4171da13e33057781). Please let me know what you think about it.

bobrippling: Looks great, thanks!

FabianK-Dev:
Great! Thanks for merging!

PR #4208, opened 2026-04-11
rescalc: use shared color picker for band selection

stweedo:
Updates Resistor Calculator to use @RKBoss6’s shared color picker module for selecting resistor band colors.
 
~~This also updates the color picker module to support an emptyColor tile rendered as an empty box with a red X. Resistor Calculator uses that for the band 4 no-band / 20% tolerance option.~~

Tested in the app and bin/sanitycheck.js passes. If the shared colorpicker changes look good to @RKBoss6 and everyone else, this should be ready for review.

### DIFF
--- a/apps/rescalc/ChangeLog
+++ b/apps/rescalc/ChangeLog
@@ -2,3 +2,4 @@
 0.02: Fixes colors not matching user input from color menu in some cases, 3 bands are now shown larger, various code improvements.
 0.03: Use transparent icon with better visibility on dark backgrounds, new resistor img with darker outlines
 0.04: Fix capitalization. Improve decimal handling.
+0.05: Use the shared color picker for band selection.
--- a/apps/rescalc/app.js
+++ b/apps/rescalc/app.js
@@ -17,6 +17,7 @@ let colorData = {
   Silver: { multiplier: 0.01, tolerance: 10, hex: '#C0C0C0' },
   None: { tolerance: 20 },
 };
+let validTolerances = [1, 2, 0.5, 0.25, 0.1, 0.05, 5, 10];
 
 function clearScreen() { // Except Back Button
   g.clearRect(24, 0, 176, 24);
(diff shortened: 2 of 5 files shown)


stweedo:
If the changes to color picker for empty color are too niche it could be pulled back out.

bobrippling:
Looks good, shall give @RKBoss6 some time for a gander too

RKBoss6:
Thanks! This looks good to me - although I'm not sure of the use of the empty color function? Is it like a transparent color selection or something else?

stweedo:
Yeah, I am not sure either TBH. I really only added it as a way to select no color at all in the rare case where someone accidentally selected a color for band 4 and had no way to undo it besides backing out and trying again. I'm leaning towards it isn't necessary, as it's not that hard to use the back/menu button to reopen the same menu again with no selections.

On a side note, when i pick a color and it returns to the original menu, it currently resets the scroll position to the top. I'll want to fix that before merging as well.

stweedo:
I fixed a bug in the color picker module where both back and remove were wired to the same function, causing picker state to clear and options.back() to fire during UI teardown, which then triggered E.showMenu/setUI from inside a remove handler. 

I removed the `emptyColor` addition from this PR, so the shared `colorpicker` changes are now just the bugfix needed for this flow.

Scroll menu position is also preserved when selecting colors.

I think this is ready to merge if there are no objections.

bobrippling:
Sounds good to me - shall merge in the next day or so, just to give @RKBoss6 chance for another look

stweedo:
Thanks! And thank you @RKBoss6 for the module, it works really well for rescalc and even speeds up color selection.

RKBoss6:
Thank you @stweedo, everything looks good to me! One thing to note - currently the color picker module only supports touch, so there's no bangle.js 1 support. If you feel up for it, you can add the button support, or you can just leave it be, I might add support soon

stweedo:
No problem! I do not have a bangle.js 1 to test with. Maybe just make the top and bottom buttons cycle through the colors and highlight the selection with a border? Middle btn or touch selects whichever is highlighted. Though that might conflict with multi selection which i have not used yet.

bobrippling:
I'll merge and then we've got a clean stale for B1 support :)

And thanks both!

PR #4212, opened 2026-04-16
recorder: add track previews, accel chart, and popup improvements

stweedo:
This PR only affects apps/recorder/interface.html.

Changes:

- Add an approximate start-location preview in each track header when a valid GPS coordinate ~~pair~~ is found within the first 100 lines of the track file, as discussed in
  https://github.com/espruino/BangleApps/pull/3906
- Cache cumulative step totals for point popups so longer tracks do not rescan the full track on every click
- Add a dedicated acceleration chart
- Add acceleration values to the point popup
- Improve battery voltage chart readability by changing its color from yellow to blue

Before:

image

image

After:

image

image

image

### DIFF
--- a/apps/recorder/interface.html
+++ b/apps/recorder/interface.html
@@ -4,22 +4,26 @@
     <link rel="stylesheet" href="../../css/spectre-icons.min.css">
     <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css" />
     <style>
-      .leaflet-map { width: 100%; max-width: 600px; aspect-ratio: 1/1; margin: 0 auto; }
-      .chart-canvas { width: 100%; height: 100%; display: block; }
+      .leaflet-map { width: 100%; max-width: 600px; aspect-ratio: 1/1; margin: 0 auto; overflow: hidden; flex: none; }
+      .leaflet-map.leaflet-container img,
+      .leaflet-map .leaflet-tile,
+      .leaflet-map .leaflet-marker-icon,
+      .leaflet-map .leaflet-marker-shadow {
+        max-width: none !important;
+        max-height: none !important;
+      }
+      .chart-canvas { position: absolute; top: 0; left: 0; width: 100%; height: 100%; display: block; }
       .chart-wrapper { position: relative; width: 100%; max-width: 600px; aspect-ratio: 1/1; margin: 0 auto 10px; }
       .leaflet-map, .chart-container {
         border-radius: 4px;
         box-shadow: 0 2px 4px rgba(0,0,0,0.1);
       }
-      .chart-canvas { position: absolute; top: 0; left: 0; width: 100%; height: 100%; }
       .track-stats { font-size: 0.9em; color: #666; margin-top: 10px; padding: 8px; background: #f8f9fa; border-radius: 4px; }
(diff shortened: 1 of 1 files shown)


stweedo:
image

apps/recorder/interface.html
```diff
@@ -131,11 +169,28 @@
...
 function calculateDistance(lat1, lon1, lat2, lon2) {
   const R = 6371000, dLat = (lat2 - lat1) * Math.PI / 180, dLon = (lon2 - lon1) * Math.PI / 180;
   const a = Math.sin(dLat/2) ** 2 + Math.cos(lat1 * Math.PI / 180) * Math.cos(lat2 * Math.PI / 180) * Math.sin(dLon/2) ** 2;
   return R * 2 * Math.atan2(Math.sqrt(a), Math.sqrt(1 - a));
 }
+
+function renderTrackPreview(trackData) {
+  const clamp = (value, min, max) => Math.min(max, Math.max(min, value));
+  // Project the first valid GPS point into an OpenStreetMap tile for the compact list preview.
+  const lat = clamp(parseFloat(trackData.Latitude), -85.05112878, 85.05112878);
```
bobrippling: What's the clamp for?

apps/recorder/interface.html
```diff
@@ -742,16 +815,18 @@
...
-var lIdx = headers.split(",").indexOf("Latitude");
-if (lIdx >= 0) {
+var headerList = headers.split(",");
+var latIdx = headerList.indexOf("Latitude");
+var lonIdx = headerList.indexOf("Longitude");
+if (latIdx >= 0 && lonIdx >= 0) {
   var tries = 100;
   var l = data;
-  while (l && l.split(",")[lIdx]==="" && tries--)
+  while (l && (l.split(",")[latIdx]==="" || l.split(",")[lonIdx]==="") && tries--)
```
bobrippling: Do we need to check both? I think it's a fair assumption that if we have lat, we'll have lon and if we don't, it's a bug in the recorder, not here

bobrippling (commented):
Nice idea, the preview looks great! I have a few questions

apps/recorder/interface.html
stweedo: That is very true. I'll revert it back to just lIdx only.

apps/recorder/interface.html
stweedo: This checks if the coordinates fall into a valid latitude range. It does seem overly defensive and could be removed.

bobrippling: I'll leave that up to you, maybe some people have an artic trip planned!

stweedo:
I corrected the review findings and did a couple other minor reverts. Does the accel popup text look okay? Should it be at the bottom of the list maybe?

bobrippling:
Good question, it does look a little packed. Non-breaking spaces don't help as it gets shunted to the next line:

image

Tried to group together but there's probably not enough space:

image

So maybe it's good as-is?

---------

Also - can the checkboxes be vertically centered too?

image

stweedo:
Fixed. On mobile, the date drops to a new line. Does that look okay?

image

bobrippling:
LGTM, happy to merge if you are?

stweedo:
Yeah, it's ready to merge. Thank you for looking it over!

bobrippling:
In that case, thanks for the cool new features!

PR #4218, opened 2026-05-10
Add new app, countdown to desired date

rawyler:
Hello dear BangleApps community,

As you can see in the screenshots, the app is pretty simple and basic.
It just displays a live countdown in days, hours, minutes and seconds until the chosen date.
The date can be selected by opening the menu via `BTN`. It will be stored on the device.

Thanks in advance for a review
Thomas


thyttan:
Hi, thanks for the submission!

I wonder if you have considered offloading some of the logic to the scheduling library `sched`? That way you can also get a prompt when the date arrives even if the app is not running. And I think we would share code/lessen logic duplication across the apps on the watch. I haven't thought it through entirely but I believe so.

What do you think? 🙂

https://github.com/espruino/BangleApps/blob/master/apps/sched/README.md

https://github.com/espruino/BangleApps/blob/master/apps/sched/lib.js

rawyler:
Hi @thyttan,

Sure, I'll give it a try. I didn't know about `sched`, excuse me.

thyttan:
No worries at all! As I said thanks for submitting. If you arrive at the conclusion that it doesn't make sense just say. I wont hold it back over that point - but I think it would probably make sense.

rawyler:
Hey @thyttan,

I've had a look at `sched` and some BangleApps already using it. If I'm not mistaken, it's mainly used to register alarms. To register the goal date as an alarm seems a bit overkill.
Please let me know, if I misunderstood the `sched` library completely.

All the best,
Thomas

thyttan:
Ok! I'll test it out tomorrow and merge if everything works 🙂

thyttan:
Seems to work nicely - thanks! 😎

PR #4227, opened 2026-05-22
[Clock BG] Add gradient background

RKBoss6:
This PR adds a vertical gradient background that runs in 50-60 ms (average of 10 timed tries), and adds settings menu for the gradient.
Settings also uses the colorpicker module for plasma, gradient, and random colors, to get any custom color combination you want directly from the watch.

dd


RKBoss6:
This is good to go now!

thyttan:
I'll ask you @gfwilliams to review this one.

gfwilliams:
Gradients sound like a great idea - but there's some slightly crazy stuff in the code here (messing around with the colour string and parsing it base16) that I'm not super happy with.

* I believe you should be able to handle pretty much everything with `g.toColor` to parse the colours, and then `g.blendColor` if you need to get a color that sits somewhere between other colors.
* It also looks like some stuff is now disabled or for BANGLEJS2 only (what's the reasoning for this? colorpicker not working on Bangle.js 1?). I like the ability to choose colours but I think we should still have presets - for a lot of things where you have a range of colours, having a rainbow or gradient easily accessible seems really handy. 
* I also really want to avoid `if BANGLEJS2` stuff everywhere, if only because we're going to have to go through and change it all when Bangle.js 3 comes along. If colorpicker doesn't work on Bangle.js 1 maybe we need to fix that (just by making it work but it a very basic way).
* While I guess 50ms is acceptable, I think you should be able to do gradients a lot faster and tidier:

Just do kind of what you're doing, but pre-create a 4bpp greyscale image 
gradient
 (use https://www.espruino.com/Image+Converter and output as "Image Object") then add the palette.

And can work out the palette with blendColor super quickly: `bg.palette=(new Uint16Array(16)).map((n,i)=>g.blendColor("#f00","#0f0",i/16));`

```
bg = {
  width : 16, height : 16, bpp : 4,
 palette : (new Uint16Array(16)).map((n,i)=>g.blendColor("#f00","#0f0",i/16)),
  buffer : atob("AAAAAAAAAAARERERERERESIiIiIiIiIiMzMzMzMzMzNERERERERERFVVVVVVVVVVZmZmZmZmZmZ3d3d3d3d3d4iIiIiIiIiImZmZmZmZmZmqqqqqqqqqqru7u7u7u7u7zMzMzMzMzMzd3d3d3d3d3e7u7u7u7u7u//////////8=")
}
```

Above uses a simple 16x16x4bpp gradient with no noise - but I see you added some randomness so you could do that in the image too. Worth noting that on Bangle.js 3 you'll have 6bpp with dithering which'll be able to render this stuff much better.

... and maybe since it's then much faster, you could use more than 16x16px for the image which might make the dither look better?

RKBoss6:
Thanks for all the pointers! Right now, the color picker module doesn't support bangle.js 1, so perhaps i'll get that in so we don't limit it by device. But for both devices, it still keeps the presets there and just adds a 'custom' option above them. Out of curiosity, how are you planning to make the pixel mapping? Is it going to be where 0,0 is an imaginary point off of the screen where the rectangle of the height and width meet, or will it be in the center when not in compatibility mode or something else entirely?

gfwilliams:
> Right now, the color picker module doesn't support bangle.js 1

It feels like you could have a variable for the currently selected item, and then use the `btn` handler to increment/decrement it (if the first argument==1 or ==3) and then accept if ==2. You wouldn't even have to check what Bangle you were on since on Bangle.js 2 the first argument won't be 1,2 or 3.

> But for both devices, it still keeps the presets there

oh ok, I just saw this bit here and it looked like that wasn't the case: https://github.com/espruino/BangleApps/pull/4227/changes#diff-76ca0acb9bf7ae1fc186131a1205c12d36890916228b56f4b223f5d24beb0cf4R106-R120

> Out of curiosity, how are you planning to make the pixel mapping? Is it going to be where 0,0 is an imaginary point off of the screen where the rectangle of the height and width meet

On Bangle.js 3? Yes, that's the plan. If you get a touch event and you draw something at the touch coordinates, it should be where your finger is - I think that's the only sensible option.

In the compatibility mode there's a 176x176 square of graphics you use and the touch coordinates are nudged to fit.

If you go into the developer console when on the web IDE and paste in ` Espruino.Config.set("EMU_BETA", 1);` then when you connect to the Emulator you'll get an option for Bangle.js 3 if you want to try things out. It's still very easrly days as I don't have any kind of real hardware yet though.

RKBoss6:
Just wanted to say - thanks for all the suggestions, I tried them out and it's an average of 18 seconds! Much better than the ~60s before. For dithering, my plan is to have 2-3 grayscale images with the dithering baked in, and then just randomly choose between them for the semblance of randomness.

I do like the 16px look though - and it matches with squares, plasma, etc.

RKBoss6:
Just a comparison between the faster version and the slower older one - the faster one shows less of each true color at the top and bottom, and more of the gradient. Not sure if it matters to anyone, but just putting it out there.
New: 
download (9)
Old:
dd

gfwilliams:
Great! You could probably adjust the input image if you wanted to make it more even at the top. It's possible it's actually an issue with the Bangle.js 2 dithering not rounding correctly though, in which case I guess it could be fixed in the firmware.

Just putting it out there but there's no reason to have a single vertical gradient - you could have horizontal, diagonal, radial - even totally different patterns. But if doing that I don't think we want the images in the library file or it'll hurt load times for all backgrounds - you could actually store the image to use as a base in the settings file.

RKBoss6:
Settled on 5 styles, and it works great! Here's some of the styles:

download (10)
download (11)

download (133)

All of them run in less than 20 ms, and uses precomputed images in settings :)

apps/clockbg/settings.js
```diff
@@ -9,10 +9,71 @@
...
     if (settings.style!="color")
       delete settings.color;
-    if (!["randomcolor","squares","plasma","rings","tris","blobs"].includes(settings.style))
+    if (!["randomcolor","squares","plasma","rings","tris","blobs","gradient"].includes(settings.style))
       delete settings.colors;
     require("Storage").writeJSON("clockbg.json", settings);
   }
+  
+function colorRange(from, to, steps) {
+    function parse(hex) {
```
gfwilliams, line 18: Please can you replace this whole thing with:

```
function colorRange(from, to, steps) {
  var result = [];
  for (var i = 0; i < steps; i++) 
    result.push(g.blendColor(from,to,i / (steps - 1));
  return result;
}
```

I'm pretty sure that's all you need since the colors don't have to be strings - they can just be the int values that blendColor creates.

apps/clockbg/lib.js
```diff
@@ -66,7 +66,18 @@ exports.reload = function() {
...
     settings.imgOpt = {scale:g.getWidth()/16};
-  } else if (settings.style=="blobs") { // ~25ms
+    
+  } else if (settings.style=="gradient") { // ~20ms
+    settings.style = "image";
+    let c = settings.colors.reverse();
+    let bg = {
+      width: 16, height: 16, bpp: 4,
+      palette: (new Uint16Array(16)).map((n, i) => g.blendColor(c[0], c[1], i / 15)),
+      buffer: E.toArrayBuffer(atob(settings.image||"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAERESISEREhEiMyMjIzMjMkRERERERERUVWZWVVVWZVZ3d2d3dnd3dpiIiYmIiIiImamqmpqqqamru7u7u7u7u8zdzN3czczM3u3e7d7t7e3///////////////////////////////8="))
```
gfwilliams, line 76: Is there ever a time when we can get gradient set but not `settings.image`? Maybe we can just replace it with `E.toArrayBuffer(atob(settings.image))`?

gfwilliams:
> Settled on 5 styles, and it works great! Here's some of the styles:

That looks awesome, thanks! It'll be nice to get this in - I think it might end up being my new clock background!

I can have a look at making the color picker work on Bangle.js 1 if you want?

RKBoss6:
Yeah actually - that would be great! I've just been unsure of how to make the multi-select option work, so of that's something you want to do, please feel free!

gfwilliams:
> Yeah actually - that would be great! I've just been unsure of how to make the multi-select option work

Just fixed in 4470f9facc9ad2ecc6135e89a23058c387293eae - in multiselect mode I remove the brown color and replace it with a back icon (I think this makes some sense on Bangle.js 2 as well). I also show the index of selected colors as well, as that could make a difference in some cases (eg which direction the fading runs in): 
image

I've just tweaked this and merged it in now.

Just a thought but as colorpicker supports multi-select, you could add an option to colorpicker to limit the max number of colours (eg to 2) and then you could avoid needing the customMenu function with selection of foreground/background?

RKBoss6:
Yeah, thanks for all the changes, that actually sounds really nice - perhaps in a new PR!

gfwilliams:
Just to add I'm really loving the gradient backgrounds, and using them on my main watch now.

RKBoss6:
:) me too! Hopefully with the Bangle.js 3, the gradients will also appear more striking with all the colors it can display!

PR #4230, opened 2026-05-23
widadjust: fix text display

KTibow:
Current rendering is utterly glitched, at least in dark mode
download
This PR switches to sane defaults and bumps the version


bobrippling:
Nice, thanks!

Issue #4217, opened 2026-05-10
[Feature request] [Alarms] Put hidden alarms in submenu


thyttan:
### Description
Add a submenu to the Alarms app called "Hidden alarms" where all alarms with the `hidden` attribute set are relegated to.

### Background

**EDIT:** I realized I should not add 27 Alarms but just keep adjusting/snoozing a singel alarm instead. Making this feature request not feel as important.

I'm trying out refactoring the `twenties` app (vibrate the watch every 20 mins to stand up and rest your eyes) to use the scheduling library to setup vibration intervals. Instead of using it's own `setTimeout` calls. ~~This adds 27 new "hidden" alarms that should be set and forget. Today that makes the main menu of the Alarms app crowded and noticeably slower to load. Filtering out hidden alarms would help with accessing alarms I actually want to interact with on a more regular basis. And hopefully speed up loading the menu too?~~

Diff against master here: https://github.com/espruino/BangleApps/compare/master...thyttan:BangleApps:twenties?expand=1

### Considerations

~~Is the way of using the scheduling library as I describe above in the background section a good or bad idea? Will adding a bunch of alarms that way lead to unwanted slowdowns when parsing alarms in other places (e.g. during boot b/w loads)? Maybe it's better to just keep using setTimeout to not introduce complexity in a bunch of places?~~

Instead of using the `hidden` attribute (that today just hides that alarms widget AFAIK) should a new attribute be introduced for hiding it from the main menu?

I'm curious to hear your thoughts on this @gfwilliams when you have the time.

thyttan:
For the `twenties` app I realized I should not add 27 Alarms but just keep adjusting/snoozing a single alarm instead. Making this feature request not feel as important.

gfwilliams:
Yes, definitely having the one alarm you reschedule sounds good!

But as you say it looks like it would be a good idea to make `hidden` hide alarms/timers from the menu (if it's set). It looks like the only app that actually uses `hidden` is `smpltimer` and it makes sense to hide it in that case?

I'd say maybe make `showMainMenu` hide alarms by default, and then add an option to show them in showAdvancedMenu? https://github.com/espruino/BangleApps/blob/master/apps/alarm/app.js#L563

Also looking at it, there is an option to show alarm `groups` which could help in the case you mention. Although looking at the code there was a bug plus it's undocumented in `sched` and defaults to `false` so I wonder how well the option actually works! I've tried to put a fix in though

[...]

PR #4231, opened 2026-05-25
Twenties: refactor to use the scheduling library instead of its own timeouts

thyttan:
Work in progress.

Try the current iteration at: https://thyttan.github.io/BangleApps/?id=twenties

### DIFF
--- a/apps/twenties/ChangeLog
+++ b/apps/twenties/ChangeLog
@@ -5,3 +5,4 @@
 0.05: Improve energy efficiency
 0.06: Align with whole thirds of the hour.
 0.07: Fix TypeError on scheduling for the next day.
+0.08: Refactor to use the scheduling library.
--- a/apps/twenties/README.md
+++ b/apps/twenties/README.md
@@ -17,6 +17,7 @@ Only vibrates during work days and hours.
 Add settings page.
 Add setting for active days and hours.
 Add setting for buzz strength/pattern.
+Add entry in settings to re-run setup.
 
 ## Contributors
 
--- a/apps/twenties/README.md
+++ b/apps/twenties/README.md
@@ -12,6 +12,7 @@ Vibrates to remind you to stand up and look away for healthy living.
 
 Only vibrates during work days and hours.
 
-## Creator
+## Contributors
 
-[@splch](https://github.com/splch/)
+[@splch](https://github.com/splch/) (Creator)
+[@thyttan](https://github.com/thyttan/)
--- a/apps/twenties/README.md
+++ b/apps/twenties/README.md
@@ -12,6 +12,12 @@ Vibrates to remind you to stand up and look away for healthy living.
 
 Only vibrates during work days and hours.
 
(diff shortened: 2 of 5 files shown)


elfehr:
Thanks for working on that!
Maybe you could add `msg: "Twenties"` to the alarm, to make it clear where it comes from so users are less likely to delete it? I wonder if there is a way to hide it completely or make it undeletable. Or maybe recreating it on boot if it has been deleted?

For curiosity, how are you testing the code in the IDE? On my side, it doesn't read the file from storage correctly, and if I upload the unminified javascript (either to RAM or to storage, replacing the original), I have an error that `sched` is not found and `exports` is not defined.

thyttan:
> if I upload the unminified javascript (either to RAM or to storage, replacing the original), I have an error that `sched` is not found and `exports` is not defined.

I get those as well, but it works anyways. When the code runs on the watch it will find the `sched` library file and use it (if you installed it before). But it's probably good to get that error when uploading so you know to get the library if you haven't already.

The `exports` not being defined is similar in that when running on the watch it will work. Maybe the Web IDE should be made to not fire that error? @gfwilliams

thyttan:
> I wonder if there is a way to hide it completely or make it undeletable.

Yes I think that makes sense. Gordon and I discussed it some in #4217

> Or maybe recreating it on boot if it has been deleted?

I did that before. I prefer to avoid it. But maybe it makes sense.

gfwilliams:
> Or maybe recreating it on boot if it has been deleted?

I think if we can hide it with `hidden:true` in the alarm, it would pretty much solve the problem without having to worry about that.

> Maybe the Web IDE should be made to not fire that error? 

Well, it's only a warning. It's hard because in most non-Bangle.js devices it's pretty helpful.

It's a bit of a pain to work around, but in the dev IDE (https://espruino.github.io/EspruinoWebIDE/) I've just added a change that means it'll now search `Storage` for modules when you connect to a device, which should fix this at the expense of a slightly longer connection time.

elfehr:
Thanks for all your work on that app! I've been running it from your version of the App Loader for the last two weeks and it runs perfectly.
The only thing I noticed is that it doesn't buzz at 18:00:20. Since 18:00 is not work time anymore, the next buzz is directly scheduled for 08:00.

thyttan:
People must think it weird seeing you look out the window from 6pm to 8am the next morning every night! 😉

Thanks for testing it out. I'll merge this once the separate PR for alarm app changes is merged. 🙂

thyttan:
@elfehr I just merged the PR. It got some changes since you installed it from my app loader. So please reinstall from here now: https://espruino.github.io/BangleApps/?id=twenties

Also, I left this as is for now: 
> The only thing I noticed is that it doesn't buzz at 18:00:20. Since 18:00 is not work time anymore, the next buzz is directly scheduled for 08:00.

Feel free to do a PR with a change/fix if you want 🙂

PR #4236, opened 2026-06-01
[twenties] Fix TypeError on scheduling for the next day

elfehr:
Twenties v0.06 introduced a bug that prevents it to set a timer for the next work day.
This error message appears in the IDE after the last buzz of the day:
```
Uncaught TypeError: Variables of type Object are not supported in date constructor
    at .boot0:52:38
            const next = new Date(now);
```

### DIFF
--- a/apps/twenties/ChangeLog
+++ b/apps/twenties/ChangeLog
@@ -4,3 +4,4 @@
 0.04: Convert to boot code
 0.05: Improve energy efficiency
 0.06: Align with whole thirds of the hour.
+0.07: Fix TypeError on scheduling for the next day.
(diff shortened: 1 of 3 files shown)


thyttan:
Thanks! I presume this is tested and works well?

elfehr:
Thank!
The error message is gone, but now I wonder if I haven't found another bug. 
`next` is set to today 8am, which is included in work time, so the date is not incremented. With print statements in the IDE, I see that `next - now` is negative, `setTimeout(scheduleNext, next - now)` seems to timeout immediately, and the function is called constantly.

thyttan:
Hm. Since I'm refactoring twenties to use the scheduling library anyways would you mind trying out that one instead: #4231

elfehr:
Alright, having a look now.

Otherwise, `Date(now+864e5)` (now+24h) seems to fix it, and for some reason it doesn't trigger the same TypeError as `Date(now)` does.

thyttan:
Feel free to open another PR if you want 🙂

PR #4237, opened 2026-06-01
[twenties] Schedule for next day

elfehr:
After work hours, schedule for tomorrow 8am rather than today 8am. Otherwise, the negative timeout seems to trigger instantly and the function `scheduleNext` is constantly running in the background until the next work day.


thyttan:
Thanks!

PR #4238, opened 2026-06-01
Add auto-populated authors

RKBoss6:
Adds auto-populated authors by using the first commit author of each app. Some apps have an author who's username was changed, or the account was deleted, so those apps were ignored and unchanged. More info can be found in the [discussion](https://github.com/orgs/espruino/discussions/7968#)


gfwilliams:
As we're doing this, please could you try and get the `author` field a bit futher up the JSON rather than right at the end?

When @bobrippling did this for https://github.com/espruino/BangleApps/pull/3994/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 he managed to get it after version/before description which looks good.

There's the script he used in there which does it already, so we could use that (I think the issue was just GitHub API rate limiting?)

RKBoss6:
Should be good to go now!

gfwilliams:
Great - thanks!

PR #4239, opened 2026-06-01
Add check to make sure we have an author field in metadata

RKBoss6:
Add error handling for missing author field in app metadata.


gfwilliams:
Does this need rebasing now all the authors are in? It seems like it's failing with errors for apps that do have an author in now

RKBoss6:
I think it's still failing for apps that don't have an author - even with the rebase, because not all apps will have an author if, say, the username couldn't be found. Will it do this for every pr, or does this one just have all apps because no app was changed directly?

gfwilliams:
Just added a tweak to stop the duplicate errors.

No, it'd be in every PR, so we can't merge this right now. We could add exceptions but I think really first we just need to add the author field to all the apps.

I'd have thought you could modify your original script to do this? Just `git blame apps/widpa/ChangeLog` for example gives you a pretty good idea who the app's main author is - I'd just choose the person with most lines

RKBoss6:
Well most of the apps that don't have authors after #4238 just don't exist on github anymore, so the commit id associated with the commit is no longer present, and rather than putting `unknown` it makes more sense to leave it blank. Is there a way to only have that part of the check fire on any app whose files have been changed in the PR?

gfwilliams:
Well, we can add exceptions for the old apps, but I think in the app loader either we're assume we always have author, or we have to deal with the fact it may not exist - and if we're coping with it not existing maybe we shouldn't really be requiring it of new submissions.

What if we allow it to be an email as well? surely we can pull the email address out of GitHub commits?

RKBoss6:
I don't know - in my mind it's best to have the author field as strict to one type of ID as possible. If we allow emails too, then it might snowball into actual names, etc, and nobody will know what to search for when they want to search for an author. If we add `author: undefined` into the apps with no author, will that get around this issue, or is it better to come up with another solution?

thyttan:
Just chiming in I think allowing email is sensible to me. 

To me the main thing is getting a hold of someone that has a degree of knowledge about the workings of an app to help review changes. So they don't break things or change the user experience in a way that the app becomes less useful to other users. Having an email in the author field helps with this in the same way as a Github username does.

Also if BangleApps were to move on to another hosting website it might end up making sense to switch out __ALL__ Github usernames to emails in an effort to keep in touch with contributors. But that's probably not likely for a while at least.

gfwilliams:
Good point about not being Github-specific actually. Even if we stay with Github we do occasionally have users that want to contribute stuff but don't want a Github login. It's rare though...

...but then, some people (probably the same ones that won't sign up to Github) don't want their emails publically available, so :shrug: 

Actually for now, let's just do: `"author":"espruino"`

RKBoss6:
Should be good to go with #4260, if we could get that merged in.

Issue #4220, opened 2026-05-14
[sched/alarm] Timers trigger immediately if they would expire after midnight (possible regression of  #2712)


ticalc-travis:
### Affected hardware version

Bangle 2

### Your firmware version

2v29

### The bug

Unfortunately, it does appear that fixing issue #2712 has introduced a new issue which I have somehow only just now run into. If I set a countdown timer that ends after midnight (for instance, it's currently 23:50 and I set a timer for 15 minutes), it now triggers an alarm right away instead of when it should. Perhaps *that* is what the check for whether the alarm was not 60 seconds in the past was for. :)

Any ideas what the best way to address this would be? Maybe make sure when we create a timer to set the `.last` and/or `.date` fields to make sure the timer will take midnight into account?

thyttan:
Tagging @gfwilliams

gfwilliams:
Thanks for the report! Yes, just tried here and it's super easy to reproduce. I'll get a fix in now.

I reckon setting `last` is a good call. Looks like we do actually attempt to do that in resetTimer, so the code there must be broken

edit: forget that - while setting `last` would fix it, if you were to set a timer 2 days in the future it would still break - so let's do `date`

ticalc-travis:
Thanks, good call. I wasn't even aware that timers wouldn't have handled being longer than 24 hours correctly before, so handling `date` feels like the right thing to do.

PR #4240, opened 2026-06-02
Stop new alarms set before current time of day from triggering…

### LINKED ISSUE #4232: [sched] Setting an alarm for an earlier time of day makes the alarm fire immediately.
### Affected hardware version

Bangle 2

### Your firmware version

2v29.75

### The bug

Setting an alarm using the sheduling library, for a time that already passed today makes the alarm fire immediately.

I am not sure if this is actually a bug. It may be the intended behavior.

My expectation would have been for the alarm to fire at the set time the next day instead.

Setting an alarm at such a time through the alarms app does indeed set the alarm for the next day instead. I am unsure of how this is achieved.

Here're the lines responsible for the behavior: 
https://github.com/espruino/BangleApps/blob/b9038c87970ed833695e60d2475701aa8e774eed/apps/sched/boot.js#L19-L21

Here's a diff with a fix:
https://github.com/espruino/BangleApps/compare/master...thyttan:sched?expand=1

However that fix seems to go against recent changes to fix other bugs in issues #2712, #4220.

I wonder if you @ticalc-travis have any immediate thoughts regarding this since you seem to have thought about the scheduling library quite a bit not that long ago.

### Reproduce

Use the Web IDE to upload the following script to RAM.

```js
{
  S = require("sched");
  const TIME_AT_NEXT_BUZZ = 10 * 1000; //10 seconds after midnight.
  S.setAlarm("test", {
    on: true,
    t: TIME_AT_NEXT_BUZZ,
    //dow: 0b1111111, // all days of the week.
    //last: Date().getDate(), // Don't fire until tomorrow.
  });
  S.reload();
}
```

### Background

I'm working on making the `twenties` app use the scheduling library instead of setting it's own timeouts. Draft PR at #4231. That's when I happened upon this behavior. Notably I set and make use of the `.js` attribute when scheduling the alarm.

ticalc-travis:
instantly after #4220 fix

Proposal to fix #4232. May need some review and testing.

### DIFF
--- a/apps/sched/lib.js
+++ b/apps/sched/lib.js
@@ -36,6 +36,10 @@ exports.setAlarm = function(id, alarm) {
     if (alarm.on!==false) alarm.on=true;
     if (alarm.timer) { // if it's a timer, set the start time as a time from *now*
       exports.resetTimer(alarm);
+    } else { // If it's an alarm, default to triggering tomorrow if time < current time of day
+      if (alarm.last===undefined) {
+        alarm.last = alarm.t < require("time_utils").getCurrentTimeMillis() ? new Date().getDate() : 0;
+      }
     }
     alarms.push(alarm);
   }
--- a/apps/sched/ChangeLog
+++ b/apps/sched/ChangeLog
@@ -40,3 +40,4 @@
 0.37: Ensure we always execute an alarm, even if it was >1 minute in the past. This fixes missed alarms when >1 alarm is scheduled next to each other (fix #2712)
 0.38: Fix timer going off immediately if it was long enough it went to the next day (#4220)
 0.39: Allow a press of the button to stop/snooze the alarm when it's gone off
+0.40: Prevent time-of-day alarms set via setAlarm() that have a time earlier than the current time triggering immediately after #4220 fix; schedule them for tomorrow by default to match original behavior
--- a/apps/sched/metadata.json
+++ b/apps/sched/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "sched",
   "name": "Scheduler",
-  "version": "0.39",
+  "version": "0.40",
   "author": "gfwilliams",
   "description": "Scheduling library for alarms and timers",
   "icon": "app.png",


apps/sched/lib.js
```diff
@@ -36,6 +36,10 @@ exports.setAlarm = function(id, alarm) {
     if (alarm.on!==false) alarm.on=true;
     if (alarm.timer) { // if it's a timer, set the start time as a time from *now*
       exports.resetTimer(alarm);
+    } else { // If it's an alarm, default to triggering tomorrow if time < current time of day
+      if (alarm.last===undefined) {
+        alarm.last = alarm.t < require("time_utils").getCurrentTimeMillis() ? new Date().getDate() : 0;
```
thyttan, line 41: This library seem to avoid requiring time_utils and do 
```
  var time = new Date(), // current time
      currentTime = (time.getHours()*3600000)+(time.getMinutes()*60000)+(time.getSeconds()*1000);
```
instead (line 47 (old) / 51 (new)).

Maybe that's better from a performance perspective not needing to pull in another library? (time_utils adds ~60 % more code to load, 80 lines)

If that's the case maybe we make a small internal helper function?: 
```
const getCurrentTimeMillis = ()=>{
  var time = new Date(); // current time
  return (time.getHours()*3600000)+(time.getMinutes()*60000)+(time.getSeconds()*1000);
}
```

gfwilliams: Yes, I wanted to avoid dragging in all that extra code for one call.

A helper function sounds like a good idea. Maybe pass in `time` as an argument though since in some of the functions where we do this we may not always be using the current time

gfwilliams:
Thanks! So am I right in thinking this wasn't actually something that was broken after https://github.com/espruino/BangleApps/issues/4220  - it's just been like it all the time?

This seems like the right way to fix it (but maybe we avoid calling into the time_utils lib here to save us dragging in all that code) - as in @thyttan's comment above.

Please could we also remove the (now) duplicated code from the alarms app at the same time though?

ticalc-travis:
I believe this likely did start becoming a problem after the #4220 fix. This situation was likely prevented before by the check in the boot code that ignored alarms that were set to trigger before the current time. But since that also caused it to ignore alarms that happened to become overdue somehow and needed triggering ASAP (which #4220 fixes), this means that alarms without the `last` attribute that are already “overdue” when set now fire immediately instead of carrying that assumption to wait until the next day. Automatically adding the `last` attribute on `setAlarm` to ensure that the alarm rolls over to the next day if the time of day is before now is how we're correcting this in this PR without losing the #4220 fix.

Do you think removing the duplicated code from Alarms now will cause problems in case someone updates one of Alarms/Sched but not the other? Or is it safe to assume that most people always update everything at once?

I plan to look into this and the other comments once I get a chance.

gfwilliams:
Ahh, yes - of course - that makes sense about removing the early alarm check.

> Or is it safe to assume that most people always update everything at once?

Yes, I think so. I think it's fair that we say to people that if you have newer versions of some things and older versions of others you can expect issues.

gfwilliams:
Just made some tweaks and merged this, but we can't remove the code form the alarms app yet: https://github.com/espruino/BangleApps/issues/4241

Issue #4241, opened 2026-06-04
[alarm/android/tevtimer/others] apps using `require("sched").setAlarms` may suffer from alarm/timer bugs


gfwilliams:
Related to https://github.com/espruino/BangleApps/issues/4232

fyi: @thyttan and @ticalc-travis

Basically there are some subtleties around putting an alarm/timer in the `alarms` array, which are now covered by `require("sched").updateAlarm`.

For instance:

* Ensuring that timers fire on the next day if `currentTimeOfDay + timer > 24`
* Ensuring alarms < current time fire on the next day

We handle this in `require("sched").updateAlarm`, which is called automatically from `require("sched").setAlarm`.

But it's NOT called from `require("sched").setAlarms`, because we don't know which alarms and new and which are old.

So right now we're in this situation where apps don't end up using `updateAlarm`, but work around the bugs in their own unique ways which may/may not work reliably.  

- [ ] `alarm` has `prepareXForSave`
- [ ] `android` has a line like `last = (event.d[j].h * 3600000 + event.d[j].m * 60000 < currentTime) ? (new Date()).getDate() : 0`
- [ ] `tevtimer` doesn't appear to do any workarounds so I think will have issues if you snooze a timer near mignight
... and probably others that use setAlarms

The safe thing is to use `setAlarm` where possible, but I don't think it makes sense for the `alarms` app - I think we need to call `updateAlarm` before we stick an alarm in the `alarms` array, and ensure we remove existing workarounds from the codebase

ticalc-travis:
I can confirm that tevtimer has a problem snoozing timers near midnight after these recent changes, so I'm working on adding a call to `updateAlarm`. This almost works, but for some weird reason the alarm seems to trigger again well before the snooze interval, even though the listed time for the timer in the Alarms app after snoozing shows up correctly.

sched also currently has a problem with snoozed alarms. In addition to adding a call to `updateAlarm`, we'll need to do something like delete the `last` attribute from the alarm on snooze so that `updateAlarm` will update it accordingly (if it's already defined, `updateAlarm` leaves it alone, so a call to `updateAlarm` alone doesn't fix the problem). With these changes, snoozing alarms over midnight looks like it would work properly for sched's alarm handling.

I plan to make a PR for these issues once I finish testing and figuring out why snoozing in tevtimer with the fix still behaves kind of weird.

gfwilliams:
> I plan to make a PR for these issues

Thanks! That would be amazing.

[...]

PR #4242, opened 2026-06-04
[alarm] tweak hidden logic

thyttan:
alarm change: 
> Hide hidden alarms from main menu by default, change in settings, see readme.
> Hide the `Hidden` setting from "New..." menu by default, change in settings.

~~sched change:~~
> ~~If alarm.hidden is set and not alarm.group then set group to "Hidden".~~

These changes are part of https://github.com/espruino/BangleApps/pull/4231/

Asking @gfwilliams for review when you get the time. Especially line 94-98 in alarm app.js.


apps/alarm/app.js
```diff
@@ -90,8 +91,11 @@ function showMainMenu(scroll, group, scrollback) {
   const getIcon = (e)=>{return e.on ? (e.timer ? iconTimerOn : iconAlarmOn) : (e.timer ? iconTimerOff : iconAlarmOff);};
 
   alarms.forEach((e, index) => {
-    const showAlarm = !settings.showGroup || (group ? e.group === group : !e.group);
-    if(showAlarm) {
+    const showAlarmInMainMenu = (!e.hidden===true || settings.showHiddenAlarms) &&
+      !(group && settings.showGroup);
+    const showAlarmInGroupMenu = settings.showGroup  &&
+      (group ? e.group === group : false);
+    if(showAlarmInMainMenu || showAlarmInGroupMenu) {
```
thyttan: I spent an unreasonable amount of time thinking about this logic - and I'm not sure if I managed to get it right even then. So please take extra care when reviewing.

thyttan: The idea is `showAlarmInMainMenu` and `showAlarmInGroupMenu` should not be true at the same time.

Also that hidden alarms will be hidden from main menu, but not from group menus.

gfwilliams:
Thanks! I'll try and look through this in a lot more detail later!

But for `sched`, do we need that change about setting the group if it's hidden? Because I feel like the less stuff we put in the alarm JSON if it can be implied the better - and due to https://github.com/espruino/BangleApps/issues/4241 we can't even be sure it'll always get set correctly (it definitely won't be set right after someone updates the alarm app). 

Worst case the setting should go into `.updateAlarm` - but ideally we just do without it, and in alarms we do: `function getGroup(alarm) { return e.group||(e.hidden?"Hidden:undefined) }` or something like that?

thyttan:
> but ideally we just do without it, and in alarms we do: function getGroup(alarm) { return e.group||(e.hidden?\"Hidden:undefined) } or something like that?

Yes, nice solution 👍

apps/alarm/app.js
```diff
@@ -90,8 +91,12 @@ function showMainMenu(scroll, group, scrollback) {
   const getIcon = (e)=>{return e.on ? (e.timer ? iconTimerOn : iconAlarmOn) : (e.timer ? iconTimerOff : iconAlarmOff);};
 
   alarms.forEach((e, index) => {
-    const showAlarm = !settings.showGroup || (group ? e.group === group : !e.group);
-    if(showAlarm) {
+    e.group = e.group||(e.hidden?"Hidden":undefined);
```
thyttan: I just did it like this without making a function first, since it seemed like it's not needed anywhere else?

gfwilliams: Yes, no need to be a function if it doesn't need to - but maybe just store it in a var rather than setting `e.group` as any changes made here will be saved back if someone changes an alarm as far as I know?

apps/alarm/app.js
```diff
@@ -107,7 +112,7 @@ function showMainMenu(scroll, group, scrollback) {
         format: v=>getIcon(e)
       };
     } else if (getGroups) {
-      groups[e.group] = undefined;
+      groups[E_GROUP] = undefined;
```
thyttan: Not sure, but does this become redundant now?

apps/alarm/app.js
thyttan: > but maybe just store it in a var rather than setting `e.group`

Done!

apps/alarm/app.js
gfwilliams: I'm pretty sure we need this - I think it's used to get a list of groups that are then used to create a menu later on.

apps/alarm/app.js
gfwilliams: The logic looks good to me I think. Does it seem to work ok?

apps/alarm/settings.js
```diff
@@ -33,14 +34,18 @@
         value : !!settings.showConfirm,
         onchange : v => { settings.showConfirm=v; save();}
       },
-      /*LANG*/'Show Menu Hidden': {
-        value : !!settings.showHidden,
-        onchange : v => { settings.showHidden=v; save();}
+      /*LANG*/'Show Menu Hidden Toggle': {
```
gfwilliams: I feel like `Show Menu Hidden Toggle` doesn't make that much sense to a user? Maybe keep it as `Show Menu Hidden`?  I'm not sure it's better but it's not worse, and it's then the same as it was :)

Perhaps we should rename `Show Menu X` to `Show X Option`? That whole thing is a bit confusing...

thyttan: I think I undid that name change everywhere now at least. Not sure what's best for 
> rename Show Menu X to Show X Option?

gfwilliams: Thanks! Well, no point complicating this PR - we can just do that another time. 

I'm not sure if I really see the need to turn them on/off individually anyway. Maybe at some point we just change it to "Advanced Alarm Options" and use that to show/hide all this stuff (or even just put it in an `Advanced` menu for every alarm and always have the option there).

... anyway, that's definitely something for later :)

gfwilliams:
This all looks good to me! Assuming it's tested and works ok - with hidden alarms, and alarms in different groups - then I'm happy to merge

thyttan:
I need to test some more 👍

thyttan:
Still work in progress.

thyttan:
I did some changes after thinking some more. 

E.g. merged `showHiddenAlarms` into the `showHidden` setting.

It now works like this (and is tested to verify):

### showHidden == false && showGroups == false:
Hidden alarms will not show at all.
Grouped alarms will show in main menu.

### showHidden == false && showGroups == true:
Hidden alarms will not show anywhere (the "Hidden" group will not be accessible, unless a non-`.hidden` alarm explicitly states it belongs to it).
Grouped alarms will show in group menus (if they are not `.hidden`).
A group will not get a menu if it only has `.hidden` alarms.


### showHidden == true && showGroups == false:
Hidden alarms will show in main menu.
Grouped alarms will show in main menu.

### showHidden == true && showGroups == true:
Ungrouped hidden alarms will show in the "Hidden" group menu.
Grouped alarms will show in group menus, also if they are `.hidden`.

thyttan:
I'll keep this in draft for a day or two if I should notice or come up with anything more.

gfwilliams:
Thanks! That sounds good!

Another option to consider: What if we don't have an option to show hidden alarms, but instead just have a menu item under `Advanced` called `Hidden Alarms` and then any hidden alarms are shown under there?

Then most people won't ever go down to the bottom of the Advanced menu, but if you do want to look it's there without having to mess with settings?

thyttan:
> Another option to consider: What if we don't have an option to show hidden alarms, but instead just have a menu item under Advanced called Hidden Alarms and then any hidden alarms are shown under there?

In that case, should the "Show Menu Hidden" do the same as before and still show/hide the "Hidden" option when creating a new alarm in the app? Or do we remove that setting altogether?

gfwilliams:
> should the "Show Menu Hidden" do the same as before

If we did that then yes, I think so. Those options are just about what options are shown to the user when editing.

I think there's another question about whether we hide all those options for alarms (menu, group, etc) in an `Advanced` menu rather than making them switchable - but I think that's something for another time!

thyttan:
> Another option to consider: What if we don't have an option to show hidden alarms, but instead just have a menu item under `Advanced` called `Hidden Alarms` and then any hidden alarms are shown under there?
> 
> Then most people won't ever go down to the bottom of the Advanced menu, but if you do want to look it's there without having to mess with settings?

I had a go at this. While I think it sounds like a better user experience, so far it's been cumbersome for me to refactor the logic to get it to work. While it feels like it should be somewhat straight forward, there are many variables to juggle in the whole flow for me at least 🙃

A diff between this PR and my current attempt here: https://github.com/thyttan/BangleApps/compare/alarm-sched-tweak-hidden-logic...thyttan:BangleApps:alarm-hidden-in-advanced?expand=1

So I'm not too eager to work more on that at the moment. But maybe we can come back to it later on if we feel like it?

gfwilliams:
Thanks for having a go with it! Yeah, maybe it's not as easy as I thought!

But yeah, this PR looks pretty nice and clean now. You say you want to wait a few days on it?

Assuming it all seems to work ok I'm happy to merge whenever you are!

thyttan:
> A diff between this PR and my current attempt here: https://github.com/thyttan/BangleApps/compare/alarm-sched-tweak-hidden-logic...thyttan:BangleApps:alarm-hidden-in-advanced?expand=1

I got somewhere with it after taking another stab at it. So if your interested you can look at that diff again now.

The logic gets a bit more involved, but maybe it can be made simpler.

But also, it builds on this PR, so merging this here now doesn't hurt the prospect of moving on to that other code.

> Assuming it all seems to work ok I'm happy to merge whenever you are!

Yes, I think we can merge this PR now.

gfwilliams:
Thanks! Ahh, yes, it's annoying having to remove the `hidden` group. I'll have a better look around that in a bit.

And yes, let's merge this one now!

PR #4243, opened 2026-06-04
boxclk: use clockbg module in interface.html, fix day suffix

stweedo:
Bumps Box Clock to v0.14 for a watch-side date suffix fix. When suffixes were enabled, dates such as 1st, 2nd, 3rd, and 4th displayed correctly in the browser editor but not on the watch.

The editor now loads apps/clockbg/lib.js directly and renders clockbg previews through a small Bangle.js graphics shim. This removes the copied pattern list from interface.html, so new clockbg patterns like gradients can preview without updating Box Clock’s editor code.


thyttan:
Thanks!

PR #4250, opened 2026-06-09
[Notify] Update README with all options

RKBoss6:
Updated usage options in README to reflect new properties and formatting. I don't know when it was last updated, but was perusing through the code and noticed a bunch of properties that weren't in the readme docs.


bobrippling:
LGTM, thanks for the improvement!

PR #4252, opened 2026-06-17
runapptests.js: Workaround for incomplete emulator reset

halemmerich:
This works around the incomplete reset in `emu.factoryReset()` by a full re-init for every test run.

I have poked around in the emulator itself, but could not get it working correctly. This is a bit slower than it was, but at least the `Uncaught SyntaxError: Got [ERASED] expected EOF` errors seem to be gone for now. I think it is somehow related to the internal state of `Storage` that does not get fully reset on factory reset.

The tests do not run in this repo because no apps changed, but you can have a look at the actions output here: [https://github.com/halemmerich/BangleApps/actions/runs/27714695139/job/81984461701](https://github.com/halemmerich/BangleApps/actions/runs/27714695139/job/81984461701)

### DIFF
--- a/bin/runapptests.js
+++ b/bin/runapptests.js
@@ -423,13 +423,17 @@ function runTest(test, testState) {
         if (test.description)
           console.log(`"${test.description}`);
         console.log(`==============================`);
-        emu.factoryReset();
-        console.log("> SENDING APP "+test.app);
-        emu.tx(command);
-        if (verbose)
-          console.log("> SENT APP");
-        emu.tx("reset()\n");
-        console.log("> RESET");
+
+        emu.stopIdle();
+        return emu.init(EMU_OPTIONS).then(() => {
+          console.log("> SENDING APP " + test.app);
+          emu.tx(command);
+          if (verbose)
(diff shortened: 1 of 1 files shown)


bobrippling:
I'm afraid I don't know enough about the emulator - probably one for @gfwilliams

halemmerich:
This is essentially a workaround like #4160 . It is a bit less complex by just doing a complete new initialization of the whole emulator instead of adding logic for waiting for finished commands. I think that should be fixed in the emulators reset methods to prevent internal state surviving.
For now I have removed the testindex stuff from this PR to make this easy to revert when the emulator has been fixed. At least it gets the CI tests working again.

gfwilliams:
Yes - I guess ideally we'd fix `factoryReset`. I think the issue is that while sending `"reset()"` the emulator is still running the old code while characters are sent and interpreted, right until the point the command is run. I think ideally we'd export the `jswrap_interface_reset` function from the emulator, but until then if this solution works then great - after all it's just for tests so it's not goingto effect the end user

PR #4254, opened 2026-06-18
fix: shorten sv_SE speed unit and improve usage stats submission reliability

MatrixNeoKozak:
This PR addresses two technical issues:

1. The 'sv_SE' speed unit was defined as 'km/tim' (6 characters), which exceeded the 5-character limit enforced by the sanity check script, leading to a known error. It has been updated to 'km/h' (4 characters), satisfying the validation requirement.

2. The 'sendUsageStats' function in 'loader.js' used the deprecated 'XMLHttpRequest' API and incorrectly updated the 'submittedUsageInfo' state even when the POST request failed. This prevented retries of failed analytics submissions. The function has been modernized to use 'fetch' with async/await, and it now only updates 'submittedUsageInfo' upon successful completion of the request.

### DIFF
--- a/apps/locale/locales.js
+++ b/apps/locale/locales.js
@@ -288,7 +288,7 @@ var locales = {
     lang: "sv_SE",
     decimal_point: ",",
     thousands_sep: ".",
-    speed: "km/tim",
+    speed: "km/h",
     distance: { 0: "m", 1: "km" },
     temperature: "°C",
     ampm: { 0: "fm", 1: "em" },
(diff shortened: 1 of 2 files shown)


bobrippling:
LGTM, and your fix has been validated by the [CI run](https://github.com/espruino/BangleApps/actions/runs/27778684462/job/82231658709?pr=4254) :)

Shall let @gfwilliams approve since this touches core files

gfwilliams:
Thanks! Yes, this looks good!

In general I wouldn't go around changing XMLHttpRequest to `fetch` if it's not broken though. It may be deprecated but I'm pretty sure it's not going anywhere anytime soon because of the sheer volume of old sites still using it

PR #4257, opened 2026-06-20
messagesoverlay: Better Robustness, less leaks

halemmerich:
This fixes several leaks and handler/watch backup/restore issues that in summary caused the watch to seemingly hang in certain conditions after longer periods of not restarting an switching between launcher and clock. This currently works a lot better than it has before and I could not find issues while testing. Now also includes tests for the fixed leaks and some improvements for runapptests.js that are used in those tests.

### DIFF
--- a/apps/messagesoverlay/lib.js
+++ b/apps/messagesoverlay/lib.js
@@ -278,8 +278,9 @@ const showCall = function(msg) {
   if (!isQuiet()) {
     if (msg.new) {
       msg.new = false;
-      if (callBuzzTimer) clearInterval(callBuzzTimer);
-      callBuzzTimer = setInterval(function() {
+      LOG("Setting call buzz interval")
+      callBuzzInterval = setInterval(function() {
+        LOG("Buzzing now");
         Bangle.buzz(500);
       }, 1000);
 
@@ -307,11 +308,12 @@ const next = function() {
   return true;
 };
 
-let callBuzzTimer = null;
+let callBuzzInterval = null;
 const stopCallBuzz = function() {
-  if (callBuzzTimer) {
-    clearInterval(callBuzzTimer);
-    callBuzzTimer = undefined;
+  if (callBuzzInterval) {
+    LOG("Stopping call buzz interval");
+    clearInterval(callBuzzInterval);
+    callBuzzInterval = undefined;
   }
 };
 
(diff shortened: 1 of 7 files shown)


apps/messagesoverlay/lib.js
```diff
@@ -278,8 +258,9 @@ const showCall = function(msg) {
   if (!isQuiet()) {
     if (msg.new) {
       msg.new = false;
-      if (callBuzzTimer) clearInterval(callBuzzTimer);
-      callBuzzTimer = setInterval(function() {
+      LOG("Setting call buzz interval")
+      callBuzzInterval = setInterval(function() {
```
bobrippling, line 262: Do we need to clear the old interval still?

halemmerich: I don't think so but I will check. Since the fix for the test harness has been merged I want to rebase the tests onto this anyway.

halemmerich: The stopCallBuzz called a few lines before clears the interval and undefines the callBuzzInterval variable.

bobrippling: Great, thanks for checking - I think the other changes pushed are just the merge/update, so I'll merge this in if you're happy?

halemmerich:
The failed tests become successful when using at least commit #dae980a (2v28.78) in EspruinoWebIDE. @gfwilliams Is there a reason why the web IDE repo is pinned to #a62becee (2v28) ?

apps/messagesoverlay/lib.js
halemmerich: Let's wait for Gordon to chime in on the emulator version. I suspect we could just use a newer 2v28 oder even the 2v29 emulator, but maybe there is a reason to stick with the initial 2v28.

gfwilliams:
> Is there a reason why the web IDE repo is pinned to #[a62becee](url) (2v28) ?

You mean at https://github.com/espruino/BangleApps/blob/master/.github/workflows/nodejs.yml#L48 ?

Well, that was @elijahr not me - but I assume the reasoning was that it was the latest IDE at that point, and it avoided tests inexplicably changing/failing if there was no change to the BangleApps repo.

So I think maybe we keep the pinning? But I see no reason why we shouldn't move to the most recent IDE commit. You might as well add it in this PR?

halemmerich:
I just bumped the version to the commit that introduces the 2v29 emulator since that will be what most people are using. Locally all tests pass and in the github action the tests for messagesoverlay do. I think this is good for now.

gfwilliams:
Great - thanks! Let's merge then!

PR #4260, opened 2026-06-23
Add author as 'Espruino' for apps that don't have a viable author

RKBoss6:
Stems from #4239, adds a line in metadata:
`author: "Espruino"` to apps that don't have an author.


gfwilliams:
Thanks!

PR #4261, opened 2026-06-23
[angles] Restore display orientation on exit

elfehr:
The Angles app rotates the display sideways, and it remains in that orientation after going back to the clock. This adds an E.kill event to restore the preferred orientation on exit.


bobrippling:
Thanks!

PR #4262, opened 2026-06-23
[C25K] Remember the day and week of program

elfehr:
The main screen of C25K proposes choosing a day and week, then starting the running program for that day. Currently it always starts at week 1 day 1, and it's the user who has to remember their progress and set the correct day. 

This adds a settings file to remember the day and week, and increments it automatically.

`incrementDay()` is called when the running program is started from the app's main menu, rather than when the exercise finishes, so that the function isn't triggered by custom exercises.


bobrippling:
Nice, thanks

PR #4266, opened 2026-06-25
[twenties] Fix missing last buzz at 18:00:20

elfehr:
Following up on the bug I reported in https://github.com/espruino/BangleApps/pull/4231, this changes how `isWorkTime()` is calculated to ensure than 18:00:00 is considered work time and 18:00:20 isn't.

Rather than stacking conditions like `d.getHours() < 18 || d.getHours() == 18 && d.getMinutes() == 0 && d.getSeconds() < 20`, I introduced new `Date` objects for the work day boundaries and compare the objects directly. That will also make it easier to eventually make the active hours user-configurable.

### DIFF
--- a/apps/twenties/lib.js
+++ b/apps/twenties/lib.js
@@ -6,8 +6,9 @@
   // - Add entry in settings to re-run setup.
 
   const getTimeAtNextBuzz = () => {
-    const isWorkTime = (d) =>
-      d.getDay() % 6 && d.getHours() >= 8 && d.getHours() < 18;
+    let workStart = new Date().setHours(8, 0, 0, 0);
+    let workdEnd = new Date().setHours(18, 0, 20, 0);
+    const isWorkTime = (d) => d.getDay() % 6 && d >= workStart && d < workdEnd;
     const isLookAwayTime = (d) =>
       d.getMinutes() % 20 === 0 && d.getSeconds() < 20;
     const NOW = new Date();
--- a/apps/twenties/lib.js
+++ b/apps/twenties/lib.js
@@ -6,13 +6,15 @@
   // - Add entry in settings to re-run setup.
 
   const getTimeAtNextBuzz = () => {
-    let workStart = new Date().setHours(8, 0, 0, 0);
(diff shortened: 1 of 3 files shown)


thyttan:
This looks good! Is it tested? 🙂

Please also bump version in metadata and changelog.

apps/twenties/lib.js
```diff
@@ -6,8 +6,9 @@
   // - Add entry in settings to re-run setup.
 
   const getTimeAtNextBuzz = () => {
-    const isWorkTime = (d) =>
-      d.getDay() % 6 && d.getHours() >= 8 && d.getHours() < 18;
+    let workStart = new Date().setHours(8, 0, 0, 0);
+    let workdEnd = new Date().setHours(18, 0, 20, 0);
```
thyttan: Typo `workdEnd`?

apps/twenties/lib.js
thyttan: Maybe the `workStart` variable you introduce should be used here as well? So they are logically bound if/when settings for start and end times are introduced.

thyttan: so 
```js
let t = [something with `workStart`];
```

apps/twenties/lib.js
elfehr: Oops, well spotted!

apps/twenties/lib.js
elfehr: Good idea!
It made me discover that annoyingly, `new Date().setHours()` returns a number and not a date object.

elfehr:
> This looks good! Is it tested? 🙂

It is! I installed it yesterday and it buzzed when it should :)

apps/twenties/lib.js
thyttan: Yes, iIrc it's the time since epoch?

thyttan: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setHours#return_value

thyttan:
I tried it a bit now - works! Thanks 😀🎆

PR #4267, opened 2026-06-26
[Health] Use theme colors and show units in graph.

RKBoss6:
Uses theme colors (bg2 and bgH) for the bar graph, and displays units (steps,bpm, etc), for each datum. Also uses g.findFont, so newer and more modern fonts are used, and everything fits nicely into the display.


bobrippling:
Thanks! Tagging @gfwilliams with this being a core app

gfwilliams:
Looks good - thanks!

PR #4269, opened 2026-06-27
[BlueWatch iOS Integration] New App!

RKBoss6:
At long last, my iOS integration app was finalized and approved by apple, and is [officially live](https://apps.apple.com/us/app/bluewatch-for-bangle-js/id6769088559)! Thank you to @gfwilliams for your support throughout the process!

That being said, this is the bangle.js companion to that app. I've also modified the `ios integration` app's description to let users know that they can use this app to gain additional features. I've been testing this pretty much for the last month or so while I finished up with Apple's strict review guidelines, so this should be well tested and ready to go without any bugs ;)


gfwilliams:
This is great - thanks for all your work! This should be a real help for iOS users.

RKBoss6:
Thanks!

PR #4272, opened 2026-06-27
New App: Line Dash - A swipeable analog clock and stats dashboard 

pagnotta:
This PR introduces **Line Dash**, a new clock face based on the beautiful [Line Clock](https://github.com/espruino/BangleApps/tree/master/apps/line_clock) by deepDiverPaul. The original minimalist line aesthetic was expanded into a swipeable suite of interactive dashboard gauges. 

**Features of the new app:**

- **Analog Clock:** Clean face that automatically inherits the system's 12/24h preference. Swipe up/down to briefly show the date (weekday and day/month, following the system locale).
- **Swipeable Dashboards:** Swipe left/right (or tap) to cycle between the gauges; every screen can be disabled individually in the settings. The app remembers the active screen across reloads — get a message while checking your heart rate mid-run and you return straight to the HR gauge, with the sensor re-enabled.
- **Steps Gauge:** A sweeping dial tracking your daily steps.
- **Distance Meter:** Distance covered today, based on steps and a configurable stride length, plus a persistent trip meter: swipe up to start a trip ("count from here"), swipe down to return to the day total — the trip keeps running in the background. Resetting a trip requires a confirmation (RESET? popup); trips survive reloads and end automatically at midnight.
- **Heart Rate Gauge:** Dynamic, color-coded tick marks indicating your current HR zone (based on an age-derived max HR). The sensor only runs while this screen is shown; optional live updates with a configurable interval to save power.
- **Battery Gauge:** A 180-degree fuel gauge with fixed color zones on the dial (green above 30%, a yellow warning band down to 15%, and a red reserve below — native 3-bit colors, no dithering). Automatically switches to this dashboard when the charger is connected and returns to the clock when unplugged.
- **Barometer:** A 300-degree 950-1050 hPa dial with 1 hPa subticks. Swipe up/down for the exact reading and an estimated altitude. The reading can be calibrated to sea level in the settings, so it matches local weather reports. The sensor is only powered while the screen is visible and unlocked (a single reading per minute while locked).

The app is intentionally full-screen and does not display widgets. An attribution to the original author has been placed in the `LICENSE` and `README.md` files.

app-screenshot-clock
app-screenshot
app-screenshot-step

app-screenshot-bat


apps.json
bobrippling, line 1: Is this used?

bobrippling (commented):
Looks great - just one question

apps.json
pagnotta: Hi Rob, thanks! No, that was by mistake. I've reset apps.json, package.json and package-lock.json to the original, so they are not affected anymore by the PR, it only adds the apps/line_dash folder.

bobrippling: Great, in that case, thanks for the app!

PR #4270, opened 2026-06-27
[quicklaunch] use launch_utils to allow fast loading

thyttan:
WIP

PR #4274, opened 2026-06-27
launch_utils: break out widgets check to separate method

thyttan:
The method takes an array of apps to check, then updates their entries in `launch.cache.json` and returns that updated launcher cache. The following example shows how this can be used to update widget info for all apps at once.

```
var launchCache = require("launch_utils").cache({showClocks:true,showLaunchers:true});
launchCache = require("launch_utils").cacheWidgetsCheck(launchCache.apps);
```


thyttan:
I use this in #4270 to update the settings object of quicklaunch to facilitate fast loading.

gfwilliams:
Looks good to me - thanks!

PR #4277, opened 2026-06-30
Pure Numerals Clock Version 0.04

peterbuehler:
Improved battery display with ringbuffer.
Blinking red dot when device is alive.


peterbuehler:
Title of pull request is wrong: it's version 0.04

bobrippling:
Thanks!

Issue #92, opened 2020-02-07
Installed settings from app cause Syntax error


MaBecker:
Hi, 

does https://banglejs.com/apps/ has the latest version?

Uncaught SyntaxError: Expecting a valid value, got ID
 at line 1 col 110
...false,"clock":null,"12hour":galse,"distance":"kilometer"}
                              ^
>require('Storage').list()
=[
  ".boot0",
  ".bootcde",
  "+boot",
  "+launch",
  "-launch",
  "+miclock",
  "-miclock",
  "*miclock",
  "+setting",
  "-setting",
  "@setting",
  "*setting"
 ]
>require('Storage').read("@setting")
="{\"ble\":true,\"dev\"" ... "nce\":\"kilometer\"}"
>console.log(require('Storage').read("@setting"))
{"ble":true,"dev":true,"timeout":10,"vibrate":true,"beep":true,"timezone":0,"HID":false,"clock":null,"12hour":galse,"distance":"kilometer"}
=undefined

gfwilliams:
Can you try re-installing?

I don't understand what happened here because `galse` isn't mentioned anywhere in the repository and if you look at the default settings file there's obviously no mention of it: https://banglejs.com/apps/apps/setting/settings-default.json

It's *possible* it's an internal error somewhere which cases an out of bounds JsVar lock/reference, so if you can find a way to reliably reproduce I can track it down.

MaBecker:
got a new one

 2v04.341 (c) 2019 G.Williams
Uncaught SyntaxError: Got '>' expected ','
 at line 1 col 12
{"ble":true>"dev":true,"timeout":10,"vibrate":true,"beep":tr...

gfwilliams:
Are you using the absolute newest firmware? I really need steps to reproduce this.

MaBecker:
> Are you using the absolute newest firmware?

Yes, just build this morning

```
=undefined
>process.env
={
  VERSION: "2v04.341",
  GIT_COMMIT: "1c3547a09",
  BOARD: "BANGLEJS",
  FLASH: 524288, RAM: 65536,
  SERIAL: "c7fa9624-75c07c51",
  CONSOLE: "Bluetooth",
  MODULES: "Flash,Storage,hea" ... "ow,graphical_menu",
  EXPTR: 536883676 }
> 
```

I will eraseAll() and start over again

MaBecker:
So after installing the apps

```
>require('Storage').list()
=[
  ".boot0",
[...]

PR #4279, opened 2026-06-30
Add required firmware version to README and sanitycheck

RKBoss6:
From App loader core [#92](https://github.com/espruino/EspruinoAppLoaderCore/pull/92), adds the `requiredFw` field to readme and `sanitycheck.js`

Also adds requiredFw to `loadanim` and removes the description message about the firmware, as the app loader now handles it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,7 @@ and which gives information about the app for the Launcher.
   "provides_modules" : ["messageicons"] // optional, this app provides a module that can be used with 'require'
   "provides_widgets" : ["battery"] // optional, this app provides a type of widget - 'alarm/battery/bluetooth/pedometer/message'
   "provides_features" : ["welcome"] // optional, this app provides some feature, used to ensure two aren't installed at once. Currently just 'welcome'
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -296,6 +296,7 @@ and which gives information about the app for the Launcher.
   "provides_modules" : ["messageicons"] // optional, this app provides a module that can be used with 'require'
   "provides_widgets" : ["battery"] // optional, this app provides a type of widget - 'alarm/battery/bluetooth/pedometer/message'
   "provides_features" : ["welcome"] // optional, this app provides some feature, used to ensure two aren't installed at once. Currently just 'welcome'
+  "requiredFw" : "2v20", // optional, specify a minimum firmware version your app needs to work properly
```
thyttan, line 299: Some bikesheding on my part: 

Does `requires_firmware` follow the style of `provides_modules` and `provides_features` that sits just before and is that better in that case?

gfwilliams: that's very true - it does match better with other stuff. I'll change and merge this along with https://github.com/espruino/EspruinoAppLoaderCore/pull/92

PR #4281, opened 2026-07-07
[Layout] Round buttons for consistency, use img scale when calculating btn size

RKBoss6:
This PR makes the layout buttons rounded to match the styles of `E.showPrompt` and `E.showAlert`, for better consistency throughout the system.
It also makes buttons take `img` scale into account when calculating the height/width

### DIFF
--- a/modules/Layout.js
+++ b/modules/Layout.js
@@ -175,23 +175,12 @@ Layout.prototype.render = function (l) {
     }, "btn":function(l){"ram";
       var x = l.x+(0|l.pad), y = l.y+(0|l.pad),
           w = l.w-(l.pad<<1), h = l.h-(l.pad<<1);
-      var poly = [
-        x,y+4,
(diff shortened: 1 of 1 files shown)


bobrippling:
I'll tag Gordon - core module and I suppose it'll affect all apps that use layout

gfwilliams:
Thanks! looks good - I'll merge and update Layout.min.js (and the one built into Bangle.js) as well.

Just FYI there are quite a few strange things going on here to try and get Layout to run at a usable speed, so you have to be really careful with any changes (but it looks like you were!) or performance will completely tank.

gfwilliams:
Actually I should have tested this myself before merging. We did have some tests for this in the `tests` folder and while you'd expect them to change because of the rounding, there changes made the buttons 1px wider and taller which pushed them offscreen in some cases:

image

I've now fixed this I hope.

PR #4282, opened 2026-07-08
taglaunch: fastload

nxdefiant:
Fast load apps similar to launcher_utils library

We can't just use launcher_utils because it does not store the tags we need.


nxdefiant:
should be fine now

bobrippling:
Nice, thanks!

PR #4288, opened 2026-07-16
line_dash v0.8: distance and trip as sub-views of the steps gauge

pagnotta:
Follow-up to #4272.

  * Distance and the trip meter are now sub-views of the steps gauge instead of
  a separate screen, matching the barometer/altimeter pattern: swipe up climbs
  steps → distance → trip, swipe down climbs back. Trip start and reset
  confirmation behave as before.
  * The "Show Distance" setting is gone — the sub-views come with the steps
  gauge. Saved state from the old distance screen migrates automatically.
  * Internal: app.js refactored into per-gauge sections with some code
  deduplication; no functional change.


bobrippling:
Thanks!

PR #4289, opened 2026-07-18
line_dash 0.9: timer sub-view on the clock gauge and altimeter drift sync

pagnotta:
This adds two features to Line Dash:

  **Timer** — a kitchen timer as a sub-view of the clock gauge (swipe up):

  * A tap opens a native menu with hours and minutes rows (up to 23:59); Start begins the countdown on a one-revolution-per-hour dial. Beyond an hour the dial wraps like the altimeter and the circle shows the remaining hours.
  * Swipe up pauses a running timer and asks to reset a paused one (RESET? popup); swipe down resumes and returns to the clock. Runs in the clock's white/red colors, magenta while paused or stopped.
  * The ringing is delegated to the system scheduler (new `scheduler` type dependency), so the timer fires with the standard alarm screen even while another app is in the foreground, and survives reboots.
  * The clock face carries a bezel-style marker at the hand position where a running timer will fire (hollow while paused), and the countdown stays live while the watch is locked (seconds during the final minute).

  **Altimeter sync** — a SYNC? popup on the altimeter view (swipe up):  confirming re-bases the altitude reference on the current reading, cancelling the weather drift (~8 m per hPa) without a trip to the settings menu.


bobrippling:
Thanks!

PR #4291, opened 2026-07-19
fix ac_ac customizer textbox width

DanielDecker:
For the app "A Configurable Analog Clock" the textboxes in the customizer in the section complications are wider than the configuration window. Entering a long URL into these results in a left shift of the UI elements. The result is, that the upload button at the end of the page is not visible any more.
This RP reduces the width for the mentioned textboxes from 50 to 40px, to fix the described problem.

Tested with Chromium and Firefox.
[Link to my BangleApps repo](https://danieldecker.github.io/BangleApps/?c=&q=ac_ac).

### DIFF
--- a/apps/ac_ac/Customizer.html
+++ b/apps/ac_ac/Customizer.html
@@ -673,7 +673,7 @@ console.log(AppSource)
     <tr>
       <td></td>
       <td>custom URL:</td>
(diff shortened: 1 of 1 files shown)


bobrippling:
Thanks!

PR #4174, opened 2026-02-16
Add a new app, the F3 Blue Moon Widget

mdoggydog:
Subject says it all:  adding a new moon-phase widget to the app catalog.

PR #4292, opened 2026-07-21
Add a new app, the F3 Blue Moon Widget

mdoggydog:
Subject says it all: adding a new moon-phase widget to the app catalog.

This is a resubmission of #4174, with all comments addressed.  I have removed all mention of the Mozilla Public License from the files.

(I would still prefer the MPL or other copyleft license, but five months later, I am just less worried about adhering to that principle for this code.)


bobrippling:
Thanks! Like with the other, with you being on the previous review, anything to add @gfwilliams ?

gfwilliams:
Thanks! Looks good!

PR #4198, opened 2026-03-23
Add a new app, the F3 Widget Tuner.

mdoggydog:
Subject says it all: adding a new widget-tweaking app to the app catalog.

PR #4293, opened 2026-07-21
Add a new app, the F3 Widget Tuner.

mdoggydog:
Subject says it all: adding a new widget-tweaking app to the app catalog.

This is a resubmission of #4198, with all comments addressed.  I have removed all mention of the Mozilla Public License from the files.

(I would still prefer the MPL or other copyleft license, but five months later, I am just less worried about adhering to that principle for this code.)


bobrippling:
Thanks, LGTM - @gfwilliams did you have anything you wanted to carry over from the prior PR (#4198) before I merge?

gfwilliams:
Looks fine to me as well - thanks @mdoggydog!

Totally understood about the license - thanks for being willing to make the change. I know it seems like it should be a minor thing but I just want to keep everything as simple as humanly possible.

PR #4294, opened 2026-07-21
[BlueWatch] Update for new iOS Bluewatch app version

RKBoss6:
Update for the iOS app version 1.1, which will be released when this is merged. 
Uses new device handshake method to allow for more redundant connections, resulting in less rejected connections between the two devices


RKBoss6:
Update: Apple has approved the iOS app's update, and right now, I'm holding it so it doesn't release the update yet, as this PR isn't merged yet. However, it would be amazing if we could try to get this in quick, as I don't know whether a release hold can expire for iOS, @thyttan or @bobrippling if you could help get this merged before it has a chance to expire, that would be absolutely amazing.

Please let me know, thanks all.

bobrippling:
Looks good, merged

PR #4297, opened 2026-07-24
[messagegui] Add support for notification actions

kidneyhex:
**GadgetBridge PR:** https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/6468

Link: https://kidneyhex.github.io/BangleApps/?id=messagegui

This adds support for Android notification "custom actions". 

GB sends an array of "actions" with just title properties, and then the index from that array is the ID sent back to GB with the INVOKE_ACTION command, so it triggers the action on the notification.

Examples where this is used:
* Task apps that have a "Perform", "Skip", or "Fail" (Do It Now, Habitica)
* Email apps that have "Trash", "Archive", "Spam" buttons (FairEmail, Gmail, Thunderbird)

What the actions array looks like:
```json
{
  "t": "notify",
  "id": 1234567890,
...
  "positive": true,
  "negative": true,
  "actions": [
    { "title": "Archive" },
    { "title": "Mark as read" },
    { "title": "Snooze" }
  ]
}
```

What the message back to GB looks like:
```json
{ "t":"notify", "n":"INVOKE_ACTION", "id":1234567890, "action":0 }
```

### Preview 

If there are actions, it shows a new swipe icon with the three vertical dots for "more actions" (a kebab menu icon):

screenshot_20260723-082051

If you swipe right, it then shows a menu with each action. If there's a "Reply" option, that will be next after the actions. If there's a "positive" option, that will be last as "Open".

screenshot_20260723-091600

"Actions" is also added to the message menu (when clicking the envelope icon in the message header).

### Questions
1) Should this be optional, with a settings parameter to either enable or disable it?

2) Should a "Dismiss" option that performs the "negative" action be added to the actions list?

3) Unicode: I have not tested what happens if action titles contain non-english text.

### DIFF
--- a/apps/messagegui/ChangeLog
+++ b/apps/messagegui/ChangeLog
@@ -125,3 +125,4 @@
       Use E.on("kill" event to ensure we always save messages before changing app
       Simplify message handling now messages lib automatically applies Bangle.messages
 0.92: Add support for displaying the time since the message was received using the 'date' parameter when pushing a message.
+0.93: Add support for notification actions.
--- a/apps/messagegui/app.js
+++ b/apps/messagegui/app.js
@@ -6,6 +6,10 @@
     body,
     sender,
     tel:string,
+    reply:bool,
+    positive:bool,
+    negative:bool,
+    actions: [{title:string}],
     new:true // not read yet
   }
 */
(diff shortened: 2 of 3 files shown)


thyttan:
This sounds cool! Crossing my fingers it can integrate good with anki notifications :)

For the UI. If there are many different 'positive' actions, maybe the swipe opens a menu to choose between them. Or a new screen with swipe options left, right, up and down?

kidneyhex:
> For the UI. If there are many different 'positive' actions, maybe the swipe opens a menu to choose between them. Or a new screen with swipe options left, right, up and down?

Yep, that's the "Actions" screenshot. Just a standard menu with each action listed, along with adding "Reply" and "Open" if either are also available.

The original person with an old PR in gadgetbridge (https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/4145) was going to add buttons directly in the message, but I didn't want to open that can of worms.

gfwilliams (approved):
This looks really good - thanks! Should be nicely backwards compatible with old Gadgetbridge too. I like the idea of the swipe+choose.

I don't think we need to have it switchable at the moment. Maybe if we get push-back from users (which i doubt!) we can do that later.

Only thing I'd ask is if we have `Reply` please can it be at the top of the list of actions? I feel like as that was the previous default action, it'll be a less confusing change if it's right at the top.

kidneyhex:
Yep, I've updated it to have Reply then the actions then Open

gfwilliams:
Great! I'll merge now then

PR #4298, opened 2026-07-24
[GBMediaRemote] Add new gadgetbridge media control app

cyberarm:
Adds a new app for controlling media through gadgetbridge.

The UI/UX deviates significantly from the existing [gbmusic](https://github.com/espruino/BangleApps/tree/master/apps/gbmusic) app hence creating a new app instead of updating the existing one.

screenshot_two


bobrippling (commented):
Looks great, thanks! If you're interested in a possible speed up, you can change `instance` in the methods to `this`, since they're equivalent and Espruino will find `this` quicker than it finds the global `instance`

Let me know if you want to make any changes, if not I'll merge

cyberarm:
I think I'll hold off on optimizing more for the time being.

bobrippling:
Cool, in that case - thanks again

PR #4299, opened 2026-07-27
[Daily Color Clock] Use colorpicker module and pad hours

RKBoss6:
Uses colorpicker module for settings and color selection. Also always pads 0 at the start of hours for better consistency and balance in the ui.


bobrippling:
Thanks!

PR #4300, opened 2026-07-28
tt: Create true time application

pavelmachek:
It shows time + uncertainty by default and precise time from GPS on request.


bobrippling:
Very cool, thanks!

PR #4304, opened 2026-07-29
[Launch] Use Bangle.haptic for haptics if available

RKBoss6:
Use Bangle.haptic for haptics if available. Currently it always calls the haptic function, but since it's a core app, it might be better to have it as a setting as well?

### DIFF
--- a/apps/iconlaunch/ChangeLog
+++ b/apps/iconlaunch/ChangeLog
@@ -28,3 +28,4 @@
       Ensure 'oneClickExit' is the default
 0.21: Display placeholder icons at start and render line by line to the screen to make loading seem faster
 0.22: Use launcher_utils library for app cacheing  and fast load
+0.23: Use Bangle.haptic if available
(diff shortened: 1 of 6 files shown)


bobrippling:
> Use Bangle.haptic for haptics if available. Currently it always calls the haptic function, but since it's a core app, it might be better to have it as a setting as well?

Good question - I've tagged Gorden, feel free to in your descriptions (in future) if you think it's relevant too :)

gfwilliams:
Thanks! Can you do `Bangle.haptic("touch")` instead please? Then potentially haptics can be disabled just for touch, or drag, or buttons. I wouldn't bother with an option.

RKBoss6:
Ah - that's a smart idea, I'll add that in right now

gfwilliams:
Looks great - thanks! I guess it might be good to add to other launchers too?

RKBoss6:
Sounds good! Should we add the different haptics to settings for people to choose specifics then, or just let apps override it?

gfwilliams:
>  Should we add the different haptics to settings for people to choose specifics then, or just let apps override it?

What do you mean? like `touch`, `btn`, `drag`? For now I thought we could have it as an app the overwrites `Bangle.haptic` - let's not overcomplicate it unless someone actually asks for the feature :)

RKBoss6:
I've added the haptics to `iconlaunch` as well, most other launchers have their own haptics implementations that I didn't feel like touching/removing :)

RKBoss6:
This looks good to go now I think, whenever you guys are ready to merge :)

gfwilliams:
Great, thanks!

PR #4306, opened 2026-08-03
[Apple Battery Widget] Create new app

RKBoss6:
Creates a new app based on the style of the iOS battery icon.


RKBoss6:
This looks good to go now!

thyttan:
[If you're making a widget please start the name with wid to make it easy to find!](https://github.com/espruino/BangleApps#example-applications)

( @bobrippling I guess we should add this to the PR merge check list? https://github.com/espruino/BangleApps/wiki/App-Contribution )

thyttan:
To clarify, the appid should start with wid. Not the user facing name.

RKBoss6:
Thanks, just pushed. We could also add it to `sanitycheck.js`, by checking if type is widget, making sure it has prefix "wid"

thyttan:
> We could also add it to `sanitycheck.js`, by checking if type is widget, making sure it has prefix "wid"

I think that's a good idea!

bobrippling:
> ( @bobrippling I guess we should add this to the PR merge check list? https://github.com/espruino/BangleApps/wiki/App-Contribution )

Good call, [how's this?](https://github.com/espruino/BangleApps/wiki/App-Contribution/_compare/5f2f5c2009093b12880170cfa5516c8431f3b1d2...e5d886e243f5196fe218b4660f3e45dd9648708d)

RKBoss6:
> I think that's a good idea!

Just made a PR for it, #4309

RKBoss6:
Now that #4309 is merged in, shall we merge this in as well?

thyttan:
Yes, thanks!

PR #4307, opened 2026-08-03
stormwatch: Clock with weather forecast

pavelmachek:
This shows clock and weather prediction graphs, current weather, or warnings about upcoming bad weather. It uses http over gadgetbridge, so should be easy to set up.


apps/stormwatch/stormwatch.app.js
```diff
@@ -0,0 +1,420 @@
...
+// TODO: logarithmic precipation
+// show precipation probability?
+// something to do with cape?
+
+const LOCATION_FILE = "mylocation.json";
+let pos;
+
+// requires the myLocation app
+function loadLocation() {
+  pos = require("Storage").readJSON(LOCATION_FILE,1)||{"lat":50,"lon":14.75,"location":"Czechia"};
```
bobrippling, line 12: Maybe emit an error instead of the default, which might confuse some users?

apps/stormwatch/stormwatch.app.js
```diff
@@ -0,0 +1,420 @@
...
+
+function msg(s) {
+  print("msg", s);
+  g.reset().clear();
+  g.setFont("Vector", 48);
+  g.drawString(s, 10, 10);
+  g.flip();
+}
+
+msg("uploading\...");
```
bobrippling: Is this meant to be here?


```suggestion
msg("uploading...");
```

pavelmachek:
Updated, should fix both issues. There's more work to do, but let me do that in v0.02.

bobrippling:
Sure, thanks for the clock!

PR #4309, opened 2026-08-10
[SanityCheck] Check for widget ids starting with 'wid'

RKBoss6:
From what was discussed in #4306, this adds checks to make sure there's a 'wid' and a 'clkinfo' prefix before ids of their respective types.

### DIFF
--- a/bin/sanitycheck.js
+++ b/bin/sanitycheck.js
@@ -250,8 +250,8 @@ apps.forEach((app,appIdx) => {
   }
   if (!app.description) ERROR(`App ${app.id} has no description`, {file:metadataFile});
   if (!app.icon) ERROR(`App ${app.id} has no icon`, {file:metadataFile});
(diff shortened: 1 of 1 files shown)


RKBoss6:
Looks like there are some apps that don't start with their respective prefixes, and I can't see us changing all the app IDs and causing reinstall/delete problems in the app loader... We should probably add exceptions for these apps instead.

RKBoss6:
Added exceptions, seems good to me.

thyttan:
Background: 

>[If you're making a widget please start the name with wid to make it easy to find!](https://github.com/espruino/BangleApps#example-applications) (BangleApps README)


_Originally posted by @thyttan in https://github.com/espruino/BangleApps/pull/4306#issuecomment-5219556942_

bobrippling:
LGTM, thanks!

PR #4313, opened 2026-08-13
ateatimer: add fastloading to/from app

thyttan:
I think this is in good shape - but will test a couple of days before I mark as ready. 

Tested for memory leaks.

Tagging @vovcia-clug as original author.

Can be tested at https://thyttan.github.io/BangleApps/?q=ateatimer .


vovcia-clug:
Tested on latest firmware, looks good to me

thyttan:
Thanks for testing!

PR #4314, opened 2026-08-15
fix(tally): jump to settings when no tally types are configured

### LINKED ISSUE #3853: [tally] By default Tally app is entirely empty
### Affected hardware version

Bangle 2

### Your firmware version

2v26

### The bug

This looks really handy, but when first installed I feel like the first thing I want do do is click on Tally app, and when I click on it I literally just see an empty menu saying `Tallies`

It'd be great if this either detected when there weren't any tallies and went straight to the `Settings` app, or maybe it always has a menu item at the bottom of the app which goes to settings, so you can easily add/change them?

Also, would it be nice if the clockinfo could actually tell you what the current tally was? That one might be easier said than done though by the look of it

### Installed apps

_No response_

xhon-pelushi:
Fixes #3853

On first install (or after removing all tally types), the app opened
straight to an empty "Tallies" list with no way to tell what to do next
— nothing to tally, and no visible path to add a type.

`app.js` now checks `tallycfg.json` (the configured tally types) before
showing the list; if it's empty, it loads the app's own settings screen
instead, using the same `eval(storage.read("<id>.settings.js"))(back)`
pattern already used elsewhere in this repo (e.g. `alarm`, `zambretti`).

Bumped to 0.03 with a ChangeLog entry.

## Test plan
- [x] Rebuilt `app.js` from `app.ts` via the real `typescript/build.sh`
      (`tsc`), rather than hand-editing the compiled output.
- [x] `node bin/sanitycheck.js` — 0 errors, 0 warnings (31 known/
      pre-existing errors and 13 known warnings elsewhere, unrelated).
- [x] `npx eslint apps/tally --max-warnings 0` — clean.
- [ ] Not tested on real Bangle.js 2 hardware or the emulator (none
      available in my environment) — would appreciate a maintainer or
      the reporter confirming the on-device behavior.

### DIFF
--- a/apps/tally/ChangeLog
+++ b/apps/tally/ChangeLog
@@ -1,2 +1,3 @@
 0.01: New app!
 0.02: Handle duplicate menu entries in the app
+0.03: Go straight to settings when no tally types are configured yet, instead of an empty list
--- a/apps/tally/app.js
+++ b/apps/tally/app.js
@@ -103,4 +103,10 @@ function tfmt(tally) {
     return "".concat(d.getHours(), ":").concat(pad2(d.getMinutes()));
 }
 var pad2 = function (s) { return ("0" + s.toFixed(0)).slice(-2); };
-showTallies(readTallies());
+var tallyCfg = storage.readJSON("tallycfg.json", 1) || [];
+if (tallyCfg.length === 0) {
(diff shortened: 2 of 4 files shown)


bobrippling:
Nice, thanks!

PR #4318, opened 2026-08-16
harden: wrap URL query params in encodeURIComponent in airqualityci clkinfo.js

anupamme:
## Summary

Harden URL construction in `apps/airqualityci/clkinfo.js` by wrapping the city, state, country, and API key query parameters in `encodeURIComponent()`.

## Motivation

The current code interpolates user-configured values directly into a URL template string. While the values come from the device's own settings (not arbitrary user input), `encodeURIComponent()` is the correct way to build URL query strings: it ensures that any spaces or special characters in a location name (e.g. "New York", "São Paulo") don't malform the request URL.

This is defensive URL/input handling — not a critical security fix. The API key is transmitted over HTTPS and `encodeURIComponent()` does not prevent it from appearing in server or proxy logs.

## Changes

- `apps/airqualityci/clkinfo.js` — four `encodeURIComponent()` calls added to URL construction (line 11)
- `apps/airqualityci/metadata.json` — version bumped to 0.02
- `apps/airqualityci/ChangeLog` — entry added for 0.02
- `apps/airqualityci/test.json` — emulator test added (see Testing below)

## Testing

Tested using the repo's emulator-based test runner (per TESTING.md):

```
node bin/runapptests.js --id airqualityci --verbose
```

Result: **SUCCESS** — all 6 assertions passed, including:
- `city=New%20York%20City` (space encoded)
- `state=New%20York` (space encoded)
- `country=US` (unchanged)
- `key=test%20key%26special` (space and `&` encoded)
- Negative assertion: raw `city=New York` does NOT appear in the URL

### DIFF
--- a/apps/airqualityci/ChangeLog
+++ b/apps/airqualityci/ChangeLog
@@ -1,1 +1,2 @@
 0.01: Initial clockinfo release
+0.02: Harden URL construction with encodeURIComponent for city, state, country, and API key params
--- a/apps/airqualityci/clkinfo.js
+++ b/apps/airqualityci/clkinfo.js
@@ -8,7 +8,7 @@
       config.rows.forEach((row) => {
         if (row.url) {
           let el = row.url.replace(/\/$/, "").split("/").reverse();
-          let url = `https://api.airvisual.com/v2/city?city=${el[0]}&state=${el[1]}&country=${el[2]}&key=${config.apiKey}`;
+          let url = `https://api.airvisual.com/v2/city?city=${encodeURIComponent(el[0])}&state=${encodeURIComponent(el[1])}&country=${encodeURIComponent(el[2])}&key=${encodeURIComponent(config.apiKey)}`;
           data[row.url] = data[row.url] || {};
           // If neither attempt nor time are set, then we have never tried
           // If attempt was set more than 1 minute ago, try again
--- a/apps/airqualityci/metadata.json
+++ b/apps/airqualityci/metadata.json
@@ -1,6 +1,6 @@
 { "id": "airqualityci",
   "name": "Air Quality Clockinfo",
-  "version": "0.01",
+  "version": "0.02",
   "author": "chuckwagoncomputing",
   "description": "Create clockinfos for air quality and temperature",
   "icon": "app.png",
--- a/apps/airqualityci/test.json
+++ b/apps/airqualityci/test.json
@@ -1,12 +1,13 @@
 {
   "app": "airqualityci",
+  "allowCustomApp": true,
   "setup": [
     {
       "id": "default",
       "steps": [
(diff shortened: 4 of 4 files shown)


thyttan:
@gfwilliams can you take a look?

If this is the case for airqualityci maybe the same thing applies to other user entered api-keys?

Not sure myself if this is a problem.

gfwilliams:
Got so say I'm not convinced this is a big problem either. 

It's an API key to an air quality API, and it's sent over HTTPS. I'm not sure what can be done with it that warrants `Severity: HIGH`, or what malicious thing the air quality website is going to do with the API key when they see it in their logs.

Storing an API key for something like this in localStorage is OK IMO too. Maybe if it's for a password manager, GitHub, email, etc we might want to reconsider, but not this.

Has @anupamme tested the fix on a Bangle to ensure this still works, as we ask all contributors to do? Can you confirm @anupamme?

If they have, great, let's merge. If not, I don't think it's worth the risk of breaking things for something that doesn't really pose a threat.

anupamme:
Thanks for taking a look. I agree that I overstated the severity in the original description.

The intent of this change is better described as defensive URL/input handling rather than preventing API-key exposure in logs. The API key is necessarily sent to the IQAir API over HTTPS, so `encodeURIComponent()` doesn't prevent the key from appearing in server/proxy logs.

I've tested the change using the repo's emulator test runner (per TESTING.md) and all assertions pass — including a location name with spaces ("New York City") and an API key with special characters. The emulator confirms the URL is correctly percent-encoded. I've also added `test.json` to the app directory so this is covered going forward.

I've also cleaned up the PR description to remove the incorrect Node.js/downstream-consumer threat model and the HIGH severity classification.

Thanks for the clarification!

gfwilliams:
Thanks - but it's not actually been tested with the custom.html or on a watch? Do you have any real hardware to test on?

My concern was that the `custom.html` could have inserted something in the JSON that was pre-escaped, and then escaping it again would break that.

anupamme:
No, I do not have real hardware to test. I can only use an emulator.

gfwilliams:
Sorry, I think that despite your test, this is actually breaking things. Any URL from https://www.iqair.com that has anything pre-urlencoded will break with your changes.

* custom.html  writes `row.url` as `https://www.iqair.com/usa/new-york/new-york` (as the example and the default URL in the input).
* The variable `el` is made from that URL **which is already urlencoded, because it's a URL**

So you absolutely shouldn't re-encode with encodeURIComponent.

Maybe the API key could be encoded, but it's alphanumeric. If you put non-alpha stuff into the key then it's not a proper API key and it'd break anyway.

I'm all for trying to discover issues, but at the end of the day, if you're submitting untested code that doesn't solve a real vulnerability and that's actually breaking things, that's not helping at all.

anupamme:
Ack. I will take this as feedback and add layers of verification to ensure the fix we are proposing works and does not break the existing functionalty.

PR #4322, opened 2026-08-17
[BlueWatch] Update for app version 1.4

RKBoss6:
Fixes bug with saving mylocation, and adds support for the iOS shortcuts app to send/emit the message.
Also adds support for sending calories data if the Calories app is installed.


bobrippling:
Thanks!

PR #4325, opened 2026-08-21
Fix:  In LCARS clock, do not delete the timer alarm when leaving clock.

mdoggydog:
The TIMER_IDX alarm, used for the "1-tap timer" feature of the LCARS clock, should not be deleted by the UI remove() method.  Otherwise, the timer is lost whenever another app takes control of UI, e.g., when a message is displayed by Messages app.

This commit brings back the intended behavior:  when leaving the clock, an active timer remains active; and when returning to the clock, the clock will take control of the timer (display its current state, etc) again.


mdoggydog:
FYI:  The build/build failure is due to linter warnings that have nothing to do with the change I made.  I.e., those warnings apply to the code already in BangleApps.

I do not know what to do about that.

thyttan:
I fixed the lint warnings and pushed to your branch and this PR. Should be unproblematic but could you please test it once and I'll merge then? :)

mdoggydog:
> I fixed the lint warnings and pushed to your branch and this PR. Should be unproblematic but could you please test it once and I'll merge then? :)

From a quick test, yes, it appears to still be in good `:works-for-me:` shape.  Thanks!

thyttan:
Thanks!

PR #4332, opened 2026-08-31
notify: only `stopEventPropagation()` in event handlers

### LINKED ISSUE #4327: [notify] Calling `require("notify").hide()` outside of an event handler crashes
### Affected hardware version

Bangle 2

### Your firmware version

2v29

### The bug

When `require("notify").hide()` is called outside of an event handler, it throws an exception because "`E.stopEventPropagation()` called when not handling events".

I'm not sure if that's something to just ignore, or wrap in a try/catch, or what...

Recreate by running in REPL:
``` js
require("notify").show({ title: "Test", body: "Hello World" });
require("notify").hide();
```

Result:
```
>require("notify").hide();
Uncaught Error: E.stopEventPropagation() called when not handling events
    at hide (notify:9:222)
...gation&&E.stopEventPropagation();function anim(){pos+=4;if(pos>0){pos=...
                                 ^
    at REPL (:1:24)
require("notify").hide();
```

## Workaround

Call it in a setTimeout:

``` js
require("notify").show({ title: "Test", body: "Hello World" });
setTimeout(() => require("notify").hide(), 1000);
```

### Installed apps

_No response_

### Author

_No response_

bobrippling:
Closes #4327

### DIFF
--- a/apps/notify/ChangeLog
+++ b/apps/notify/ChangeLog
@@ -11,3 +11,4 @@
 0.12: Add Bangle.js 2 support with Bangle.setLCDOverlay
 0.13: Add a default title background for the dark theme
 0.14: Stop event propagation on notification tap
+0.15: Only stop event propagation during events
--- a/apps/notify/metadata.json
+++ b/apps/notify/metadata.json
@@ -2,7 +2,7 @@
   "id": "notify",
   "name": "Notifications (default)",
   "shortName": "Notifications",
-  "version": "0.14",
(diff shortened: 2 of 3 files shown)


gfwilliams:
Looks good to me! I think we just need to bump the version. Thanks!

bobrippling:
Ah doh! Done

gfwilliams:
Great, thanks!

PR #4334, opened 2026-09-03
[IMC] Fix icon not being visible in launchers

RKBoss6:
Changed the icon to be inverted, so it's not a white icon on a white background, or vice versa.


bobrippling:
Thanks!

PR #4335, opened 2026-09-03
[GridLaunch] Fix last row not showing

RKBoss6:
This is a really silly mistake, but when testing, I had mistook the blank row caused by the invisible [imc icon](https://github.com/espruino/BangleApps/pull/4334) to mean that there was an extra row present. I've fixed that now, so the last row of apps will be shown.


thyttan:
Thanks!

PR #4337, opened 2026-09-06
[Weather] Redesign app for more info/readability

RKBoss6:
Changes the weather app to have a cleaner design, and show rain chance/high & low temperatures along with all the other data. 
Got inspiration from Pebble's new weather app, which uses new fonts and looks cleaner, and thought it might look good for Bangle as well. 
Since it's a pretty impactful change, any suggestions are welcome, and it doesn't have to get merged at all - It's just a design that I think looks pretty good.

Try it out at [my app loader](https://rkboss6.github.io/BangleApps/?id=weather)

I replaced the smaller vector fonts with the '14' font, as that's much more readable when small.

**New**:
screen

**Old**:
image


bobrippling:
Thanks!

PR #4342, opened 2026-09-08
[Weather] Fix widget redrawing bug

RKBoss6:
Fixes widget redrawing background using a previously set background from a different app.

### DIFF
--- a/apps/weather/widget.js
+++ b/apps/weather/widget.js
@@ -71,6 +71,7 @@
     return t;
   }
   function draw(){
+    g.reset("widget");
     g.clearRect(this.x, this.y, this.x+this.width-1, this.y+23);
     const w = weather.get();
     if (setting("hide")||!w){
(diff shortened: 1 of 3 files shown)


apps/weather/widget.js
```diff
@@ -71,14 +71,15 @@
     return t;
   }
   function draw(){
+    g.reset("widget");
```
bobrippling, line 74: `reset()` doesn't take any args, right?


```suggestion
    g.reset();
```

RKBoss6: I believe in 2v30+ or 2v29 cutting edge, resetting with parameter `widget` allows widgets to reset to show different widget bar colors: https://github.com/espruino/BangleApps/commit/c8920cfbfe12944165f9f40f568e5d66cb51723b,
https://github.com/espruino/Espruino/commit/835a481b92ead675e1af04e6128aafc6bbb6d06e

bobrippling: Ah nice! Thanks for the links too

PR #4345, opened 2026-09-10
widbt_notify: Fix hide widget when connected behavior

### LINKED ISSUE #4343: [widbt_notify / Bangle.drawWidgets] Redraw does not happen when width is toggled to 0
### Affected hardware version

Bangle 2

### Your firmware version

2v29.277

### The bug

The widget does not hide on connection when `hideConnected` setting is `true`.

The widget reacts to the NRF connection event and sets the width to zero which should prompt the subsequent Bangle.drawWidgets call to hide the widget. But the internal logic of Bangle.drawWidgets says to `continue` if the widgets width is zero and so it is not redrawn. If I understand correctly.

https://github.com/espruino/BangleApps/blob/555f074ae859678653ebd021c9ebddabffa39e2e/apps/widbt_notify/widget.js#L44

[Bangle.drawWidgets L11](https://github.com/espruino/Espruino/blob/d8322cec957a81e7a927b969a271568dad475c46/libs/js/banglejs/Bangle_drawWidgets_Bangle2.js#L11):
```
    if (!p || wd.width == 0) continue;
```

https://github.com/espruino/Espruino/commit/64958d3e65c3542fc86031959e35a67f8834840a was intended to "Only clear if widgets with width>0 exist" but that fails to account for this case where the widget did have a width on the previous draw but doesn't anymore.

Right now I'm inclined to update `Bangle.drawWidgets` but it's possible it's better to leave it as is and work around it in `widbt_notify`, e.g. like this: https://github.com/espruino/BangleApps/pull/3735 (similarly to how widbtstates does it now https://github.com/espruino/BangleApps/pull/4141)

@bobrippling what do you think?

This relates to e.g.:
https://github.com/espruino/BangleApps/pull/4141
https://github.com/espruino/Espruino/issues/2702
https://github.com/espruino/BangleApps/pull/3735

### Installed apps

_No response_

### Author

_No response_

thyttan:
Main author: @gfwilliams in https://github.com/espruino/BangleApps/issues/4343#issuecomment-5615703371

Closes #4343

### DIFF
--- a/apps/widbt_notify/ChangeLog
+++ b/apps/widbt_notify/ChangeLog
@@ -16,3 +16,4 @@
 0.17: Minor display fix
 0.18: Use notify lib to stop this widget clearing the screen of the running clock/app
 0.19: Adding .reset("widget") to support different colour widget bars
+0.20: Fix hide widget when connected behavior.
--- a/apps/widbt_notify/metadata.json
+++ b/apps/widbt_notify/metadata.json
@@ -1,7 +1,7 @@
 {
   "id": "widbt_notify",
   "name": "Bluetooth Widget with Notification",
-  "version": "0.19",
+  "version": "0.20",
   "author": "KungPhoo",
   "description": "Show the current Bluetooth connection status with some optional features: show message, buzz on connect/loss, hide always/if connected.",
   "icon": "widget.png",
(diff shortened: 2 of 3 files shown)


gfwilliams:
Great - thanks! Just checking - did you test this? Because I didn't check the code I posted :upside_down_face:

thyttan:
It seems to work as intended, yes!

PR #4347, opened 2026-09-11
Examen-New App

Camroc007:
Added a new app called Examen!


Camroc007:
I've updated this PR to also include the Liturgical Calendar app, as it complements the Examen app. I've fixed the metadata and linter warnings for both. If you would prefer these to be in separate Pull Requests, please let me know and I'll be happy to split them. Thanks!

bobrippling:
Thanks!

PR #4350, opened 2026-09-16
openstmap 0.34: upload maps along a GPX route

rambonette:
I take my Bangle.js 2 with me on hikes, and I usually plan a trip by downloading the GPX track and following it offline on my phone. I wanted something similar on the watch, so I don't have to keep pulling the phone out.

I tried GiPy first. Uploading a GPX file there is really handy, but I found the map hard to read on the watch's screen. OpenStreetMap looks much nicer, but covering a whole route meant adding maps one area at a time, which got tedious quickly. So I tried to get the best of both.

This adds a **GPX route** mode to the OpenStreetMap uploader: load a GPX file and upload the maps along the route, instead of picking areas one at a time.

image

Brief list of changes:

- Choose how far either side of the route to cover (±250m to 2km) and the level of detail (zoom 14 to 17)
- The maps are split into areas of up to 8x8 tiles on a shared grid, so they join up. No changes needed in the watch app
- The route is drawn onto the maps in magenta
- Before downloading, the estimated size is compared with the free space on the watch (free + trash, since Storage compacts itself when needed). It's checked again before uploading, which also covers **Get Map**
- Maps from one route are listed and deleted together

**Map area** mode is unchanged. The first commit only moves the Get Map rendering into functions so both modes can share it.

Tested on a Bangle.js 2 with a ~19km track.

Possible follow-up: draw the route as an overlay on the watch that can be turned on and off, instead of drawing it into the map images.

App loader: https://rambonette.github.io/BangleApps/?id=openstmap

P.S.: this is my first PR in this project, I've read the contribution wiki and I hope I got everything right :smile: sorry in advance if I messed up something :pray:


bobrippling:
Looks good to me - tagging @gfwilliams as the original author

bobrippling:
> App loader: https://rambonette.github.io/BangleApps/?id=openstmap
> P.S.: this is my first PR in this project, I've read the contribution wiki and I hope I got everything right 😄 sorry in advance if I messed up something 🙏

Oh and you nailed it, even with the extra link to an app loader - so thanks again

rambonette:
@bobrippling thanks for the feedback 🙏 

> Possible follow-up: draw the route as an overlay on the watch that can be turned on and off, instead of drawing it into the map images.

also regarding this, I know that the better solution would've been to draw the GPX track as overlay in the watch (I'll actually attempt that in the near future) so the user can hide/show the track, but I didn't want the PR to be too big. While I'm pretty confident in my web dev skills, I'm still in a tinkering phase with the actual Espruino JS subset 😄

bobrippling:
Sounds good, I'm sure a follow up PR is a nice way to break up those changes :)

gfwilliams:
This looks awesome, thank you! I'm very happy to merge this in.

One thing I'd say for the future is maybe you could detect areas that are next to each other and merge them. In your example you have a 6 long row that could be 48x8 instead of 6x8x8. I could be wrong and maybe you tried it already, but it feels like the less separate map areas you have the faster it'll be.

On the GPX overlay front, I feel like your pre-rendered solution is quite sensible and works well - maybe it's nice to have an option not to render it though? We do currently have some code in there that overlays a track from the GPS recorder, and it does it by rendering that in the background so that you can scroll quickly and it'll update over time. You could use that kind of thing to draw a GPX trace (but maybe you'll want to simplify the GPX track as much as you can first to make it render quicker). On the cutting edge firmwares I also just added a variable line width (`g.lineTo(x,y,width)`) which could make drawing the track a lot more sensible too.

PR #4351, opened 2026-09-16
popconlaunch: Recognize launches made with launch_utils and `Bangle.load`

ticalc-travis:
This updates Popcon so that it works with recent launchers that use `launch_utils` (which in turn uses `Bangle.load`). Previously Popcon only piggybacked the global `load` to watch for launches. This patch makes it do the same for `Bangle.load` as well. (I also threw in a `console.debug` call to make it easy to confirm that it is seeing and recording the launches.)


bobrippling:
Nice, thanks for sorting this out! One request - would you be able to port the changes to [the typescript source file](https://github.com/espruino/BangleApps/blob/3a076e24c14b72010a5fae0189d5e814cab95c37/apps/popconlaunch/boot.ts) and if you can, regenerate the javascript from that (`(cd typescript && ./build.sh)`)? If not, let me know and I can do either of those

https://github.com/espruino/BangleApps/blob/3a076e24c14b72010a5fae0189d5e814cab95c37/apps/popconlaunch/boot.ts#L99-L115

ticalc-travis:
I pushed a commit. Hopefully I did it right. I don't know anything about TypeScript and had to do some research to get it working.

bobrippling:
Great, looks spot on. Made a minor tweaks to remove the casts, but thanks!

PR #4352, opened 2026-09-17
[New app] ToyBox

tonykakuuu:
This is a new app that lets you (or a toddler within reach) draw lines and make waves on your watch. It's a collection of demos/toys and I might add more in the future. The app contains its own launcher to select a toy, long-pressing the button takes you back to the watch face.


bobrippling:
Nice idea, thanks for the app!

PR #4354, opened 2026-09-19
Backup: Include date in filename

RKBoss6:
Includes the date in the backup.zip file name:
Looks something like this: `Bangle.js Backup 09_19_2026.zip`


thyttan:
What's the reason for the change? Had any trouble yourself?

If we do this I suggest to conform to https://en.wikipedia.org/wiki/ISO_8601

RKBoss6:
ISO formatting is a good idea, yes. I usually name all my backups with the date, and I know other users do too, and it would be a good thing for the system to do automatically - especially since most people backup before they change something significant/backup periodically.

pavelmachek:
For the record, yes, I'd like to see date in the filename, too.

gfwilliams:
Yes, adding date sounds good - I usually end up adding it manually. But could you do it ISO please?

As far as I can tell this:

```
new Intl.DateTimeFormat(undefined, {
      month: '2-digit',
      day: '2-digit',
      year: 'numeric'
    }).format(new Date())
'21/09/2026'
```

Produces slashes which on non-windows machines are directory separators - not a great idea for filenames!

Just do `(new Date()).toISOString().substr(0,10)`

RKBoss6:
Just changed - I also added it to the android backup name, I don't think there should be any problems with it as it's saved as `Banglejs backup 2026-09-22.zip`.

gfwilliams:
Thanks! Just made a minor tweak to use the exact same filename for both - merging now

